PR #2, opened 2023-09-18
Add release workflow

alfeilex:
This PR extend the release workflow to build native images for each major OS and then deploy it on nexus

Related issue: #4


cli/pom.xml
```diff
@@ -87,6 +94,39 @@
           </archive>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
```
hohwille, line 99: this must not be in the `<build>` section but needs to be moved to a profile.
That profile should have the `<id>` named `deploy` as this is our profile we already have and trigger for building and deploying a release:
https://github.com/devonfw/maven-parent/blob/b910467901c8b12e218ef0f3ac5bc8e9f60758c6/pom.xml#L294

cli/pom.xml
```diff
@@ -87,6 +94,39 @@
...
+          <execution>
+            <id>attach-artifacts</id>
+            <phase>package</phase>
+            <goals>
+              <goal>attach-artifact</goal>
+            </goals>
+            <configuration>
+              <artifacts>
+                <artifact>
+                  <file>cli-${version}-Linux.tar.gz</file>
```
hohwille, line 111: Shouldn't we use `${project.artifactId}` instead hard-coding `cli`?
Also we should align the OS with the syntax we are using in [ide-urls](https://github.com/devonfw/ide-urls/tree/master/java/java/17.0.8_7) and also in custom software repositories what is:
* `linux`
* `mac`
* `windows`

.github/workflows/test-release.yml
```diff
@@ -0,0 +1,76 @@
+name: GraalVM Native Image builds
```
hohwille, line 1: Will we rename the file to just `release.yml`?
Also the name should be something like `Build and deploy a release including native-image`

hohwille (commented):
Sorry, to make things complicated but can we also reuse the workflow to build the same without the tagging and staging to run every night (if something has changed) to build a SNAPSHOT release and deploy it to OSSRH?
Maybe the easiest approach would be to first get this release workflow right and then just copy the YML file to a nightly SNAPSHOT workflow and adapt it accordingly...

.github/workflows/test-release.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Build native images with maven plugin
+        shell: bash
+        run: |
+          maven_config="$(cat .mvn/maven.config)"
+          current_version="${maven_config/#*-Drevision=}"
+          current_version="${current_version/ */}"
+          cd cli
+          mvn -B -ntp -Pnative -DskipTests=true package
+          cd target
+          tar -cvf cli-${current_version}-${{ runner.os }}.tar cli*
```
hohwille, line 32: IMHO it would be better to solve all such things in maven rather than in workflows (see also my comment that we create a copy of this workflow and would then have such logic redundantly).

cli/pom.xml
alfeilex: done

cli/pom.xml
alfeilex: I created the tar.gz in the workflow and used the `${{ runner.os }}` for the name, which capitalises `Windows`

done

hohwille: Thanks. Great work 👍 However, you are IMHO still using `macos` instead of `mac`.

hohwille:
@alfeilex awesome improvements and progress 👍 Thanks!

cli/pom.xml
alfeilex: done

cli/pom.xml
```diff
@@ -117,18 +157,54 @@
               </execution>
             </executions>
             <configuration>
+              <quickBuild>true</quickBuild>
```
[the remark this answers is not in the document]
alfeilex: done

cli/pom.xml
```diff
@@ -120,15 +160,69 @@
               <agent>
                 <enabled>true</enabled>
               </agent>
+              <imageName>
+                ${imageName}-${os.detected.classifier}
+              </imageName>
               <buildArgs>
                 <arg>--enable-url-protocols=https</arg>
-                <arg>--static</arg>
+                <arg>-H:IncludeResources=".*"</arg>
```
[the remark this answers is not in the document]
hohwille: I guess, we actually do not need resources like `devon.properties`.
All the defaults have already been build into the Java code (see `IdeVariables`).
For other files like e.g. templates such as [project.xml](https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/templates/jenkins/project.xml), I would propose not to use java resources but instead simply bundle those via maven assembly plugin into the resulting release package (if we come to the point that we need those at all).

cli/src/main/resources/nls/en_US.properties
```diff
@@ -0,0 +1,26 @@
+usage=Usage:
```
hohwille, line 1: why do we need such file? IMHO this should be deleted it is also missing the `ide_` prefix and therefore seems to be pointless.

cli/src/main/resources/nls/ide_en.properties
```diff
@@ -0,0 +1,26 @@
+usage=Usage:
```
hohwille, line 1: I do not see that we ever want to support different texts for `en` vs. `en_US`.
The `ide.properties` will be used as fallback for any language that is not supported more explicitly.
We could add `ide_es.properties` and `ide_fr.properties` some day in the future...

hohwille (approved):
@alfeilex thanks for your excellent work on this very complex story and workflows. You did a very great job 👍 
I had changed the `groupId` and `artifactId` so we might need some minimal adoption here, but it looks very good already.
Also we surely need more content for the assembly like the `ide` and `ide.bat` scripts and also I guess some script for auto-completion. But all that can also be done after the merge.

.github/workflows/release.yml
```diff
@@ -58,7 +98,7 @@ jobs:
           git push
           git push --tags
           noDotVersion="${current_version//.}"
-          gh release create "release/${current_version}" --title "${current_version}" --notes "# Download
+          gh release create "release/${current_version}" ./cli/*.tar.gz --title "${current_version}" --notes "# Download
           https://repo1.maven.org/maven2/com/devonfw/tools/ide/devonfw-ide-cli/${current_version}/devonfw-ide-cli-${current_version}.tar.gz
```
hohwille, line 102: That link seems to be outdated. I guess it would be like this:
https://repo1.maven.org/maven2/com/devonfw/tools/IDEasy/ide-cli/${current_version}/ide-cli-${current_version}.tar.gz

.github/workflows/release.yml
```diff
@@ -23,8 +57,14 @@ jobs:
           cache: 'maven'
           gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }} # Value of the GPG private key to import
           gpg-passphrase: GPG_PASSPHRASE # env variable for GPG private key passphrase
+      
+      - uses: actions/download-artifact@v3
       - name: Publish to Apache Maven Central
         run: |
+          mv ./natives/* ./cli/
```
hohwille, line 64: ah, this is where this is comming from.
Shouldn't we follow the same best-practices like maven and keep artifacts generated during the build only in `target` folders?
The assembly-plugin will also bundle other static files that we need to include into the release. Those will also come from the `target` folder where the `maven-resources-plugin` will have copied them (after potential filtering).

cli/src/main/assembly/exec.xml
```diff
@@ -0,0 +1,28 @@
...
+        <exclude>*/**</exclude>
+      </excludes>
+    </fileSet>
+    <fileSet>
+      <directory>${project.build.directory}</directory>
+      <outputDirectory>./</outputDirectory>
+      <includes>
+        <include>${imageName}-windows.exe</include>
+        <include>${imageName}-linux</include>
+        <include>${imageName}-mac</include>
```
hohwille, line 24: I am not sure, if I get this correctly:
Are we building one generic *.tar.gz that contains the binaries for all operating systems?
Or do we build a different *.tar.gz release artifact per OS?
I thought we would do the latter and maybe we do.

Some further thoughts for future improvements if we do the latter:
* The binary inside the package should be named just `bin/ide[.exe]`.
* The download file of the release should be rather `ide-cli-«os»-«arch».tar.gz` where `arch` is most probably `x64` but for `mac` we could think of providing `arm64`.

cli/src/main/resources/nls/en_US.properties
alfeilex: I forgot to remove these files because I tried something to fix the last error that was fixed by adding the resources

cli/src/main/resources/nls/ide_en.properties
alfeilex: Done by removing

.github/workflows/release.yml
alfeilex: Done

cli/src/main/assembly/exec.xml
alfeilex: Currently we build three different `*.tar.gz `. 

We can also create one big `tar.gz` with all binaries, but should not we also provide OS specific downloads, specifically on GitHub? Maybe we can provide a generic tar.gz for Maven Central and upload OS specific tar.gz as assets to the GitHub Release?

cli/pom.xml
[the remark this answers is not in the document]
alfeilex: So I will change this by just including the `nls` folder with `H:IncludeResources`?

.github/workflows/release.yml
alfeilex: Done

cli/pom.xml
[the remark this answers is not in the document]
alfeilex: I just included the `nls` folder in the `H:IncludeResources`

cli/src/main/assembly/exec.xml
hohwille: All fine. I just wanted to understand and clarify. What confused me is that in the same assembly declaration we have includes for all 3 OS types.

Creating downloads per OS is exactly what we had discussed and decided so far. So you did this perfectly correct.
I also see the advantage that this way we only have a single `ide[.exe]` binary. Therefore our bash wrapper script will just call that binary and life is easy. What if we would create one big release containing `ide` binary for linux and for mac x64 and mac aarm64? Would make all more complicated.
A little bit interesting will be the case that one is working on windows but also wants to use IDEasy inside WSL sharing the same installation (esp. `IDE_ROOT`). Maybe then such user can manually install the linux binary aside so he has a `bin` folder with both `ide.exe` and linux elf `ide` binary and we can make this work properly from the `ide` bash wrapper script. However, nothing to solve right now. Considered done.

.github/workflows/release.yml
hohwille: I do not see this being changed yet. However, it is more a suggestion or nice-to-have improvement.
We can also improve after merge, so I stop blocking this PR with this detail.

PR #53, opened 2023-09-28
#8: implemented progress bar

jan-vcapgemini:
Addresses: #8 

Implements:
* added me.tongfei.progressbar 0.10.0 to dependencies
* replaced regular Files copy with progress bar copy
* added test infrastructure and 1st test


cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -83,6 +91,92 @@ public void download(String url, Path target) {
...
+   * @param response {@link HttpResponse} to use
+   */
+  private void downloadFileWithProgressBar(String url, Path target, HttpResponse<InputStream> response) throws IOException {
+
+    String contentLength = response.headers().firstValue("content-length").orElse("undefined");
+    if (contentLength.equals("undefined")) {
+      throw new IllegalStateException("Content-Length was not provided by download source: " + url);
+    }
+
+    long size = Long.parseLong(contentLength);
```
hohwille, line 108: Please use `firstValueAsLong` directly instead of manual parsing. Also consider that the server could send an invalid content header like `31MB`. I would prefer to be as robust as possible and rather catch such exceptions falling back to some default behavior (e.g. assume a download size of 10MB and having a progress bar that does not work smooth or even skip the process bar entirely).

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -83,6 +91,92 @@ public void download(String url, Path target) {
...
+
+    byte[] data = new byte[1024];
+    boolean fileComplete = false;
+    int count;
```
hohwille, line 112: To be pedantic, a `long` would be the correct value type here instead of `int`.
I do not really assume that IDEasy will ever download files bigger than 2 terra-bytes but I want to teach developers to avoid accidentally creating such unnecessary limitations. In a customer project a 2 TB limit could cause a production bug one day :)

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -83,6 +91,92 @@ public void download(String url, Path target) {
...
+        ProgressBar pb = pbb.build()) {
+      while (!fileComplete) {
+        count = body.read(data, 0, 1024);
+        if (count <= 0) {
+          fileComplete = true;
+        } else {
+          bufferedOut.write(data, 0, count);
+          pb.stepBy(count);
+        }
+      }
```
hohwille, line 127: Nice solution. What we IMHO will need is some abstraction as in tests or in the GUI the `ProgressBar` implementation has to be exchanged with something else.
This abstraction should IMHO be provided in (Abstract)IdeContext.

hohwille (changes requested):
@jan-vcapgemini Thanks for this PR. Great solution. 👍 
However, I left some review comments to get the last things right.
Could you please have a look?

cli/pom.xml
```diff
@@ -69,6 +69,11 @@
       <version>2.35.1</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>me.tongfei</groupId>
+      <artifactId>progressbar</artifactId>
```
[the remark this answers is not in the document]
jan-vcapgemini: I've added it to our LICENSE now.
Don't know how to check the GraalVM compatibility yet. Open for hints.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -83,6 +91,92 @@ public void download(String url, Path target) {
...
+   * @param url the url to download
+   * @param target Path of the target directory
+   * @param response {@link HttpResponse} to use
+   */
+  private void downloadFileWithProgressBar(String url, Path target, HttpResponse<InputStream> response) throws IOException {
+
+    String contentLength = response.headers().firstValue("content-length").orElse("undefined");
+    if (contentLength.equals("undefined")) {
+      throw new IllegalStateException("Content-Length was not provided by download source: " + url);
+    }
```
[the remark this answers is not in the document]
jan-vcapgemini: I've replaced the exception with a log warning now.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
jan-vcapgemini: I've added a fallback to 10MB now.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
jan-vcapgemini: I've refactored the ProgressBar to IdeContext now.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
hohwille: Sorry, my mistake. This is not the counter of the total filesize but only the size of a chunk that is anyway limited to the array lenth (1024).
Can you avoid redundantly specifying the buffer size `1024` as literal (use `data.length` instead of repeating 1024).

cli/pom.xml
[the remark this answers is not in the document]
hohwille: Still not so easy. The nightly build & the release process are also not yet working smooth yet. We will have to do a KT session how to build the native image locally and run + test it (on Windows).
Considered resolved for now.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
hohwille: IMHO the updates we have done together will resolve all this. I will wait until those are pushed and then we will most likely be able to merge this PR.

hohwille (approved):
@jan-vcapgemini Thanks for completing this complex PR. Great result and ready for merge 👍

PR #54, opened 2023-09-28
Remove test workflow

alfeilex:
This PR removes the `test-release.yml` file that was created to test the release and does not work on the original repo.


hohwille (approved):
@alfeilex Excellent. Thanks for the update. Can directly be merged 👍

Issue #23, opened 2023-09-22
Implement ToolCommandlet for Gradle


hohwille:
Create `ToolCommandlet` for Gradle
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/gradle

PR #59, opened 2023-09-28
#23: implemented tool commandlet for gradle

aBega2000:
Addresses: #23  

added class Gradle.java
added cmd-gradle in Ide.properties
imported Gradle class and called the add method for Gradle in class CommandletManagerImpl.java


hohwille (approved):
@aBega2000 thanks for your PR. Looks like you got all correct. Great work 👍 
FYI: You did not add the German translation for `cmd-gradle` to `Ide_de.properties`.
We can discuss in the next team meeting if we better let this outdate for the moment to concentrate on dev progress and later do I18N/L10N in one go or if we always want to keep that in sync...

PR #60, opened 2023-09-29
#5: fixed assertLogMessage

moritzLanger:
#5 fixed bug where assertLogMessage would only check the Warning level


hohwille (approved):
@moritzLanger thanks for this PR & fix. 👍

Issue #28, opened 2023-09-22
Implement ToolCommandlet for Kotlin


hohwille:
Create `ToolCommandlet`s for Kotlin
Old commandlets:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/kotlinc
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/kotlinc-native

PR #64, opened 2023-10-02
#28 Implement ToolCommandlet for Kotlin

CREITZ25:
#28 Implement ToolCommandlet for Kotlin


hohwille (approved):
@CREITZ25 thanks for your PR. The two commandlets have also been straight forward in bash. 👍 Seems that your IDE setup is working correct and you do not have the UTF-8 encoding errors that I have seen in other PRs today.
We could add a `TAG_COMPILER` that would make sense here. But nice to have that can also be done later. Ready for merge...

Issue #29, opened 2023-09-22
Implement ToolCommandlet for Node.js


hohwille:
Create `ToolCommandlet` for Node.js
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/node

PR #68, opened 2023-10-05
#29: implemented tool commandlet for node.js

aBega2000:
Addresses: #29 

added class Node.java
added cmd-node in Ide.properties and in Ide_de.properties
imported Node class and called the add method for Node in class CommandletManagerImpl.java


hohwille (approved):
@aBega2000 thanks for your PR. All looking good. 👍 
We could add a TAG_JAVASCRIPT and assign it but that is nice to have and can be done after the merge as well. I approve this as it is complete and functional.
Please note that the bash node commandlet in devonfw-ide did add specific features for devon4node that we do not implement intentional to decouple IDEasy from `devon4*` stacks that are also deprecated so all is fine.

PR #73, opened 2023-10-09
#5: added Tests for VersionCommandlets

moritzLanger:
#5 added test for the following Classes: 

1. [VersionGetCommandlet](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/commandlet/VersionGetCommandlet.java)
2. [VersionSetCommandlet](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/commandlet/VersionSetCommandlet.java)
3. [VersionListCommandlet](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/commandlet/VersionListCommandlet.java)
4. [HelpCommandlet](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/commandlet/HelpCommandlet.java)

fixed minor bug, where the a "Not Operator" was at the wrong position in [HelpCommandlet](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/commandlet/HelpCommandlet.java)


cli/src/test/java/com/devonfw/tools/ide/commandlet/HelpCommandletTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Test of {@link HelpCommandlet} does not require home.
+   */
+  @Test
+  public void testThatHomeIsNotReqired() {
+
+    // arrange
+    String path = "workspaces/foo-test/my-git-repo";
+    IdeContext context = newContext("basic", path, false);
```
hohwille, line 21: for such simple tests, you can also use `IdeTestContextMock.get()`

cli/src/test/java/com/devonfw/tools/ide/commandlet/HelpCommandletTest.java
```diff
@@ -0,0 +1,93 @@
...
+    assertLogMessage(context, IdeLogLevel.INFO, "-v | --version  Print the IDE version and exit.");
+  }
+
+  /**
+   * Assertion for the IDE-Logo that should be displayed.
+   */
+  public void assertLogoMessage(IdeContext context) {
+
+    assertLogMessage(context, IdeLogLevel.INFO,
+        "__       ___ ___  ___\n" + "\\ \\     |_ _|   \\| __|__ _ ____ _\n" + " > >     | || |) | _|/ _` (_-< || |\n"
```
hohwille, line 89: you can add some empty comment markers to for IDEs not to reformat. Also in this case, we could simply make `HelpCommandlet.LOGO` package visible (remove `private` keyword) and directly access it from here using `split("\n")`.

cli/src/test/java/com/devonfw/tools/ide/commandlet/VersionGetCommandletTest.java
```diff
@@ -0,0 +1,54 @@
...
+      String path = "workspaces/foo-test/my-git-repo";
+      IdeContext context = newContext("basic", path, false);
+      VersionGetCommandlet versionGet = context.getCommandletManager().getCommandlet(VersionGetCommandlet.class);
+      versionGet.tool.setValueAsString("java");
+      // act
+      try {
+        versionGet.run();
+      } catch (CliException e) {
+        //assert
+        assertThat(e).isInstanceOf(Exception.class)
```
hohwille, line 30: `isInstanceof(Exception.class)` does not really make sense here as `CliException` extends `RuntimeException` what extends `Exception` so that assertion is a static `true` that can never fail.

cli/src/test/java/com/devonfw/tools/ide/commandlet/HelpCommandletTest.java
moritzLanger: Done

cli/src/test/java/com/devonfw/tools/ide/commandlet/VersionGetCommandletTest.java
```diff
@@ -0,0 +1,54 @@
...
+            .hasMessageContaining("Tool java is not installed!");
+        return;
+      }
+      //just a false case if versionGet.run() does not throw the Exception
+      assertThat(true).isFalse();
```
hohwille, line 35: The pattern for this is to call `failBecauseExceptionWasNotThrown(CliException.class)` at the end of the `try { ... }` block.

cli/src/test/java/com/devonfw/tools/ide/commandlet/VersionSetCommandletTest.java
```diff
@@ -0,0 +1,29 @@
...
+    //arrange
+    String path = "workspaces/foo-test/my-git-repo";
+    IdeContext context = newContext("basic", path, true);
+    VersionSetCommandlet versionSet = context.getCommandletManager().getCommandlet(VersionSetCommandlet.class);
+    versionSet.tool.setValueAsString("mvn");
+    versionSet.version.setValueAsString("1.0.0");
+    //act
+    versionSet.run();
+    //assert
+    assertThat(versionSet.version.getValueAsString()).isEqualTo("1.0.0");
```
hohwille, line 27: This is not the proper assertion here. Actually you need to read the `ide.properties` file and assert that it now contains `MVN_VERSION=1.0.0`.
BTW: Better use a version that you have added (e.g. `3.1.0`). We later might add a feature that will ask the user to confirm if the version that he selected does not exist.

moritzLanger: I tried using 
    assertThat(context.getSettingsPath().resolve("ide.properties")).hasContent("MVN_VERSION=3.1.0");
    
   which seems to work as well, however i receview the following error: (same result with your call)
   
   
Extra content at line 1:
  ["ide.properties",
   "ide.properties",
   "ide.properties",
   "ide.properties",
   "ide.properties",
   "ide.properties",
   "ide.properties",
   "ide.properties",
   "ide.properties",
   "ide.properties",
   "ide.properties",
   "ide.properties"]


	at com.devonfw.tools.ide.commandlet.VersionSetCommandletTest.testVersionSetCommandletRun(VersionSetCommandletTest.java:28)

moritzLanger: When setting the context copyForMutation flag false and looking at IDE.properties i see that the hole file was overwritten and now looks as followed 

ide.properties
ide.properties
ide.properties
ide.properties
ide.properties
ide.properties
ide.properties
ide.properties
ide.properties
ide.properties
ide.properties
ide.properties
MVN_VERSION=3.1.0

moritzLanger: Debug this error, the file gets overwriten [here in ToolCommandlet](https://github.com/moritzLanger/IDEasy/blob/test/improveTestCoverage/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java#L527C1-L527C1)

cli/src/test/java/com/devonfw/tools/ide/commandlet/VersionGetCommandletTest.java
moritzLanger: implemented

cli/src/test/java/com/devonfw/tools/ide/commandlet/HelpCommandletTest.java
moritzLanger: implemented

cli/src/test/java/com/devonfw/tools/ide/commandlet/VersionGetCommandletTest.java
moritzLanger: is it enough to assertThat(e).isInstanceOf(CliExeption.class) ?

cli/src/test/java/com/devonfw/tools/ide/commandlet/VersionSetCommandletTest.java
hohwille: The problem is not that the file gets overwritten adding the MVN_VERSION line because that is exactly what we want.
Your assertion is using `hasContent` and that is expecting the given String to be the entire content of the file. What you really want is to check that your given String is equal to one line in that file. I could not find an assertion in AssertJ doing this. However, you can easily do `Files.readAllLines` on the file and then assert with `contains` on that.

Howerver, if I am not mistaken this reveals that the content of the `ide.properties` file contains only lines with `ide.properties`. So great: You found a serious bug!
IMHO when writing the properties file back to disc we are IMHO using the `toString()` representation of each `VariableLine`. Seems that for the existing variables read from the file this somehow gives us the name of the properties file instead of the variable assignment (key=value).

hohwille: https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java#L131

moritzLanger: Using Files.readAllLines works, and the test now passes.

moritzLanger: However i still believe the fact, that the files is overwritten is a problem, as it originally contains more than just the MVN_VERSION and after unsing the VersionSetCommandlet vital Information is gone. E. G. the IDEcontextTest Failes if run after the VersionSetCommandletTest, because ide.properties no longer contains :

`BAR=bar-${SOME}`

moritzLanger: Found the Problem and i belive i fixed it with with the last Commit

cli/src/test/java/com/devonfw/tools/ide/commandlet/VersionGetCommandletTest.java
hohwille: It is enough to just assert with `hasMessageContaining` and simply remove `isInstanceOf`.
So:
```
CliException e = new CliException("");
if (e instanceof CliException) {
  ...
}
```
Is as meaningful as:
```
if (true) {
  ...
]
```

cli/src/test/java/com/devonfw/tools/ide/commandlet/VersionSetCommandletTest.java
hohwille: Great. I can confirm that you found and fixed two serious bugs. Thanks 👍

hohwille (approved):
@moritzLanger thanks for your PR. You did not only add JUnit tests and improve the coverage. You also found severe bugs this way and also fixed them. Perfect work 👍 
Ready for merge.

Issue #70, opened 2023-10-05
Complete build logic for native-image


hohwille:
With #4 we have implemented the complex workflows for building the native images.
However, this does not yet work properly so some last rework is needed.
The native-images do not arrive here:
https://s01.oss.sonatype.org/content/repositories/snapshots/com/devonfw/tools/IDEasy/ide-cli/2024.01.001-SNAPSHOT/
PR #62 did not fix the problem.
IMHO there are different profiles but those attaching the native image artifacts are not activated in the nightly build.
Please check my review suggestions where I already gave:
https://github.com/devonfw/IDEasy/pull/2#discussion_r1329812445

hohwille:
Still failing:
```
[ERROR] Failed to execute goal org.sonatype.plugins:nexus-staging-maven-plugin:1.6.13:deploy (injected-nexus-deploy) on project ide-cli: Failed to install artifact com.devonfw.tools.IDEasy:ide-cli:tar.gz:linux:2024.01.001-SNAPSHOT: /home/runner/work/IDEasy/IDEasy/cli/ide-cli-linux-x64 (No such file or directory) -> [Help 1]
```

hohwille:
I was able to download the windows package with the ide.exe. I could extract it normally with 7-Zip and also I can run the binary (e.g. print the help). Some things do not seem to work in the native image, but that is a different story so I will close this issue now.
p.s.: When I have investigated things that work in JVM mode but do not work in native-image, I will create individual bug issues for that.

PR #76, opened 2023-10-10
#70: Complete build logic of native-image

alfeilex:
This PR completes the build logic of native-image workflow. 

Related issue: #70

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -120,12 +120,12 @@
                     <classifier>linux</classifier>
                   </artifact>
                   <artifact>
-                    <file>target/${releaseName}</file>
+                    <file>target/${releaseName}.tar.gz</file>
                     <type>tar.gz</type>
                     <classifier>windows</classifier>
                   </artifact>
                   <artifact>
-                    <file>target/${releaseName}</file>
+                    <file>target/${releaseName}.tar.gz</file>
                     <type>tar.gz</type>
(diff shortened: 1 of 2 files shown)


cli/pom.xml
```diff
@@ -115,17 +115,17 @@
               <configuration>
                 <artifacts>
                   <artifact>
-                    <file>${releaseName}</file>
+                    <file>target/${project.artifactId}-linux-x64.tar.gz</file>
```
alfeilex, line 118: This should fix the current error of the workflow. I'm not sure if entering the filenames statically is the best way, but the last deployment job in the workflow is running on Ubuntu and the classifier is not working.

alfeilex:
> @alfeilex thanks for your PR. It seems that `${releaseName}` is the trick that fixes this. You also have removed the cancellation if nothing changed. I am fine with this as it simplifies things and also is good if the previous run failed, because then we did not get a new SNAPSHOT version. Regarding #70 this seems to be unrelated and changing the order of the OS matrix was also some kind of fix? If so I do not yet get it. Anyway ready for merge.

I have reverted the changes of the workflow because I only changed some things there for testing purposes. Should we remove the cancellation anyway?

cli/pom.xml
hohwille: I am happy with this approach for now. We can still improve, if we think this is problematic.

Issue #69, opened 2023-10-05
Add support for executing scripts


hohwille:
As a developer of IDEasy, I do not want to worry what I want to execute with `ProcessContext` so that like in bash, I can just run any executable command.
However, if the command is a bash script (e.g. `catalina.sh` for tomcat), then IMHO `ProcessBuilder` will not be able to run it directly.
If I am not mistaken the problem and solution is explained here:
https://www.baeldung.com/run-shell-command-in-java

We also need to implement a way how to determine if an executable is a (bash)script or not. My assumption would be:
* Filename is `*.cmd` or `*.bat` -> run using Windows CMD (only if OS is windows)
* Filename is `*.sh` or the first 2 bytes of the file content are `#!` -> run using Bash - on Windows we also have to find bash. If I just enter `bash` in a `CMD` shell, I get `C:\Windows\System32\bash.exe` that is not our Git-Bash but a very limited (and IMHO kind of buggy Bash implementation from Microsoft that is not working on my computer). So like we did in devonfw-ide, we might need to auto-detect and find where bit-bash is installed (most likely `"C:\Program Files\Git\git-bash.exe"` but we need a more generic way - see old bash scripts to see how it can be done). Further, we might not want to open a new Git-Bash window but run that inside our existing shell (that may be a git-bash, CMD, or PowerShell - all three standalone or inside Windows-Terminal).

hohwille:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/devon.bat#L38-L69

hohwille:
Done with PR #86

PR #86, opened 2023-10-13
#69: add support for executing scripts

salimbouch:
addresses: #69 
implements:

- Method findPathFor that returns the path of the bash for the given String (GitForWindows or Cygwin), otherwise null if it isnt't installed.
- Method hasSheBang that returns true if the file starts with '#!', otherwise false (in case file doesn't have a sheBang or isn't readable)
- Code in the run Method that checks whether the executable is a bash script, if the operating system is windows, it opens it with a bash. I explicitely use the bin\bash.exe and not the git-bash.exe since git-bash.exe always opens a new window and runs the script inside it, whereas the bin\bash.exe runs the script inside the existing window.


cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -201,4 +218,53 @@ private String createCommandMessage(String suffix) {
...
+  private boolean hasSheBang(String filePath) {
+    try(BufferedReader reader = new BufferedReader(new FileReader(filePath))) {
+      String firstLine = reader.readLine();
+      return firstLine != null && firstLine.startsWith("#!");
+    } catch (IOException e) {
+      return false;
+    }
+  }
+
+  private String findPathFor(String bash) {
```
hohwille, line 230: a better name could be `findPathInWindowsRegistry`

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -201,4 +218,53 @@ private String createCommandMessage(String suffix) {
...
+    String[] registryKeys = {"HKEY_LOCAL_MACHINE","HKEY_CURRENT_USER"};
+    String regQueryResult = null;
+    String toolValueName = ("GitForWindows".equals(bash)) ? "InstallPath" : "rootdir";
+
+    for (String registryKey : registryKeys) {
+      String command = "reg query " + registryKey + "\\Software\\" + bash + "  /v " + toolValueName + " 2>nul";
+
+      try {
+        Process process = new ProcessBuilder("cmd.exe", "/c", command).start();
+        BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
```
hohwille, line 240: For all streams (all instances of `AutoClosable`) please use [try-with-ressource pattern](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html).

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -201,4 +218,53 @@ private String createCommandMessage(String suffix) {
...
+        int exitCode = process.waitFor();
+        if (exitCode != 0) {
+          String msg = createCommandMessage(" failed with exit code " + exitCode + "!");
+          throw new CliException(msg, exitCode);
+        }
+        regQueryResult = output.toString();
+        break;
+      } catch (Exception e) {
+        throw new IllegalStateException(createCommandMessage(" failed: Could not start process!"), e);
+      }
```
hohwille, line 257: Maybe you could return `null` in case of an error, if we have other fallbacks (at least we should try the default `%ProgramFiles%\Git\git-bash.exe`) to find bash and only throw an exception if we could not find bash anywhere.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -115,6 +117,21 @@ public ProcessResult run(boolean capture) {
...
+    boolean isBashScript = "sh".equals(fileExtension) || hasSheBang(executableName);
+    if(isBashScript && this.context.getSystemInfo().isWindows()) {
+      String bashPath = findPathFor("GitForWindows");
+      if (bashPath == null) {
+        bashPath = findPathFor("Cygwin\\setup");
+      }
+      if (bashPath == null) {
+        throw new IllegalStateException("Could not find bash. Please install git for Windows and rerun.");
+      }
+      this.arguments.add(0, bashPath);
```
hohwille, line 132: You need to also add `-c` option (`bash.exe -c my-script.sh myarg1 myarg2`).
Try this in your git-bash:
```
$ bash.exe -c "echo hi"
hi
```

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -115,6 +117,21 @@ public ProcessResult run(boolean capture) {
     }
     // pragmatic solution to avoid copying lists/arrays
     this.arguments.add(0, this.executable.toString());
+
+    String executableName = this.executable.toString();
+    String fileExtension = FilenameUtil.getExtension(this.executable.toString());
```
hohwille, line 122: use `executableName` here

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
hohwille: As the entire "command" has to be one after argument to `bash -c`, I guess that you have to do something like this:
```
this.arguments.add(0, this.executable.toString());
List<String> args = this.arguments;
if (isBashScript) {
  // new logic goes here
  args = new ArrayList();
  String bash = "bash";
    // ... your code to find bash on windows goes here ...
    bash = findBashOnWindows();
  }
  args.add(bash);
  args.add("-c");
  args.add(this.arguments.join(" "));
}
this.processBuilder.command(args);
```

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
salimbouch: done

salimbouch:
The method findBashOnWindows now first checks whether git bash is installed on the default location (`Program Files\Git\bin\bash.exe`), if the bash exists, it returns it, otherwise it looks for it using the windows registry, first it looks for git for windows then for Cygwin, if none of them are found, it throws an exception. I honestly believe even if it doesn't find a bash anywhere it should still just return null and  print a warning that git bash should be installed since I check the returned value of this method and if it is null, it uses the plain bash that seems to work just fine too.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
salimbouch: Apparently in ProcessBuilder `-c` is only used when running a bash command, and not a bash script. 
`processBuilder.command("bash", "-c", "ls")` works, but `processBuilder.command("bash", "-c", pathToScript)` returns command not found.
With that said, I went back to my old variation of using `processBuilder.command(pathToBash, pathToScript, args)` which works perfectly.

hohwille (approved):
@salimbouch thanks for your rework. You addressed all review comments nicely 👍 
Ready for merge.

p.s.: We could still add some nice-to-have improvements in the future:
* detection of bash could be done in IdeContext and the result could be stored so that we do not have to repeat registry lookup if multiple scripts are called in the same CLI call.
* the behavior of `findBashOnWindows` is now a little inconsistent: If nothing is found in the registry at all we throw an exception while if the registry lookup fails, we return `null`. My thought was to return `null` in the detection and have some place calling this detection and handling the case if nothing was found that then throws the exception.

Issue #89, opened 2023-10-13
copy of MSI file is not deleted after installation


MattesMrzik:
### Expected behavior

The copy of the .msi file created by msiexec should get deleted after installation of that .msi file.

### Actual behavior

The copy does not get deleted.

### Steps to reproduce

1.  install a toolCommandlet that uses a .msi file.

### Comments/Hints:

A wrong path is passed to the  fileAccess.delete method in ToolCommandlet.java

### Affected version:

* OS: Windows

hohwille:
Fixed with PR #90

PR #90, opened 2023-10-13
#89: fix deletion of the copied msi file

MattesMrzik:
When we extract an MSI file with `msiexec` for whatever reason that creates a copy of the MSI file in the target folder.
We never found a CLI option to prevent `msiexec` from doing this. Therefore we just have to delete that copied MSI file afterwards.
However, the code to delete it was bug and that is now fixed with this PR.


hohwille (approved):
@MattesMrzik thanks for tracing down and fixing this bug 👍 
LOL. What an obvious and stupid bug from me - I should stop coding late at night :)

Issue #79, opened 2023-10-10
Simplify multi-keyword commandlets and make usage consistent


hohwille:
From #5: 
> Currently `ide set mvn version 3.9.3` sets the maven version to 3.9.3.
Each commandlet must have a unique name that should also give an idea what the commandlet is about.
Currently I named this commandlet `set-version` what seems to make sense.
Also this would allow later to also have a `ide set intellij edition ultimate` that would be handled by a `set-edition` commandlet.
As a result you would need to type `ide help set-version` to get the help of `set-version` commandlet and `ide help set-edition` to get the help of `set-edition` commandlet. Would it then be better to also change the command syntax to `ide set-version mvn 3.9.3` and `ide set-edition intellij ultimate`? IMHO this would avoid some user-confusion.

As I did not get any feedback, I now decide that we should go for this approach as it will make our life easier and the help and usage more consistent. A little drawback is that `ide set mvn version 3.9.3` is a little bit more natural language than `ide set-version mvn 3.9.3`. However, if we later think this is relevant, we can still add some CLI alias syntax concept that allows to have `ide set mvn version 3.9.3` as a kind of alias to `ide set-version mvn 3.9.3` what would still be the primary syntax.
FYI: Feel free to also change `VersionSetCommandlet` to `SetVersionCommandlet`, etc. if that makes more sense for you then.

PR #92, opened 2023-10-16
#79: Simplified multi-keyword commandlets

aBega2000:
Addresses: #79 

- Simplified multi-keyword commandlets and made usage consistent.

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/commandlet/VersionSetCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/commandlet/VersionSetCommandlet.java
@@ -27,7 +27,7 @@ public class VersionSetCommandlet extends Commandlet {
   public VersionSetCommandlet(IdeContext context) {
 
     super(context);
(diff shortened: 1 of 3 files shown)


cli/src/main/java/com/devonfw/tools/ide/commandlet/VersionSetCommandlet.java
```diff
@@ -27,9 +27,8 @@ public class VersionSetCommandlet extends Commandlet {
   public VersionSetCommandlet(IdeContext context) {
 
     super(context);
-    addKeyword("set");
+    addKeyword("set-version");
```
hohwille, line 30: correct. This is what was planned. For further simplification just use `getName()` instead of repeating as redundant String literal. Still in draft mode so I guess you are aware that there are other commandlets to change in the same way...

hohwille (approved):
@aBega2000 thanks for your PR. Looks good 👍 
Ready for merge.

Issue #91, opened 2023-10-16
Generalize unzip and untar behaviour and move IDEasy tool unpacking specific behaviour to ToolCommandlet


MattesMrzik:
The unzip and untar methods of FileAccessImpl.java currently extract to a temporary directory, collapse single top folders and move the result to the target directory.
The behavior of these methods should be generalized, such that these simply extract the file to the passed target directory. The collapsing of the single top folders and the moving from the temporary directory to the actual target directory should be implemented in the extract method of ToolCommandlet.java.

PR #95, opened 2023-10-16
 #91: Collapsing of single top folders moved to ToolCommandlet

MattesMrzik:
Removed collapsing of single top level folders when unpacking files using a archive input stream to make methods unzip and untar more general. The collapsing of single top level folders during tool installation is now implemented in ToolCommandlet::extract.

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
@@ -277,7 +277,7 @@ public abstract class ToolCommandlet extends Commandlet implements Tags {
     if (installedVersion == null) {
       this.context.success("Successfully installed {} in version {}", this.tool, resolvedVersion);
     } else {
-      this.context.success("Successfully installed {} in version {} replacing previous version {]", this.tool,
+      this.context.success("Successfully installed {} in version {} replacing previous version {}", this.tool,
           resolvedVersion, installedVersion);
     }
     postInstall();
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
@@ -1,5 +1,6 @@
 package com.devonfw.tools.ide.tool;
 
+import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -287,7 +288,7 @@ public abstract class ToolCommandlet extends Commandlet implements Tags {
    * custom post intallation logic.
    */
   protected void postInstall() {
-    
+
     // nothing to do by default
   }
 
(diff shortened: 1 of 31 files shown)


cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -481,6 +482,31 @@ protected void extract(Path file, Path targetDir) {
...
         throw new IllegalStateException("Unknown archive format " + extension + ". Can not extract " + file);
       }
+
+      // collapse single parent directories
+      File currentDir = new File(targetDir.toString());
+      String[] currentSubFiles = currentDir.list();
+      if (currentSubFiles == null) {
+        throw new IllegalStateException("Extracted file does not contain any file or directories");
+      }
+      // TODO add wbin as well? (in case of msi azure installation)
```
hohwille, line 492: Not needed here `az` has a `CLI2` folder containing various files and folders including `wbin`.
These exclusions listed here are to avoid that we omit a folder that is required.
E.g. if your archive only contains
* bin
   * tool.exe
   * tool.dll

Then we still want to keep the `bin` folder in the `targetDir`.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -481,6 +482,31 @@ protected void extract(Path file, Path targetDir) {
...
+      while (currentSubFiles != null && currentSubFiles.length == 1 && !currentDir.toString().equals("bin")) {
+        currentDir = new File(String.valueOf(currentDir.toPath().resolve(currentSubFiles[0])));
```
hohwille, line 494: Stop using `File` and instead use `Path`. It needs some time to get used to it but `File` should be considered deprecated just like other JDK crap such as `Date`, `Calendar`, etc.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -429,22 +430,44 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+   *         path (including {@code s}) are the sole item in their respective directory and {@code s} is not named
+   *         "bin".
+   */
+  private Path getProperInstallationSubDirOf(Path path) {
+
+    try (Stream<Path> stream = Files.list(path)) {
+      Path[] subFiles = stream.toArray(Path[]::new);
+      if (subFiles.length == 0) {
+        throw new IllegalStateException("The directory " + path + " does not contain any files.");
+      } else if (subFiles.length == 1 && !subFiles[0].toString().equals("bin") && Files.isDirectory(subFiles[0])) {
```
hohwille, line 445: `java.nio.file.Path` is better than `java.io.File` but also has some pitfalls:
`Path.toString()` will return the entire path of the file (`C:\Projects\...`) and not just its "filename" (`bin`).
Please use `!subFiles[0].getFileName().toString().equals(IdeContext.FOLDER_BIN)` instead.

How about the other special exclusion rules from the old bash code?
https://github.com/devonfw/ide/blob/af67497e06fb82e2e50a4d9ffa10e3f9dda44aa1/scripts/src/main/resources/scripts/functions#L949

For that it is better to rewrite your code using a local variable
```
} else if (subFiles.length == 1) {
  String filename = subFiles[0].getFileName().toString();
  if (!filename.equals(IdeContext.FOLDER_BIN) && !filename.equals(IdeContext.FOLDER_CONTENTS) && !filename.endsWith(".app") && Files.isDirectory(subFiles[0])) {
    return getProperInstallationSubDirOf(subFiles[0]);
  }
}
```

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -457,30 +480,32 @@ protected void extract(Path file, Path targetDir) {
...
-        fileAccess.delete(tmpDir);
+        pc.executable("xar").addArgs("-C", tmpDirPkg, "-xf", file).run();
+        Path contentPath = fileAccess.findFirst(tmpDirPkg, p -> p.getFileName().toString().equals("Payload"), true);
+        fileAccess.untar(contentPath, tmpDir, TarCompression.GZ);
+        fileAccess.delete(tmpDirPkg);
       } else {
         throw new IllegalStateException("Unknown archive format " + extension + ". Can not extract " + file);
       }
+      this.context.getFileAccess().move(getProperInstallationSubDirOf(tmpDir), targetDir);
+      this.context.getFileAccess().delete(tmpDir);
```
hohwille, line 508: simply use `fileAccess` instead of `this.context.getFileAccess()` :)

hohwille (approved):
@MattesMrzik thanks for your PR. Great work 👍 
I left some last comments to fix and improve.
Can you address them so we can merge?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -429,22 +430,44 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+   * @return the deepest subdir {@code s} of the passed path such that all directories between {@code s} and the passed
+   *         path (including {@code s}) are the sole item in their respective directory and {@code s} is not named
+   *         "bin".
+   */
+  private Path getProperInstallationSubDirOf(Path path) {
+
+    try (Stream<Path> stream = Files.list(path)) {
+      Path[] subFiles = stream.toArray(Path[]::new);
+      if (subFiles.length == 0) {
+        throw new IllegalStateException("The directory " + path + " does not contain any files.");
```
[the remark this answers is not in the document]
MattesMrzik: Thanks for your remarks. I have implemented them now. But I got no errors during the installation of `oc`. I did
`git fetch upstream` and `git merge upstream/main` and then run `Ide oc`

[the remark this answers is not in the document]
hohwille: BTW: You can test this by going to your download cache directory (e.g. `C:\Users\«yourlogin»\Downloads\ide\default`) and replacing an archive with an empty zip or tar.gz file. Then delete the software version from your shared software repo (IDE_ROOT/_ide/software/...) and reinstall that exact tool version.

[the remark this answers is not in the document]
hohwille: Thanks. All resolved now.

Issue #14, opened 2023-09-22
Implement ToolCommandlet for Azure CLI


hohwille:
Create `ToolCommandlet` for Azure CLI
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/az

Issue #78, opened 2023-10-10
Improve writing changes to ide.properties


hohwille:
In case we are changing the value of a property and saving the changes back to disc, I want to have that the order of the lines in the `ide.properties` file remains stable so that only the new value is changed in the line where the property was declared before.

Assuming we have this `ide.properties` file with some legacy history:
```
#here we define the maven version for both the old and the new release to avoid incompatiblities...
MVN_VERSION=3.9.4
MAVEN_VERSION=${MVN_VERSION}
#here we define to use commercial ultimate edition for Intellij as all team members have a license...
INTELLIJ_EDITION=ultimate
INTELLIJ_EDITION_TYPE=u
```

Now if we run `ide version-set mvn 3.9.5` (see #79) we should get:

```
#here we define the maven version for both the old and the new release to avoid incompatiblities...
MVN_VERSION=3.9.5
MAVEN_VERSION=${MVN_VERSION}
#here we define to use commercial ultimate edition for Intellij as all team members have a license...
INTELLIJ_EDITION=ultimate
INTELLIJ_EDITION_TYPE=u
```

So the following would be undesired:
```
#here we define the maven version for both the old and the new release to avoid incompatiblities...
MAVEN_VERSION=${MVN_VERSION}
#here we define to use commercial ultimate edition for Intellij as all team members have a license...
INTELLIJ_EDITION=ultimate
INTELLIJ_EDITION_TYPE=u
MVN_VERSION=3.9.5
```

Even more wrong would be:
```
#here we define the maven version for both the old and the new release to avoid incompatiblities...
MVN_VERSION=3.9.4
MAVEN_VERSION=${MVN_VERSION}
#here we define to use commercial ultimate edition for Intellij as all team members have a license...
INTELLIJ_EDITION=ultimate
INTELLIJ_EDITION_TYPE=u
MVN_VERSION=3.9.5
```

MattesMrzik:
I have looked at the code and also tried some value changes of properties and saved the changes back to disc. I set some arbitrary helm versions.
`ide set-version helm 1`
then after some additional manual changes the file settings/ide.properties looked like: 
```shell
...
HELM_VERSION=1
a=1
b=2
c=3
``` 
Then I ran `ide set-version helm 1.1` and the file then looked just as desired:
```shell
...
HELM_VERSION=1.1
a=1
b=2
c=3
```

MattesMrzik:
[...]

PR #99, opened 2023-10-17
#14: Implement ToolCommandlet for Azure CLI

MattesMrzik:
#14: Implemented ToolCommandlet for Azure CLI.


cli/src/main/java/com/devonfw/tools/ide/tool/az/Azure.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  @Override
+  public void postInstall() {
+
+    super.postInstall();
+
+    EnvironmentVariables variables = this.context.getVariables();
+    EnvironmentVariables settingsVariables = variables.getByType(EnvironmentVariablesType.WORKSPACE);
+    // TODO oder EnvironmentVariablesType.CONF
```
hohwille, line 34: yep, needs to go to `conf/ide.settings`.

cli/src/main/java/com/devonfw/tools/ide/tool/az/Azure.java
```diff
@@ -0,0 +1,41 @@
...
+    // TODO I think there is nothing written to a file, just the env var exported and thereby only active in current
+    // session. Should this env var be written to a file?
```
hohwille, line 36: Simply do the change and then call `save` method on the according variables.
FYI: Only `EnvironmentVariablesPropertiesFile` implements `save` properly. So currently doing `variables.save()` would have no effect. Maybe we should throw an `UnsupportedOperationException` in this case so the developer does not get mistaken...

hohwille: See also #78

MattesMrzik:
Why hasn't this:
`[hohwille](https://github.com/hohwille) linked an issue [yesterday](https://github.com/devonfw/IDEasy/pull/99#event-10681228055) that may be closed by this pull request`
happened automatically? It seems that I can't manually link these as I have no gear-symbol at the Development section on the right.

hohwille:
> Why hasn't this: ... happened automatically? 

Read out Teams channel, where I linked the solution.

hohwille (approved):
@MattesMrzik Thanks for your PR. Solved nicely and perfect 👍

Issue #82, opened 2023-10-11
Softlink creation requires admin privileges during software installation 


MattesMrzik:
### Actual behavior

Running _com.devonfw.tools.ide.cli.Ide_ as non admin and passing an argument to install new software (such as gradle, helm, ...)  installs software in _ide_ directory and then creates a softlink. See [code](https://github.com/devonfw/IDEasy/blob/75425f7e6f38b5bd56d38bb2c5411d834ad3d11a/cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java#L291 ). 
This link creation throws an exception: 
`Failed to create a symbolic link C:\projects\devonfw\software\gradle pointing to C:\projects\_ide\software\default\gradle\gradle\8.4`

### Expected behavior

IDEasy should work for users without admin privileges as well. 
In the old bash code, we had this workaround:
https://github.com/devonfw/ide/blob/5a3f41627adcb13254f80f841f0fbdeca5e7bdf0/scripts/src/main/resources/scripts/functions#L1418-L1428
Either we should always do this on windows in our `symlink` method or we may be able to somehow determine if the user has admin rights or only try the fallback in a catch block if the regular `createSymbolicLink` failed and we are on Windows.

### Complete error msg

`An unexpected error occurred!
We are sorry for the inconvenience.
Please check the error below, resolve it and try again.
If the error is not on your end (network connectivity, lack of permissions, etc.) please file a bug:
https://github.com/devonfw/ide/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=IllegalStateException%3A+Failed+to+create+a+symbolic+link+C%3A%5Cprojects%5Cdevonfw%5Csoftware%5Cgradle+pointing+to+C%3A%5Cprojects%5C_ide%5Csoftware%5Cdefault%5Cgradle%5Cgradle%5C8.4
java.lang.IllegalStateException: Failed to create a symbolic link C:\projects\devonfw\software\gradle pointing to C:\projects\_ide\software\default\gradle\gradle\8.4
    at com.devonfw.tools.ide.io.FileAccessImpl.symlink(FileAccessImpl.java:293)
    at com.devonfw.tools.ide.tool.ToolCommandlet.doInstall(ToolCommandlet.java:267)
    at com.devonfw.tools.ide.tool.ToolCommandlet.install(ToolCommandlet.java:239)
    at com.devonfw.tools.ide.tool.gradle.Gradle.install(Gradle.java:28)
[...]

PR #105, opened 2023-10-19
#82: Softlink creation requires admin privileges during software installation

moritzLanger:
#82 Softlink creation requires admin privileges during software installation


cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -288,7 +290,12 @@ public void symlink(Path source, Path targetLink) {
         this.context.debug("Deleting symbolic link to be re-created at {}", targetLink);
         Files.delete(targetLink);
       }
-      Files.createSymbolicLink(targetLink, source);
+      if (this.context.getSystemInfo().getOs().equals(WINDOWS)) {
```
hohwille, line 293: a simplification is `this.context.getSystemInfo().isWindows()` :)

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -288,7 +290,12 @@ public void symlink(Path source, Path targetLink) {
...
+        Runtime rt = Runtime.getRuntime();
+        rt.exec("cmd //c 'mklink /D /J $(cygpath -w " + source + ") $(cygpath -w" + targetLink + ")");
```
hohwille, line 295: IMHO it is better to do that in a dedicated catch block:
```
} catch (FileSystemException e) {
  info("Failed to create native symlink due to lack of permissions. See https://github.com/devonfw/IDEasy/blob/main/documentation/symlinks.asciidoc for further details. Error was {}", e.getMessage());
  // workaround goes here...
}
```

For the workaround we are using `ProcessBuilder` that we typically wrap via `ProcessContext`.
Currently `FileAccessImpl` has no direct access to `IdeContext` and therefore `ProcessContext` (as I initially wanted to avoid cyclic dependencies as architectural best-practice), but we could easily change this.
Howerver, it is fully fine to use `ProcessBuilder` directly here. But just like `File` also `Runtime` should be considered deprecated as it has some bugs addressed by the new construct (`ProcessBuilder`).

hohwille: Well after rethinking this: Will the output of the `mklink` command go to nirvana (`/dev/null`) and potential errors (exit code other than 0) will be ignored then?
All this is already handled by `ProcessContext` so it may be the best option to use that.

moritzLanger:
TODO : 
 https://github.com/devonfw/IDEasy/blob/main/documentation/symlinks.asciidoc  has yet to be created.

hohwille (approved):
@moritzLanger thanks for your update. Now perfect and ready for merge 👍 
Should I wait until you add the documentation or can you simply set this as ready for review and provide the documentation in a separate PR?

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
hohwille: As now everything is solved, I assume this question is already cleared. However, the executable is just a `Path` but can be any command so also just `cmd` or `mklink`.

moritzLanger:
> @moritzLanger thanks for your update. Now perfect and ready for merge 👍 Should I wait until you add the documentation or can you simply set this as ready for review and provide the documentation in a separate PR?

I set the PR as ready for review. I will add the documentation in a separate PR in the coming days.

hohwille (approved):
@moritzLanger Thanks for the helpful documentation and screenshots.
Now ready for merge.

Issue #106, opened 2023-10-19
Setting properties does not update export flag when passing identical value


MattesMrzik:
### Expected behavior

EnvironmentVariablesPropertiesFile::set should update "export" flag even if the value of the property is not updated.

### Actual behavior
EnvironmentVariablesPropertiesFile::set only updates the "export" flag if the new value differs from the old value.

### Steps to reproduce

```
# ide.properties
var1=1
export var2=2
``` 

```java
Path propertiesFilePath = Path.of("ide.properties");
AbstractEnvironmentVariables parent = null;
EnvironmentVariablesType type = EnvironmentVariablesType.SETTINGS;
EnvironmentVariablesPropertiesFile variables = new EnvironmentVariablesPropertiesFile(parent, type,
        propertiesFilePath, LOGGER);

variables.set("var1", "1", true);
variables.set("var2", "2", false);
variables.save();
``` 

### Related/Dependent Issues
#78 

### Affected version:

* OS: Windows

PR #107, opened 2023-10-19
#78: the properties file stays stable. Wrote test

MattesMrzik:
The test confirms that the properties file remains stable even if values are updated. The test further revealed, that the set method of EnvironmentVariablesPropertiesFile does not update the "export" flag if the old and new value are identical (see issue #106 )

### DIFF
--- a/cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
+++ b/cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
@@ -1,6 +1,5 @@
 package com.devonfw.tools.ide.environment;
 
-import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
@@ -43,7 +42,7 @@ class EnvironmentVariablesPropertiesFileTest extends Assertions {
   }
 
   @Test
-  void testSave() {
+  void testSave() throws Exception {
 
     // arrange
     List<String> linesToWrite = new ArrayList<>();
@@ -64,20 +63,10 @@ class EnvironmentVariablesPropertiesFileTest extends Assertions {
     linesToWrite.add("var9=9");
 
     Path propertiesFilePath = Path.of("target/tmp-EnvironmentVariablesPropertiesFileTest-ide.properties");
-    try {
-      Files.write(propertiesFilePath, linesToWrite, StandardOpenOption.CREATE_NEW);
-    } catch (IOException e) {
-      System.err.println("File already exists or an error occurred: " + e.getMessage());
-    }
+    Files.write(propertiesFilePath, linesToWrite, StandardOpenOption.CREATE_NEW);
     // check if this writing was correct
-    try {
-      List<String> lines = Files.readAllLines(propertiesFilePath);
-      for (int i = 0; i < lines.size(); i++) {
-        assertThat(lines.get(i)).isEqualTo(linesToWrite.get(i));
-      }
-    } catch (IOException e) {
-      System.err.println("An error occurred: " + e.getMessage());
-    }
+    List<String> lines = Files.readAllLines(propertiesFilePath);
+    assertThat(lines).containsExactlyElementsOf(linesToWrite);
 
     AbstractEnvironmentVariables parent = null;
     EnvironmentVariablesType type = EnvironmentVariablesType.SETTINGS;
@@ -117,20 +106,10 @@ class EnvironmentVariablesPropertiesFileTest extends Assertions {
     linesAfterSave.add("var9=9");
     linesAfterSave.add("var10=10");
     linesAfterSave.add("export var11=11");
-    try {
-      List<String> lines = Files.readAllLines(propertiesFilePath);
-      for (int i = 0; i < lines.size(); i++) {
-        assertThat(lines.get(i)).isEqualTo(linesAfterSave.get(i));
-      }
-    } catch (IOException e) {
-      System.err.println("An error occurred: " + e.getMessage());
-    }
 
+    lines = Files.readAllLines(propertiesFilePath);
+    assertThat(lines).containsExactlyElementsOf(linesAfterSave);
     // clean up
-    try {
-      Files.delete(propertiesFilePath);
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
+    Files.delete(propertiesFilePath);
   }
 }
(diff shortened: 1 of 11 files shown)


MattesMrzik:
The PR tests should succeed if PR #108 is merged.

cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
```diff
@@ -37,4 +42,96 @@ public void testLoad() {
...
+    linesToWrite.add("export var5=5");
+    linesToWrite.add("export var6=6");
+    linesToWrite.add("# third comment");
+    linesToWrite.add("var7=7");
+    linesToWrite.add("var8=8");
+    linesToWrite.add("# 4th comment");
+    linesToWrite.add("# 5th comment");
+    linesToWrite.add("var9=9");
+
+    Path propertiesFilePath = Path.of("src/test/resources/com/devonfw/tools/ide/env/var/ide.properties");
```
hohwille, line 66: never write to `src` in tests. This is causing changes that will then appear in git and may be added and committed accidentally. Mutable testdata shall be in target or temp-folder.

MattesMrzik: I tried using _context.getFileAccess().createTempDir()_ but since this does not seem to work when git-actions runs the tests I will be using the target dir.

cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
```diff
@@ -37,4 +42,95 @@ public void testLoad() {
...
+    linesToWrite.add("var8=8");
+    linesToWrite.add("# 4th comment");
+    linesToWrite.add("# 5th comment");
+    linesToWrite.add("var9=9");
+
+    Path propertiesFilePath = Path.of("target/tmp-EnvironmentVariablesPropertiesFileTest-ide.properties");
+    try {
+      Files.write(propertiesFilePath, linesToWrite, StandardOpenOption.CREATE_NEW);
+    } catch (IOException e) {
+      System.err.println("File already exists or an error occurred: " + e.getMessage());
```
hohwille, line 70: Even though this should never happen, it is not very reasonable exception handling.
For ultimate simplicity in JUnits, you can just add `throws Exception` to the test method or you could simply do `throw new IllegalStateException("Failed to write file" + propertiesFilePath, e);` in the `catch` block.

cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
```diff
@@ -37,4 +42,95 @@ public void testLoad() {
...
+    }
+    // check if this writing was correct
+    try {
+      List<String> lines = Files.readAllLines(propertiesFilePath);
+      for (int i = 0; i < lines.size(); i++) {
+        assertThat(lines.get(i)).isEqualTo(linesToWrite.get(i));
+      }
+    } catch (IOException e) {
+      System.err.println("An error occurred: " + e.getMessage());
```
hohwille, line 79: same here.

cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
```diff
@@ -37,4 +42,95 @@ public void testLoad() {
...
+    try {
+      Files.write(propertiesFilePath, linesToWrite, StandardOpenOption.CREATE_NEW);
+    } catch (IOException e) {
+      System.err.println("File already exists or an error occurred: " + e.getMessage());
+    }
+    // check if this writing was correct
+    try {
+      List<String> lines = Files.readAllLines(propertiesFilePath);
+      for (int i = 0; i < lines.size(); i++) {
+        assertThat(lines.get(i)).isEqualTo(linesToWrite.get(i));
```
hohwille, line 76: actually you are not asserting our code here but the fundamentals of Java.
While this is OK, it feels a little schizophrenic to write tests that do not trust that Java is working...

cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
```diff
@@ -37,4 +42,95 @@ public void testLoad() {
...
+    linesAfterSave.add("# 4th comment");
+    linesAfterSave.add("# 5th comment");
+    linesAfterSave.add("var9=9");
+    linesAfterSave.add("var10=10");
+    linesAfterSave.add("export var11=11");
+    try {
+      List<String> lines = Files.readAllLines(propertiesFilePath);
+      for (int i = 0; i < lines.size(); i++) {
+        assertThat(lines.get(i)).isEqualTo(linesAfterSave.get(i));
+      }
```
hohwille, line 124: Why not just doing `assertThat(lines).containsExactlyElementsOf(linesAfterSave)`?
We are using `AssertJ` because it is expressive and makes our life easier.
You can also do `assertThat(propertiesFilePath).hasContent(EXPECTED_PROPERTIES_FILE_CONTENT)` and define the file-content as multi-line String. However, I am not aware how this will deal with newline characters on different operating-systems.

cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
```diff
@@ -37,4 +42,95 @@ public void testLoad() {
...
+    } catch (IOException e) {
+      System.err.println("An error occurred: " + e.getMessage());
+    }
+
+    // clean up
+    try {
+      Files.delete(propertiesFilePath);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
hohwille, line 133: maybe the easiest solution is to add `throws Exception` to the test method, so you can remove all `try { ... } catch (...) {...}` blocks here and make your life simple. If an exception will occur the test will fail.

hohwille (approved):
@MattesMrzik thanks for adding this test. You created a nice test-scenario that should cover most  relevant cases. 👍 
You could even add a case that removes a variable by settings its value to `null` to cover that as well.
I left a few general comments on writing great tests. Please have a look. After resolving these comments, we can merge.

cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
MattesMrzik: yep, just wanted to make sure the file is exactly like I want it to be:)

hohwille (approved):
@MattesMrzik thanks for the rework. Now everything is perfectly nailed - great work 👍

PR #108, opened 2023-10-19
#106: fixed properties update export-flag bug

MattesMrzik:
Fixed the bug. Now the flag is updated even if the value remains unchanged.


hohwille (approved):
@MattesMrzik thanks for the fix. Looks good 👍

Issue #85, opened 2023-10-13
getToolBinary() may return incorrect result


hohwille:
### Expected behavior

On Windows we have a specific requirement when executing a binary `«tool»` to execute the following files if they exist in the given order:
1. `«tool».exe`
2. `«tool».cmd`
3. `«tool».bat`
4. `«tool».msi`
5. `«tool».ps1` ??? Maybe first test in bash and cmd if that is actually executed properly if I just run `«tool»` before implementing
6. `«tool»`

Actually this should be a global behavior in IDEasy to behave like any shell on Windows whenever we execute something so this logic shall be build into `ProcessContext`

### Actual behavior

We currently have implemented this in `ToolCommandlet` and the implementation is incorrect as it executes the first match so the order is more or less random:
https://github.com/devonfw/IDEasy/blob/4896faa477c3fe7fffde058abbaed3a6a332a470/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java#L115-L125

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. create a bin folder with all of the 6 given binaries where `«tool»` is just `tool`
2. use ProcessContext to execute `tool` when that bin folder is on your PATH. You may even implement a dummy `ToolCommandlet` for testing with toolname `tool` and have your `bin` folder from 1. into `software/tool/bin` to test `getToolBinary()`.

### Comments/Hints:

Actually Java's `Process` already iterates the `PATH` to find and execute a binary, in case only the name of the tool is given without path. In other words, if you execute `mvn` it goes through your `PATH` to actually determine that you want to run `C:\Projects\devonfw\software\mvn\bin\mvn`. Initially I did not want to reinvent the wheel and re-implement this in our own Java-Code. However, in order to behave in the same way as `bash` used to work, we are more or less forced to implement this as we actually need to call `C:\Projects\devonfw\software\mvn\bin\mvn.cmd` in that case with higher priority. And if for some reason we need to execute something that is not the tool binary itself, it will not work in such case on Windows.
The good news is that I meanwhile have implemented `SystemPath` that allows you to easily iterate the path and can be retrieved from `IdeContext`.
Maybe the needed functionality for this story could even be implemented in this class. So `ProcessContext` can simply use `context.getSystemPath().findBinary(command)` and all will be implemented in that central place where is seems to suite well.
Also this would be "design for testability" as it can easily be tested (without actually executing commands and `tool.bat` or `tool.exe` files 

### Affected version:

[...]

PR #111, opened 2023-10-23
#85: Fixed the executing order of the tools

aBega2000:
Addresses: #85 

I have found a solution to the issue #85. The order of the executing files is now in the same order as requested by the issue. But in order to do so, I had to change the interface of the executable() method, I had to add the String tool so that the tool name is passed to this method as an argument, in order to work with the name and the extension of the tool. Since I changed the interface, I had to make small changes to the Eclipse.java class as well, and as far as I tested it works correctly. 

If this solution is not clean, I have another option in mind, but that would require that we keep the getToolBinary() method and make changes to it. I am keeping this PR as a draft until it is decided that it is a correct solution.

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
@@ -4,6 +4,7 @@ import java.io.BufferedReader;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.IOException;
+import java.io.File;
 import java.lang.ProcessBuilder.Redirect;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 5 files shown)


cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -72,34 +73,31 @@ public ProcessContext directory(Path directory) {
...
-          if (Files.exists(cmd)) {
-            exe = cmd;
-          } else {
-            cmd = parent.resolve(filename + ".bat");
-            if (Files.exists(cmd)) {
-              exe = cmd;
-            }
-          }
+
+    String[] possibleExtensions = { ".exe", ".cmd", ".bat", ".msi", ".ps1", "" };
```
hohwille, line 82: use a constant for things like this:
```
private List<String> EXTENSION_PRIORITY = List.of(".exe", ".cmd", ".bat", ".msi", ".ps1", "");
```

aBega2000:
I have found another solution to the issue where I keep the interface in its original form.

hohwille (approved):
@aBega2000 Thanks for the update 👍 Ready for merge.
It would be great, if you can add some more trace and debug logging and also write a JUnit test for `SystemPath.findBinary` in a new PR.

Issue #56, opened 2023-09-28
Implement auto-completion


hohwille:
As a IDEasy user, I want to have command-line auto-completion so that I can be more productive and do not always have to read the help pages.

We had implemented this in devonfw-ide with manually bash scripts and the bash completion infrastructure.
My vision is that we pass the arguments to auto-complete directly into our Java main program.
So more or less `Ide complete «args»` would be called and we solve this with a `CompleteCommandlet`.
By finding the proper commandlet to invoke we can find its arguments and depending on their type automatically determine the possible options. That is why I prepared according specific `Property` sub-classes so that for a `VersionProperty` we know that a version-number is to be passed and from out tool-commandlet to be invoked, we can figure out the possible and matching versions.

BTW: This story is not easy and therefore not good for beginners. Also I learned that debugging auto-completion while using it is tricky as you can not use debug output to the console while pressing [Tab] key. It is therefore good to have an understanding of how auto-completion in bash works and then first have JUnits covering the different cases (completion of a single word where a prefix of that word is already given vs. a new word is started, etc.). If still some debugging is needed it is better to temporary write a log-file with the input and output of the auto-completion to later understand what went wrong and why... Good luck with this story :)

jan-vcapgemini:
@MattesMrzik found out that there seems to be an issue with the WIndows Terminal on Windows 11.
The jline3 autocompletion does not work at all on Windows 11. Currently if you hit tab, the terminal just adds a tabstop character to the prompt.

PR #112, opened 2023-10-23
#56 implement autocompletion

jan-vcapgemini:
Addresses/Fixes: #56 

Implements:
* Jline3 autocompletion
* some examples
* test framework
* some tests
* jansi for Windows Terminal support
* checks for invalid input patterns e.g. `get-version mvn -`


cli/src/main/java/com/devonfw/tools/ide/cli/CommandletRegistry.java
```diff
@@ -0,0 +1,184 @@
...
+
+  @Override
+  public List<String> commandInfo(String command) {
+
+    List<String> out = new ArrayList<>();
+    // TODO: take command and get help from Ide
+    Commandlet helpCommandlet = context.getCommandletManager().getCommandlet("help");
+    Property<?> property = new CommandletProperty(command, false, "");
+    // helpCommandlet.add(property);
+    helpCommandlet.run();
```
hohwille, line 88: If you really want to run the help for the commendlet then do this:
```
HelpCommandlet helpCommandlet = context.getCommandletManager().getCommandlet(HelpCommandlet.class);
helpCommandlet.commandlet.setValueAsString(command);
helpCommandlet.run();
```
However, this will print the help to the logger. While you could mock it and retrieve its output here, it would IMHO be better to refactor `HelpCommandlet` and `Ide` to allow better reuse here...

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @Override
+  public void complete(LineReader reader, ParsedLine commandLine, List<Candidate> candidates) {
+
+    assert commandLine != null;
+    assert candidates != null;
+    String word = commandLine.word();
+    List<String> words = commandLine.words();
+    // TODO: implement rest of this
+    Commandlet sub = findSubcommandlet(words, commandLine.wordIndex());
```
hohwille, line 42: I am aware that this is all WIP but just to give some guidance here:
We would need to work on `Property` rather than on `Commandlet` here to do 2nd, 3rd, ... level completions.
1st level is easy: list all commandlet names/keywords.
Next levels get more tricky:
e.g. 1st = "install"
for 2nd: find this property in `InstallCommandlet` as 2nd arg property:
https://github.com/devonfw/IDEasy/blob/6b7c97cbdf578888df867ef769e1e337e5a91d31/cli/src/main/java/com/devonfw/tools/ide/commandlet/InstallCommandlet.java#L17
As it is a `ToolProperty` it can only complete to commandlets that are actually instances of `ToolCommandlet`.
for 3rd: you would find this property:
https://github.com/devonfw/IDEasy/blob/6b7c97cbdf578888df867ef769e1e337e5a91d31/cli/src/main/java/com/devonfw/tools/ide/commandlet/InstallCommandlet.java#L20
And now that we have to complete versions.

The most maintainable approach for implementing all this is making use of polymorphism and delegate the completion to the property itself.
However, therefore you will need to create some completion context containing the `IdeContext` together with the information that has already been gathered for the completion (the matching `commandlet` and its arguments).
Then a `VersionProperty`, `ToolProperty`, etc. can implement some method `autocomplete(CompletionContext, List<Candidate>)` accordingly.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CompleteCommandlet.java
```diff
@@ -0,0 +1,126 @@
...
+  @Override
+  public String getName() {
+
+    return "complete";
+  }
+
+  @Override
+  public void run() {
+
+    AnsiConsole.systemInstall();
```
hohwille, line 59: It is great if we can reuse our auto-completion infrastructure that we implemented for jline.
However, this seems to go in the wrong direction as it will also include undesired side-effects of jline of outputting ANSI control characters and printing additional help where we only need plain auto-completion as pure ASCII.

cli/pom.xml
```diff
@@ -75,6 +75,21 @@
...
+    <dependency>
+      <groupId>org.fusesource.jansi</groupId>
+      <artifactId>jansi</artifactId>
+      <version>2.4.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jline</groupId>
+      <artifactId>jline-terminal-jansi</artifactId>
+      <version>3.23.0</version>
+    </dependency>
```
hohwille, line 92: Great stuff with this awesome libs that give us amazing features.
Can you group the dependencies per `groupId` and use a variable `${jline.version}` to avoid redundancy and simplify maintenance (upgrades)?

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
```diff
@@ -94,6 +111,11 @@ public int run(String... args) {
    */
   public int runOrThrow(String... args) {
 
+    if (args.length == 0) {
+      initializeJlineCompletion();
+      return 1;
```
hohwille, line 116: an exit code different from 0 means an error...

jan-vcapgemini: Adjusted now.

cli/pom.xml
jan-vcapgemini: Removed.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CompleteCommandlet.java
jan-vcapgemini: Removed.

cli/src/main/java/com/devonfw/tools/ide/cli/CommandletRegistry.java
jan-vcapgemini: Adjusted now.

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
jan-vcapgemini: Implemented in a "static" way.

jan-vcapgemini:
I've kept the CommandletRegistry class because it will offer support for various other features like command descriptions, tailtipwidgets etc.
I just need to disable the internal help command from jline3 somehow.

jan-vcapgemini:
I've had to re-add jansi and license additions because of the bug with the Windows Terminal.

MattesMrzik (commented):
Good work!
I have added some comments. Additionally, I have observed, that 
- in git bash when type a string `s` and then press `ctrl + c` nothing is printed to the console. Then upon the next key press `k`, `k` is appended to `s` followed by a newline and the msg `User cancled with CTRL+C`. 
 - In both Windows Terminal and git bash. When I press `ctrl + d` completions are suggested. But when continuing to type the word you want, the completions aren't updated anymore. Test with `set-version k` -> `ctrl + d` -> ` otlincnat` or `set-version az 2.5` -> `ctrl + d` -> `4`
 - when running `get-version` on a tool that is not installed, no msg is printed and the prompt is exits. Running Ide directly with this command prints `Tool ... is not installed!`

cli/src/main/java/com/devonfw/tools/ide/cli/CommandletRegistry.java
```diff
@@ -0,0 +1,178 @@
...
+
+import com.devonfw.tools.ide.commandlet.Commandlet;
+import com.devonfw.tools.ide.commandlet.ContextCommandlet;
+import com.devonfw.tools.ide.commandlet.HelpCommandlet;
+import com.devonfw.tools.ide.context.IdeContext;
+
+/**
+ * Implements the {@link CommandRegistry} for jline3. Inspired by: picocli
+ */
+public class CommandletRegistry implements CommandRegistry {
```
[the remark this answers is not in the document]
jan-vcapgemini: Yes, but I've kept it for reference and possible implementation of further features. 
This class can be used as soon as I've found out how to disabled the internal help from jline3.

cli/pom.xml
```diff
@@ -75,6 +76,16 @@
...
     </dependency>
+    <dependency>
+      <groupId>org.jline</groupId>
+      <artifactId>jline</artifactId>
+      <version>${jline.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.fusesource.jansi</groupId>
+      <artifactId>jansi</artifactId>
+      <version>2.4.0</version>
```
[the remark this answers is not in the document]
jan-vcapgemini: Sure, looks like I've forgotten to re-add the version variable.

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
```diff
@@ -94,8 +109,109 @@ public int run(String... args) {
...
+          } catch (UserInterruptException e) {
+            // Ignore
+            context.warning("User canceled with CTRL+C", e);
+          } catch (EndOfFileException e) {
+            context.warning("User canceled with CTRL+D", e);
+            return;
+          } catch (Exception e) {
+            throw new RuntimeException("An error occurred while using autocompletion", e);
+          } finally {
+            AnsiConsole.systemUninstall();
```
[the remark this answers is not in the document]
jan-vcapgemini: Yes, I've basically taken this from the jline3 implementation example, should be fine.

cli/src/test/java/com/devonfw/tools/ide/context/AutocompletionReaderTestSupport.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+    reader.list = false;
+    reader.menu = false;
+
+    in.setIn(new ByteArrayInputStream(buffer.getBytes()));
+
+    // run it through the reader
+    // String line;
+    // while ((line = reader.readLine((String) null)) != null) {
+    // System.err.println("Read line: " + line);
```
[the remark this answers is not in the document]
jan-vcapgemini: Yes, these can be removed, I've c&p the code from jline3 and missed that part.

cli/src/main/java/com/devonfw/tools/ide/cli/CommandletRegistry.java
```diff
@@ -0,0 +1,178 @@
...
+    return out;
+  }
+
+  // For JLine >= 3.16.0
+  @Override
+  public Object invoke(CommandRegistry.CommandSession session, String command, Object[] args) throws Exception {
+
+    List<String> arguments = new ArrayList<>();
+    arguments.add(command);
+    arguments.addAll(Arrays.stream(args).map(Object::toString).collect(Collectors.toList()));
```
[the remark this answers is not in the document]
jan-vcapgemini: Adjusted now.

cli/pom.xml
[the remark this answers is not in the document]
jan-vcapgemini: Adjusted now.

cli/src/test/java/com/devonfw/tools/ide/context/AutocompletionReaderTestSupport.java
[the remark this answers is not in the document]
jan-vcapgemini: Removed the commented code.

cli/src/test/java/com/devonfw/tools/ide/context/AutoCompletionTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+  }
+
+  @Test
+  public void testIdeCompleterPreventsOptionsAfterCommand() throws IOException {
+
+    ContextCommandlet contextCommandlet = new ContextCommandlet();
+    IdeTestContext ideContext = new IdeTestContext(Paths.get(""), "");
+    reader.setCompleter(new IdeCompleter(contextCommandlet, ideContext));
+    assertBuffer("-t -f get-version - ", new TestBuffer("-t -f get-version - ").tab());
```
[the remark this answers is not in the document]
jan-vcapgemini: I've added another check for invalid input patterns and adjusted the test.

jan-vcapgemini:
I've found a huge bug.

If `-v` was executed once it will always trigger: 
```
if (contextCommandlet.version.isTrue()) {
      return null;
}
```
and completely break the ide e.g. only return the current version.

The old version of the code before my refactorings can be found here:
https://github.com/devonfw/IDEasy/blob/489c8d6e6775d1f7be3e4b7dc591c12896f074d9/cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java#L151-L154

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
```diff
@@ -0,0 +1,201 @@
...
+import com.devonfw.tools.ide.tool.ToolCommandlet;
+import com.devonfw.tools.ide.version.VersionIdentifier;
+
+/**
+ * Implements the {@link Completer} for jline3 autocompletion. Inspired by picocli
+ */
+public class IdeCompleter extends ArgumentCompleter implements Completer {
+
+  /** Pattern which should stop autocompletion e.g. 'get-version mvn -' (prevents options after tool commands) */
+  private static final String INVALID_PATTERN = "([\\s][a-z]*[\\s][\\-](\\s)*.*$)|([a-z][\\-][a-z]*[\\s][\\-](\\s)*.*$)|(^[a-z]*(\\s)[\\-](\\s)*.*$)";
```
MattesMrzik, line 37: Can you pls add a test for this regex (with words that contain "-" in middle of the word, e.g. "foo-bar")? Maybe change [a-z] to \w or \S to make this regex also work when the words contain chars like [A-Z], _, üöüÜÖÄ

MattesMrzik (changes requested):

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
jan-vcapgemini: I've added another test which checks these cases now.
Further checks for more cases should be added to a new issue I think.

MattesMrzik (commented):
Thanks for the fixes. The error stream is not yet displayed. Is this something you want to implement in this PR or do it later in a different PR?

jan-vcapgemini:
> Good work! I have added some comments. Additionally, I have observed, that
> 
>     * in git bash when type a string `s` and then press `ctrl + c` nothing is printed to the console. Then upon the next key press `k`, `k` is appended to `s` followed by a newline and the msg `User cancled with CTRL+C`.
> 
>     * In both Windows Terminal and git bash. When I press `ctrl + d` completions are suggested. But when continuing to type the word you want, the completions aren't updated anymore. Test with `set-version k` -> `ctrl + d` -> ` otlincnat` or `set-version az 2.5` -> `ctrl + d` -> `4`
> 
>     * when running `get-version` on a tool that is not installed, no msg is printed and the prompt is exits. Running Ide directly with this command prints `Tool ... is not installed!`

Thanks for your advanced tests. I've adjusted the autocompleter now to catch all exceptions and be able to continue if an internal error occurred. CTRL+C and CTRL+D should work fine too now.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CompleteCommandlet.java
jan-vcapgemini: Had to re-add `jansi` because of issues with Windows Terminal.

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
```diff
@@ -179,7 +179,7 @@ private void initializeJlineCompletion() {
           } catch (EndOfFileException e) {
             // CTRL+D
             return;
-          } catch (CliException e) {
+          } catch (Exception e) {
             context.error("An error occurred while using autocompletion: {}", e);
```
MattesMrzik, line 183: You could maybe split the try block. So you can catch auto completions error first, then have a different try block for `runCommand()`

MattesMrzik (changes requested):
Can you maybe split the try block to differentiate between autocompletion and CLI errors?

cli/src/main/java/com/devonfw/tools/ide/cli/CommandletRegistry.java
```diff
@@ -0,0 +1,178 @@
...
+    // }
+    // if (o.arity().max() > 0) {
+    // key += "=" + o.paramLabel();
+    // }
+    // options.put(key, val);
+    // }
+    // return new CmdDesc(main, ArgDesc.doArgNames(Arrays.asList("")), options);
+    // TODO: implement this
+    return new CmdDesc(main, ArgDesc.doArgNames(Arrays.asList("description")), options);
+  }
```
MattesMrzik, line 177: Was this method also removed in JLine 3.16.0 and kept it in case this component is used with an older version of JLine?
If not, could we use more descriptive strings than "test" and "description"?

MattesMrzik (commented):
Ready to be reviewed by @hohwille

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
hohwille: Seems to be addressed. @MattesMrzik can we resolve this?

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
```diff
@@ -94,8 +109,113 @@ public int run(String... args) {
...
+            AnsiConsole.systemUninstall();
+          }
+        }
+
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+    } catch (Throwable e) {
+      throw new RuntimeException(e);
+    }
```
hohwille, line 198: What is the point in doing this?
A `Throwable` can be an `OutOfMemoryError` and creating a `RuntimeException` with another `StackTrace` may in the worst case again end in an out of memory causing a JVM crash.
If you want to handle checked exceptions and wrap them as `RuntimeExecption` find but this outer try - catch block seems pointless to me. If you have something like that catching `Exception` will still be sufficient and if you do generic exception wrapping after all, you can use the chance to enrich the exception message with contextual information ("Unexpected error during interactive auto-completion").

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
```diff
@@ -124,13 +244,12 @@ public int runOrThrow(String... args) {
     return 1;
   }
 
-  private CliArgument initContext(CliArgument first) {
+  private CliArgument retrieveCliArgument(CliArgument first, ContextCommandlet contextCommandlet) {
```
hohwille, line 247: The signature refactoring fully makes sense to me. However, the renaming IMHO does not improve the understanding and readability of the method. The main purpose of this method is to run the `ContextCommandlet` to initialize the context/ to apply the contextual CLI arguments (`--debug`, etc.) to the `CliContext`.

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
```diff
@@ -146,6 +265,22 @@ private CliArgument initContext(CliArgument first) {
       }
       current = current.getNext(true);
     }
+    return current;
+  }
+
+  private CliArgument retrieveCliArgumentByContext(CliArgument first, ContextCommandlet contextCommandlet) {
+
+    CliArgument current = retrieveCliArgument(first, contextCommandlet);
+    if (contextCommandlet.version.isTrue()) {
```
hohwille, line 274: FYI: Maybe it would have been easier to create a separate `VersionCommandlet` that can do nothing but print the version and exit.

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
```diff
@@ -0,0 +1,217 @@
...
+
+  /** Checks for invalid pattern e.g. 'get-version foobar - foobar' */
+  private static final String INVALID_PATTERN_PART3 = "([a-z][\\-][a-z]*[\\s][\\-](\\s)*.*$))";
+
+  /** Checks for invalid pattern e.g. 'foo-bar - foobar' */
+  private static final String INVALID_PATTERN_PART4 = "(^[a-z]*(\\s)[\\-](\\s)*.*$)";
+
+  /** Pattern which should stop autocompletion e.g. 'get-version mvn -' (prevents options after tool commands) */
+  private static final String INVALID_PATTERN = INVALID_PATTERN_PART1 + "|" + INVALID_PATTERN_PART2 + "|" + INVALID_PATTERN_PART3
+  + "|" + INVALID_PATTERN_PART4;
```
hohwille, line 50: If I get it right what is checked here altogether is that a hyphen may not occur in isolation.
* Can't this be done with `(^|[\\s])-($|[\\s])` much simpler and much more efficient?
* What is the impact if the user does this? I know it is an edge-case but we already support GNU `--` option in our CLI parser (that allows to provide a file named `-` or `-f` without parsing it as a CLI option). At least for commands like `rm` this is important as if you by accident create a file named `-` or `-f` you could never delete it again :)

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
```diff
@@ -0,0 +1,217 @@
...
+
+    for (Commandlet commandlet : commandletCollection) {
+      // TODO: add more logic to remove unused keyword
+      commandlets.add(commandlet.getName());
+      commandlets.add(commandlet.getKeyword());
+    }
+
+    for (Commandlet commandlet : commandletCollection) {
+      if (commandlet instanceof ToolCommandlet) {
+        // TODO: add more logic to remove unused keyword
```
hohwille, line 89: better create the next story instead of adding TODOs to `main` branch (what we do with the merge as I guess we do not want to wait for perfection with the merge as others already rely on this being merged).

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
```diff
@@ -0,0 +1,217 @@
...
+        toolCommandlets.add(commandlet.getName());
+        toolCommandlets.add(commandlet.getKeyword());
```
hohwille, line 91: do we have any example where `name` can be used via CLI and is different from `keyword`?

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
```diff
@@ -0,0 +1,217 @@
...
+              for (VersionIdentifier vi : versions) {
+                sort++;
+                String versionName = vi.toString();
+                candidates.add(new Candidate(versionName, versionName, null, null, null, null, true, sort));
+              }
+            }
+          }
+        }
+      }
+    }
```
hohwille, line 194: this seems to be a very important core part of this story.
IMHO we should do a meeting to discuss this. So far our CLI is rather simple but common cases are also things like `--some-option some-value` where `--some-option` is an option property that takes a value (instead of being a boolean flag property). Maybe we can avoid adding this complexity but this code seems to redundantly reimplement the existing CLI parsing but based on JLine3 code. If we have bugs or changes in one place, we will always also have to change the same in the other place.
Further you make parsing code more readable and maintainable if you create a "state object" that wraps the input (e.g. `List<String> words` or in other cases `Reader` or whatever) together with the current parsing position allowing to consume data (here `words`) from that "state object" as a stream of tokens. That was the idea behind `CliArgument` or if you want to have a really complex example to look at see e.g. https://github.com/m-m-m/scanner/blob/ecb9a47d1a19796fd26f89dfb984f54477ae96fd/core/src/main/java/io/github/mmm/scanner/CharStreamScanner.java

As discussed earlier my suggestion and expectation would be to delegate the computation of the auto-completion candidates to the property via some new method. You are not utilizing the power of OOP/polymorphism if you do `if (property instanceof ToolProperty) { ... } else if (property instanceof ...`.

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  private void addCandidates(List<Candidate> candidates, Iterable<String> cands, String preFix, String postFix,
+      boolean complete) {
+
+    for (String s : cands) {
+      candidates
+          .add(new Candidate(AttributedString.stripAnsi(preFix + s + postFix), s, null, null, null, null, complete));
+    }
+  }
```
hohwille, line 216: I do not see any invocation using the method directly so why having `preFix` and `postFix`feature if we never use it?

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -116,6 +118,7 @@ public abstract class AbstractIdeContext implements IdeContext {
   public AbstractIdeContext(IdeLogLevel minLogLevel, Function<IdeLogLevel, IdeSubLogger> factory, Path userDir) {
 
     super();
+    this.loggerFactory = factory;
     this.loggers = new HashMap<>();
     for (IdeLogLevel level : IdeLogLevel.values()) {
       IdeSubLogger logger;
```
hohwille, line 124: as you copy & pasted this code into `setLogLevel` method please replace it here with
```
setLogLevel(minLogLevel);
```

FYI: Github suggestion does not work out of reach from the actual code change...

hohwille (changes requested):
@jan-vcapgemini Wow. Thanks for this amazing PR. This is really a complex and tough story and you did very well. 👍 
Thanks for even adding licenses and the great JUnit tests that are really great to read and follow.

As always I digged in the details and left quite some comments for improvement.
Lets have a call and see how we can complete this story for merge asap.

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
```diff
@@ -94,8 +109,113 @@ public int run(String... args) {
...
+
     CliArgument first = CliArgument.of(args);
     CliArgument current = initContext(first);
+    return processCliArgument(current);
+  }
+
+  /**
+   * Initializes jline3 autocompletion.
+   */
+  private void initializeJlineCompletion() {
```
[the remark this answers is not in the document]
jan-vcapgemini: Adjusted.

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
jan-vcapgemini: I've added a new issue and linked it in the TODO comment.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
jan-vcapgemini: Adjusted now.

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
```diff
@@ -94,8 +109,113 @@ public int run(String... args) {
...
+      init.run();
+      this.context = init.getIdeContext();
+
+      // Supplier<Path> workDir = context::getCwd;
+      // set up JLine built-in commands
+      // TODO: fix BuiltIns or remove
+      // Builtins builtins = new Builtins(workDir, null, null);
+      // builtins.rename(Builtins.Command.TTOP, "top");
+      // builtins.alias("zle", "widget");
+      // builtins.alias("bindkey", "keymap");
```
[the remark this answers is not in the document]
jan-vcapgemini: Replaced with TODO with link to issue.

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
```diff
@@ -94,8 +109,113 @@ public int run(String... args) {
...
+        // Enable autosuggestions
+        autosuggestionWidgets.enable();
+
+        // TODO: implement TailTipWidgets
+        // TailTipWidgets widgets = new TailTipWidgets(reader, systemRegistry::commandDescription, 5,
+        // TailTipWidgets.TipType.COMPLETER);
+        // widgets.enable();
+        // TODO: add own KeyMap
+        // KeyMap<Binding> keyMap = reader.getKeyMaps().get("main");
+        // keyMap.bind(new Reference("tailtip-toggle"), KeyMap.alt("s"));
```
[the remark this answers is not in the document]
jan-vcapgemini: Replaced with TODO with link to issue.

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
```diff
@@ -94,8 +109,113 @@ public int run(String... args) {
    */
   public int runOrThrow(String... args) {
 
+    if (args.length == 0) {
+      try {
+        initializeJlineCompletion();
+        return ProcessResult.SUCCESS;
+      } catch (RuntimeException e) {
+        return 1;
+      }
```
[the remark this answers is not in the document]
jan-vcapgemini: Adjusted, but not sure about the returns. Might break/cancel the autocompletion with no way to start it again.

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
jan-vcapgemini: Thanks for your suggestion but this is more complicated I think.
See behavior of the autocompletion tests when I use your suggestion:
```
expected: "-t -f get-version -"
 but was: "-t -f get-version get-version"
```
and:
```
expected: "get-version -t "
 but was: "get-version -t az"
```
The autocompleter should stop when we can't provide any valid options.

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
jan-vcapgemini: Not yet I think.

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
jan-vcapgemini: I agree that this code is hard to read and needs optimizations.
Was the closest PoC I could get working, looking forward to the meeting.

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
jan-vcapgemini: Changed Throwable to Exception now.

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
jan-vcapgemini: Renamed to initContext.

cli/src/main/java/com/devonfw/tools/ide/cli/Ide.java
jan-vcapgemini: I've removed -v and --version and replaced it with VersionCommandlet now.

cli/src/main/java/com/devonfw/tools/ide/cli/CommandletRegistry.java
jan-vcapgemini: Adjusted.

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
jan-vcapgemini: Was taken from the picocli implementation example could be removed I think.

cli/src/main/java/com/devonfw/tools/ide/cli/CommandletRegistry.java
hohwille: Thanks. Still it would make much sense to properly expose the logic to get help for better reuse than actually invoking the HelpCommandlet, but lets do this after the merge. There are tons of other things to rework. I do not like merging stuff that has lots of TODOs and things we already know that we need to rework, but in this case, we need to go forward.

cli/src/main/java/com/devonfw/tools/ide/cli/IdeCompleter.java
hohwille: Then it is pointless to add both name and keyword.

hohwille:
@jan-vcapgemini thanks again for this great PR. You did solid pioneer work to figure out how to make this work. 👍 
Based on your work the following PRs had been created:
* #187
* #178 

These PRs copied your work and further improved it. Therefore, this PR can be closed as it is more or less replaced by the other PRs. Credits to you for making this happen - I guess I would have not integrated jline3 myself but now I really love your idea and see it as a extremely valuable enhancement and feature for IDEasy.

Issue #47, opened 2023-09-25
Support for global tools


hohwille:
We have implemented a lot of cross-cutting logic for tool installations in `ToolCommandlet`.
However, we have tools like `Docker` that behave very differently: Those are tools that get installed globally, that is: 
* You can not have multiple installations of the same tool on the same computer at a time.
* These tools are NOT installed into `IDE_HOME/software` but somewhere globally into the system (e.g. `C:\ProgramFiles` on Windows or `/Applications` on MacOS).
* The downloads of such tools will never be extracted (see `isExtract()` method) and actually `installInRepo` does not really make sense here.

We should discuss and decide the following design options:
* Make `ToolCommandlet` more generic, add a `protected` method `isGlobalInstallation()` that returns `false` by default that can be overridden by global commandlets like `docker`.
* Create abstract sub-classes `GlobalToolCommandlet` and `LocalToolCommandlet` derived from `ToolCommandlet` and make all existing `ToolCommandlet`s derive from `LocalToolCommandlet`. Then move logic like `installInRepo` method into `LocalToolCommandlet` and also for some other methods like `install(...)` but keep an abstract signature for the latter in `ToolCommandlet`. Then implement the `install(...)` methods in `GlobalToolCommandlet`.

hohwille:
Once this story is implemented, we can discuss what global tools we could add to IDEasy:
* `git` - actually we do not really require git-bash to be present to run our CLI that can also run directly from CMD without any bash being present (however some features like #69 may not work). So we could also add the ability to install git via IDEasy so we download the latest installer and run it.
* `keepass` (there is also a MacOs port and IMHO a linux variant)
* `subversion` (`TortoiseSVN` on Windows?)
* how about windows specific tools like `Notepad++`, `SuperPutty`/`WinSCP`/`Kitty`, ...

Issue #75, opened 2023-10-10
Consider integration of oracle XE


hohwille:
As a developer, I want as much automation for installing my required toolings as possible so that I can save time and use it for other things.
Currently devonfw-ide focuses on local sandbox tools. Only docker is some kind of excuse that is a global tool.
Why not also supporting other global tools and simplify the installation for the users.

Meanwhile Oracle XE can be downloaded without login, etc. - see e.g. https://download.oracle.com/otn-pub/otn_software/db-express/OracleXE213_Win64.zip
Already being able to automatically download the latest or a specific version and running the installer could save some time.

However, it should be considered that a better alternative for projects is to use docker in combination of a project internal container image repository where Oracle could be a prepared image. This could even allow to customize the DB setup and auto-run some scripts to create users and tune some parameters as needed for the project. Still I can not believe that XE that is designed for development only has password expiry per default. Things like this are the first options to change and every project I have seen so far does this so IDEasy could help here with further automation. Also disabling the HTTP port 8080 in XE could be automated. There are many more such things that could be considered...

hohwille:
We had long discussions about this topic and came to the following conclusion:
* instead of installing docker natively on the host, it is easier and more flexible to run it in a container (docker/podman)
* when connecting to the DB via JDBC (from the app in development) or via tools like Oracle SQLDeveloper it does not make a difference for the user whether oracle is running on the host or in a container.
* however, when the user wants to run tools like `sqlplus` or especially `impdp` or `expdp` this makes a big difference and this is where we should provide additional help, guidance and support in IDEasy.
* Oracle provides an instance-client that can be downloaded and configured to a remote DB (e.g. the one in the container). This works fine with `sqlplus` but however for `impdp` there is a bug causing error `ORA-39012: Client detached before the job started.`
[...]

PR #113, opened 2023-10-24
#47: Support for global tools

salimbouch:
addresses #47


cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -37,7 +37,7 @@ public abstract class ToolCommandlet extends Commandlet implements Tags {
   /** The commandline arguments to pass to the tool. */
   public final StringListProperty arguments;
 
-  private MacOsHelper macOsHelper;
+  protected MacOsHelper macOsHelper;
```
hohwille, line 40: do not change the visibility of the field but instead of the `getMacOsHelper()` method.
Directly using this file in a sub-class would lead to NPEs as the getter lazily initializes the field.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -291,202 +221,6 @@ protected void postInstall() {
...
-      return true;
-    }
-    return false;
-  }
-
-  /**
-   * @param file the {@link Path} to the file to extract.
-   * @param targetDir the {@link Path} to the directory where to extract (or copy) the file.
-   */
-  protected void extract(Path file, Path targetDir) {
```
hohwille, line 436: I would expect the `extract` logic to be available both for local and global tool commandlets.
While most global tools directly download an MSI, PKG, EXE, dpkg, etc. and install it without that we extract it in our Java code, there can be some edge-cases, where we will also need this for global tool commandlets in the future...

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -0,0 +1,85 @@
...
+    ToolRepository toolRepository = this.context.getDefaultToolRepository();
+    VersionIdentifier configuredVersion = getConfiguredVersion();
+    VersionIdentifier resolvedVersion = toolRepository.resolveVersion(this.tool, edition, configuredVersion);
+    Path binaryPath = getToolBinary();
+    if (Files.exists(binaryPath)) {
+      // TODO: check whether resolvedVersion is different than installedVersion, if so update the tool?
+      return false;
+    }
+    // download and install the global tool
+    Path target = toolRepository.download(this.tool, edition, resolvedVersion);
```
hohwille, line 76: this still looks like quite some redundancy between global and local tools that we could share in `ToolCommandlet`.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -102,33 +102,13 @@ public void runTool(VersionIdentifier toolVersion, String... args) {
   /**
    * @return the {@link Path} where the main executable file of this tool is installed.
    */
-  public Path getToolBinary() {
+  public abstract Path getToolBinary();
```
hohwille, line 105: There are two things to distinguish for global tools:
1. the installer binary used to install the tool globally on the system
2. the binary of the tool once it is installed on the system

According to the JavaDoc this method should refer to the 2nd one.
However, I can not see how we could implement this in a generic way.
Also we have to revisit the sense of this method after #111 - most probably this method will disappear (Sorry, when stories are implemented in parallel that have overlapping code changes it always gets kind of tricky and you will most probably also will have to deal with git conflicts. It therefore will make sense to work closer together with @aBega2000 so you both work together and not "against each other").

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
hohwille: BTW: the `isExtract()` method is there from the old design of devonfw-ide in bash and represents the parameter
`noUnpack` that was mainly used for global tool commandlets:
https://github.com/devonfw/ide/blob/af67497e06fb82e2e50a4d9ffa10e3f9dda44aa1/scripts/src/main/resources/scripts/functions#L1266

salimbouch:
@hohwille do we want to have a file tool version in global tools? if so, should this file be next to the binary? if not, how do we know what version of the global tool we have and how can users update global tools?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
salimbouch: Done, local are by default extractable and global arent but in need we can override isExtract and make it return true

salimbouch: Question now is in case we need to extract a global tool where should we extract it? In the download cache or somewhere else?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
salimbouch: Im waiting for https://github.com/devonfw/IDEasy/pull/111 to be merged then I'll resolve the conflicts and change my code according to his changes.

hohwille: I merged #111 now.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
hohwille: > Question now is in case we need to extract a global tool where should we extract it? In the download cache or somewhere else?

This shall be done in a temp folder as it will only be needed temporary to run the installer and after the installation (e.g. to `C:\Program Files`) it is not needed anymore. This is the same behavior as in devonfw-ide.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -0,0 +1,246 @@
...
+      assert (!linkDir.equals(rootDir));
+      this.context.getFileAccess().copy(toolVersionFile, linkDir, FileCopyMode.COPY_FILE_OVERRIDE);
+    }
+    return new ToolInstallation(rootDir, linkDir, binDir, resolvedVersion);
+  }
+
+  /**
+   * @return the currently installed tool version or {@code null} if not found (tool not installed).
+   */
+  protected String getInstalledToolVersion() {
```
hohwille, line 207: never duplicate such code. Either move it from ToolCommandlet here or leave it in ToolCommandlet.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -0,0 +1,101 @@
...
+
+    String path = System.getenv("PATH");
+    String[] pathDirs = path.split(File.pathSeparator);
+    for (String dir : pathDirs) {
+      Path toolPath = Paths.get(dir, getName());
+      if (Files.isExecutable(toolPath)) {
+        return toolPath;
+      }
+    }
+    return null;
```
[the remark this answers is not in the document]
hohwille: If we need this for `doInstall`, you could simply use `this.context.getPath().findBinary(getBinaryName())`.
You have removed `getBinaryName()` from `ToolCommandlet`.
IMHO this must be reverted.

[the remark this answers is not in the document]
salimbouch: I removed it since it's the same as getName(), or is there something that I oversaw?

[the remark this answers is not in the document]
salimbouch: 1. Done.
2. Done. No need for `Files.isExecutable` now 
3. Done. That's exactly what I did but my commits weren't pushed.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
salimbouch: I made getInstalledVersion abstract in ToolCommandlet, implemented it in LocalToolCommandlet and in GlobalToolCommandlet it just gives null back.
Another solution would be to remove it entirely from ToolCommandlet and GlobalToolCommandlet and only implement it in LocalToolCommandlet then change GetVersionCommandlet to: 
`

    ToolCommandlet commandlet = this.tool.getValue();
    if (commandlet instanceof GlobalToolCommandlet) {
      throw new UnsupportedOperationException("Not implemented yet!");
    }
    VersionIdentifier installedVersion = ((LocalToolCommandlet)commandlet).getInstalledVersion();
    if (installedVersion == null) {
      throw new CliException("Tool " + commandlet.getName() + " is not installed!", 4);
    }
    this.context.info(installedVersion.toString());
  }`

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
salimbouch: 1- Should I replicate the overloading of installInRepo (i.e. installGlobally) or is it enough if I the whole logic is in doInstall?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -230,183 +174,17 @@ public boolean install(boolean silent) {
...
-   */
-  protected boolean isExtract() {
-
-    return true;
-  }
-
-  /**
-   * @return the currently installed tool version or {@code null} if not found (tool not installed).
-   */
-  protected String getInstalledToolVersion() {
```
hohwille, line 371: I did not realize before that I myself have duplicated this method. It is there twice:
* `public VersionIdentifier getInstalledVersion()` - fine
* `protected String getInstalledToolVersion()` - duplicated non-sense, it seems this is not used anywhere so it can simply be removed.

Sorry for getting a little off topic - not your actual task, but it would be great if we can already clean this up.
I also just found out that @aBega2000 forgot to add JavaDoc for his new method `getBinaryName()` that would also be nice to add.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -0,0 +1,101 @@
...
+
+    super(context, tool, tags);
+  }
+
+  /**
+   * Override this if the global tool comes with a file that has to be extracted.
+   * @return {@code true} to extract (unpack) the downloaded binary file, {@code false} otherwise.
+   */
+  @Override
+  protected boolean isExtract() {
```
hohwille, line 36: Never duplicate the JavaDoc for overridden methods (that is one of the reasons for `@Override` annotation).
The JavaDoc will always be inherited and all IDEs support this and also when you generate the JavaDoc as HTML this will happen. When you copy & paste the JavaDoc and it gets changed, the copy gets outdated and nobody will notice.

You may add additional information to the JavaDoc of inherited methods. But this is only to be done in very few cases.
Please note that JavaDoc describes the API of a method and not its implementation.
If you want to explain the implementation simply use code comments:
```

@Override
protected boolean isExtract() {
  // for global tools we usually download installers and do not want to extract them (e.g. installer.msi file shall not be extracted)
  return false;
}
```

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * Installs {@link #getName() tool}, if force mode is enabled it proceeds with the installation even if the tool
+   * is already installed
+   * @param silent - {@code true} if called recursively to suppress verbose logging, {@code false} otherwise.
+   * @return {@code true} if the tool was newly installed, {@code false} if the tool was already installed before and
+   *         nothing has changed.
+   */
+  @Override
+  protected boolean doInstall(boolean silent) {
```
hohwille, line 49: remove JavaDoc

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -0,0 +1,101 @@
...
+    return null;
+  }
+
+  /**
+   * @return the currently installed tool version or {@code null} if not found (tool not installed).
+   */
+  @Override
+  protected String getInstalledToolVersion() {
+    return null;
+  }
```
hohwille, line 97: remove entirely. This class is abstract. If it can not implement this methods it should leave the task for the sub-classes to implement it.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
hohwille: This method should be entirely removed. I already left a comment that this was a mistake on my end from the start.
Further do not implement methods "wrong". Either we do not support `getInstalledVersion()` for `GlobalToolCommdlet`s or we do. If we do then `GlobalToolCommdlet` can not easily implement this in a generic way. It would need to run the binary and ask it for its version like this:
```
$ docker -v
Docker version 24.0.5, build ced0996
```
Then the method has to be implemented in `DockerCommandlet` accordingly.

IMHO this makes sense as e.g. `VersionGetCommandlet` should also work for global tools.

BTW: your suggestion with `instanceof` is not making use of polymorphism. The beauty of OOP is that it simplifies such things. So if the behavior of your example would actually be desired, it would still be better to simply have the following in `GlobalToolCommandlet`:
```
@Override
protected String getInstalledToolVersion() {
  throw new UnsupportedOperationException("Not implemented yet!");
}
```
and in `LocalToolCommandlet` keep the rest of the implementation.
The `instanceof` approach makes `ToolCommandlet` to know `GlobalToolCommandlet` what otherwise is not required and more elegant.

hohwille: And finally what would make sense is something like this (most probably in `ToolCommandlet`) as it allows proper reuse:
```
  protected VersionIdentifier getInstalledVersion(Path toolPath) {
    
    if (!Files.isDirectory(toolPath)) {
      this.context.debug("Tool {} not installed in {}", getName(), toolPath);
      return null;
    }
    Path toolVersionFile = toolPath.resolve(IdeContext.FILE_SOFTWARE_VERSION);
    if (!Files.exists(toolVersionFile)) {
      Path legacyToolVersionFile = toolPath.resolve(IdeContext.FILE_LEGACY_SOFTWARE_VERSION);
      if (Files.exists(legacyToolVersionFile)) {
        toolVersionFile = legacyToolVersionFile;
      } else {
        this.context.warning("Tool {} is missing version file in {}", getName(), toolVersionFile);
        return null;
      }
    }
    try {
      String version = Files.readString(toolVersionFile).trim();
      return VersionIdentifier.of(version);
    } catch (IOException e) {
      throw new IllegalStateException("Failed to read file " + toolVersionFile, e);
    }
  }
```
Now we can easily implement this making reuse:
```

    return getInstalledVersion(getToolPath());
  }
```

The benefit is that we can later also reuse `getInstalledVersion(Path)` to get the version of a tool from the software repository (e.g. to verify that all is sane).

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -494,6 +272,32 @@ protected void extract(Path file, Path targetDir) {
     }
   }
 
+  /**
+   * @return {@code true} to extract (unpack) the downloaded binary file, {@code false} otherwise.
+   */
+  protected boolean isExtract() {
+
+    return true;
+  }
```
hohwille, line 281: if you are moving methods within the same file, it looks like `isExtract()` or `getBinaryName()` has actually been added or removed but actually nothing has changed with these two methods at all.
I know that this quickly happens by accident and you do not notice and see this in your IDE.
However, always watch your diffs before you commit and also have a look into your diff in the PR to see that all is as expected and intended. It is not a big deal if this happens a few time. I just want to raise awareness to make life of reviewers easier in the future :)

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
hohwille: A common pattern in OOP is also to have a template implementation in an abstract class like this:
```
public abstract class ChatBase {
  public void doSomething() {
    // common beginning part
    System.out.println("Hello, how are your?");
    // special part that differs
    doSomethingSpecial();
    // common ending part
    System.out.println("It was nice talking to you, goodbye.");
  }
  
  protected abstract void soSomethingSpecial();
}

public class ChatGoodMood extends ChatBase {
  @Override
  protected void soSomethingSpecial() {
    System.out.println("It is a beautiful day and I feel great.");
  }
}

public class ChatBadMood extends ChatBase {
  @Override
  protected void soSomethingSpecial() {
    System.out.println("The wars are not ending and the climate will collapse. I feel very sad.");
  }
}
```

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
hohwille: Method is gone, resolved.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
salimbouch: Done

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
salimbouch: Done. Thanks for clarifying

salimbouch: I removed `Override it to add custom post installation logic.` from the javaDoc of the method postInstall.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
salimbouch: Done

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
salimbouch: Done

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
salimbouch: I will make sure to avoid this in my next PRs :)

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
salimbouch: Done. Thanks for the clarification.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -230,183 +178,16 @@ public boolean install(boolean silent) {
...
-    }
-    postInstall();
-    return true;
-  }
+  protected abstract boolean doInstall(boolean silent);
 
   /**
-   * This method is called after the tool has been newly installed or updated to a new version. Override it to add
-   * custom post intallation logic.
+   * This method is called after the tool has been newly installed or updated to a new version.
```
hohwille, line 184: Did you remove `Override it to add custom post intallation logic.` on purpose or by accident?

hohwille (approved):
@salimbouch Thanks for your rework. Now everything is perfect. As I had quite a list of review comments and added improvements it is especially great to see that you solved it all in one shot 👍 
Ready for merge.

The last comment is just for my curiosity and shall not block the merge...

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
hohwille: IMHO we can do such nice-to-have improvements after the merge.
It is better to have this already on main branch so it gets available for other stories already worked on like #75.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
hohwille: I just saw your explanation in the other comment...

PR #114, opened 2023-10-24
Update AssertJ to 3.24.2 (broken)

hohwille:
We had AssertJ version `2.6.0` that was entirely outdated and caused errors with Java17:
```
Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make field private ... accessible: module java.base does not "opens java.util" to unnamed module @7fbe847c
```
This is fixed with the update so we can use `SoftAssertions` and other modern features of AssertJ.
As AssertJ renamed some assertion methods for `java.time.*`, I had to fix a test and rename the method calls.


hohwille:
Replaced by #115

PR #125, opened 2023-10-30
#121 Windows version of latest rancher desktop is missing

### LINKED ISSUE #121: Windows version of latest rancher desktop is missing
### Expected behavior
For `DOCKER_EDITION=rancher` both
```
devon docker setup
```
and
```
ide install docker
```

should work.

### Actual behavior

The above fails on windows because the URLs file for windows_x64 is missing here:
https://github.com/devonfw/ide-urls/tree/master/docker/rancher/1.10.0

1. Open the URL https://github.com/devonfw/ide-urls/tree/master/docker/rancher/1.10.0
2. look for windows URLs

### Comments/Hints:

URL Updater for Rancher seems to have some issue. E.g. Rancher switched from *.exe to *.mis and maybe they now switched again with the latest release. We need to find a more robust solution rather than assuming a fixed extension that changes between each release (Actually the tool vendors should stick with strict naming conventions and common best-practices for releases. However many tool vendors don't and there is little that we can do about this so we have to get creative to circumvent those problems)
https://github.com/devonfw/IDEasy/blob/fd02c6042a9ac3ec2b6e29edf5517d1453d972b8/cli/src/main/java/com/devonfw/tools/ide/tool/docker/DockerRancherDesktopUrlUpdater.java#L45

### Affected version:

* OS: Windows

CREITZ25:
Fixes #121 DockerRancherDesktopUrlUpdater.java: Changing filetype from exe to msi for the windows download.


hohwille (approved):
@CREITZ25 thanks for this fix. Ready to merge 👍 
Fingers crossed that tonight this will add the missing windows URL.

PR #133, opened 2023-11-07
context question index bug fix

MattesMrzik:
The index that corresponds to the question options was not incremented. This fixes it.


salimbouch:
Looks good :)

hohwille (approved):
@MattesMrzik Thanks for finding and fixing this bug. 👍 
I am slightly confused as I already created the same bugfix but indeed I could not find it in the git history for whatever reason so either I never pushed it or it ended up in some PR that never got merged.

Your fix is perfectly correct and therefore I approve and merge.

PR #134, opened 2023-11-07
fileaccess delete symlink fix

MattesMrzik:
Deleting a symlink resulted in an exception due to the call to deleteRecursive. This call was removed in the case of a symlink.


salimbouch:
looks good to me.

hohwille (approved):
@MattesMrzik Thanks for finding and fixing this bug 👍 
I already reviewed it as part of your other PR. However, as you created this PR, I can already approve and merge this.

Issue #10, opened 2023-09-20
Create support for IDEs


hohwille:
Currently we have implemented `ToolCommandlet` with subclasses like `Java` or `Mvn`.
In order to implement `Eclipse`, `Intellij`, etc. we need an additional abstract class derived from `ToolCommandlet` that handles all the features required for IDEs:
* installation of plugins
* removal of plugins
* configuring the workspace (by calling our `Configurator` - this is the old code https://github.com/devonfw/ide/blob/af67497e06fb82e2e50a4d9ffa10e3f9dda44aa1/scripts/src/main/resources/scripts/functions#L1538)

For the `Configurator` we should:
* remove main method: https://github.com/devonfw/IDEasy/blob/ebb06e5d026753405d4407d143422c13f774f48e/cli/src/main/java/com/devonfw/tools/ide/configurator/Configurator.java#L247-L252
* instead of the `run` method with `String[]` args, we should add explicit parameters to the constructor and change the `run` method to be `public void run()` instead. See https://github.com/devonfw/IDEasy/blob/ebb06e5d026753405d4407d143422c13f774f48e/cli/src/main/java/com/devonfw/tools/ide/configurator/Configurator.java#L153
* It would make sense to pass `IdeContext` or `IdeLogger` to `Configurator` as Constructor arg and use that instead of the old logger: https://github.com/devonfw/IDEasy/blob/ebb06e5d026753405d4407d143422c13f774f48e/cli/src/main/java/com/devonfw/tools/ide/configurator/Configurator.java#L281

hohwille:
Was auto-closed by PR but project import and some other features not completed. Hence I do reopen.

hohwille:
As this story was very unclear (what is to be done, what is already working), I created individual new issues so we can now close this story. New tickets are:
* #516: repository import (aka project import)
* #519: dedicated plugin installation/uninstallation

PR #135, opened 2023-11-08
#10: improve IDE support and configure workspace

hohwille:
Implements most of #10:
* completely refactored configurator (only little left of the original code)
* integrated configurator into `IdeToolCommandlet` so workspace gets auto-configured before IDE is started
* pass proper arguments to eclipse
* show project name in eclipse title-bar (instead of workspace that is already shown be default)
* check if eclipse workspace is locked
* fixed bug in `AbstractIdeContextTest.newContext(String, String, boolean)`: `IdeContext` was created in original location and not in the copy - hence `VersionSetCommandletTest` was modifying git managed file (undesired side effect).
* improved some tests
* replace last `File` occurrences with `Path`
* normalize paths in String representation (replace backslash with slash)
* make paths absolute in `IdeContext` (otherwise workspace configuration is not working if paths are relative)
* moved logic to resolve variables from `EnvironmentVariablesResolved` to `EnvironmentVariables` API


cli/src/main/java/com/devonfw/tools/ide/merge/DirectoryMerger.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+    return this.fallbackMerger;
+  }
+
+  @Override
+  public void inverseMerge(Path workspace, EnvironmentVariables variables, boolean addNewProperties, Path update) {
+
+    if (Files.isDirectory(update)) {
+      if (!Files.isDirectory(workspace)) {
+        Log.warn("Workspace is missing directory: " + workspace);
```
jan-vcapgemini, line 92: Why are we using jline.Log instead of this.context.warning here?

cli/src/main/java/com/devonfw/tools/ide/merge/PropertiesMerger.java
```diff
@@ -0,0 +1,160 @@
...
+   * @return the loaded {@link Properties}.
+   */
+  public static Properties loadIfExists(Path file) {
+
+    Properties properties = new Properties();
+    if (file != null) {
+      if (Files.exists(file)) {
+        load(properties, file);
+      } else {
+        Log.trace("Properties file does not exist: " + file);
```
jan-vcapgemini, line 76: Why are we using jline.Log instead of this.context.warning here?

cli/src/main/java/com/devonfw/tools/ide/variable/VariableDefinition.java
```diff
@@ -71,6 +73,8 @@ default String toString(V value) {
 
     if (value == null) {
       return "";
+    } else if (value instanceof Path) {
+      return value.toString().replace('\\', '/');
```
jan-vcapgemini, line 77: Cool :)

cli/src/test/java/com/devonfw/tools/ide/merge/DirectoryMergerTest.java
```diff
@@ -51,84 +53,78 @@ public class ConfiguratorTest extends Assertions {
...
   public void testConfigurator() throws Exception {
 
-    // given
-    Configurator configurator = new Configurator();
-    String tmp = System.getProperty("java.io.tmpdir");
-    File tmpDir = new File(tmp);
-    File workspaceDir = createUniqueFolder(tmpDir, ".test.workspace");
-    String clientEnvHome = DEVON_IDE_HOME;
+    // arrange
+    Path workspaceDir = Files.createTempDirectory("IDEasy-DirectoryMergerTest");
```
jan-vcapgemini, line 57: Why are we not using jUnit5 @temp directory here instead?

jan-vcapgemini (changes requested):
Looks good to me.
I've added some typo fixes and fixed log messages.
Please check the questions I've added.

jan-vcapgemini (changes requested):
Looks good to me.
I've added some typo fixes and fixed log messages.
Please check the questions I've added.

cli/src/main/java/com/devonfw/tools/ide/merge/DirectoryMerger.java
hohwille: accident and mistake on my end: the old code was containing `Log.warn` and I removed out `Log` class and the static import and thought that I had replaced all these with a proper `context.warn` but missed some.
By saving the Eclipse Save Actions automatically added the import to `org.jline`.

Thank you so much for finding and fixing this "bug".

cli/src/main/java/com/devonfw/tools/ide/variable/VariableDefinition.java
hohwille: I was annoyed that we always have to add special cases for windows and IMHO all tools and everything can deal with regular slashes in 2023+.
Now JUnit testing is more simple.
Can we close this conversation or was there some "irony" in your comment?

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -138,6 +151,97 @@ public EnvironmentVariables resolved() {
...
+    if (recursion > MAX_RECURSION) {
+      throw new IllegalStateException("Reached maximum recursion resolving " + value + " for root valiable " + rootSrc
+          + " with value '" + rootValue + "'.");
+    }
+    recursion++;
+    Matcher matcher = VARIABLE_SYNTAX.matcher(value);
+    if (!matcher.find()) {
+      return value;
+    }
+    StringBuilder sb = new StringBuilder(value.length() + 8);
```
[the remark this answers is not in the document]
hohwille: I just moved this code from somewhere else in this PR.
However, you are totally right, that `8` is just some magic number here and could also be `4` or `16` or whatever.
The idea is to pick some size that even when replacing `${variable}` with `variable-value` we have some extra capacity left and it is less likely that a new `char[]` has to be allocated and the old data needs to be copied with `arrayCopy`.

I have never seen a silver bullet pattern for this problem. You could even add a specific percentage of the length of the original value. Some devs even try to make it even more accurate and scan the entire string upfront and count the number of potential replacements. Then you improve this capacity prediction more and more and end up with code that actually does the entire resolution before just to compute the capacity and then again after for the actual replacements. As a result you wanted to optimize performance but caused more overhead in computation and in code to maintain that you actually spoiled it. :)
Feel free to add a comment, replace `8` with something else, ...

hohwille:
@jan-vcapgemini thank you very much for your review and your constructive fixes of my typos and the log-statements I missed out to align. All your changes are correct and improve the code quality. 👍

cli/src/test/java/com/devonfw/tools/ide/merge/DirectoryMergerTest.java
hohwille: This was an old test-case that I quickly wanted to fix so it was working again and it was 1:18 AM (in the night) when I completed my work and created the PR so I was just too tired for further perfection (have a look at how this test was creating the temporary directory before my change :) ).
Thanks for pointing this out. I will rework it to use the `@Temp` annotation as you suggested. I assume this will also ensure it gets deleted after the test method call ended.

hohwille: BTW: if we both aim for perfection - the reverseMerge is still implemented very naive and could get lots of improvements and is also entirely not covered by JUnits (however, for the moment it can also not be used via CLI yet - what will also have to be implemented for #10 but in a new PR).

cli/src/main/java/com/devonfw/tools/ide/variable/VariableDefinition.java
jan-vcapgemini: Can be closed as this was just cool, seen issues with that lately so good that it's fixed now.

hohwille:
@jan-vcapgemini I finally found some minutes to address your review-comments.
Please check and see if we can resolve the remaining conversations and finally merge this.

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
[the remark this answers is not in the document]
hohwille: as discussed with @moritzLanger I will now merge this as others are blocked by the bugfix with the context copyForMutation feature fixed in the PR.
I addressed this review comment. Usually @jan-vcapgemini would be in charge to decide if we can resolve it but we need to move on and he is currently unavailable.

cli/src/test/java/com/devonfw/tools/ide/merge/DirectoryMergerTest.java
hohwille: Done.
As discussed with @moritzLanger I will now merge this as others are blocked by the bugfix with the context copyForMutation feature fixed in the PR.
I addressed this review comment. Usually @jan-vcapgemini would be in charge to decide if we can resolve it but we need to move on and he is currently unavailable.

Issue #127, opened 2023-11-03
Increase commandlet test coverage


moritzLanger:
The test coverage of the commandlets need to be improved.

Test for version get version set version list and the help commandlet have already been written.

PR #137, opened 2023-11-09
#127: Increase commandlet test coverage for InstallCommandlet 

moritzLanger:
https://github.com/devonfw/IDEasy/issues/127 Added 100% coverage for InstallCommandlet

Related PR #144


MattesMrzik (commented):
Looking good. Did you differentiate between Linux and Windows because the behavior of the InstallCommandlet actually differs for these OS or because the github action that checks the test is using the Linux version, and you on your local machine are using Windows? If not, could you use only a single stubFor?

moritzLanger:
> Looking good. Did you differentiate between Linux and Windows because the behavior of the InstallCommandlet actually differs for these OS or because the github action that checks the test is using the Linux version, and you on your local machine are using Windows? If not, could you use only a single stubFor?

The difference in behavior is only that Linux needs tgz and windows does zip, so for testing on a windows machine you need a different download, that's why I included two Stubs. It also has the positive side effect, of testing that both the unpacking of .zip and .tgz is working correctly.

cli/src/test/java/com/devonfw/tools/ide/commandlet/InstallCommandletTest.java
```diff
@@ -0,0 +1,115 @@
...
+  public void assertTestInstall(IdeContext context) {
+
+    assertThat(context.getSoftwarePath().resolve("java")).exists();
+    assertThat(context.getSoftwarePath().resolve("java/InstallTest.txt")).hasContent("This is a test file.");
+    assertThat(context.getSoftwarePath().resolve("java/bin/HelloWorld.txt")).hasContent("Hello World!");
+    if(context.getSystemInfo().isWindows()){
+      assertThat(context.getSoftwarePath().resolve("java/Java.cmd")).exists();
+    } else if (context.getSystemInfo().isLinux() || context.getSystemInfo().isMac()) {
+      assertThat(context.getSoftwarePath().resolve("java/Java.sh")).exists();
+    }
```
hohwille, line 110: My suggestion was to include commands following the actual tool conventions so we can not only test the installation but also the execution (there is some non-trivial logic to find the tool executable to run after the installation). Then we could actually run `Java` commandlet and if such `java/bin/java.cmd` MS-Dos script of `java/bin/java` as Bash-Script is present then the proper dummy output could be asserted depending on the OS (e.g. `Dummy Java 17.0.6 on Windows` could be asserted on Windows if you put `echo Dummy Java 17.0.6 on Windows` into `java.cmd`).

hohwille:
> The difference in behavior is only that Linux needs tgz and windows does zip, so for testing on a windows machine you need a different download, that's why I included two Stubs. It also has the positive side effect, of testing that both the unpacking of .zip and .tgz is working correctly.

This is great that we test OS dependent behaviors and also extracting both tgz and zip. However, just for the record and comming to the original question from @MattesMrzik you could however, have added `http://localhost:1111/installTest/java.tgz` to the `url` files for each OS with just one stub and it would have also worked a little simpler. Do not change this but just to make it clear where the extension is coming from.

hohwille (approved):
@moritzLanger Thanks for this PR and the great coverage. Good job 👍 
Can you still have a look at my review-comment as there was some slight misunderstanding of what we discussed earlier.

cli/src/test/java/com/devonfw/tools/ide/commandlet/InstallCommandletTest.java
moritzLanger: After further analysis, the script using echo is not writing into the system.out, and running the test manually I receive the following warning and report: 

`[WARNING] Corrupted channel by directly writing to native stream in forked JVM 1. See FAQ web page and the dump file C:\Users\molanger\ide\workspaces\main\127\IDEasy\cli\target\surefire-reports\2023-12-14T12-15-45_934-jvmRun1.dumpstream`

` Created at 2023-12-14T12:15:47.165
Corrupted channel by directly writing to native stream in forked JVM 1. Stream 'C:\Users\molanger\ide\workspaces\main\127\IDEasy\cli>echo "This is a test install file." '.

# Created at 2023-12-14T12:15:47.174
Corrupted channel by directly writing to native stream in forked JVM 1. Stream '"This is a test install file."'.

`

hohwille: I am unclear what is going on but my first guess is that maven surefire is somehow interfering here and already redirecting streams of our JUnit test that again wanted to do the same.
So lets first get this merged and then consider further perfection...

PR #140, opened 2023-11-13
#139: feature for making symlinks relative

### LINKED ISSUE #139: Feature for making symlinks relative
Sometimes it is desired that symlinks point to relative locations not to absolute ones. For example if a link in `dir1` points to some file inside `dir1` and then `dir1` is moved or renamed. Then, if the link was relative instead of absolute, it is still functioning.

MattesMrzik:
closes #139 
- changes `symlink` default behavior to work with relative instead of absolute paths. This can be toggled with a flag.
- made `symlink`'s behavior consistent between soft links and windows junctions.
- added tests for these


cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -282,22 +284,74 @@ private void copyRecursive(Path source, Path target, FileCopyMode mode) throws I
...
+        if (Files.isSymbolicLink(targetLink)) {
+          this.context.debug("Deleting symbolic link to be re-created at {}", targetLink);
+          Files.delete(targetLink);
+        } else {
+          BasicFileAttributes attr = Files.readAttributes(targetLink, BasicFileAttributes.class,
+              LinkOption.NOFOLLOW_LINKS);
+          if (attr.isOther() && attr.isDirectory()) {
+            this.context.debug("Deleting symbolic link (junction) to be re-created at {}", targetLink);
+            Files.delete(targetLink);
+          }
```
hohwille, line 335: great. I guesses something like this but did not exactly know how to do it.
I do not know if `attr.isOther()` will reliably tell us if the directory is a junction but I assume you have tested this with a regular directory and a junction and also that for a junction the `Files.delete(targetLink)` will only delete the junction itself. So this looks like a perfect solution to me. Thanks for figuring this out 👍

hohwille: Maybe we should also include a test if we are on windows in this else for the junction.
I have no idea what `attr.isOther()` could also mean on mac or linux but as junctions can only exist on windows this would be safer to avoid problems in edge-cases.

MattesMrzik: I have added this isWindows() check. And the behaviour for `Files.delete(targetLink)` is that only the junction is deleted not the linked directory. Yes, `attr.isOther()` returns false for a directory.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -282,22 +284,74 @@ private void copyRecursive(Path source, Path target, FileCopyMode mode) throws I
...
+  }
+
+  @Override
+  public void makeSymlinkRelative(Path link, boolean followTarget) {
+
+    if (!Files.isSymbolicLink(link)) {
+      throw new IllegalStateException(
+          "Can't call makeSymlinkRelative on " + link + " since it is not a symbolic link.");
+    }
+    Path linkTarget = null;
```
moritzLanger, line 300: `= null` is not needed as Object are initialized as null so `Path linkTarget` should work aswell

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -282,22 +284,74 @@ private void copyRecursive(Path source, Path target, FileCopyMode mode) throws I
...
+  }
+
+  @Override
+  public void relativeSymlink(Path link, Path source) {
+
+    Path relativeSource = link.getParent().relativize(source);
+    // to make relative links like this work: dir/link -> dir
+    relativeSource = (relativeSource.toString().isEmpty()) ? Paths.get(".") : relativeSource;
+    symlink(relativeSource, link);
+  }
```
moritzLanger, line 318: This method is currently only used by the above method, meaning it could be included to safe some lines of code.

If the need for a direct creation of relative symlinks could come up in the future this comment could be ignored.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
```diff
@@ -51,6 +51,32 @@ public interface FileAccess {
...
+   */
+  void makeSymlinkRelative(Path link, boolean followTarget);
+
+  /**
+   * Creates a symbolic relative link.
+   *
+   * @param source the source {@link Path} to link to.
+   * @param targetLink the {@link Path} where the symbolic link shall be created pointing to {@code source}.
+   */
+  void relativeSymlink(Path targetLink, Path source);
```
moritzLanger, line 78: could be removed if the implementation is deemed unnecessary, see comment in FileAccesImpl.java

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -282,22 +284,74 @@ private void copyRecursive(Path source, Path target, FileCopyMode mode) throws I
...
-        this.context.debug("Deleting symbolic link to be re-created at {}", targetLink);
-        Files.delete(targetLink);
+      if (Files.exists(targetLink)) {
+        if (Files.isSymbolicLink(targetLink)) {
+          this.context.debug("Deleting symbolic link to be re-created at {}", targetLink);
+          Files.delete(targetLink);
+        } else {
+          BasicFileAttributes attr = Files.readAttributes(targetLink, BasicFileAttributes.class,
+              LinkOption.NOFOLLOW_LINKS);
+          if (attr.isOther() && attr.isDirectory() && this.context.getSystemInfo().isWindows()) {
```
moritzLanger, line 332: `isOther()` : 
true if the file something other than a regular file, directory or symbolic link

`isDirectory()` :
true if the file is a directory
(see [Documentation](https://docs.oracle.com/javase/8/docs/api/java/nio/file/attribute/BasicFileAttributes.html))

don't these two cancel each other out and therefore this if would always be false?

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
```diff
@@ -0,0 +1,227 @@
+package com.devonfw.tools.ide.io;
+
+import static org.junit.jupiter.api.Assertions.*;
```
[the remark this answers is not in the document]
moritzLanger: they will be included in the AbstractIdeContextTest which you could extend

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
```diff
@@ -0,0 +1,227 @@
...
+    FileAccess fileAccess = new FileAccessImpl(context);
+    Path dir = tempDir.resolve("dir");
+    fileAccess.mkdirs(dir);
+
+    Path file = tempDir.resolve("file");
+    try {
+      Files.write(file, "Hello World!".getBytes());
+    } catch (IOException e) {
+      throw new RuntimeException("When preparing testSymlink writing of " + file + " failed. ", e);
+    }
```
moritzLanger, line 34: You could manually put a file in the test resource folder to avoid these lines of creating writing a file:

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
```diff
@@ -0,0 +1,227 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import com.devonfw.tools.ide.context.IdeContext;
+import com.devonfw.tools.ide.context.IdeTestContextMock;
+
+public class FileAccessImplTest extends Assertions {
+
+  @Test
+  void testSymlink(@TempDir Path tempDir) {
```
moritzLanger, line 22: not necessary but we normally declare  Test methods as public, even to the default is public, just a consistency thing across tests

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
```diff
@@ -0,0 +1,227 @@
...
+        });
+        assertThat(e2).hasMessageContaining("These junctions can only point to absolute paths");
+      } else {
+        throw new RuntimeException(
+            "Creating symbolic link with Files.createSymbolicLink failed and junctions can not be used since the OS is not windows: "
+                + e.getMessage());
+      }
+    }
+
+    // test for normal symlinks (not junctions)
```
moritzLanger, line 66: Again maybe a own test for these new parts

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
```diff
@@ -0,0 +1,227 @@
...
+          assertThat(linkInParent2.getParent().resolve(readPath)).existsNoFollowLinks();
+          assertThat(linkInParent2.getParent().resolve(readPath)).exists();
+        } catch (IOException e) {
+          throw new RuntimeException("Could not call Files.readSymbolicLink on moved relative link: " + linkInParent2,
+              e);
+        }
+      }
+    }
+  }
+}
```
moritzLanger, line 227: The test look solid from a functional standpoint but should definitely be separated in multiple Test for the sake of Clarity and ease of use later on. Normally an assert should not be followed by another act, but rather a new test for that.

moritzLanger (changes requested):
Good Work 👍 
I added some Minor suggestions for the main Code and one question on the If statement, but otherwise really well done

The Test should be a bit more separated for Clarity reasons in my opinion, as well as minor changes to be more similar to other Tests in the Project

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
MattesMrzik: okey:)

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
MattesMrzik: I have talked with @hohwille about this, and we think that relative symlinks can be more fitting in some cases than absolute ones. He even suggested making the default behavior of the symlink method to use relative paths. In my opinion, we should keep the symlink method as it is (absolute) and have in addition the method relativeSymlink. Additionally, the method makeSymlinkRelative is used in the  moveAndProcessExtraction of the AWS toolcommandlet, where AWS ships a script that creates absolute symlinks, that would break after moving the installation from the tempDir to the software dir. So the method makeSymlinkRelative is applied to make them relative and thereby still functioning after they have been moved.
But if you have strong arguments, then you can convince me.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
hohwille: Only member variables are automatically initialized as null (and array items) but not local variables.
However, as the variable gets initialized in 302 and in case of an IOException we exit with a RuntimeException the compiler is smart enough to understand it is initialized in 308 so @moritzLanger is right that the assignment is not needed.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -282,22 +284,74 @@ private void copyRecursive(Path source, Path target, FileCopyMode mode) throws I
...
+    }
+    Path linkTarget = null;
+    try {
+      linkTarget = followTarget ? link.toRealPath() : Files.readSymbolicLink(link);
+    } catch (IOException e) {
+      throw new RuntimeException("For link " + link + " the call to "
+          + (followTarget ? "toRealPath" : "readSymbolicLink") + " in method makeSymlinkRelative failed.", e);
+    }
+    this.context.getFileAccess().delete(link); // delete old absolute link
+    this.context.getFileAccess().relativeSymlink(link, linkTarget); // and replace it by the new relative link
```
hohwille, line 308: it is pointless to call yourself via `this.context.getFileAccess().` - simply remove this prefix before `delete` or `relativeSymlink`.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
hohwille: IMHO it is totally confusing that `symlink` is taking the source as the first argument and the target as second argument and you now created a very similar method where you inverted the semantic of these parameters.
Please also note that my order of args was not chosen by random: symbolic links work this way `ln -s «source» «target»` and that is also intentional as is is analog to `cp «source» «target»`.

My suggestion would be to instead add the method like this and even make relative symlinks the new default
```
void symlink(Path source, Path targetLink, boolean relative);

default symlink(Path source, Path targetLink) {
  symlink(source, targetLink, true);
}
```
In all places where we currently create symlinks (esp. in software folder) we would benefit from relative links.
If we have the need to create an absolute link, we have the ability to provide `false` for the `relative` parameter.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
moritzLanger: Ok, looking at the docu it seemed weird, but if it works 👍

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
MattesMrzik: I have implemented this `arrange act assert` structure.

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
```diff
@@ -0,0 +1,227 @@
...
+          fileAccess.symlink(file, link);
+        });
+        assertThat(e1).hasMessageContaining("These junctions can only point to directories or other junctions");
+
+        IllegalStateException e2 = assertThrows(IllegalStateException.class, () -> {
+          fileAccess.symlink(Paths.get("dir"), link);
+        });
+        assertThat(e2).hasMessageContaining("These junctions can only point to absolute paths");
+      } else {
+        throw new RuntimeException(
```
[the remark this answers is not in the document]
MattesMrzik: I have changed this. Thanks

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
```diff
@@ -0,0 +1,227 @@
...
+
+import com.devonfw.tools.ide.context.IdeContext;
+import com.devonfw.tools.ide.context.IdeTestContextMock;
+
+public class FileAccessImplTest extends Assertions {
+
+  @Test
+  void testSymlink(@TempDir Path tempDir) {
+
+    IdeContext context = IdeTestContextMock.get();
```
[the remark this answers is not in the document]
MattesMrzik: I have implemented this `arrange act assert` structure.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
MattesMrzik: Okey, I see your point. I have adjusted it to be like you suggested. Thank you both!

moritzLanger:
You can also resolve all the Conversations started by me from the previous review

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
MattesMrzik: After discussing this with you and the team I changed this a bit but still act inside a tmp dir.

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
```diff
@@ -0,0 +1,227 @@
...
+      throw new RuntimeException("When preparing testSymlink writing of " + file + " failed. ", e);
+    }
+
+    Path link = tempDir.resolve("link");
+    Path linkToLink = tempDir.resolve("linkToLink");
+
+    boolean junctionsUsed = false;
+    try {
+      Files.createSymbolicLink(link, file);
+    } catch (IOException e) { // if permission is not hold, then junctions are used instead of symlinks (on Windows)
```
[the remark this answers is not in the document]
MattesMrzik: I separated it. Thanks!

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
```diff
@@ -0,0 +1,299 @@
...
+      Path readPath = Files.readSymbolicLink(link);
+      if (Files.exists(readPath)) {
+        fail("The link target " + readPath + " (from readSymbolicLink) should not exist");
+      }
+    }
+  }
+
+  /*
+   * only pass readLinks = true when junctions are not used.
+   */
```
hohwille, line 240: If you want to use JavaDoc start with `/**` and use `@param readLinks` to be more precise.
Otherwise if you only want a code-comment simply use `//` what avoids two pointless extra lines ;)

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
```diff
@@ -0,0 +1,299 @@
...
+    assertSymlinkToRealPath(dir.resolve("d1/d11/link1"), dir.resolve("d1"));
+    assertSymlinkToRealPath(dir.resolve("d1/d11/link2"), dir.resolve("d1/d11"));
+    assertSymlinkToRealPath(dir.resolve("d1/d11/link3"), dir.resolve("d1/d11/d111"));
+    assertSymlinkToRealPath(dir.resolve("d1/d11/link4"), dir.resolve("d1/d11/d111/d1111"));
+    assertSymlinkToRealPath(dir.resolve("d1/d11/link5"), dir.resolve("d2"));
+    assertSymlinkToRealPath(dir.resolve("d1/d11/link6"), dir.resolve("d2/d22"));
+    assertSymlinkToRealPath(dir.resolve("d1/d11/link7"), dir.resolve("d2/d22/d222"));
+    assertSymlinkToRealPath(dir.resolve("d2/d22/link8"), dir.resolve("d1"));
+    assertSymlinkToRealPath(dir.resolve("d2/link9"), dir.resolve("d1"));
+    assertSymlinkToRealPath(dir.resolve("link10"), dir.resolve("d1"));
```
hohwille, line 252: I often see such pattern in JUnits ("lack of business oriented test-data/-scenario") - however it is very hard to read and understand such "technical random setup".
I figured it corresponds to `createSymlinks` but if you use more speaking names and an example that somehow makes sense it is much easier for others to read the code. Example:
Create symlink from `long-path/that/should-have/some-shortcut` to `shortcut` and users can much better understand your expected result than why `d1/d11/link4` points to `d1/d11/d111/d1111`.

Please note: This is only for your information to think about for the next time. IMHO your test is OK and you do not have to refactor this. I am already happy that you write a test and cover so many cases what is great. Just always trying to give feedback so you can learn and still improve...

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
```diff
@@ -0,0 +1,299 @@
...
+
+    Path source = dir.resolve("checkIfWindowsJunctionsAreUsed");
+    Path link = dir.resolve("checkIfWindowsJunctionsAreUsedLink");
+    context.getFileAccess().mkdirs(source);
+    try {
+      Files.createSymbolicLink(link, source);
+    } catch (IOException e) {
+      return context.getSystemInfo().isWindows();
+    }
+    return false;
```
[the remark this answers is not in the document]
hohwille: This is from your current build on linux:
```
[INFO] Running com.devonfw.tools.ide.io.FileAccessImplTest
11:05:47.217 [main] [C:] - INFO  - c.d.tools.ide.log.IdeSlf4jLogger - Can not check the Test: testWindowsJunctionsCanNotPointToFiles since windows junctions are not used.
11:05:47.220 [main] [C:] - DEBUG - c.d.tools.ide.log.IdeSlf4jLogger - Deleting symbolic link to be re-created at /tmp/junit16793612142448171343/parent/d1/d11/link1
11:05:47.224 [main] [C:] - INFO  - c.d.tools.ide.log.IdeSlf4jLogger - Can not check the Test: testSymlinkAbsoluteAsFallback since windows junctions are not used and fallback from relative to absolute paths as link target is not used.
11:05:47.226 [main] [C:] - DEBUG - c.d.tools.ide.log.IdeSlf4jLogger - Deleting symbolic link to be re-created at /tmp/junit9783774862290736358/parent/d1/d11/link1
11:05:47.231 [main] [C:] - DEBUG - c.d.tools.ide.log.IdeSlf4jLogger - Deleting symbolic link to be re-created at /tmp/junit2359642677952943407/parent/d1/d11/link1
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.020 s -- in com.devonfw.tools.ide.io.FileAccessImplTest
```
You can run your test locally on Windows and see if there is any difference.

cli/src/test/java/com/devonfw/tools/ide/version/VersionIdentifierTest.java
```diff
@@ -93,7 +93,7 @@ public void testIllegal() {
     for (String version : illegalVersions) {
       try {
         VersionIdentifier.of(version);
-        fail("Illegal verion '" + version + "' did not cause an exception!");
+        fail("Illegal version '" + version + "' did not cause an exception!");
```
hohwille, line 96: thanks for fixing my typos!

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
```diff
@@ -62,7 +62,12 @@ public interface FileAccess {
    * @param source the source {@link Path} to link to.
    * @param targetLink the {@link Path} where the symbolic link shall be created pointing to {@code source}.
    */
-  void symlink(Path source, Path targetLink);
+  void symlink(Path source, Path targetLink, boolean relative);
+
+  default void symlink(Path source, Path targetLink) {
```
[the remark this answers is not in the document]
MattesMrzik: Jep, im getting not errors due missing JavaDoc. But I have added these.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -291,27 +294,66 @@ private void copyRecursive(Path source, Path target, FileCopyMode mode) throws I
...
 
   @Override
-  public void symlink(Path source, Path targetLink) {
+  public void symlink(Path source, Path targetLink, boolean relative) {
+
+    if (!source.isAbsolute()) {
+      throw new IllegalStateException("When creating a symbolic link the source (" + source
+          + ") must be an absolute path. If you want to create a relative link, "
+          + "then pass the source as an absolute path and set the relative flag to true.");
+    }
```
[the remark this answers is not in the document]
MattesMrzik: The rationale was that I wanted to force the use of the flag "relative" instead of just passing a relative path as source. But I can change it to also work with passing realative sources.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -291,27 +294,66 @@ private void copyRecursive(Path source, Path target, FileCopyMode mode) throws I
...
+        if (Files.isSymbolicLink(targetLink)) {
+          this.context.debug("Deleting symbolic link to be re-created at {}", targetLink);
+          Files.delete(targetLink);
+        }
       }
-      Files.createSymbolicLink(targetLink, source);
+      try { // since broken junctions are not detected by Files.exists(brokenJunction)
+        BasicFileAttributes attr = Files.readAttributes(targetLink, BasicFileAttributes.class,
+            LinkOption.NOFOLLOW_LINKS);
+        if (attr.isOther() && attr.isDirectory() && this.context.getSystemInfo().isWindows()) {
```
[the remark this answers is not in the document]
MattesMrzik: Thanks for these suggestions. I implemented them.

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
MattesMrzik: Thanks! I tried to make the link names a bit more clear. I left the names of the dirs the same, because I dont want to have super long path names.

hohwille (changes requested):
@MattesMrzik sorry for the extra round. This is almost perfect now but as I found something that looks like a bug but I was not sure, if I do not get it or if my suggested fix is 100% correct, I decided to give this back to you for a final completion. Thanks in advance 👍

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -290,28 +293,101 @@ private void copyRecursive(Path source, Path target, FileCopyMode mode) throws I
...
+   * 
+   * @param source the {@link Path} to adapt.
+   * @param targetLink the {@link Path} used to calculate the relative path to the {@code source} if {@code relative} is
+   *        set to {@code true}.
+   * @param relative the {@code relative} flag.
+   * @return the adapted {@link Path}.
+   */
+  private Path adaptPath(Path source, Path targetLink, boolean relative) {
+
+    if (relative && source.isAbsolute()) {
```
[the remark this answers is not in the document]
MattesMrzik: The passed source may be absolute or relative, but will be adapted to be absolute or relative depending on the flag "relative". So if the flag says relative and the source is absolute, then the source is changed to be relative. So this `if` is correct. Nonetheless, the body of the `if` was wrong, which may have caused the confusion.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -290,28 +293,101 @@ private void copyRecursive(Path source, Path target, FileCopyMode mode) throws I
...
+      return source.toAbsolutePath();
+    }
+    if (!relative && !source.isAbsolute()) {
```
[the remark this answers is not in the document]
MattesMrzik: I think the method you are looking for is `Path::toRealPath()`. During creation of link `l1`  if the source is a symlink `l2` do we want to have `l2`s location or the target of `l2` as source of the newly created link `l1`?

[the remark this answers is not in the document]
MattesMrzik: I think you are locking for the method `toRealPath()`. I have implemented your mentioned behavior and wrote a test for it. Let's say link `l1` is about to be created. And as source for `l1` a path to another link `l2` is passed. Because I set `LinkOption.NOFOLLOW_LINKS` the source of `l1` will be `l2` instead of the source of `l2`.

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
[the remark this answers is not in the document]
MattesMrzik: Thanks.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -290,28 +293,101 @@ private void copyRecursive(Path source, Path target, FileCopyMode mode) throws I
...
+          this.context.warning(
+              "You are on Windows and you do not have permissions to create symbolic links. Junctions are used as an "
+                  + "alternative, however, these can not point to relative paths. So the source (" + adaptedSource
+                  + ") is interpreted as " + "absolute path (" + adaptedSource.toAbsolutePath() + ").");
+        }
+        if (!Files.isDirectory(adaptedSource.toAbsolutePath())) { // if source is a junction. This returns true as well.
+          throw new IllegalStateException(infoMsg
+              + "\\n These junctions can only point to directories or other junctions. Please make sure that the source ("
+              + adaptedSource.toAbsolutePath() + ") is one of these.");
+        }
```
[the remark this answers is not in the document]
MattesMrzik: I fixed this. Thanks for your suggestions.

hohwille (approved):
@MattesMrzik thanks for your nice rework. 👍 
Ready for merge.

PR #143, opened 2023-11-14
#100: implement repository commandlet

### LINKED ISSUE #100: Implement Commandlet for projects/repositories
As a IDEasy user, I want to be able to add projects/repositories so that I can get them setup automatically.
In `devonfw-ide` there is `devon project ...` and the setup automatically adds all active projects:
* https://github.com/devonfw/ide/blob/master/documentation/project.asciidoc
* https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/project

This functionality is missing in IDEasy and shall be implemented with this story.

salimbouch:
closes #100 

- implements repository commandlet (old project commandlet)
- solves bug, findBinary not finding the binary that was just installed (see setPath in SystemPath)


cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,169 @@
...
+        for (Path propertiesFile : propertiesFiles) {
+          doImportRepository(propertiesFile, forceMode);
+        }
+      }
+    }
+  }
+
+  private void doImportRepository(Path repositoryFile, boolean forceMode) {
+
+    this.context.info("Importing from {} ...", repositoryFile.toString());
```
hohwille, line 88:
```suggestion
    this.context.info("Importing repository from {} ...", repositoryFile.getFileName().toString());
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * The constructor.
+   *
+   * @param context the {@link IdeContext}.
+   */
+  public RepositoryCommandlet(IdeContext context) {
+
+    super(context);
+    addKeyword(getName());
+    add(new StringProperty("setup", true, null));
```
hohwille, line 35:
```suggestion
    addKeyword("setup");
```

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
```diff
@@ -131,4 +132,12 @@ default void copy(Path source, Path target) {
    */
   Path findFirst(Path dir, Predicate<Path> filter, boolean recursive);
 
+  /**
+   * Retrieves a list of file paths from the specified directory that satisfy the given predicate.
+   * @param dir the folder to iterate through
+   * @param filter the {@link Predicate} that determines whether a file should be included in the list.
+   * @return a list of paths that satisfy the provided {@link Predicate} or {@code null} if no match was found.
```
hohwille, line 139:
```suggestion
   * @return a list of paths that satisfy the provided {@link Predicate}. Will be {@link List#isEmpty() empty} if no match was found but is never {@code null}.
```

cli/src/main/resources/nls/Ide_de.properties
```diff
@@ -5,6 +5,7 @@ options=Optionen:
 cmd-az=Werkzeug Kommando fuer Azure Kommandoschnittstelle.
 cmd-env=Gibt die zu setztenden und exportierenden Umgebungsvariablen aus.
 cmd-get-version=Zeigt die Version des selektierten Werkzeugs an.
+cmd-repository=Richtet das vorkonfigurierte Projekt ein.
```
hohwille, line 8:
```suggestion
cmd-repository=Richtet das vorkonfigurierte Git Repository ein.
```

cli/src/main/resources/nls/Ide_de.properties
```diff
@@ -24,6 +25,8 @@ val-args=Die Kommandozeilen-Argumente zur Übergabe an das Werkzeug.
 val-tool=Das zu selektierende Werkzeug Kommando.
 val-version=Die Werkzeug Version.
 val-set-version-version=Die zu setztende Werkzeug Version.
+val-repository-project= Das vorkonfigurierte Projekt zum Einrichten. 
+
```
hohwille, line 29:
```suggestion
val-repository-repository=Der Name der Properties-Datei des vorkonfigurierten Git Repositories zum Einrichten. Falls nicht angegeben, werden alle aktiven Projekte eingerichtet. 
```

cli/src/main/resources/nls/Ide.properties
```diff
@@ -5,6 +5,7 @@ options=Options:
 cmd-az=Tool commandlet for Azure CLI.
 cmd-env=Print the environment variables to set and export.
 cmd-get-version=Get the version of the selected tool.
+cmd-repository=setup the pre-configured repository.
```
hohwille, line 8:
```suggestion
cmd-repository=setup the pre-configured git repository.
```

cli/src/main/resources/nls/Ide.properties
```diff
@@ -25,6 +26,7 @@ val-args=The commandline arguments to pass to the tool.
 val-tool=The tool commandlet to select.
 val-version=The tool version.
 val-set-version-version=The tool version to set.
+val-repository-repository=The pre-configured repository to setup, omit to setup all active repositories.
```
hohwille, line 29:
```suggestion
val-repository-repository=The name of the properties file of the pre-configured git repository to setup, omit to setup all active repositories.
```

hohwille (changes requested):
@salimbouch thanks for your PR. You integrated the feature properly in the right code places and your approach looks fine 👍 
I added several review comments for improvement. Please have a look.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
```diff
@@ -131,4 +132,12 @@ default void copy(Path source, Path target) {
...
+   */
+  List<Path> getFilesInDir(Path dir, Predicate<Path> filter);
```
hohwille, line 141: actually this method does not only return files but also folders. It should therefore be named `getChildren` or `getChildrenOfDir`.

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,169 @@
...
+  public String getName() {
+
+    return "repository";
+  }
+
+  @Override
+  public void run() {
+
+    Path repositoriesPath = this.context.getSettingsPath().resolve("repositories");
+    Path legacyRepositoriesPath = this.context.getSettingsPath().resolve("projects");
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,169 @@
...
+          return;
+        }
+      }
+      doImportRepository(repositoryFile, true);
+    } else {
+      // If no specific repository is provided, check for repositories folder
+      Path repositories = Files.exists(repositoriesPath) ? repositoriesPath :
+          Files.exists(legacyRepositoriesPath) ? legacyRepositoriesPath : null;
+
+      if (repositories == null) return;
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,169 @@
...
+
+      List <Path> propertiesFiles = this.context.getFileAccess().getFilesInDir(repositories,
+          path -> "properties".equals(FilenameUtil.getExtension(path.getFileName().toString())));
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,169 @@
...
+
+      if (propertiesFiles != null) {
```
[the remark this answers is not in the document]
salimbouch: done :) this inconsistency happened because the method used to return null when no children were found and I forgot to change this accordingly 😅

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryConfig.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public record RepositoryConfig(
+    String path,
+    String workingSets,
+    String workspace,
+    String gitUrl,
+    String gitBranch,
+    String buildPath,
+    String buildCmd,
+    String eclipse,
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
```diff
@@ -166,6 +166,7 @@ public Path getPath(String tool) {
    */
   public void setPath(String tool, Path path) {
 
+    this.paths.add(path);
```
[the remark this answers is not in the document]
salimbouch: No this is the PR where I first stumbled upon this bug.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,169 @@
...
+        properties.getProperty("workingsets"),
+        properties.getProperty("workspace"),
+        properties.getProperty("git_url"),
+        properties.getProperty("git_branch"),
+        properties.getProperty(("build_path")),
+        properties.getProperty("build_cmd"),
+        properties.getProperty("eclipse"),
+        Boolean.parseBoolean(properties.getProperty("active"))
+    );
+  }
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.List;
+import java.util.Properties;
+
+/**
+ * {@link Commandlet} to setup a repository
+ */
+public class RepositoryCommandlet extends Commandlet {
+
+  /** the repository to setup. */
+  public final PathProperty repository;
```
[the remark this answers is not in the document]
salimbouch: I implemented a RepositoryProperty but not sure if that's the right way to do it. How can I access the context from a property? For now I gave the context as a parameter to the property which doesn't seem clean, plus should the resolving of the `Path` from the string value be done in a new method or from `getValue()?`

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,169 @@
...
+      }
+      if (!Files.exists(repositoryFile)) {
+        Path legacyRepositoryFile = legacyRepositoriesPath.resolve(repositoryFile.getFileName().toString());
+        if (Files.exists(legacyRepositoryFile)) {
+          repositoryFile = legacyRepositoryFile;
+        } else {
+          this.context.warning("Could not find " + repositoryFile);
+          return;
+        }
+      }
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,136 @@
...
+    String repository = repositoryConfig.path();
+    String gitUrl = repositoryConfig.gitUrl();
+    if (repository == null || "".equals(repository) || gitUrl == null || "".equals(gitUrl)) {
+      this.context.warning("Invalid repository configuration {} - both 'path' and 'git-url' have to be defined."
+          , repositoryFile);
+      return;
+    }
+
+    this.context.debug(repositoryConfig.toString());
+    this.context.debug("Pull or clone git repository {} ...", repository);
```
hohwille, line 102: In line #113 we are calling `gitPullOrClone`. It would be better to keep this closer together as other invocations like `mkdirs` may also log on debug meanwhile.
However, since this debug message is entirely generic, why not directly having it inside `gitPullOrClone` method implementation?

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryConfig.java
```diff
@@ -0,0 +1,65 @@
...
+    } catch (IOException e) {
+      throw new IllegalStateException("Failed to read file: " + filePath, e);
+    }
+
+    Set<String> importsSet = getImports(properties);
+
+    return new RepositoryConfig(properties.getProperty("path"), properties.getProperty("workingsets"),
+        properties.getProperty("workspace"), properties.getProperty("git_url"), properties.getProperty("git_branch"),
+        properties.getProperty(("build_path")), properties.getProperty("build_cmd"), importsSet,
+        Boolean.parseBoolean(properties.getProperty("active")));
```
hohwille, line 48: fine but be aware that `Boolean.parseBoolean("true ")` will return `false` what might be surprising and can be hard to trace down.

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,136 @@
...
+      }
+      commandlet.arguments.setValue(args);
+      commandlet.run();
+    } else {
+      this.context.info("Build command not set. Skipping build for repository.");
+    }
+
+    if (!Files.exists(repositoryPath.resolve(".project"))) {
+      for (String ideCommandlet : repositoryConfig.imports()) {
+        //TODO: import repository to ideCommandlet
```
hohwille, line 132: mhm. Actual `TODO` tags should be avoided on `main`.
I fully agree that we should implement the actual project import in a separate PR later.
According to clean code it would be better to remove this if/for block and instead create a new story issue for the use-case of importing projects.

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,136 @@
...
+      for (int i = 1; i < command.length; i++) {
+        args.add(command[i]);
+      }
+      commandlet.arguments.setValue(args);
+      commandlet.run();
+    } else {
+      this.context.info("Build command not set. Skipping build for repository.");
+    }
+
+    if (!Files.exists(repositoryPath.resolve(".project"))) {
```
hohwille, line 130: Is this `.project` file the one from the `Eclipse` import? I guess so. Using constants in `IdeContext` that have some JavaDoc is always better to make the code understandable...

hohwille (changes requested):
@salimbouch Thanks for your PR and the great work 👍 
Looks good to me and you did nicely implement the existing behavior from the bash script and the new functionality. And thanks for fixing other issues like JavaDoc arg name, etc.

Surely I left several review comments for improvement.
Once resolved, you can reassign to me and I can approve and merge.
If you have questions or want to discuss something about the design feel free to get in touch.

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,136 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+
+import static com.devonfw.tools.ide.commandlet.RepositoryConfig.loadProperties;
+
+/**
+ * {@link Commandlet} to setup a repository
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,136 @@
...
+  public String getName() {
+
+    return "repository";
+  }
+
+  @Override
+  public void run() {
+
+    Path repositoriesPath = this.context.getSettingsPath().resolve(context.FOLDER_REPOSITORIES);
+    Path legacyRepositoriesPath = this.context.getSettingsPath().resolve(context.FOLDER_LEGACY_REPOSITORIES);
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/property/RepositoryProperty.java
```diff
@@ -0,0 +1,75 @@
...
+  @Override
+  public String parse(String valueAsString) {
+
+    return valueAsString;
+  }
+
+  public Path getValueAsPath(IdeContext context) {
+
+    Path repositoriesPath = context.getSettingsPath().resolve(context.FOLDER_REPOSITORIES);
+    Path legacyRepositoriesPath = context.getSettingsPath().resolve(context.FOLDER_LEGACY_REPOSITORIES);
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/property/RepositoryProperty.java
```diff
@@ -0,0 +1,75 @@
...
+      repositoryFile = repositoriesPath.resolve(repositoryFile.getFileName().toString() + ".properties");
+    }
+    if (!Files.exists(repositoryFile)) {
+      Path legacyRepositoryFile = legacyRepositoriesPath.resolve(repositoryFile.getFileName().toString());
+      if (Files.exists(legacyRepositoryFile)) {
+        repositoryFile = legacyRepositoryFile;
+      } else {
+        throw new IllegalStateException("Could not find " + repositoryFile);
+      }
+    }
```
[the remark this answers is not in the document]
salimbouch: done, implemented method findExistingFile in FileAccess.

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
salimbouch: done :)

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryConfig.java
salimbouch: added `trim()` which should remove all leading and ending white spaces and should resolve that problem.

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
salimbouch: Done, removed that if block and will create a new story for importing repositories

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
salimbouch: Done, removed this if block entirely.

cli/src/main/java/com/devonfw/tools/ide/property/RepositoryProperty.java
```diff
@@ -0,0 +1,75 @@
+package com.devonfw.tools.ide.property;
+
+import com.devonfw.tools.ide.context.IdeContext;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.function.Consumer;
+
+public class RepositoryProperty extends Property<String> {
```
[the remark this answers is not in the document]
salimbouch: So I can just call `parse` in the RepositoryCommandlet instead? Since Properties have no context attribute how can I access it?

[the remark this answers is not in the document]
hohwille: Fine for now, but in PR #178 I refactored it to have the IdeContext. I will now also be able to remove the static hacks entirely.

hohwille:
@salimbouch thanks for your reworks. I resolved most of the review comments. 👍 
PR/feature-branch needs to be rebased and conflicts resolved.
Ideally you can even address the final suggestions for `RepositoryProperty`.
As we meanwhile have Coveralls complaining you will need to add a JUnit test covering your new feature.

Once merged, we can further discuss to modernize the API as `ide repository setup` probably does not make so much sense anymore. Will we support other operations than adding (cloning and building) a repository? Either `add` and `remove` would make sense or if we decide to only ever support adding repositories (deletion of a folder can also be done by the user manually) we could even skip the additional keyword (currently it is hardcoded as required what does not make much sense as you cannot use `ide repository ...` with anything else then `setup ...`).

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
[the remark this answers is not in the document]
salimbouch: Done

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -626,4 +626,37 @@ private Path findFirstRecursive(Path dir, Predicate<Path> filter, boolean recurs
...
+  public Path findExistingFile(String fileName, List<Path> searchDirs) {
+
+    for (Path dir : searchDirs) {
+      Path filePath = dir.resolve(fileName);
+      try {
+        if (Files.exists(filePath)) {
+          return filePath;
+        }
+      } catch (SecurityException e) {
+        throw new IllegalStateException("SecurityException while checking file existence.");
```
hohwille, line 656: https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.asciidoc#catching-and-handling-exceptions

cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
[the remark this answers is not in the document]
hohwille: The "fix" is still incorrect. It will add a path entry and cause duplicates.
Assuming you run the same method twice:
```
systemPath.setPath("mvn", "/d/projects/foo/software/mvn");
systemPath.setPath("mvn","/d/projects/bar/software/mvn");
```
This will cause the `PATH` variable to contain both `/d/projects/foo/software/mvn` and `/d/projects/bar/software/mvn`.
You think this scenario sounds unrealistic?
This is exactly what will happen if I call `ide` in project `foo` and then switch to project `bar` and call `ide` there.
Do we have a bug ticket for this problem itself?

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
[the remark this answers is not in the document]
hohwille: Thanks. We still need to implement auto-completion for `RepositoryProperty`. Feel free to do so by overriding the `completeValue` method or to create a new story issue for that so we dont forget this.

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,130 @@
...
+
+    this.context.debug(repositoryConfig.toString());
+
+    String workspace = repositoryConfig.workspace() != null ? repositoryConfig.workspace() : "main";
+    Path workspacePath = this.context.getIdeHome().resolve("workspaces").resolve(workspace);
+    this.context.getFileAccess().mkdirs(workspacePath);
+
+    if (repositoryConfig.gitBranch() != null && !repositoryConfig.gitBranch().isEmpty()) {
+      gitUrl = gitUrl + "#" + repositoryConfig.gitBranch();
+    }
```
hohwille, line 109: See https://github.com/devonfw/IDEasy/pull/155/files#diff-ca40b72302007b2636f83de91be5b8e9616aab3ba14e6bff7aa7f8d9705d4fb9R57-R60

salimbouch: Discussed with @jan-vcapgemini, I'll wait for his PR to be merged then change this block accordingly.

cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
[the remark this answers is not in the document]
salimbouch: I just saw your comment right now and couldn't address it in time, good that @MustaphaOuchen took care of it 👍

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
salimbouch: I also noticed the property "build_path" isn't taken into consideration in my implementation. In order to make it work we'd have to change the directory where the process is executed which is not doable since we just call the run() method of the build tool. My suggestion would be we overload the method run() in ToolCommandlet.

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * The constructor.
+   *
+   * @param context the {@link IdeContext}.
+   */
+  public RepositoryCommandlet(IdeContext context) {
+
+    super(context);
+    addKeyword(getName());
+    addKeyword("setup");
```
hohwille, line 36: FYI: In devonfw-ide we used the term `setup` as generic keyword for "install":
* `devon mvn setup` installs maven (and Java)
* `devon project setup` installs all configured projects
* We have some logic in the setup process that runs `setup` on all the according commandlets.

However, omitting this keyword would result in `ide repository` what is also rather confusing or not self-explanatory.
We could have the freedom now to change this keyword to anything more suitable.
But I currently do not see a much better name: `ide repository install` might also be kind of confusing.
Potentially we could simply use `add` and then later also have the ability to offer `remove` as well.
Anyway, I will just set this comment to resolved as this seems to be fine and if we come to some better idea, we can still add another PR after merge.

cli/src/main/java/com/devonfw/tools/ide/property/RepositoryProperty.java
```diff
@@ -0,0 +1,62 @@
...
+   *
+   * @param name the {@link #getName() property name}.
+   * @param required the {@link #isRequired() required flag}.
+   * @param alias the {@link #getAlias() property alias}.
+   * @param mustExist the {@link #isPathRequiredToExist() required to exist flag}.
+   */
+  public RepositoryProperty(String name, boolean required, String alias, boolean mustExist) {
+
+    super(name, required, alias, mustExist);
+  }
```
hohwille, line 23:
```suggestion
   */
  public RepositoryProperty(String name, boolean required, String alias) {

    super(name, required, alias, true);
  }
```

cli/src/main/java/com/devonfw/tools/ide/property/RepositoryProperty.java
```diff
@@ -0,0 +1,62 @@
...
+   * @param name the {@link #getName() property name}.
+   * @param required the {@link #isRequired() required flag}.
+   * @param alias the {@link #getAlias() property alias}.
+   * @param mustExist the {@link #isPathRequiredToExist() required to exist flag}.
+   * @param validator the {@link Consumer} used to {@link #validate() validate} the {@link #getValue() value}.
+   */
+  public RepositoryProperty(String name, boolean required, String alias, boolean mustExist, Consumer<Path> validator) {
+
+    super(name, required, alias, mustExist, validator);
+  }
```
hohwille, line 37:
```suggestion
   * @param validator the {@link Consumer} used to {@link #validate() validate} the {@link #getValue() value}.
   */
  public RepositoryProperty(String name, boolean required, String alias, Consumer<Path> validator) {

    super(name, required, alias, true, validator);
  }
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
```diff
@@ -0,0 +1,130 @@
...
+    this.repository = add(new RepositoryProperty("", false, "repository", true));
```
hohwille, line 37:
```suggestion
    this.repository = add(new RepositoryProperty("", false, "repository"));
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
hohwille: I still think it is better to explicitly pass the branch as separate argument to `GitContext` instead of concatenating url and branch, pass it, and then parse that combined string back into URL and branch. Feels more like a hack than a solution.
But surely it is not wrong and will work this way...

hohwille (approved):
@salimbouch thanks for your rework and maintaining the feature-branch with all the conflict resolutions.
This now seems fine to me and can be merged.
I left some comments. If you have some time to accept the proposals and maybe even improve on the last suggestion this would be great. I would merge this on Thursday. If you don't address the last point till then, I can merge it and we can still improve this in a separate PR after merge.

cli/src/main/java/com/devonfw/tools/ide/commandlet/RepositoryCommandlet.java
salimbouch: Done. also solved bug of `git clone` not working, basically changed `git clone URL#BRANCH` to `git clone URL` then git `checkout BRANCH`. GitUrl which was responsible for parsing the combined string back into URL and branch is now kind of useless, imho we can just delete it.

hohwille: Thanks @salimbouch: You brought this to perfection now 👍

Issue #131, opened 2023-11-07
Support for tool dependencies


aBega2000:
As a devonfw-IDEasy user, I want to have tools being setup with their correct dependencies so that it just works and I do not have to worry.

E.g. several tools have java as dependency:

- mvn
- eclipse
- intellij
- jenkins
- sonar
- tomcat #36 
- ...

However, they may all require different versions of java ending up in an inconsistent situation that **devonfw-ide** currently cannot resolve. As a solution I see the following ideas:

- In ide-urls in each tool/edition version folder we will introduce a new (optional) **dependencies** file.

- This file will contain a list of 1-n tools with their required version or version range (similar to a dependency in maven, gradle, or npm).

- We can still discuss if that should be in JSON format (and then being named **dependencies.json**) or just a plain file with one dependency per line (TBD)

- If no such file is present, there are no dependencies required. Otherwise these dependencies have to be setup as well to launch the tool.

- In order to avoid conflicts between different versions of the same tool, we make [the shared software repository](https://github.com/devonfw/ide/blob/master/documentation/software.asciidoc#shared) the new default.

- The default java version for the build (e.g. for maven or gradle) will still be found in **software/java** but as a symlink to the configured JAVA_VERSION.

- However tools like eclipse/intellij/jenkins/sonar/... can require their own Java version via the **dependencies** file. If the default Java (JAVA_VERSION) does not match, then an additional java version is found or installed in the shared software repository and used instead by setting JAVA_HOME to it and adding it to the beginning of the PATH but only locally in the commandlet before launching the java based tool.

- We will analyze if we can find ways to automatically determine **dependencies** for specific tool versions. Where not possible, we will simply put the latest default into the according UrlUpdater class for that tool/edition and have to change it whenever a new version comes out that has different dependencies. If we notice too late (what might be the regular case), we have to change existing dependency version file(s) manually and also change the UrlUpdater then for future versions.

- All existing dependencies will need to be added via manual crafting what can be quite some effort.

aBega2000:
3 Options how a Json file could look like to match the versions between tomcat and Java:

Option 1:
{
  "tomcat 7*": "Java 7*",
  "tomcat 8*": "Java 7*",
  "tomcat 9*": "Java 8*",
  "tomcat 10*": "Java 11*",
  "tomcat 11*": "Java 21*"
}

Option 2 (nested Objects) :
{
  "Tomcat 7": {
    "javaVersion": "Java 7"
  },
  
[...]

PR #145, opened 2023-11-20
#131: Support for tool dependencies

aBega2000:
Fixes: #131  

This is the part of the solution that is implemented in the ToolCommandlet. It automatically detects if there are dependencies to be installed, and installs them in the Repository.

The Json file to be read has the following format: 

{
  "dependencies": {
    "(10.0.27,10.1.18]": [
      {
        "tool": "java",
        "versionRange": "[11,21_35]"
      }
    ],
    "(9.0.85,10.0.27]": [
      {
        "tool": "java",
        "versionRange": "[8,11)"
      }
    ]
  }
}

A guide to how to create a Json file for dependencies will also be written.


cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -1,10 +1,13 @@
 package com.devonfw.tools.ide.tool;
 
+import java.io.File;
```
jan-vcapgemini, line 3: File?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -388,4 +393,175 @@ public void setVersion(VersionIdentifier version, boolean hint) {
...
+   *
+   * @param toolVersionToCheck the {@link String} of the version of the tool that should be checked.
+   * @return the Version of the dependency as {@link String}, for the tool that was to be checked or
+   * {@code null} if not found.
+   */
+  protected String getDependencyVersion(String toolVersionToCheck) {
+
+    try {
+      ObjectMapper objectMapper = new ObjectMapper();
+      JsonNode tomcatVersions = objectMapper.readTree(new File(getDependencyJsonPath()));
```
jan-vcapgemini, line 425: tomcatVersions?, shouldn't this be more generic?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -388,4 +393,175 @@ public void setVersion(VersionIdentifier version, boolean hint) {
...
+
+      String requiredDependencyVersion = findDependencyVersionFromJson(tomcatVersions, toolVersionToCheck);
+
+      if (requiredDependencyVersion != null) {
+        this.context.info(getName() + " version " + toolVersionToCheck +
```
jan-vcapgemini, line 430: Please use this syntax: `this.context.info("{} version {} requires at least {} version {}", getName(), toolVersionToCheck, getDependencyName(), requiredDependencyVersion);`
And maybe adjust the info message a bit.
String concatenations in loggers can be resource intensive.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -388,4 +393,175 @@ public void setVersion(VersionIdentifier version, boolean hint) {
...
+            " requires at least " + getDependencyName() + " version " + requiredDependencyVersion);
+        return requiredDependencyVersion;
+      } else {
+        this.context.info("No specific " + getDependencyName() + " version requirement found for "+ getName()
+            + " version " + toolVersionToCheck);
+        return null;
+      }
+    } catch (IOException e) {
+      e.printStackTrace();
```
jan-vcapgemini, line 439: This needs to throw or log instead. Please make sure that the default printStackTrace always gets replaced.
Also make sure that the exception is added to the log message.
e.g. `this.context.warning("An error occured", e);`

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -388,4 +393,175 @@ public void setVersion(VersionIdentifier version, boolean hint) {
...
+   * @return {@link String} the Version of the dependency with which the tool works correctly.
+   */
+  private String findDependencyVersionFromJson(JsonNode toolVersions, String toolVersionToCheck) {
+    String requiredDependencyVersion = null;
+
+    // Iterate through the fields of the Json file
+    Iterator<Map.Entry<String, JsonNode>> fields = toolVersions.fields();
+    int[] targetVersion = parseVersion(toolVersionToCheck);
+
+    OuterLoop:
```
jan-vcapgemini, line 506: ?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -388,4 +393,175 @@ public void setVersion(VersionIdentifier version, boolean hint) {
...
+
+      Map.Entry<String, JsonNode> entry = fields.next();
+      String versionKey = entry.getKey();
+      int[] foundToolVersion = parseVersion(versionKey); // Found version when iterating the Json file
+
+      if (isEqualOrLess(foundToolVersion, targetVersion)) {
+        JsonNode dependencies = entry.getValue();
+        for (JsonNode dependencyNode : dependencies) {
+          if (getDependencyName().equals(dependencyNode.get("dependency").asText())) {
+            requiredDependencyVersion = dependencyNode.get("MinVersion").asText();
```
jan-vcapgemini, line 517: Maybe move `dependency` and `MinVersion` to a constant.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -388,4 +393,175 @@ public void setVersion(VersionIdentifier version, boolean hint) {
...
+  private int[] parseVersion(String versionString) {
+
+    String[] versionComponents = versionString.split("\\.");
+
+    int[] version = new int[versionComponents.length];
+    for (int j = 0; j < versionComponents.length; j++) {
+      version[j] = Integer.parseInt(versionComponents[j]);
+    }
+    return version;
+  }
```
jan-vcapgemini, line 544: Couldn't this be done with `VersionIdentifier`?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -388,4 +393,175 @@ public void setVersion(VersionIdentifier version, boolean hint) {
...
+
+    for (int i = 0; i < Math.min(currentToolVersion.length, targetVersion.length); i++) {
+      if (currentToolVersion[i] < targetVersion[i]) {
+        return true;
+      } else if (currentToolVersion[i] > targetVersion[i]) {
+        return false;
+      }
+    }
+    return true;
+  }
```
jan-vcapgemini, line 565: Couldn't this be done with `VersionObject` e.g. `compareTo` method?

jan-vcapgemini (changes requested):
Thanks for your PR.
I've found some parts which might be implemented in `VersionIdentifier` already.
It also looks like the auto reformat (imports) was not applied properly.
Please check.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
aBega2000: Fixed. Now using Path instead.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
aBega2000: Since I started writing it with tomcat as example, made this mistake. Now fixed.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
aBega2000: Fixed and changed the info message a bit.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
aBega2000: Changed like discussed and not by using the labeled break

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
aBega2000: Fixed by deleting this method totally

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
aBega2000: Also fixed by using the VersionIdentifier and deleted this method completely.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
aBega2000: Fixed

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
aBega2000: Fixed this, but still not sure if totally correctly

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +196,154 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+        VersionIdentifier dependencyVersionToInstall = findDependencyVersionToInstall(dependencyVersionNumberFound,
+            dependency);
+
+        String DefaultToolRepositoryID = this.context.getDefaultToolRepository().getId();
+        Path dependecyPath = this.context.getSoftwareRepositoryPath().resolve(DefaultToolRepositoryID)
+            .resolve(dependency).resolve(dependencyTool.getEdition()).resolve(dependencyVersionToInstall.toString());
+
+        if (Files.exists(dependecyPath)) {
+          this.context.info("The version {} of the dependency {} is already installed in repository",
+              dependencyTool.getInstalledVersion(dependecyPath), dependency);
```
jan-vcapgemini, line 258: Small typo:
```suggestion
        Path dependencyPath = this.context.getSoftwareRepositoryPath().resolve(DefaultToolRepositoryID)
            .resolve(dependency).resolve(dependencyTool.getEdition()).resolve(dependencyVersionToInstall.toString());

        if (Files.exists(dependencyPath)) {
          this.context.info("The version {} of the dependency {} is already installed in repository",
              dependencyTool.getInstalledVersion(dependencyPath), dependency);
```

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +196,154 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+   */
+  private Map.Entry<String, JsonNode> findDependenciesFromJson(JsonNode toolVersions,
+      VersionIdentifier toolVersionToCheck) {
+
+    // Iterate through the fields of the Json file
+    Iterator<Map.Entry<String, JsonNode>> fields = toolVersions.fields();
+    while (fields.hasNext()) {
+      Map.Entry<String, JsonNode> entry = fields.next();
+      String versionKey = entry.getKey();
+      VersionIdentifier foundToolVersion = VersionIdentifier.of(versionKey);
```
jan-vcapgemini, line 316: We should use a model e.g. StatusJson and map this to the json file instead I think.

jan-vcapgemini (changes requested):
Thanks for your adjustments.
Please check my requested changes. I think we should use the json mapper with a model instead.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +196,154 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+  /**
+   * Method to get the Path of the dependencies Json file
+   * 
+   * @return the {@link Path} of the dependencies file for the tool
+   */
+  protected Path getDependencyJsonPath() {
+
+    Path path = this.context.getUrlsPath();
+    Path toolPath = path.resolve(getName()).resolve(getEdition());
+    return toolPath.resolve("dependencies.Json");
```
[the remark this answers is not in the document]
aBega2000: Done!

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +196,154 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+    String dependencyEdition = this.context.getVariables().getToolEdition(dependency);
+
+    List<VersionIdentifier> versions = this.context.getUrls().getSortedVersions(dependency, dependencyEdition);
+
+    for (VersionIdentifier vi : versions) {
+      if (vi.compareVersion(dependencyVersionNumberFound).isGreater()) {
+        return vi;
+      }
+    }
+    return null;
```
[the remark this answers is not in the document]
aBega2000: Actually, by doing this change, also the flow of the program changes a bit. So basically like it is, it returns the first version that is greater than the VersionFound. For example if the version found is 8, it can find 21 and return it since it is greater than 8. But I changed it like you have suggested, and now the flow of the program is as follows: if the version found is 8.0.1, the loop goes through all the versions that are greater than 8.0.1, and returns the one that is just one level (Version) higher, so that it's still higher than the version found but nearly the same. Please let me know what you think of this change.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +196,154 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+          dependencyTool.setVersion(dependencyVersionToInstall, false);
+          this.context.info("The version {} of the dependency {} is being installed", dependencyVersionToInstall,
+              dependency);
+          LocalToolCommandlet dependencyLocal = (LocalToolCommandlet) dependencyTool;
+          dependencyLocal.installInRepo(dependencyVersionToInstall);
+          this.context.info("The version {} of the dependency {} was successfully installed",
+              dependencyVersionToInstall, dependency);
+        }
+      }
+    } catch (NullPointerException e) {
```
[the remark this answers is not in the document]
aBega2000: Fixed!

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +196,154 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+      ObjectMapper objectMapper = new ObjectMapper();
+      Path jsonFilePath = getDependencyJsonPath();
+
+      JsonNode toolVersions = objectMapper.readTree(jsonFilePath.toFile());
+      Map.Entry<String, JsonNode> entry = findDependenciesFromJson(toolVersions, getInstalledVersion());
+
+      if (entry == null) {
+        this.context.error("The entry with the specified version was not found in the Json file");
+      } else
+        return entry.getValue();
```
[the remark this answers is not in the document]
aBega2000: Fixed!

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
aBega2000: Fixed!

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
aBega2000: Also tried fixing this. Now I read the Json file by using the MAPPER.readValue(). And then I set the result to a JsonNode since I made my program work with JsonNode. Please let me know if these changes are sufficient.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +196,154 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+  }
+
+  /**
+   * Method to install the dependency
+   */
+  protected void installDependency() {
+
+    JsonNode nodeInJson = readJson();
+    String dependencyString = "dependency";
+    String MinVersionString = "MinVersion";
```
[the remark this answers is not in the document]
jan-vcapgemini: As discussed, all strings should be moved to a class constant for easy access.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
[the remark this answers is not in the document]
jan-vcapgemini: Just some background information: A NPE is a checked exception and should never be catched / ignored.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -4,21 +4,35 @@
...
+import java.util.stream.Stream;
 
 /**
  * {@link ToolCommandlet} that is installed locally into the IDE.
  */
 public abstract class LocalToolCommandlet extends ToolCommandlet {
 
+  private static final ObjectMapper MAPPER = JsonMapping.create();
+
+  private static final String dependencyFileName = "dependencies.json";
```
jan-vcapgemini, line 34: constants need to be in upper case:
```suggestion
  private static final String DEPENDENCY_FILENAME = "dependencies.json";
```

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +201,128 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+  /**
+   * Method to get the Path of the dependencies Json file
+   * 
+   * @return the {@link Path} of the dependencies file for the tool
+   */
+  private Path getDependencyJsonPath() {
+
+    Path URLspath = this.context.getUrlsPath();
+    Path toolPath = URLspath.resolve(getName()).resolve(getEdition());
+    return toolPath.resolve(dependencyFileName);
```
jan-vcapgemini, line 213:
```suggestion
    return toolPath.resolve(DEPENDENCY_FILENAME);
```

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +201,128 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+        } else {
+          this.context.info("The version {} of the dependency {} is being installed", dependencyVersionToInstall,
+              dependencyName);
+          LocalToolCommandlet dependencyLocal = (LocalToolCommandlet) dependencyTool;
+          dependencyLocal.installInRepo(dependencyVersionToInstall);
+          this.context.info("The version {} of the dependency {} was successfully installed",
+              dependencyVersionToInstall, dependencyName);
+        }
+      }
+    } catch (Exception e) {
```
jan-vcapgemini, line 244: I'm not sure if we should catch all exceptions here, would be better to catch only exceptions which are to be expected.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +201,128 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+  }
+
+  private List<DependencyInfo> readJson() {
+
+    Path dependencyJsonPath = getDependencyJsonPath();
+
+    try (BufferedReader reader = Files.newBufferedReader(dependencyJsonPath)) {
+      DependencyJson dependencyJson = MAPPER.readValue(reader, DependencyJson.class);
+      return findDependenciesFromJson(dependencyJson.getDependencies(), getConfiguredVersion());
+    } catch (Exception e) {
```
jan-vcapgemini, line 256: I'm not sure if we should catch all exceptions here, would be better to catch only exceptions which are to be expected e.g. `JsonMappingException`

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +201,128 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+
+    List<DependencyInfo> dependencies = readJson();
+
+    try {
+      for (DependencyInfo dependencyInfo : dependencies) {
+        VersionRange dependencyVersionRangeFound = VersionRange.of(dependencyInfo.getVersionRange());
+        String dependencyName = dependencyInfo.getDependency();
+        ToolCommandlet dependencyTool = this.context.getCommandletManager().getToolCommandlet(dependencyName);
+        VersionIdentifier dependencyVersionToInstall = findDependencyVersionToInstall(dependencyVersionRangeFound,
+            dependencyName);
```
jan-vcapgemini, line 226: `findDependencyVersionToInstall` can be null, so better check this first.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -63,6 +76,12 @@ protected boolean doInstall(boolean silent) {
     // install configured version of our tool in the software repository if not already installed
     ToolInstallation installation = installInRepo(configuredVersion);
 
+    if (Files.exists(getDependencyJsonPath())) {
+      installDependency();
+    } else {
+      this.context.info("No Dependencies file found");
```
jan-vcapgemini, line 82: Maybe add more info for the user about what he has to do now.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +201,128 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
 
+  /**
+   * Method to get the Path of the dependencies Json file
+   * 
+   * @return the {@link Path} of the dependencies file for the tool
+   */
+  private Path getDependencyJsonPath() {
+
+    Path URLspath = this.context.getUrlsPath();
+    Path toolPath = URLspath.resolve(getName()).resolve(getEdition());
```
jan-vcapgemini, line 212:
```suggestion
    Path urlsPath = this.context.getUrlsPath();
    Path toolPath = urlsPath.resolve(getName()).resolve(getEdition());
```

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +201,128 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+    try {
+      for (DependencyInfo dependencyInfo : dependencies) {
+        VersionRange dependencyVersionRangeFound = VersionRange.of(dependencyInfo.getVersionRange());
+        String dependencyName = dependencyInfo.getDependency();
+        ToolCommandlet dependencyTool = this.context.getCommandletManager().getToolCommandlet(dependencyName);
+        VersionIdentifier dependencyVersionToInstall = findDependencyVersionToInstall(dependencyVersionRangeFound,
+            dependencyName);
+
+        String DefaultToolRepositoryID = this.context.getDefaultToolRepository().getId();
+        Path dependencyRepository = this.context.getSoftwareRepositoryPath().resolve(DefaultToolRepositoryID)
```
jan-vcapgemini, line 229:
```suggestion
        String defaultToolRepositoryId = this.context.getDefaultToolRepository().getId();
        Path dependencyRepository = this.context.getSoftwareRepositoryPath().resolve(defaultToolRepositoryId)
```

jan-vcapgemini (changes requested):
Thanks for your changes. I've added some small suggestions and comments.
I would also suggest to add some tests for your methods e.g. `readJson`, `findDependencyVersionToInstall` etc.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
aBega2000: Fixed

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
aBega2000: Fixed

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
aBega2000: Fixed. Now catching only IOException and throwing RuntimeException

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
aBega2000: I wrote a check for this, please let me know if it's correct

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
aBega2000: Actually, the user needs to do nothing, if no dependency file was found, it means the tool doesn't have any dependencies to be installed.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
aBega2000: Fixed

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
aBega2000: Fixed

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
jan-vcapgemini: If this message is not relevant to the user, I would move it to trace instead.
As you're already showing infos about installed dependencies, this message is not needed I think.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
jan-vcapgemini: Shouldn't the method stop then and do nothing?
The error message won't help I think.

hohwille: Yes, @jan-vcapgemini is correct. You need to skip the dependency via `continue` if `null` or throw an exception.
If a dependency is actually required but cannot be found that is indeed an error. I am unsure if you should simply continue. At least we should ask the user to continue because most probably the result will be inconsistent if the dependency is missing and support will get much harder if we have to deal with follow-up issues because a tool is not working later because a missing dependency was skipped.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
aBega2000: Fixed, moved to trace.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
aBega2000: I changed it to a continue, so that the code continues to the next dependency tool, if the version for the actual dependency tool is not found.

jan-vcapgemini: We should create a new issue for @hohwille suggestions. I will resolve this CR for now.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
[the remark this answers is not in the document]
jan-vcapgemini: This needs some unit tests I think but I will resolve this for now.

jan-vcapgemini:
As Coveralls suggests, we need to add some unit tests for these additions.

cli/src/main/java/com/devonfw/tools/ide/url/model/file/dependencyJson/DependencyJson.java
```diff
@@ -0,0 +1,19 @@
+package com.devonfw.tools.ide.url.model.file.dependencyJson;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DependencyJson {
```
hohwille, line 7: Can you please add JavaDoc to public types and methods?

cli/src/main/java/com/devonfw/tools/ide/url/model/file/dependencyJson/DependencyInfo.java
```diff
@@ -0,0 +1,15 @@
+package com.devonfw.tools.ide.url.model.file.dependencyJson;
+
+public final class DependencyInfo {
```
hohwille, line 3: Please add JavaDoc to type and getters.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +202,125 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+  private List<DependencyInfo> findDependenciesFromJson(Map<String, List<DependencyInfo>> dependencies,
+      VersionIdentifier toolVersionToCheck) {
+
+    for (Map.Entry<String, List<DependencyInfo>> map : dependencies.entrySet()) {
+
+      String versionKey = map.getKey();
+      VersionIdentifier foundToolVersion = VersionIdentifier.of(versionKey);
+
+      // if a newer (greater) version is available, that is not already in the Json file
+      if (toolVersionToCheck.getStart().compareVersion(foundToolVersion.getStart()).isGreater()) {
```
hohwille, line 318: why do we only compare the start segments of the versions? IMHO this is incorrect.

Woudn't it make sense that we use `VersionRange` that we have meanwhile created and ensure it supports single version and range as well or create a generic version parser that can return some `VersionMatcher` interface both implemented by `VersionIdentifier` and `VersionRange` exposing the `contains(VersionIdentifier)` method?

cli/src/main/java/com/devonfw/tools/ide/url/model/file/dependencyJson/DependencyJson.java
```diff
@@ -0,0 +1,19 @@
...
+  private final Map<String, List<DependencyInfo>> dependencies;
+
+  public DependencyJson() {
+
+    this.dependencies = new LinkedHashMap<>();
+  }
+
+  public Map<String, List<DependencyInfo>> getDependencies() {
```
hohwille, line 15: In general it is better to keep such structures internally to simplify the JSON mapping with Jackson but for the public API instead provide rather something like `public List<DependencyInfo> getDependencies(VersionIdentifier toolVersion)` or if you want to keep logic out of the data objects and keep them dump and stupid `public List<ToolVersionDependencyInfo> getDependencies()`. Looking at this method some new member in the team will not have a clue what the semantic of this map will be. What are the keys of type `String`? Truth is they are `VersionIdentifiers` (or are they patterns or may they also be version ranges)?

cli/src/main/java/com/devonfw/tools/ide/url/model/file/dependencyJson/DependencyInfo.java
```diff
@@ -0,0 +1,15 @@
...
+  private String dependency;
+  private String versionRange;
+
+  public String getDependency() {
+    return dependency;
+  }
```
hohwille, line 9: should we rename this to `tool`?

Reading the code reveals more semantics that should ideally already be said by the API itself:
```
String dependencyName = dependencyInfo.getDependency();
````

File: cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java [hunk 29]
```diff
@@ -182,4 +202,125 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+   * @return the {@link Path} of the dependencies file for the tool
+   */
+  private Path getDependencyJsonPath() {
+
+    Path urlsPath = this.context.getUrlsPath();
+    Path toolPath = urlsPath.resolve(getName()).resolve(getEdition());
+    return toolPath.resolve(DEPENDENCY_FILENAME);
+  }
+
+  private void installDependency() {
```
hohwille, line 217: Refactoring/rename suggestion:
```suggestion
  private void installDependencies() {
```

cli/src/main/java/com/devonfw/tools/ide/url/model/file/dependencyJson/DependencyInfo.java
```diff
@@ -0,0 +1,15 @@
...
+
+  public String getVersionRange() {
```
hohwille, line 11: This should be `VersionRange` instead of `String` to make the API expressive.

hohwille (changes requested):
@aBega2000 thanks for this PR. You did a good job to implement this rather advanced story and get the pieces together. 👍 
As @jan-vcapgemini already requested, you need to add JUnit tests that cover your new code. The tests also help to see the feature in action as then we can look at concrete examples with JSON files used by the test.
Also there are merge issues that have to be addressed (sorry for the review delay).
Please also address the review comments.

cli/src/main/java/com/devonfw/tools/ide/url/model/file/dependencyJson/DependencyJson.java
aBega2000: Done

cli/src/main/java/com/devonfw/tools/ide/url/model/file/dependencyJson/DependencyInfo.java
aBega2000: Done

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
aBega2000: I changed this by changing the Json file a bit. Now the Json file looks like this: 

"dependencies": {
    "(10.0.27,10.1.18]": [
      {
        "tool": "java",
        "versionRange": "[11,21_35]"
      }
    ]
  }
  
  This change allowed me to use VersionRange.contains() to further simplify the code I had written in this method. Now the method cheks if the foundToolVersionRange contains the toolVersionToCheck, and only if none of the Version Ranges of the tool in the Json file contain it, is null returned. To be honest I still need to write the tests to make sure that this works correctly in all the cases.

cli/src/main/java/com/devonfw/tools/ide/url/model/file/dependencyJson/DependencyJson.java
aBega2000: In my Opinion it also makes sense to only change the String to VersionRange, and that's what I've done. Please let me know if this really makes sense.

cli/src/main/java/com/devonfw/tools/ide/url/model/file/dependencyJson/DependencyInfo.java
aBega2000: I have renamed it to tool, but I actually preferred dependency, since in all this issue I have tried to keep the word "tool" for the tool to be installed centrally, and dependency for the tool to be installed as a dependency. Please let me know what you think of this.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
aBega2000: Done

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -182,4 +202,125 @@ private boolean isInstalledVersion(VersionIdentifier expectedVersion, VersionIde
...
+   * Method to search the List of versions available in the ide and find the right version to install
+   *
+   * @param dependencyVersionRangeFound the {@link VersionRange} of the dependency that was found that needs to be
+   *        installed
+   * @param dependency the {@link String} of the dependency tool
+   *
+   * @return {@link VersionIdentifier} of the dependency that is to be installed
+   */
+  private VersionIdentifier findDependencyVersionToInstall(VersionRange dependencyVersionRangeFound,
+      String dependency) {
```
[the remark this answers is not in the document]
aBega2000: Done

cli/src/main/java/com/devonfw/tools/ide/url/model/file/dependencyJson/DependencyJson.java
hohwille: Lets agree on this `Map<String, List<DependencyInfo>>` as it makes the JSON structure simpler.
But we can omit the root object (delete `DependencyJson`) and you can directly deserialize to `Map`:
https://www.baeldung.com/jackson-map#bd-3-mapltobjectobjectgt-deserialization

hohwille:
> As @jan-vcapgemini already requested, you need to add JUnit tests that cover your new code. The tests also help to see the feature in action as then we can look at concrete examples with JSON files used by the test.
Also there are merge issues that have to be addressed (sorry for the review delay).
Please also address the review comments.

For the record. JUnit test can be found in PR #250

hohwille:
All changes had also been included in PR #250 that has already been merged.
Hence, this PR is obsolete and can be closed.

PR #147, opened 2023-11-22
#104: proper evaluation of variables

### LINKED ISSUE #104: Proper evaluation of variables
In `EnvironmentVariables` we have already implemented quite some complexity to support the inheritance of `ide.properties` from different sources and also the resolution of variables as `${«variable-name»}` within variables as we can see here:
https://github.com/devonfw/IDEasy/blob/11d14e77a6b1a286f7af87d014c05cfd2c3bb121/cli/src/test/java/com/devonfw/tools/ide/context/IdeContextTest.java#L33-L34

However, a common pattern in scripting with variables is something like this:
```
PATH=${PATH};C:\Tools\extra-tool
```
Or
```
MAVEN_ARGS=${MAVEN_ARGS} --extra-option
```
The expectation would be that in this case the variable expression `${MAVEN_ARGS}` would resolve to the value of `MAVEN_ARGS` from the parent `EnvironmentVariables`. This would need to be done in a loop or recursively to allow this feature on all levels of `ide.properties`.
Resolving such variable expressions is currently implemented [here](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesResolved.java).

We should first write a test (or extend `IdeContextTest`) that adds some variable to all the `ide.properties` files of the test project:
```
TOOL_ARGS=${TOOL_ARGS} «env-type»
```
Where «env-type» is the [environment-type](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesType.java).
However, the type `system` refers to the global environment variables and needs to be ignored here.

The result of this `TOOL_ARGS` variable when evaluated should be asserted as `" user settings workspace conf"`.
I would expect this test to fail.
Then the implementation needs to be fixed such that the test gets green.
Please note that if `${TOOL_ARGS}` gets resolved in type `user` and the parent (`system`) does not define it then it should resolve to the empty string (Currently we keep variable expressions as `${«variable-name»}` if its value is undefined).

MattesMrzik:
closes #104.
Environment vars like VAR=${VAR} can now be resolved. Added a test for it as well.


cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -140,7 +140,7 @@ protected void collectVariables(Set<String> variables) {
 
   /**
    * @param propertiesFilePath the {@link #getPropertiesFilePath() propertiesFilePath} of the child
-   *        {@link EnvironmentVariables}.
```
MattesMrzik, line 143: I am note sure what the correct code style is. I think I loaded the `settings/eclipse/ ide-code-style.xml` corretly. So does this style comply with the agreed upon?

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -149,9 +149,25 @@ public AbstractEnvironmentVariables extend(Path propertiesFilePath, EnvironmentV
     return new EnvironmentVariablesPropertiesFile(this, type, propertiesFilePath, this.context);
   }
 
+  private EnvironmentVariablesType incrementType(EnvironmentVariablesType type) {
+
+    EnvironmentVariables current = this;
+    while (current.getType() != type) {
+      current = current.getParent();
+      if (current == null) {
+        throw new IllegalStateException("Could not find type " + type + " in " + this);
```
hohwille, line 158: can this happen? If so it will break our variable resolution and therefore prevent the user from starting the IDE causing a full blocker issue.
If we cannot resolve a variable it is better to leave it unresolved and e.g. log a warning but not fail hard.

MattesMrzik: Okey, I changed it.

alfeilex (commented):
Well done, I have no improvements to make

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariables.java
```diff
@@ -23,6 +23,18 @@ public interface EnvironmentVariables {
    * @param name the name of the environment variable to get.
```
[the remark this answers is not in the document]
MattesMrzik: added it.

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -161,16 +179,31 @@ public EnvironmentVariables resolved() {
   @Override
   public String resolve(String string, Object src) {
 
-    return resolve(string, src, 0, src, string);
+    return resolve(string, src, 0, src, string, EnvironmentVariablesType.CONF);
   }
 
-  private String resolve(String value, Object src, int recursion, Object rootSrc, String rootValue) {
+  public String resolve(String value, Object src, int recursion, Object rootSrc, String rootValue) {
```
hohwille, line 185:
```suggestion
  private String resolve(String value, Object src, int recursion, Object rootSrc, String rootValue) {
```

Why should we make this internal method public? And if you do, you should add JavaDocs.

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -149,6 +149,24 @@ public AbstractEnvironmentVariables extend(Path propertiesFilePath, EnvironmentV
...
+            + "be resolved correctly", rootSrc, type);
+        return null;
+      }
+    }
+    EnvironmentVariables parent = current.getParent();
+    if (parent == null) {
+      return null;
+    }
+    return parent.getType();
+  }
```
hohwille, line 168: I still do not get why you added this method.
Did you consider using the existing `getByType` method?
This seems to do almost the same thing and has a JavaDoc explaining what it does.
Then you could also call the `resolve` method on the returned `EnvironmentVariables` and also you would not need to extend the resolve method with the additional `EnvironmentVariablesType` parameter.
Does this make sense or am I missing something?
Does this make sense or

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -181,35 +214,56 @@ private String resolve(String value, Object src, int recursion, Object rootSrc,
     StringBuilder sb = new StringBuilder(value.length() + EXTRA_CAPACITY);
     do {
       String variableName = matcher.group(2);
-      String variableValue = getValue(variableName);
-      if (variableValue == null) {
-        this.context.warning("Undefined variable {} in '{}={}' for root '{}={}'", variableName, src, value, rootSrc,
-            rootValue);
+      String variableValue = "";
+      if (src.toString().equals(variableName)) {
```
hohwille, line 218: We can not rely on `src` this way if you read the contract (JavaDoc) of the `resolve` method because `src` can be anything and is only for debugging and error reporting.

hohwille (commented):
@MattesMrzik thanks for this PR. Looking at your test, you did a great job solving this story. 👍 Also thanks for fixing typos in existing code, etc.

For the implementation I have started to leave some review comments. IMHO it would be best, if we can have a short meeting together to discuss this further, as doing that async with github comment ping-pong does not seem efficient.

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
MattesMrzik: Thansk I didn't know this. I fixed it.

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
MattesMrzik: I changed the method, please have a look.

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
MattesMrzik: Thanks, I changed it.

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -161,36 +163,64 @@ public EnvironmentVariables resolved() {
...
+      String variableValue = resolvedVars.getValue(variableName);
       if (variableValue == null) {
         this.context.warning("Undefined variable {} in '{}={}' for root '{}={}'", variableName, src, value, rootSrc,
             rootValue);
-      } else {
-        String replacement = resolve(variableValue, variableName, recursion, rootSrc, rootValue);
+        continue;
+      }
+      EnvironmentVariables lowestFound = findVariable(variableName); // for DirectoryMergerTest this returned false when
+                                                                     // resolving IDE_HOME
```
hohwille, line 195:
```suggestion
      EnvironmentVariables lowestFound = findVariable(variableName);
```

Looks like a personal note. JUnit-test behavior should be documented in JUnit and not in productive code. At this point for other developers this is more confusing than helpful. Besides "false" does not make sense to me in this context at all, so I assume it is an outdated leftover.

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
hohwille: We are getting closer. However, IMHO `resolvedVars` is never changed. Hence, it is the same as `this`. Hence, it is a pointless argument that can be omitted making the code much better to read and understand.
I might be mistaken. This is a complex story and understanding the code in the browser is not trivial. We can also have a video-chat together to quickly clarify this. But we almost reached the finishing line for merging. Thanks for your great work solving this complex problem 👍

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariables.java
```diff
@@ -19,10 +19,6 @@ public interface EnvironmentVariables {
   /** Filename of the legacy variable configuration file. {@value} */
   String LEGACY_PROPERTIES = "devon.properties";
 
-  /**
-   * @param name the name of the environment variable to get.
-   * @return the value of the variable with the given {@code name}. Will be {@code null} if no such variable is defined.
-   */
   default String get(String name) {
```
hohwille, line 22: did you remove this JavaDoc by accident?

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -161,36 +163,64 @@ public EnvironmentVariables resolved() {
   @Override
   public String resolve(String string, Object src) {
 
-    return resolve(string, src, 0, src, string);
+    return resolve(string, src, 0, src, string, this);
   }
 
-  private String resolve(String value, Object src, int recursion, Object rootSrc, String rootValue) {
+  private String resolve(String value, Object src, int recursion, Object rootSrc, String rootValue,
+      AbstractEnvironmentVariables resolvedVars) {
```
hohwille, line 170: `resolvedVars` is passed as `this` but never changed. Recursive invocations also always pass the same instance so if I am not mistaken, this parameter is pointless.

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
hohwille: > String replacement = ((AbstractEnvironmentVariables) next).resolve(next.getFlat(variableName), variableName, recursion, rootSrc, rootValue, resolvedVars);

OK. Not easy to see. Maybe then we need some doc comment for other to understand this...

MattesMrzik: I added javadoc to explain how `resolvedVars` is used and why it is necessary.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariables.java
MattesMrzik: jep, thanks for noticing.

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
MattesMrzik: I added javadoc to explain how resolvedVars is used and why it is necessary.

hohwille (approved):
@MattesMrzik all great. Thanks for your PR 👍 
Will merge asap.

Issue #101, opened 2023-10-17
Implement ability to setup IDE instance


hohwille:
As a IDEasy user, I want to be able to setup my project with a single command so that I can automate my setup.
This is the core feature of IDEasy but it is not implemented yet. For devonfw-ide see here:
* https://github.com/devonfw/ide/blob/af67497e06fb82e2e50a4d9ffa10e3f9dda44aa1/scripts/src/main/resources/scripts/command/ide#L235
* https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/setup.bat
* https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/setup

NOTE: I am thinking of a different way to setup IDEasy compared to devonfw-ide. Therefore, there may only be a single installation of our CLI in `$IDE_ROOT` instead of multiple ones in each `$IDE_HOME`. With devonfw-ide we have this redundantly in `$IDE_HOME/scripts` per instance. Therefore the new setup to be implemented here should be designed like [devon ide create](https://github.com/devonfw/ide/blob/master/documentation/ide.asciidoc#create).
So a natural way to implement this story would be to add a `CreateCommandlet` so we can have the same CLI API without the `devon` as prefix.

ATTENTION: This issue is blocked by #100 and it is closely related to #102

PR #150, opened 2023-11-28
#102: implement update commandlet

### LINKED ISSUE #102: Implement ability to update IDE instance
As a IDEasy user, I want to be able to update my current project (`$IDE_HOME`) so that I am in sync with my dev-team and all tools and configs are up-to-date.

In devonfw-ide this corresponds to the command `devon ide update`. 
https://github.com/devonfw/ide/blob/af67497e06fb82e2e50a4d9ffa10e3f9dda44aa1/scripts/src/main/resources/scripts/command/ide#L517-L522

It does the following things:
1. update settings (git pull)
2. update all tools that are already installed (in Java we could loop over all tool commandlets and delegate this task to each of them)
3. install/update all custom tools (`ide-custom-tools.json` use `context.getCustomToolRepository()`)
4. install/update all regular tools (`$IDE_TOOLS`)
5. update projects/repositories (lacking in devonfw-ide?) - see #100 

Here we shall also address proper error handling. That is if we iterate over tools in 2.,3., and 4. and some tool update fails with an error, we should not abort the entire process and prevent other tools to get updated. Instead in Java we can now properly catch and collect the errors and include a report at the end of the process.
In devonfw-ide we had issues with such error handling and could only implement a [hack](https://github.com/devonfw/ide/commit/84d675da8e66ffd730add90b8b91d82cad0b09ae).

It would be natural to support this feature via `ide update` and therefore create an `UpdateCommandlet`.
What needs to be discussed is that there is also the old `devon ide update scripts` command that updates our own software product. It would be somewhat natural to invoke this via `ide update` as it updates the ide product that is behind the `ide` command (IDEasy).

ATTENTION: This story is closely related to #101

salimbouch:
closes #102


cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+  }
+
+  private void updateSoftware() {
+
+    Set<ToolCommandlet> toolCommandlets = new HashSet<>();
+    Map<String, String> tool2Exception = new HashMap<>();
+
+    // installed tools in IDE_HOME/software
+    List<Path> softwares = this.context.getFileAccess().getFilesInDir(this.context.getSoftwarePath(), p -> true);
```
aBega2000, line 64: I think it's better not to use single letter variables, like "p" in this case. Maybe something more meaningful.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,110 @@
...
+      this.context.success("Successfully updated settings repository.");
+    } else {
+      throw new IllegalStateException("Cannot find settings repository.");
+    }
+  }
+
+  private void updateSoftware() {
+
+    Set<ToolCommandlet> toolCommandlets = new HashSet<>();
+    Map<String, String> tool2Exception = new HashMap<>();
```
aBega2000, line 61: Maybe the name tool2Exception is not totally clear. As I can see this stands for the tools that couldn't be installed successfully, so maybe "toolUnsuccessful" or "toolNotInstalled", or something similar in the context.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -458,4 +458,24 @@ private Path findFirstRecursive(Path dir, Predicate<Path> filter, boolean recurs
     return null;
   }
 
+  @Override
+  public List<Path> getFilesInDir(Path dir, Predicate<Path> filter) {
+    List<Path> files = null;
```
aBega2000, line 463: Can't "files" be directly set to new ArrayList<>()? As it will always be null in the beggining, so why check if it's null and then set it to the ArrayList?

aBega2000 (commented):
Just some suggestions about naming of two variables and the setting of a variable

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
salimbouch: True, changed to path.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
salimbouch: It makes sense to return an empty arraylist instead of null in case there are no paths are satisfy the predicate, thanks for pointing that out :)

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
salimbouch: I moved the initialization of this variable close where it's used and I think now it is easier to know what it is for :)

aBega2000 (approved):

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,109 @@
...
+
+    this.context.info("Updating settings repository ...");
+    Path settingsPath = this.context.getSettingsPath();
+    if (Files.isDirectory(settingsPath)) {
+      // perform git pull on the settings repo
+      this.context.gitPullOrClone(settingsPath, SETTINGS_REPO_URL);
+      this.context.success("Successfully updated settings repository.");
+    } else {
+      throw new IllegalStateException("Cannot find settings repository.");
+    }
```
hohwille, line 55: If the `settings` are not already cloned, you would need to ask the user for the settings git URL. This will allow to reuse the same also for creation of a new project.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  private void updateSoftware() {
+
+    Set<ToolCommandlet> toolCommandlets = new HashSet<>();
+
+    // installed tools in IDE_HOME/software
+    List<Path> softwares = this.context.getFileAccess().getFilesInDir(this.context.getSoftwarePath(), path -> true);
```
hohwille, line 63:
```suggestion
    List<Path> softwares = this.context.getFileAccess().getFilesInDir(this.context.getSoftwarePath(), Files::isDirectory);
```

It might make sense to add a variant of `getFilesInDir` method without the `Predicate` parameter that gives you all children. However, here you only want to get children that are directories.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,109 @@
...
+    for (Path software : softwares) {
+      String toolName = software.getFileName().toString();
+      try {
+        toolCommandlets.add(this.context.getCommandletManager().getToolCommandlet(toolName));
+      } catch (IllegalArgumentException e) {
+        //tool is a custom tool, ignore ...
+      }
```
hohwille, line 70: It is not good to use exceptions. If we need this here, we can add `CommandletManager.getToolCommandletOrNull(String)` so we do not have to throw and then catch exceptions if we then ignore the exception as creating exceptions is an expensive operation in Java so it should always be avoided in regular cases and only happen if something is really exceptional.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    // regular tools in $IDE_TOOLS
+    String regularToolsString = this.context.getVariables().get("IDE_TOOLS");
+    String[] regularTools = regularToolsString.split(",\\s");
```
hohwille, line 75:
```suggestion
    List<String> regularTools =  IdeVariables.IDE_TOOLS.get(this.context);
```

Please note that this is not only more elegant - it also includes the required backward compatibility for `DEVON_IDE_TOOLS`.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,109 @@
...
+        tool2exception.put(toolCommandlet.getName(), e.getMessage());
+      }
+    }
+
+    if (tool2exception.isEmpty()) {
+      this.context.success("All tools were successfully installed.");
+    } else {
+      this.context.warning("Following tools could not be installed:");
+      for (String toolName : tool2exception.keySet())
+      this.context.warning("{}: {}", toolName, tool2exception.get(toolName));
```
hohwille, line 102: always use curly braces for code-blocks - omitting them is considered an anti-pattern in Java and makes the code harder to read.
Further, it is great that you do not make the update stop at the first error but continue as I always wanted to have this already for devonfw-ide (where it fails fast unfortunately), but if we have an error, we should not end with status code 0 (success) at the end. So most probably we should instead create some generic container that allows to start steps and also end them with either success or an error (`Throwable`). That container could then do all the step-logging including the success and error logging plus a method to complete at the end that will log a summary at the end and if not everything was successful raise a CliException (e.g. with message like `1 step(s) failed (10%) and 9 step(s) succeeded (90%) out of 10 step(s)!`). This generic container could be reused for other situations that we will have in IDEasy where we run multiple steps.

hohwille:
This PR is more or less an extension of PR #143 that I just reviewed. So first we should get this merged before I continue with this PR.
However, I can already confirm that the changes in `CustomToolRepositoryImpl` look like a nice fix of bugs that I created writing untested code ;) Also I already left some review comments but I will do a final review once PR #143 has been merged and this PR was updated to that state.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
salimbouch: done, created suggested method.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    // update/install the toolCommandlets
+    Map<String, String> tool2exception = new HashMap<>();
+    for (ToolCommandlet toolCommandlet : toolCommandlets) {
+      try {
+        this.context.step("Setting up {}", toolCommandlet.getName());
+        toolCommandlet.install(false);
+      } catch (Exception e) {
+        tool2exception.put(toolCommandlet.getName(), e.getMessage());
```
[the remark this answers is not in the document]
salimbouch: done, by the generic container, which logs the error and gives summary at the end

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
salimbouch: done, implemented a genertic container class.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,167 @@
...
+  public String getName() {
+
+    return "update";
+  }
+
+  @Override
+  public void run() {
+
+    updateSettings();
+    this.context.getFileAccess().mkdirs(this.context.getWorkspacePath());
```
hohwille, line 48: You should already create the main workspace:
```suggestion
    this.context.getFileAccess().mkdirs(this.context.getWorkspacePath().resolve(IdeContext.WORKSPACE_MAIN));
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,167 @@
...
+import com.devonfw.tools.ide.property.StringProperty;
+import com.devonfw.tools.ide.repo.CustomTool;
+import com.devonfw.tools.ide.tool.CustomToolCommandlet;
+import com.devonfw.tools.ide.tool.ToolCommandlet;
+import com.devonfw.tools.ide.variable.IdeVariables;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.*;
```
hohwille, line 14: avoid start-imports. Just call `devon ide update` before you launch your IDE the next time.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,167 @@
...
+
+    return "update";
+  }
+
+  @Override
+  public void run() {
+
+    updateSettings();
+    this.context.getFileAccess().mkdirs(this.context.getWorkspacePath());
+    setupConf(this.context.getSettingsPath().resolve("devon"), this.context.getIdeHome());
```
hohwille, line 49: Oh, I forgot about this one. With `IDEasy` we want to get rid of `devon` on the long run.
Therefore we should have a new folder name and use `devon` only as fallback if the new name is not present.
Typically we simply use `ide` instead of `devon`.
Here it would actually make sense use `templates` instead what would better explain what the folder is actually about.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,167 @@
...
+    updateSoftware();
+    updateRepositories();
+  }
+
+  private void setupConf(Path template, Path conf) {
+
+    try (Stream<Path> childStream = Files.list(template)) {
+      Iterator<Path> iterator = childStream.iterator();
+      while (iterator.hasNext()) {
+        Path child = iterator.next();
```
hohwille, line 59: meanwhile you could also use `this.context.getFileAccess().listChildren(template, f -> true)` to simplify.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,167 @@
...
+        String basename = child.getFileName().toString();
+
+        if (!basename.equals(".") && !basename.equals("..") && !basename.equals("*")) {
```
hohwille, line 62: This is a prefect "transcription" of what was implemented in bash.
But in Java, we now have proper ways to list the children of a directory without hacks.
In Java we will never get `.`, `..` or `*` back here unless there really exists a file called `*`.
So you can simply remove this `if` check.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,167 @@
...
+          Path confPath = conf.resolve(basename);
+
+          if (Files.isDirectory(child)) {
+            if (!Files.exists(confPath) || !Files.isDirectory(confPath)) {
```
hohwille, line 66:
```suggestion
            if (!Files.isDirectory(confPath)) {
```
This is sufficient as `isDirectory` will only return `true` if the path exists and is a directory.

hohwille: Technically you do not need the `if` at all since `mkdirs` will ensure the directory exists.
If it already exists, the method will be idempotent and do nothing.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,167 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.*;
+import java.util.stream.Stream;
+
+/**
+ * {@link Commandlet} to update settings, software and repositories
+ */
+public class UpdateCommandlet extends Commandlet {
```
hohwille, line 20: See PR #186: You will need an abstract base class for both `UpdateCommandlet` and `CreateCommandlet` that share lots of methods like `updateSettings`, `updateSoftware`, `setupConf`, etc.
IMHO your life will get easier if you combine both to a single PR after PR #143 is merged.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
salimbouch: this already includes the main workspace. [here](https://github.com/devonfw/IDEasy/blob/70ea7c4638d1d9452cf882d29047e5fc63ff98b3/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L180)

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
salimbouch: Once #143 is merged I will delete this PR and keep working on #186

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
salimbouch: @hohwille you think we should abort the whole updating process when there is no `templates` nor `devon` folder in settings? the odds of that happening are pretty slim since that would only happen if `git clone` of settings repo goes wrong which means we wouldn't get to `setupConf` in the first place. Right now I just `return` when no `templates` nor `devon` exists.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
salimbouch: Why would we want to use `updateSettings`, `updateSoftware`, `setupConf` for `CreateCommandlet?` I thought `CreateCommandlet` is for creating the new instance and making it recognizable as an instance then update basically sets up settings and all the rest.

salimbouch: I still don't get why we would need for `CreateCommandlet` to have `updateSettings`, `updateSoftware`, `setupConf `etc. 
If I call these methods in `CreateCommandlet `the `context` will be refering to the old instance and not the new, and changing the whole `context` is probably problematic. Can't we just create a folder with the necessary folders inside and then call `ide update`?

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
```diff
@@ -0,0 +1,169 @@
...
+          repository = DEFAULT_SETTINGS_REPO_URL;
+        } else {
+          this.context.info("Missing your settings at {} and no SETTINGS_URL is defined.", settingsPath);
+          this.context.info("Further details can be found here:");
+          this.context.info("https://github.com/devonfw/IDEasy/blob/main/documentation/settings.asciidoc");
+          this.context.info("Please contact the technical lead of your project to get the SETTINGS_URL for your project.");
+          this.context.info("In case you just want to test IDEasy you may simply hit return to install the default settings.");
+          this.context.info("");
+          this.context.info("Settings URL [{}]:", DEFAULT_SETTINGS_REPO_URL);
+          repository = this.context.readLine();
```
hohwille, line 112: We need to add a dedicated `question` method for this.
You need to be aware that we will later build a JavaFX GUI for IDEasy.
A method like `readLine()` does not make sense there as we need to render a popup containing the entire info text with an input field for the settings URL with the default value as placeholder.

cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
```diff
@@ -177,6 +181,11 @@ default boolean question(String question) {
   @SuppressWarnings("unchecked")
   <O> O question(String question, O... options);
 
+  /**
+   * @return the input from the end-user (e.g. read from the console).
+   */
+   String readLine();
```
hohwille, line 187: as already stated this method should not be part of the public API.

cli/src/test/resources/ide-projects/update/_ide/urls/java/java/17.0.6/linux_x64.urls
```diff
@@ -0,0 +1 @@
+http://localhost:1111/installTest/linux
```
hohwille, line 1: seems all like copy&paste stuff that is not needed anywhere in your tests that work entirely without wiremock.
So please omit the entire `urls` folder here if it is not needed. We will fully lose control of maintenance if we add tons of  clutter and nobody knows what is actually needed and relevant and why something is there...

cli/src/test/resources/ide-projects/update/project/settings/ide.properties
```diff
@@ -0,0 +1,22 @@
...
+BAR=bar-${SOME}
+
+TEST_ARGS1=${TEST_ARGS1} settings1
+TEST_ARGS4=${TEST_ARGS4} settings4
+TEST_ARGS5=${TEST_ARGS5} settings5
+TEST_ARGS6=${TEST_ARGS6} settings6
+TEST_ARGS7=${TEST_ARGS7} settings7
+TEST_ARGS8=settings8
+TEST_ARGS9=settings9
+TEST_ARGSb=${TEST_ARGS10} settingsb ${TEST_ARGSa} ${TEST_ARGSb}
```
hohwille, line 21: Please avoid windows line endings.
Also what are these `TEST_ARGS*` variables used for? Also looks like pure copy&paste.

cli/src/test/resources/ide-projects/update/project/settings/eclipse/plugins/anyedit.properties
```diff
@@ -0,0 +1,3 @@
+plugin_url=https://raw.githubusercontent.com/iloveeclipse/plugins/latest/
```
hohwille, line 1: these plugins are not used by the tests as far as I can see so only add what you really need.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * The constructor.
+   *
+   * @param context the {@link IdeContext}.
+   */
+  public CreateCommandlet(IdeContext context) {
+
+    super(context);
+    addKeyword(getName());
+    newInstance = add(new StringProperty("", false, "newInstance"));
```
hohwille, line 27: `newInstance` sounds rather technically and will not really help the end-user what this property/argument is about.
One could also guess wrong that this is a flag set to `true` or `false` configuring if a new instance is created or not...
After we have clarified that the git urls to clone into workspaces are called `repository` we also agreed that this here can now be called `project` without being ambiguous. I am not sure if you had already been on the old teams group where this discussion took place with the excel sheet asking how to uniquely name the things on the different leveles like `IDE_ROOT`, `IDE_HOME` and the sub-folders in the workspaces.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  @Override
+  public void run() {
+
+    String newInstanceName = newInstance.getValue();
+    Path newInstancePath;
+
+    if (newInstanceName == null) {
+      newInstancePath = this.context.getCwd();
```
hohwille, line 43: maybe this comes from the old bash code of devonfw-ide but IMHO this does not really make sense anymore.
Either we use `ide create «project-name»` or we `cd` to the project folder and use `ide update`.
I would simply make the project name a required property and avoid this case.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -0,0 +1,75 @@
...
+    } else {
+      newInstancePath = this.context.getIdeRoot().resolve(newInstanceName);
+      this.context.getFileAccess().mkdirs(newInstancePath);
+    }
+
+    this.context.info("Creating new IDEasy instance in {}", newInstancePath);
+    if (!this.context.getFileAccess().isEmptyDir(newInstancePath)) {
+      this.context.askToContinue("Directory is not empty, continue?");
+    }
```
hohwille, line 52:
```suggestion
    }

    this.context.info("Creating new IDEasy instance in {}", newInstancePath);
    if (this.context.getFileAccess().exists(newInstancePath)) {
      this.context.askToContinue("Directory " + newInstancePath + " already exists. Do you want to continue?");
    } else {
      this.context.getFileAccess().mkdirs(newInstancePath);
    }
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -0,0 +1,75 @@
...
+    } else {
+      this.context.warning("Could not create IDEasy Instance.");
+    }
+  }
+
+  private void initializeInstance(Path newInstancePath) {
+
+    FileAccess fileAccess = this.context.getFileAccess();
+    fileAccess.mkdirs(newInstancePath.resolve(IdeContext.FOLDER_SOFTWARE));
+    fileAccess.mkdirs(newInstancePath.resolve(IdeContext.FOLDER_UPDATES));
```
hohwille, line 69: This folder is now created automatically in `$IDE_ROOT/_ide/updates` and not anymore per IDE project:
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -0,0 +1,75 @@
...
+    fileAccess.mkdirs(newInstancePath.resolve(IdeContext.FOLDER_PLUGINS));
+    fileAccess.mkdirs(newInstancePath.resolve(IdeContext.FOLDER_WORKSPACES));
```
hohwille, line 71: If you create the workspaces folder here, then directly create the `main` workspace:
```suggestion
    fileAccess.mkdirs(newInstancePath.resolve(IdeContext.FOLDER_WORKSPACES).resolve(IdeContext.WORKSPACE_MAIN));
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -0,0 +1,75 @@
...
+    fileAccess.mkdirs(newInstancePath.resolve(IdeContext.FOLDER_SETTINGS));
+
```
hohwille, line 73: I would not create this folder as it shall be created by `git` and if git fails the folder should be missing what makes debugging and support easier.
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -0,0 +1,75 @@
...
+    addKeyword(getName());
+    newInstance = add(new StringProperty("", false, "newInstance"));
+  }
+
+  @Override
+  public String getName() {
+
+    return "create";
+  }
+
```
hohwille, line 35: You need to override this so I can can call this commandlet outside of a project (IDE_HOME):
```suggestion

  @Override
  public boolean isIdeHomeRequired() {

    return false;
  }

```

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -0,0 +1,75 @@
...
+    initializeInstance(newInstancePath);
+    ProcessContext pc = this.context.newProcess().executable("ide");
+    pc.addArgs("update");
+    pc.directory(newInstancePath);
+    if (pc.run() == ProcessResult.SUCCESS) {
+      this.context.success("IDEasy Instance successfully created in {}", newInstancePath);
+    } else {
+      this.context.warning("Could not create IDEasy Instance.");
+    }
```
hohwille, line 62: Do not call ourselves as process. If IDEasy is already launched what should we call another process that needs to boostrap and then do the things we already have in our Java code next door (next method)?
That is exactly why I suggested to create a super-class that both the `CreateCommandlet` and the `UpdateCommandlet` extend. Then simply add methods like `updateSettings` and `setupConf` there as protected methods and just call them from the sub-classes. If I am not mistaken, you could even orchestrate the entire process of the update (calling `updateSettings`, creating folders, `setupConf`, etc.) there in a single reusable method.
The only thing to consider is that you need to supply the path to IDE_HOME somehow or you would need to re-initialize the `IdeContext` on the fly.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
hohwille: We either need to re-initialize the context for that or you can simply provide the `Path` to the project and resolve things like `settings` from there manually.
However, it will not be working by calling ourselves as an external process.
Please be aware that IDEasy can also be called from a JavaFx GUI in the future and then when calling `ide` script as an external process will simply cause bugs and not work and then questions in that sub-process will not be asked in the JavaFX GUI but in the headless background process in the nirvana.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
hohwille: Correct. I updated my comment. I was not aware that you could already see my review comments before I submit the review... Interesting effect on github...

cli/src/test/java/com/devonfw/tools/ide/commandlet/UpdateCommandletTest.java
```diff
@@ -0,0 +1,57 @@
...
+    IdeTestContext context = newContext(PROJECT_UPDATE);
+    UpdateCommandlet uc = context.getCommandletManager().getCommandlet(UpdateCommandlet.class);
+    // act
+    uc.run();
+
+    // assert
+    assertLogMessage(context, IdeLogLevel.SUCCESS, "Successfully updated settings repository.");
+    assertThat(Files.exists(context.getConfPath()));
+    assertThat(Files.exists(context.getSettingsPath().resolve("java")));
+    assertThat(Files.exists(context.getSettingsPath().resolve("mvn")));
```
hohwille, line 30: Seems that our Knowledge-Transfer is still not working.
Things like proper usage of AssertJ was already discussed two times in KT session but if people miss the sessions it seems the knowledge never flows to them.
Please make advance of the rich capabilities that AssertJ offers.
In this case simply do `assertThat(file).exists();`
Please also be aware that `assertThat(«expression»);` is always wrong as it does not check anything.
It will just create an `AbstractAssert` but you have to call an actual assertion method on it so in case of a `boolean` assert something like `isTrue()`.

hohwille: Further just from reading the code I am quite sure that this test will not work as expected and the settings git repo is not pulled and there will be no `java` or `mvn` subfolder in `settings`.

cli/src/test/resources/ide-projects/update/project/home/.ide/ide.properties
```diff
@@ -0,0 +1,15 @@
...
+FOO=foo-${BAR}
+
+TEST_ARGS1=${TEST_ARGS1} user1
+TEST_ARGS2=${TEST_ARGS2} user2
+TEST_ARGS3=${TEST_ARGS3} user3
+TEST_ARGS7=user7
+TEST_ARGS10=user10
+TEST_ARGSb=userb
+TEST_ARGSc=${TEST_ARGS1} userc
+TEST_ARGSd=${TEST_ARGS1} userd
```
hohwille, line 15: Also Windows Line endings in mixed EOL style.
This file is IMHO pointless for your test and should be removed.

hohwille (changes requested):
@salimbouch thanks for your rework and great that you merged implementation of #101 and #102 into this single PR what makes it a lot easier. 👍 
I left various review comments for improvement or fixes. Also we need to have a test for `CreateCommandlet` as I can only see that `UpdateCommandlet` is tested so far.
As this PR has quite some complexity and we have done some review rounds already, could you create a teams meeting for Thursday or Friday so we can align and complete the last steps together to get this to the finishing line and avoid further review ping-pongs?
You have already done a very great job and this is close to the end but it will make things much easier, if we discuss face-to-face.

cli/src/main/java/com/devonfw/tools/ide/common/StepContainer.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A utility class to manage and log the progress of steps in a process.
+ * Each step can be started, ended with success or failure, and the overall completion
+ * status can be checked.
+ * @throws CliException if one or more steps fail.
+ */
+public class StepContainer {
```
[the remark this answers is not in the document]
hohwille: I will refactor this after the merge as this is not really related to this PR and needed for merge.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpdateCommandlet.java
hohwille: Solved. Thanks for the nice enhancement.

cli/src/test/resources/ide-projects/update/_ide/urls/java/java/17.0.6/linux_x64.urls
salimbouch: done

cli/src/test/resources/ide-projects/update/project/settings/eclipse/plugins/anyedit.properties
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
salimbouch: done, renamed to `project`

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
salimbouch: removed this block, still not implemented reinitializing of the context.

cli/src/test/java/com/devonfw/tools/ide/commandlet/UpdateCommandletTest.java
salimbouch: done

hohwille (approved):
@salimbouch thanks for your excellent rework. Now everything looks good and is ready for merge. 👍

PR #151, opened 2023-12-01
#20: Implement ToolCommandlet for GCViewer 

### LINKED ISSUE #20: Implement ToolCommandlet for GCViewer
Create `ToolCommandlet` for GCViewer
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/gcviewer

alfeilex:
This PR implements GCViewer and adds some code to handle unpacked jar files. Furthermore, it adds a workaround for sourceforge urls. 

Fixes #20


cli/src/main/java/com/devonfw/tools/ide/util/FilenameUtil.java
```diff
@@ -28,6 +28,12 @@ public static String getExtension(String path) {
       lastSlash = 0;
     }
     int lastDot = path.lastIndexOf('.');
+
+    // workaround for sourceforge urls
+    if (path.contains("sourceforge") && path.endsWith("download")) {
+      return path.substring(lastDot + 1, lastDot + 4);
```
aBega2000, line 34: I don't quite understand what the numbers 1 and 4 stand for, so I suppose it's magic numbers in this case.

alfeilex: The numbers define the range of extraction to obtain the extension of the downloaded file from the last dot.
For example if the `url www.sourceforge.com/testfile.jar/download` is given, then the substring `jar` is extracted.

However, I replaced `lastDot + 4` with `lastSlash` to avoid conflicts, when the extensions are more than 3 characters long.

aBega2000 (approved):

cli/src/main/java/com/devonfw/tools/ide/util/FilenameUtil.java
aBega2000: Thank you for your changes.

cli/src/main/java/com/devonfw/tools/ide/util/FilenameUtil.java
```diff
@@ -28,6 +28,12 @@ public static String getExtension(String path) {
       lastSlash = 0;
     }
     int lastDot = path.lastIndexOf('.');
+
+    // workaround for sourceforge urls
+    if (path.contains("sourceforge") && path.endsWith("download")) {
+      return path.substring(lastDot + 1, lastSlash);
+    }
```
hohwille, line 35: I am not sure, if I fully got this but it seems we are using this method with an URL instead of a `path`.
It was not really designed for this but if we have a sourceforge URL and the actual download filename is somehow a folder in the path I am fine with this workaround. Thanks for adding the comment to make it more clear. Probably add an example URL to make it even more obvious what is going on here.

hohwille (changes requested):
@alfeilex Thanks for this PR. Looks good to me 👍 
I left some comments for improvements, one is most probably a bug that needs to be fixed before we can merge.

cli/src/main/java/com/devonfw/tools/ide/util/FilenameUtil.java
hohwille: Sorry for perfection but we could additionally check that path starts with `https://` to make this clear and avoid errors in strange edge cases.

cli/src/main/java/com/devonfw/tools/ide/tool/gcviewer/GcViewer.java
```diff
@@ -0,0 +1,42 @@
...
+  protected boolean isExtract() {
+
+    return false;
+  }
+
+  @Override
+  public void run() {
+
+    install(true);
+    ProcessContext pc = new ProcessContextImpl(this.context);
```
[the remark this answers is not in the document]
alfeilex: done

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -270,7 +270,13 @@ protected void extract(Path file, Path targetDir) {
...
     } else {
       this.context.trace("Extraction is disabled for '{}' hence just moving the downloaded file {}.", getName(), file);
-      fileAccess.move(file, targetDir);
+
+      try {
+        Files.createDirectories(targetDir);
+      } catch (IOException e) {
+        throw new IllegalStateException("Failed to create folder " + targetDir);
+      }
+      fileAccess.move(file, targetDir.resolve(file.getFileName()));
```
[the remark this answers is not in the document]
alfeilex: Where do you see the additional folders situation? Because currently the software tools are organized in `_ide/software/default/eclipse/eclipse` anyway or have we already changed that?

I installed `eclipse` with the the unchanged behaviour and there was no difference in the folder structure. But I still think your suggestion makes more sense and I have added it.

cli/src/main/java/com/devonfw/tools/ide/util/FilenameUtil.java
alfeilex: I have added the check with `startsWith(https://)`. We could also take `https://sourceforge` as expression, but I don't if there are subdomains that would fail the check, so I keep `sourceforge` as separate expression.

hohwille (approved):
@alfeilex Excellent! Thanks for making this perfect 👍 
Ready for merge.

PR #141, opened 2023-11-13
#93: Support for Maven Plugins

alfeilex:
This PR implements the possibility that maven plugins are downloaded and moved to `lib/ext` folder. 

fixes #93 

The tool currently extends `IdeToolCommandlet`, but should be replaced by `PluginBasedToolCommandlet` in the future (see #142)

PR #153, opened 2023-12-11
#142 and #94: Plugin based commandlet feature and PluginBasedCommandlet

### LINKED ISSUE #94: Ability to define plugins in user config
As IDEasy user, I want to be able to define a plugin that I always want to use independent of what the project settings git has configured so that I do not always have to manually post-install it if I want to have it for personal favor.
So lets assume I have a personal favor for plugin `MagicPlugin` for the IDE `Cool-IDE`. Further, lets assume, that I use IDEasy to manage and work in 10 different projects. For all these projects I want to use `Cool-IDE` with `MagicPlugin`. However, the 10 projects have never heard of `MagicPlugin` or do not like it for some reason so there is no chance to have it pre-configured in `settings/Cool-IDE/plugins/MagicPlugin.properties`.

Therefore, I want to add an ability to configure plugins also per user independent of the `settings`.
As the `conf` folder is also per project and we need a central place for such global configuration, we will use this:
 `~/.ide/settings/«tool»/plugins/«plugin».properties`

That means that when a plugin based tool is configured, then plugins are taken from both `$IDE_HOME/settings/«tool»/plugins/«plugin».properties` and the new location in the users home directory.
If the same `«plugin».properties` exists in both locations, we have to chose a priority, which file to use.
IMHO letting users home config override project settings would make sense even though this is opposite to the behavior for `ide.properties`.

This story is related to #10 and #93.

See also here: https://github.com/devonfw/IDEasy/blob/df453f52513702d04be41a3bbb00e3f5e29a8ffc/cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java#L42

### LINKED ISSUE #142: Implement PluginBasedCommandlet
Currently, we use `IdeToolCommandlet` to handle plugin based commandlets like `eclipse`, `vscode`, `intellij`. But there are also other plugin based tools like `mvn`. This tool is not an `ide` tool and does not need functionalities like installing plugins etc. that are defined in `IdeToolCommandlet`.

This story should implement an abstract **PluginBasedCommandlet** which extends **LocalToolCommandlet.**  Functions like `getPluginsMap` or `getPluginsConfigPath` should be moved from **IdeToolCommandlet** to the new class. 

Finally, the classes should be in this relationship:
  ```
  PluginBasedToolCommandlet extends LocalToolCommandlet
        Mvn extends PluginBasedToolCommandlet
        IdeToolCommandlet extends PluginBasedToolCommandlet
            Eclipse extends IdeToolCommandlet
            ...
```

There are related issues that may be in conflict with this issue:
#93, #94

### PR DESCRIPTION
**Pull request by diiinesh**
fixes #142:

- created an abstract class "PluginBasedCommandlet" with all the plugin releated functions

fixes #94 

- modified getPluginsMap function, such that plugins can be processed with higher priority in user home directory

### CODE REVIEW

File: cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java [hunk 1]
```diff
@@ -1,32 +1,25 @@
...
 /**
  * {@link ToolCommandlet} for an IDE (integrated development environment) such as {@link Eclipse}, {@link Vscode}, or
  * {@link Intellij}.
  */
-public abstract class IdeToolCommandlet extends LocalToolCommandlet {
-
-  private Map<String, PluginDescriptor> pluginsMap;
+public abstract class IdeToolCommandlet extends PluginBasedCommandlet {
 
   private Collection<PluginDescriptor> configuredPlugins;
```
hohwille, line 24: This was also moved so it needs to be deleted here (otherwise you have it twice)

cli/src/main/java/com/devonfw/tools/ide/tool/PluginBasedCommandlet.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+    return this.configuredPlugins;
+  }
+
+  /**
+   * @return the {@link Path} where the plugins of this {@link IdeToolCommandlet} shall be installed.
+   */
+  public Path getPluginsInstallationPath() {
+
+    // TODO add edition???
```
hohwille, line 110: This is a good question!
However, our current approach in devonfw-ide was:
* when we install or update a tool, we simply delete the plugins folder for that tool and reinstall all plugins from scratch. Please also note that IntelliJ supports getting the plugins based on the IntelliJ version so we might get different plugin binaries for different IntelliJ versions. For this reason we do not keep plugins but simply delete and reinstall them when the product changes.
* we can only have one installation of a tool inside a devonfw-ide instance. Therefore we can not have multiple editions of the same tool in the same instance and there is no conflict.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
hohwille: Seems to apply for many other methods - please move code instead of copying it to the super-class without removing it in the sub-class what is causing duplicates.

hohwille (commented):
@diiinesh I am aware this is still in draft. However, due to a question I was asked to answer in this PR I already looked at your code.
You did a good job and are very close to the final result.
IMHO you can address my review comments and then take this PR out of draft mode and move it to the review so it can get merged.

cli/src/main/java/com/devonfw/tools/ide/tool/PluginBasedCommandlet.java
```diff
@@ -0,0 +1,161 @@
...
+import com.devonfw.tools.ide.io.FileAccess;
+import com.devonfw.tools.ide.tool.ide.IdeToolCommandlet;
+import com.devonfw.tools.ide.tool.ide.PluginDescriptor;
+import com.devonfw.tools.ide.tool.ide.PluginDescriptorImpl;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.*;
```
hohwille, line 15: your IDE settings are outdated/incorrect. Star-Imports are discouraged. According to our IDE settings this should also be auto-configured correctly for Intellij (maybe ask @jan-vcapgemini if it is not working for you) and for Eclipse it was correct from the beginning.

cli/src/test/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandletTest.java
```diff
@@ -67,38 +64,4 @@ public void testNormalizeWindowsLine() {
...
-    assertLogMessage(context, IdeLogLevel.INFO, "IDE_TOOLS=mvn,eclipse");
-    assertLogMessage(context, IdeLogLevel.INFO, "ECLIPSE_VERSION=2023-03");
-    assertLogMessage(context, IdeLogLevel.INFO, "FOO=foo-bar-some-${UNDEFINED}");
-    assertLogMessage(context, IdeLogLevel.INFO, "JAVA_VERSION=17*");
-    assertLogMessage(context, IdeLogLevel.INFO, "INTELLIJ_EDITION=ultimate");
-    assertLogMessage(context, IdeLogLevel.INFO, "DOCKER_EDITION=docker");
-  }
-  /**
-   * Test of {@link EnvironmentCommandlet} does not require home.
-   */
```
hohwille, line 95: your commit is removing and reverting a lot of things that seems to be undesired.
I already reverted some things in the browser but that is rather tedious.
IMHO you did something wrong with git here to lose/revert changes that had made by others as improvements.
See also entire removal of `ConextCommandletTest.java` what is IMHO a mistake.
I already reverted the removal of the coverall stuff added by Moritz.

cli/src/test/java/com/devonfw/tools/ide/tool/AbstractPluginBasedCommandletTest.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  @Test
+  public void testGetPluginsMap() {
+
+
+
+
+
+  }
```
hohwille, line 36: does not seem to be complete.
Why is a @Test annotation inside a Commandlet? This does not make sense.
Either this is a test so it may extend `Assertions` or other test-classes or it is a commandlet but not a test.

pom.xml
```diff
@@ -49,26 +49,6 @@
...
-          </execution>
-          <execution>
-            <id>report</id>
-            <phase>prepare-package</phase>
-            <goals>
-              <goal>report</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
```
hohwille, line 71: was added by Moritz on purpose. Do not remove.

cli/src/test/java/com/devonfw/tools/ide/commandlet/HelpCommandletTest.java
```diff
@@ -24,7 +24,7 @@ public void testThatHomeIsNotReqired() {
     // act
     HelpCommandlet help = new HelpCommandlet(context);
     // assert
-    assertThat(help.isIdeHomeRequired()).isFalse();
+    assertThat(help.isIdeHomeRequired()).isEqualTo(false);
```
hohwille, line 27: This was also an improvement that you are reverting.

README.adoc
```diff
@@ -10,7 +10,6 @@ image:https://img.shields.io/github/license/devonfw/IDEasy.svg?label=License["Ap
 image:https://img.shields.io/maven-central/v/com.devonfw.tools.ide/ide-cli.svg?label=Maven%20Central["Maven Central",link=https://search.maven.org/search?q=g:com.devonfw.tools.ide]
 image:https://github.com/devonfw/IDEasy/actions/workflows/build.yml/badge.svg["Build Status",link="https://github.com/devonfw/IDEasy/actions/workflows/build.yml"]
 image:https://github.com/devonfw/IDEasy/actions/workflows/update-urls.yml/badge.svg["Update URLS Status",link="https://github.com/devonfw/IDEasy/actions/workflows/update-urls.yml"]
-image:https://coveralls.io/repos/github/devonfw/IDEasy/badge.svg?branch=main["Coverage Status",link="https://coveralls.io/github/devonfw/IDEasy?branch=main"]
```
hohwille, line 13: please keep this badge added by Moritz for the coverage.

hohwille (changes requested):
@diiinesh thanks for your PR.
It seems that git did play some trick on you and something went wrong in this PR reverting lots of changes/improvements that others have done by accident.
Please have a look. See also PR #141

hohwille:
To avoid double effort, I will close this PR and we concentrate on PR #141 that includes the changes of this PR and already fixed some of the problems.

Issue #148, opened 2023-11-23
Offer git restore if urls were changed locally


jan-vcapgemini:
As an IDEasy user, I want IDEasy to detect if I've modified my local `urls` repository.
It would be nice if IDEasy could provide a notification if `urls` is in an inconsistent state and offer an automatic `git restore`.

Currently IDEasy just checks if the `urls` folder exists and if not, it does a `git checkout`.
If the folder exists, IDEasy attempts only a `git pull` which can not detect any changes e.g. deletions.

If the user is offline and force mode is active, the user needs to be asked if he wants to repair (and lose) his local changes on the urls repository.

hohwille:
See also #173 
We can detect conflicts if not force is active for git update and in such case ask the user (with warning to lose all changes) if he wants to revert all his changes.

jan-vcapgemini:
Can be closed because for urls we want to enforce a git reset.
Will be continued if other processes using git reset should wait for user input.

PR #155, opened 2023-12-12
#173: use force update mode for ide-urls

jan-vcapgemini:
Implements: #173 

* add force parameter to `gitPullOrClone`
* if force is true and online do `git pull` followed with `get reset` and `git cleanup`
* minor cleanups
* added extra checks for changed and untracked files
* introduced new `GitUtils` class (handles git commands and error messages)
* added capture and warnings to all git commands
* added tests and mocks for all git commands

Does not implement:
* detect and fix changes if offline and force is active (has to be added in #148)


salimbouch (commented):
Looks good to me :)

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -178,7 +180,6 @@ public AbstractIdeContext(IdeLogLevel minLogLevel, Function<IdeLogLevel, IdeSubL
         if (Files.isDirectory(rootPath)) {
           if (!ideRootPath.equals(rootPath)) {
             warning("Variable IDE_ROOT is set to '{}' but for your project '{}' would have been expected.");
-            ideRootPath = rootPath;
```
hohwille, line 181: warning message is lacking parameter arguments.
Indeed the assignment you removed does not make sense as it is overridden after the `if` block, but to me this still looks buggy or does not make sense. The assignment in #184 will also have no effect since it is the same as in #173.
We set `root` to env variable `$IDE_ROOT`. So only in the case that this is set (that is most probably untested as nobody in our team has this variable set) we actually want to set `ideRootPath` to `$IDE_ROOT` (`rootPath`) and do some sanity checks as verification.
IMHO the assignment `ideRootPath = rootPath` is what we really want no matter if `rootPath` equals `ideRootPath` or not.
Also the assignment in #187 should be in question. As a result IDEasy will not be functional.
Stay with the fallback from #173 would still be better. The `warning` in #186 should maybe be logged as error.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -611,7 +608,19 @@ public void gitPullOrClone(Path target, String gitRepoUrl) {
         askToContinue(message);
```
hohwille, line 608: Just since I am pointed to all this code one more remark: `*This* is a local git repo` is IMHO not very helpful for end-users. We should add `target` path to the message so the user knows what `This` actually is about.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -611,7 +608,19 @@ public void gitPullOrClone(Path target, String gitRepoUrl) {
...
+        if (isOnline()) {
+          result = pc.addArg("fetch").addArg("origin").addArg("master").run(false);
+          if (!result.isSuccessful()) {
+            warning("Git failed to fetch from origin master");
+          }
+        }
+        if (force) {
+          result = pc.addArg("reset").addArg("--hard").addArg("HEAD~").run(false);
+          if (!result.isSuccessful()) {
+            warning("Git failed to reset to head");
```
hohwille, line 620: same here. See #625 where `target` is added to the end-user message to be more helpful.

hohwille (changes requested):
@jan-vcapgemini thanks for your PR with analysis and fix. 👍 
Great that you also addressed some code-cleanups.
I like your solution and also I think it is a very good idea that you added an explicit `force` flag instead of (mis)using our `--force` mode for this as this more depends on the Git use-case and for `ide-urls` this is exactly desired while for `ide-settings` this could end in a disaster as end-users may loose lots of valuable configuration work.

So in general I am happy with your PR and would like to merge it.
However, since you stumbled over some odd stuff in our code-base and started improving, I would love to take the opportunity to really fix that even though not related to the actual story.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -178,7 +180,6 @@ public AbstractIdeContext(IdeLogLevel minLogLevel, Function<IdeLogLevel, IdeSubL
         if (Files.isDirectory(rootPath)) {
           if (!ideRootPath.equals(rootPath)) {
             warning("Variable IDE_ROOT is set to '{}' but for your project '{}' would have been expected.");
```
hohwille, line 182:
```suggestion
            warning("Variable IDE_ROOT is set to '{}' but for your project '{}' the path '{}' would have been expected.", root, this.ideHome.getFileName(), ideRootPath);
```

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -178,7 +180,6 @@ public AbstractIdeContext(IdeLogLevel minLogLevel, Function<IdeLogLevel, IdeSubL
...
-            ideRootPath = rootPath;
           }
           ideRootPath = this.ideHome.getParent();
```
hohwille, line 184:
```suggestion
          ideRootPath = rootPath;
```

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
hohwille: aha, and wenn the internet connection is interrupted shouldn't we log some warning at least?

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -611,7 +608,19 @@ public void gitPullOrClone(Path target, String gitRepoUrl) {
         askToContinue(message);
       } else {
         pc.errorHandling(ProcessErrorHandling.WARNING);
-        result = pc.addArg("pull").run(false);
+        if (isOnline()) {
+          result = pc.addArg("fetch").addArg("origin").addArg("master").run(false);
```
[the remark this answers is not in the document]
hohwille: I am still not 100% sure if `git fetch ...` can reliable replace `git pull` nor am I really convinced that the source of our bug #6 is `git pull` itself but instead more something with Java invoking `git` and some stream redirection issue ending up in a blocked read operation from the process output.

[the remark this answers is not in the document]
jan-vcapgemini: I'm pretty sure that git pull is the issue because if it needs to merge, it currently prompts for an input with a yes/no question and hangs the process. A solution for this could be to set a merge strategy beforehand. But I think the git fetch and especially reset should be more reliable here.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
jan-vcapgemini: Not sure if we can detect that properly without using capture.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
jan-vcapgemini: I've added the target path now.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
jan-vcapgemini: I've added the target path now.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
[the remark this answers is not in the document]
jan-vcapgemini: I've adjusted the fetch to `master:master` now.

hohwille (approved):
@jan-vcapgemini thanks for your rework. Now all looks good. 👍 
I will however have to test this manually before I will merge.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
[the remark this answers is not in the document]
hohwille: I will test this once more, but also have to test the version with `force=false` what is tricky, as clone or updating of settings is not yet implemented/merged.

hohwille:
Tested but not working as expected:
```
Running command 'C:\Program Files\Git\cmd\git.exe' with arguments 'fetch' 'origin' 'master:master' failed with exit code 128!
Git failed to fetch from origin master
Failed to update git repository at D:\projects\_ide\urls
See above error for details. If you have local changes, please stash or revert and retry.
```

Seems that also the stackoverflow page was wrong:
```
$ git fetch origin master:master
fatal: refusing to fetch into branch 'refs/heads/master' checked out at 'D:/projects/_ide/urls'
```

However, your original approach succeeds:
```
$ git fetch origin master
From https://github.com/devonfw/ide-urls
 * branch              master     -> FETCH_HEAD
```
But IMHO it is not updating but just fetching so also not a solution.

hohwille:
Your fetch approach is not working I tested with changes missing locally:
```
$ git fetch origin master
From https://github.com/devonfw/ide-urls
 * branch              master     -> FETCH_HEAD

hohwille@CE49454 MSYS /d/projects/_ide/urls (master)
$ git pull
Updating b6e6e776..e88b5eea
Fast-forward
 .../android-studio/2023.3.1.1/linux_x64.urls       |  1 +
 .../2023.3.1.1/linux_x64.urls.sha256               |  1 +
 .../android-studio/2023.3.1.1/mac_arm64.urls       |  1 +
 .../2023.3.1.1/mac_arm64.urls.sha256               |  1 +
 .../android-studio/2023.3.1.1/mac_x64.urls         |  1 +
 .../android-studio/2023.3.1.1/mac_x64.urls.sha256  |  1 +
 .../android-studio/2023.3.1.1/status.json          | 25 ++++++++++++++++++++++
 .../android-studio/2023.3.1.1/windows_x64.urls     |  1 +
 .../2023.3.1.1/windows_x64.urls.sha256             |  1 +
 9 files changed, 33 insertions(+)
 create mode 100644 android-studio/android-studio/2023.3.1.1/linux_x64.urls
 create mode 100644 android-studio/android-studio/2023.3.1.1/linux_x64.urls.sha256
 create mode 100644 android-studio/android-studio/2023.3.1.1/mac_arm64.urls
 create mode 100644 android-studio/android-studio/2023.3.1.1/mac_arm64.urls.sha256
 create mode 100644 android-studio/android-studio/2023.3.1.1/mac_x64.urls
 create mode 100644 android-studio/android-studio/2023.3.1.1/mac_x64.urls.sha256
 create mode 100644 android-studio/android-studio/2023.3.1.1/status.json
 create mode 100644 android-studio/android-studio/2023.3.1.1/windows_x64.urls
 create mode 100644 android-studio/android-studio/2023.3.1.1/windows_x64.urls.sha256
```

So only with `git pull` I got the required changes so this PR breaks the existing behavior and adds new bugs even if git pull does not hang.

cli/src/main/java/com/devonfw/tools/ide/context/GitUtils.java
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * Runs a git pull or a git clone.
+   *
+   * @param force boolean true enforces a git hard reset and cleanup of added files.
+   * @param gitRepoUrl String of repository URL.
+   */
+  protected void runGitPullOrClone(boolean force, String gitRepoUrl) {
+
+    this.processContext = this.context.newProcess().directory(this.targetRepository).executable("git")
+        .withEnvVar("GIT_TERMINAL_PROMPT", "0");
```
hohwille, line 50: here ProcessContext is created but it is also used in other methods. All of these methods are protected (none is public).
If one of the other methods is used before this one, it will fail with a NullPointerException.
I would consider creating a private method to access `processContext` that creates it lazily and all methods access it via the private method. It is also possible to not store it as member variable and always create it per method, but reuse can make sense...

hohwille (changes requested):
@jan-vcapgemini thanks for your PR and initiative to bring our git usage further 👍 
I like that you extract git usage that also grows in complexity to its own class and stop blowing up our `IdeContext` "god object".
As already announced this slightly collides/conflicts with #163 that has meanwhile been merged so you need to resolve conflicts and rebase.
Would't it make sense to have the API of `GitUtils` public and remove the git stuff from `IdeContext` but rather make `GitUtils` available via `context.getGitUtils()`?

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -11,11 +11,7 @@
 import com.devonfw.tools.ide.cli.CliArgument;
 import com.devonfw.tools.ide.cli.CliArguments;
 import com.devonfw.tools.ide.cli.CliException;
-import com.devonfw.tools.ide.commandlet.Commandlet;
-import com.devonfw.tools.ide.commandlet.CommandletManager;
-import com.devonfw.tools.ide.commandlet.CommandletManagerImpl;
-import com.devonfw.tools.ide.commandlet.ContextCommandlet;
-import com.devonfw.tools.ide.commandlet.HelpCommandlet;
+import com.devonfw.tools.ide.commandlet.*;
```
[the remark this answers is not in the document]
jan-vcapgemini: Looks like my new IDE did not import the settings properly.
We might have to re-check our 999 import condition, it seems that it was not to applied for my new project.
I had to set this manually in my settings.

cli/src/main/java/com/devonfw/tools/ide/context/GitUtils.java
jan-vcapgemini: In my new test scenarios I actually ran into this problem already.
I agree that we need to re-initialize this before each git command.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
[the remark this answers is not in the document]
jan-vcapgemini: I've fixed this now and adjusted the dev-ide-settings.

jan-vcapgemini:
I'm still working on the new tests (simulating git commands).

jan-vcapgemini:
I'm a bit stuck here. The tests are working locally, but not in github actions.
This PR is missing one last test/mock for a git reset.

jan-vcapgemini:
Last test was added and github actions isOffline check was fixed.
Ready for review now.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -599,78 +599,30 @@ public DirectoryMerger getWorkspaceMerger() {
...
+   * Checks if the Git repository in the specified target folder needs an update by inspecting the modification time of
+   * a magic file.
    *
    * @param urlsPath The Path to the Urls repository.
    * @param repoUrl The git remote URL of the Urls repository.
+   * @param force boolean true enforces a git hard reset and cleanup of added files.
    */
-
-  private void gitPullOrCloneIfNeeded(Path urlsPath, String repoUrl) {
+  private void gitPullOrCloneIfNeeded(Path urlsPath, String repoUrl, boolean force) {
```
hohwille, line 625: great that you are following `SoC` and extracting git-logic out of `IdeContext` into its own git helper component.
But to do this consistent also this method needs to go there (making it `public`).

cli/src/test/java/com/devonfw/tools/ide/util/GitUtilsProcessResultMock.java
```diff
@@ -0,0 +1,48 @@
+package com.devonfw.tools.ide.util;
+
+import java.util.List;
+
+import com.devonfw.tools.ide.process.ProcessResult;
+
+/**
+ * Mocks a process result with adjustable params
+ */
+public class GitUtilsProcessResultMock implements ProcessResult {
```
hohwille, line 10: It would be smarter to add a `GitHelper` interface and Refactor `GitUtils` to a `GitHelperImpl` implementation of it.
Then you can change this class to a `GitHelperMock` implementing `GitHelper` interface.
This way our test implementation of `IdeContext` could use `GitHelperMock` without loosing the ability to call processes via `ProcessContext`.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
jan-vcapgemini: Adjusted.

cli/src/test/java/com/devonfw/tools/ide/util/GitUtilsProcessResultMock.java
jan-vcapgemini: Adjusted, added new GitContext interface and implementation. This class was removed now.

cli/src/main/java/com/devonfw/tools/ide/context/GitContext.java
```diff
@@ -0,0 +1,82 @@
...
+   * @param repoUrl the git remote URL to clone from. May be suffixed with a hash-sign ('#') followed by the branch name
+   *        to check-out.
+   * @param targetRepository the {@link Path} to the target folder where the git repository should be cloned or pulled.
+   *        * It is not the parent directory where git will by default create a sub-folder by default on clone but the *
+   *        final folder that will contain the ".git" subfolder.
+   * @param remoteName the remote server name.
+   * @param branchName the name of the branch.
+   * @param force boolean true enforces a git hard reset and cleanup of added files.
+   */
+  void pullOrCloneIfNeeded(String repoUrl, Path targetRepository, String remoteName, String branchName, boolean force);
```
hohwille, line 25: we should discuss this.
* Do we really always need this in our signatures here?
* What if I want to clone the default branch of a repository but I do not know the name of that default branch? That currently does not really be possible.
* Looking at the implementation these 2 new parameters are however ignored unless `force` is `true`. That is completely unclear for the API user reading the JavaDoc.
* Should we remove the `force` flag and these 2 new parameters entirely and instead add a new method signature for the `pullOrFetchAndResetIfNeeded` usecase? Or is there a better idea?

cli/src/main/java/com/devonfw/tools/ide/context/GitContext.java
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * Runs a git pull.
+   *
+   * @param targetRepository the {@link Path} to the target folder where the git repository should be cloned or pulled.
+   *        * It is not the parent directory where git will by default create a sub-folder by default on clone but the *
+   *        final folder that will contain the ".git" subfolder.
+   * @param remoteName the remote server name.
+   * @param branchName the name of the branch.
+   */
+  void pull(Path targetRepository, String remoteName, String branchName);
```
hohwille, line 60: The two now arguments are ignored and therefore only cause confusion to the API.
```suggestion
   */
  void pull(Path targetRepository);
```

cli/src/main/java/com/devonfw/tools/ide/context/GitContext.java
hohwille: The `reset` method can stay as it is. A `pullOrFetchAndResetIfNeeded` may have an optional `remoteName` argument and use `"origin"` as default. Then all would be consistent and clean.

jan-vcapgemini: Thanks for your suggestions. I've now implemented them, please check again.

cli/src/main/java/com/devonfw/tools/ide/context/GitContext.java
jan-vcapgemini: Adjusted.

cli/src/main/java/com/devonfw/tools/ide/context/GitContextImpl.java
```diff
@@ -0,0 +1,286 @@
...
+      } else {
+        this.processContext.errorHandling(ProcessErrorHandling.WARNING);
+
+        if (!this.context.isOffline()) {
+          pull(targetRepository);
+        }
+      }
+    } else {
+      String branch = "";
+      int hashIndex = gitRepoUrl.indexOf("#");
```
hohwille, line 115: this is what I suggest to put into `GitUrl`. However, nice to have that we can improve later.

hohwille (approved):
@jan-vcapgemini Thanks for completing this complex PR. Great job - you have reached a great state & result 👍 
Ready for merge.

cli/src/main/java/com/devonfw/tools/ide/context/GitContextImpl.java
hohwille: I will resolve this for now so we can merge. We should still improve this later...

PR #159, opened 2023-12-18
#158: VersionRange with open boundaries

### LINKED ISSUE #158: VersionRange with open intervals
The VersionRange currently only allows closed intervals that include the borders, i.e. [x,y] for x<=y.
This class should be modified to also allow open and half open intervals, i.e. (x,y), (x,y], [x,y), that can exclude the borders.
The String representation should be similar to the current VersionRange, where the two VersionIdentifiers are separated by `>`. If `x` is not set, then it is treated as `-infinity`, and missing `y` as `infinity`. So (1.2.3.>] and (1.2.3.>) both represent all versions that are greater than 1.2.3 , but not including version 1.2.3.

MattesMrzik:
Extended the VersionRange to also model open boundaries that do not include the specified value. Also wrote tests for this class.
Closes #158.


cli/src/main/java/com/devonfw/tools/ide/version/VersionRange.java
```diff
@@ -92,10 +204,29 @@ public String toString() {
...
+      value = value.substring(1);
+    }
+    if (value.endsWith(")")) {
+      rightIsExclusive = true;
+      value = value.substring(0, value.length() - 1);
+    }
+    if (value.endsWith("]")) {
+      value = value.substring(0, value.length() - 1);
+    }
+
```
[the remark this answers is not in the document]
MattesMrzik: Thanks for your remark. I implemented it.

aBega2000 (approved):

cli/src/main/java/com/devonfw/tools/ide/version/VersionRange.java
```diff
@@ -69,33 +187,72 @@ public int compareTo(VersionRange o) {
...
+    if (this.min == null) {
+      return o.min == null && this.max.equals(o.max) && this.rightIsExclusive == o.rightIsExclusive;
+    }
+    if (this.max == null) {
+      return this.min.equals(o.min) && o.max == null && this.leftIsExclusive == o.leftIsExclusive;
+    }
+    return this.min.equals(o.min) && this.leftIsExclusive == o.leftIsExclusive && this.max.equals(o.max)
+        && this.rightIsExclusive == o.rightIsExclusive;
+
+  }
```
hohwille, line 221:
```suggestion
  public boolean equals(Object obj) {

    if (this == obj) {
      return true;
    } else if ((obj == null) || (getClass() != obj.getClass())) {
      return false;
    }
    VersionRange o = (VersionRange) obj;
    if (this.min == null && this.max == null) {
      return o.min == null && o.max == null;
    } else if (this.min == null) {
      return o.min == null && this.max.equals(o.max) && this.rightIsExclusive == o.rightIsExclusive;
    } else if (this.max == null) {
      return this.min.equals(o.min) && o.max == null && this.leftIsExclusive == o.leftIsExclusive;
    }
    return this.min.equals(o.min) && this.leftIsExclusive == o.leftIsExclusive && this.max.equals(o.max)
        && this.rightIsExclusive == o.rightIsExclusive;
  }
```

Nothing wrong here, but always use curly braces for `if` blocks and use `else if` for more compact code.

cli/src/main/java/com/devonfw/tools/ide/version/VersionRange.java
```diff
@@ -46,11 +158,17 @@ public VersionIdentifier getMax() {
...
         return false;
       }
     }
     if (this.max != null) {
+      if (this.max.equals(version)) {
+        return !this.rightIsExclusive;
+      }
       if (version.isGreater(this.max)) {
         return false;
       }
```
hohwille, line 174:
```suggestion
    if (this.min != null) {
      VersionComparisonResult compareMin = version.compareVersion(this.min);
      if (compareMin.isLess()) {
        return false;
      } else if (compareMin.isEqual() && !this.leftIsExclusive) {
        return false;
      }
    }
    if (this.max != null) {
      VersionComparisonResult compareMax = version.compareVersion(this.max);
      if (compareMax.isGreater()) {
        return false;
      } else if (compareMax.isEqual() && !this.rightIsExclusive) {
        return false;
      }
```

In case min equals, you did not check max anymore. This is a bug. It is always good to extend JUnits to cover these boundaries (https://en.wikipedia.org/wiki/Boundary-value_analysis).
Further if not equals the comparison was done twice that can be slightly optimized, by calling `compareVersion` only once for min/max.

hohwille (approved):
@MattesMrzik thanks for this PR and rework. 👍 
I added 2 suggestions. If you can agree and accept, we can merge this immediately.

hohwille (approved):
@MattesMrzik thanks for the fast rework and fixing the bug in my suggested code 👍 
Ready for merge now.

PR #163, opened 2024-01-03
#118: smart git pulls

### LINKED ISSUE #118: Smarter git pull on urls
As a IDEasy user, I want to get feedback fast so that I feel the tool is usable.
However, in devonfw-ide when I type `devon java version set ` and hit the `[Tab]` key, the auto-completion performs a git pull what involves HTTPS communication and is slow. Even if I then hit `[Tab]` again, I have to wait again for the same time.
However, it is very unlikely that another update has arrived in that second especially since we know that we have a github action performing the updates that runs only once per night.

Therefore this story should implement a smarter approach:
Whenever we want to do a `git pull` on urls, we should first check the local timestamp of a magic file like e.g. `urls/.git/HEAD`.
If that file has an mtime newer than some given delta (say 5 minutes), we can assume that our urls are already up-to-date and we can omit the `git pull`. However, if we do the `git pull`, we should update the mtime of that magic file (just like `touch urls/.git/HEAD` would do). This way auto-completion will get more responsive and the user has to wait only for the first time.

salimbouch:
closes #118


cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -652,6 +653,47 @@ public void gitPullOrClone(Path target, String gitRepoUrl) {
...
+      // Get the modification time of the magic file
+      long fileMTime;
+      try {
+        fileMTime = Files.getLastModifiedTime(magicFilePath).toMillis();
+      } catch (IOException e) {
+        throw new IllegalStateException("Could not read " + magicFilePath, e);
+      }
+
+      // Check if the file modification time is older than the delta threshold
+      if (currentTime - fileMTime > deltaThreshold) {
```
moritzLanger, line 682: not sure you need to define deltaThreshold above, as we only use it once, instead > 300* 1000

moritzLanger: In addition, as the origin where we would pull from only changes 1 per night at max, the threshold could be way higher imho

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -652,6 +653,47 @@ public void gitPullOrClone(Path target, String gitRepoUrl) {
...
+        gitPullOrClone(urlsPath, repoUrl);
+        try {
+          Files.setLastModifiedTime(magicFilePath, FileTime.fromMillis(currentTime));
+        } catch (IOException e) {
+          throw new IllegalStateException("Could not read " + magicFilePath, e);
```
moritzLanger, line 687: The error could also be a lack of writing permission, so I would change the String to something like "Could not read or write in" + magicFilePath

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -652,6 +653,47 @@ public void gitPullOrClone(Path target, String gitRepoUrl) {
...
+   * Checks if the Git repository in the specified target folder needs an update by
+   * inspecting the modification time of a magic file.
+   *
+   * @param urlsPath The Path to the Urls repository.
+   * @param repoUrl The git remote URL of the Urls repository.
+   */
+
+  private void gitPullOrCloneIfNeeded(Path urlsPath, String repoUrl) {
+
+    Path magicFilePath = urlsPath.resolve(".git").resolve("HEAD");
```
moritzLanger, line 666: One could change the Order and reuse one resolve of .git like this:

Path gitDirectory = urlsPath.resolve(".git");
Path magicFilePath = gitDirectory.resolve("HEAD")

moritzLanger (approved):
Looks good 👍 
I just added some small suggestions that could be worked in.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
salimbouch: true, changed it, thanks!

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
salimbouch: this looks cleaner, thanks

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
salimbouch: I agree it should be a bit higher than 5 mins, changed it to 30, however I think its better for readability to keep it in a variable with a comment so that whoever reads it knows what its about.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -652,6 +653,47 @@ public void gitPullOrClone(Path target, String gitRepoUrl) {
...
+   *
+   * @param urlsPath The Path to the Urls repository.
+   * @param repoUrl The git remote URL of the Urls repository.
+   */
+
+  private void gitPullOrCloneIfNeeded(Path urlsPath, String repoUrl) {
+
+    Path gitDirectory = urlsPath.resolve(".git");
+    Path magicFilePath = gitDirectory.resolve("HEAD");
+    long deltaThreshold = 30 * 60 * 1000; // 30 minutes in milliseconds
```
hohwille, line 668: use a constant for this as it has to be computed only once.
```
private static final long GIT_PULL_CACHE_DELAY_MILLIS = 30 * 60 * 1000;
```

Maybe even consider using `java.time.Duration` what makes the code more readable:
```
private static final long Duration GIT_PULL_CACHE_DELAY = Duration.ofMinutes(30);
// Duration.ofSeconds(30*60)
// Duration.ofMillis(30*60*1000)
```

Also my suggestion was to cache `5` minutes. Is `30` minutes maybe too much? What do the others think?

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -652,6 +653,47 @@ public void gitPullOrClone(Path target, String gitRepoUrl) {
...
+   * inspecting the modification time of a magic file.
+   *
+   * @param urlsPath The Path to the Urls repository.
+   * @param repoUrl The git remote URL of the Urls repository.
+   */
+
+  private void gitPullOrCloneIfNeeded(Path urlsPath, String repoUrl) {
+
+    Path gitDirectory = urlsPath.resolve(".git");
+    Path magicFilePath = gitDirectory.resolve("HEAD");
```
hohwille, line 667: only create this `Path` inside the `if` as otherwise it is not needed.

hohwille (approved):
@salimbouch thanks for your PR and the nice working solution. 👍 
Sorry for the long review delay.
It is great that you added a new `gitPullOrCloneIfNeeded` method to add the new behavior on top. Your implementation is functionally perfectly correct.
I added some comment for improvement. Can you resolve merge conflict and address the review comments? Then we are ready to merge this nice change.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
salimbouch: Done

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
salimbouch: Done

salimbouch: Now when force mode is enabled git pull will be performed, still at 30 min.

PR #170, opened 2024-01-08
#158: improved VersionRange

hohwille:
Improvement for Version Range (#158, #159):
* Consistency (min = null means negative infinity and requires leftExclusive, max = null means positive infinity and requires rightExclusive, min must not be greater than max)
* Syntax "[min>max]" does not really make sense (as it turns out this seems to come from https://github.com/devonfw/ide/issues/1054) and I changed `>` to `,` what is rather common (same in Maven, etc.).
* Improved maintainability of code


MattesMrzik (approved):
All good.

PR #172, opened 2024-01-08
#138: added updater for mvn 4.x versions

### LINKED ISSUE #138: maven 4.x releases not available in ide-urls
### Expected behavior

As a user of devonfw-ide or IDEasy, I want to be able to use latest versions of maven so that I can test new features or check if my project build is still working with that version.

### Actual behavior

[Maven 4.x releases](https://maven.apache.org/download.cgi#alpha-4-x-release) are not available in [ide-urls](https://github.com/devonfw/ide-urls/tree/master/mvn/mvn) and therefore the installation of such versions fail.

1. `devon mvn version set 4.0.0-alpha-8`
2. `devon mvn setup`

Or with IDEasy:
1. `ide install mvn 4.0.0-alpha-8`

### Related/Dependent Issues

ATTENTION: This issue is somehow blocked by issue https://github.com/devonfw/ide/issues/1369

MattesMrzik:
Fixes #138. Maven updates for the 4.x releases are distributed via a different URL than the other releases, so this updater has been added to retrieve the 4.x releases from the new website. This may only be necessary temporarily, as these 4.x versions will later be listed on the original URL as well.


cli/src/main/java/com/devonfw/tools/ide/tool/mvn/Mvn4UrlUpdater.java
```diff
@@ -0,0 +1,42 @@
...
+  @Override
+  protected String getVersionUrl() {
+
+    return "https://archive.apache.org/dist/maven/maven-4/";
+  }
+
+  @Override
+  protected Pattern getVersionPattern() {
+
+    return Pattern.compile("(\\d\\.\\d\\.\\d-[a-z]*?-\\d{1,2})");
```
aBega2000, line 40: How do you test that this is the correct pattern?

aBega2000 (commented):
Except the one question I have asked, everything else seems fine and correct

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/Mvn4UrlUpdater.java
MattesMrzik: I have run the `UpdateInitiator` and the correct versions are fetched and their urls added to the url repo.

aBega2000 (approved):

hohwille:
Note: Great work, looks all good, but we need to have https://github.com/devonfw/ide/pull/1374 out in the wild as otherwise we are causing more harm than good.
So my suggestion would be to release the PR with devonfw-ide this month and then still wait ~1 month for it to spread out before we merge this PR and populate the maven 4.x-alpha versions.

hohwille:
> So my suggestion would be to release the PR with devonfw-ide this month and then still wait ~1 month for it to spread out before we merge this PR and populate the maven 4.x-alpha versions.

3 month later should be safe for merge.
If people are using outdated versions of devonfw-ide and therefore get maven updated to `4.*-alpha` they need to run `devon ide update scripts` to prevent this undesired behavior.

Issue #175, opened 2024-01-09
IDEasy throws a NPE if a version was not found


jan-vcapgemini:
### Expected behavior

If a version number was not found IDEasy should throw an exception with a specific message.

### Actual behavior
IDEasy throws a NPE without a message.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Adjust ide.properties (add MVN_VERSION=3.)
2. run `install mvn`

### Comments/Hints:
Easy to fix with a null check.

### Affected version:

* IDEasy dev-SNAPSHOT

PR #176, opened 2024-01-09
#175: fixed NPE for missing version

jan-vcapgemini:
Fixes: #175

Implements:

* added null check and proper exception message


hohwille (approved):
@jan-vcapgemini thanks for this fix 👍

Issue #52, opened 2023-09-28
Adjusting Intellij settings in ide-settings


CREITZ25:
As an Intellij user, I want to use Intellij with defaults that make sense to us. To do this, the required settings should be stored in the ide-settings repository.

* [x] Settings - Editor - Code Style - Java - Tab "Imports": 
* [x] Class count to use import with '\*' : 999
* [x] Names count to use static import with '*': 999
* [x] Set line length to at least 160 characters (I cannot imagine that as of today any developer does not get this number of characters on his screen without the need to scroll - the restrictions limiting to 80 or later 120 characters per line come from an ancient time and the reformatting line wraps added by intellij contribute to our diff-war)
* [x] Ensure we get a Warning if JavaDoc is missing or incomplete for public and protected elements (Types, Fields, Constructors, Methods including parameters).
* [x] Activate "save actions" in Intellij (like in Eclipse to ensure the code-style is always applied and cannot be forgot on commit)
* [x] Eclipse reformat plugin should not be used anymore - we want to configure IntelliJ natively (otherwise you get errors if you never installed and configured Eclipse and our team is not using eclipse) See here
https://github.com/devonfw/ide-settings/blob/d1a7dd1f1ae817b3fb659f92a8fb54eb567b3b74/intellij/plugins/EclipseCodeFormatter.properties#L2
* [x] `Optional`: ~FolderCompact plugin could be pre-installed too~ (I initially liked this plugin but it has some severe downsides - if it is installed, you do not see folders like `src` anymore and if you have ` src/main/assembly` or anything not common you get unable to see this in IntelliJ. This renders the plugin useless even though it would otherwise be great. See my rating comment [here](https://plugins.jetbrains.com/plugin/18189-foldercompact))
* [x] Maven Builder settings need to be pre-configured
* [x] Encoding for resource bundle properties to UTF-8
* more?

jan-vcapgemini:
For `Adapter for Eclipse Code Formatter` we need to use these settings:

![image](https://github.com/devonfw/IDEasy/assets/59438728/0032b6c5-a632-4b13-b809-ab486d8de43b)

jan-vcapgemini:
For the `Maven Runner` I'd suggest these settings:

![image](https://github.com/devonfw/IDEasy/assets/59438728/cb7b8f64-5006-4310-a8ce-266ea2b0e32c)

jan-vcapgemini:
For `Maven Running Tests` I'm using:

[...]

PR #184, opened 2024-01-16
#183: removed isInstalledVersion()

### LINKED ISSUE #183: Duplicate message if tool is already installed
### Expected behavior

When installing a tool that is already installed I want to get a single message.

### Actual behavior

I get (for example):
```
Version v3.11.1 of tool helm is already installed at C:\projects\_ide\software\default\helm\helm\v3.11.1
Version v3.11.1 of tool helm is already installed
``` 

### Steps to reproduce 

Run `ide install <tool>` twice

MattesMrzik:
Fixes #183.  `isInstalledVersion()` was only used once and caused the duplicate msg if the tool was already installed.

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
@@ -2,6 +2,7 @@ package com.devonfw.tools.ide.tool;
 
 import com.devonfw.tools.ide.context.IdeContext;
 import com.devonfw.tools.ide.io.FileAccess;
+import com.devonfw.tools.ide.log.IdeLogLevel;
 import com.devonfw.tools.ide.process.ProcessContext;
 import com.devonfw.tools.ide.process.ProcessErrorHandling;
 import com.devonfw.tools.ide.repo.ToolRepository;
@@ -31,7 +32,9 @@ public abstract class GlobalToolCommandlet extends ToolCommandlet {
 
   @Override
   protected boolean isExtract() {
-    // for global tools we usually download installers and do not want to extract them (e.g. installer.msi file shall not be extracted)
+
+    // for global tools we usually download installers and do not want to extract them (e.g. installer.msi file shall
+    // not be extracted)
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


hohwille (changes requested):
Indeed we have this log message duplicated:
* https://github.com/devonfw/IDEasy/blob/f75bfa6e38ae77ba02238703754dae974e86cca2/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java#L136
* https://github.com/devonfw/IDEasy/blob/f75bfa6e38ae77ba02238703754dae974e86cca2/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java#L45
* https://github.com/devonfw/IDEasy/blob/f75bfa6e38ae77ba02238703754dae974e86cca2/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java#L178

So thanks @MattesMrzik for noticing and taking action to improve. 👍 
However, using the `silent` flag to determine the log-level is an intended behavior we have taken from `devonfw-ide`.
The idea is as following:
```
$ devon mvn setup
Version 3.9.6 of tool mvn is already installed
```
This is the intended default behavior. But also consider this:
```
$ devon --debug mvn setup
Version 17.0.8 of tool java is already installed
Version 3.9.6 of tool mvn is already installed
```
By default the user would not expect to get this info about java and that would come every time we run `mvn` if we log it statically on `info` or it would come never by default if we log statically on `debug` so `devon java setup` would give no feedback at all by default what would be confusing:
```
$ devon java setup
$
```
Your change is fine, but you should also add the dynamic logging part from `isInstalledVersion` method to `ToolCommandlet` as a protected method that is used by Local/GlobalToolCommandlet. I missed this on the review of #113 

UPDATE: I just see that in GlobalToolCommandlet we do not have a `version` so you could still do what I suggested but allow `GlobalToolCommandlet` to pass `null` as version parameter and log accordingly.

salimbouch (commented):
Just like @hohwille said we need this dynamic logging otherwise we don't get any output when not in debugger mode. If I'm not mistaken you can move isInstalledVersion to ToolCommandlet and use it both in Global and Local, maybe call it isInstalled since Global Tools don't have any version.

salimbouch (approved):
Looks good to me 👍

hohwille (approved):
OK, PR does not avoid the redundancies but functionally now correct and ready for merge.
I commented on #52 because of the pointless line-wrapping changes, but this can now be merged. Thanks again 👍

PR #185, opened 2024-01-17
#132: untar preserves unix file permissios

### LINKED ISSUE #132: Extracting files on Linux doesn't preserve file permissions
### Expected behavior

On Linux: using the methods untar and unzip in ToolCommandlet to extract files preserves file permissions.

### Actual behavior

These methods do not preserve the file permissions. In particular, the tool binaries are not executable.

### Steps to reproduce

Install a tool like helm or aws on a Linux machine.

### Affected version:

* OS: Linux

MattesMrzik:
Partly fixes #132.
The untar bug of #132 is fixed. Unzipping while preserving file permissions is still missing.
Also wrote a test for untar.


MattesMrzik:
Currently, zip archives are unpacked using `ZipArchiveInputStream`. To preserve file permissions on Unix systems, the method `ZipArchiveEntry.getExternalAttributes()` must be used, but `ZipArchiveInputStream` [is unable to fill this field](https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/archivers/zip/ZipArchiveEntry.html#getExternalAttributes()), you must use ZipFile if you want to read entries using this attribute. 

So either we refactor the `unpack()` method to use [ZipFile](https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/archivers/zip/ZipFile.html) instead of `ZipArchiveInputStream`. Or check if the file system uses posix file permissions and call a method that uses `ZipFile` to extract the permissions.

hohwille (approved):
@MattesMrzik Thanks for this PR. Great work. You nailed it from the start. 👍 
Ready for merge.

hohwille:
@MattesMrzik now the workaround (was it in `AwsCommandlet`) could be removed, correct?

MattesMrzik:
> @MattesMrzik now the workaround (was it in AwsCommandlet) could be removed, correct?

Not yet. Aws uses zip instead of tar and permissions are not yet conserved when extracting zip on unix.

PR #186, opened 2024-01-18
#101: implement create commandlet

### LINKED ISSUE #101: Implement ability to setup IDE instance

salimbouch:
closes #101 

- implements CreateCommandlet class, that creates a new IDEasy instance with the necessary folders to make it recognizable as an IDEasy instance, it then calls update (through the wrapper script) that clones settings, creates workspaces/main and updates software/project
- this PR contains #143 and #150


moritzLanger (approved):
Looks good to me. 

I would suggest adding more javadoc for both public and private Methods. 

The testcoverage is currently decreasing by a big margin and need to be brought up.

PR #200, opened 2024-02-13
#9: background process

### LINKED ISSUE #9: Implement solution for launching background process and emulating start command
In IDEasy we use `ProcessContext` to launch external programs that under the hood uses `ProcessBuilder`.
We need to add support to run a process in "background" such that it runs independent of our IDEasy process (JVM or native-image) so IDEasy CLI can terminate while the launched process will keep running.
This is required e.g. to launch graphical tools such as IDEs (Eclipse, Intellij, VSCode, Android-Studio, GCViewer, JMC, etc.).
Further from bash we are using the windows `start` command to launch some processes like Eclipse to avoid extra CMD windows from opening.
All this has to be solved and implemented in Java and it turns out that such fundamental basics are hard to find as regular searches and even specific tutorials about Process and ProcessBuilder do not say a single word about these aspects.

MustaphaOuchen:
Closes #9


cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -109,72 +110,47 @@ public ProcessResult run(boolean capture, boolean isBackgroundProcess) {
...
-        } else if (this.errorHandling == ProcessErrorHandling.WARNING) {
-          level = this.context.warning();
-        } else {
-          level = this.context.error();
-          level.log("Internal error: Undefined error handling {}", this.errorHandling);
-        }
-        level.log(message);
+
+      int exitCode;
+      if (isBackgroundProcess) {
```
MustaphaOuchen, line 142: This point probably is disputable

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -272,4 +248,90 @@ private String findBashOnWindows() {
...
+
+      this.arguments.clear();
+
+      List<String> newArgs = new ArrayList<>(windowsArgs);
+      newArgs.add(currentCommandToRunInCmd);
+
+      this.arguments.addAll(newArgs);
+
+    } else {
+      // just assume unix system for know
```
salimbouch, line 332:
```suggestion
      // just assume unix system for now
```

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -272,4 +248,90 @@ private String findBashOnWindows() {
...
+      String currentCommandToRunInCmd = this.arguments.stream().map(Object::toString).collect(Collectors.joining(" "));
+      currentCommandToRunInCmd += " | more";
+
+      this.arguments.clear();
+
+      List<String> newArgs = new ArrayList<>(windowsArgs);
+      newArgs.add(currentCommandToRunInCmd);
+
+      this.arguments.addAll(newArgs);
+
```
salimbouch, line 330: Maybe we can do it like this and avoid 2 new local variables? let me know what you think: 
```
      addArgs("|", "more");
      List<String> windowsArgs = List.of("cmd.exe", "/c", "start", "cmd.exe", "/k");
      for (int i = windowsArgs.size() - 1; i >= 0; i--) {
        this.arguments.add(0, windowsArgs.get(i));
      }
```

salimbouch (approved):
Looks good! I added a couple comments, have a look.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
MustaphaOuchen: I would rather prefer simple to understand code than trying reducing variables. I think a reversed loop would make it harder to understand what is done.

salimbouch: You're right, the loop makes it ugly, maybe like this?: 
```
      addArgs("|", "more");
      List<String> windowsArgs = List.of("cmd.exe", "/c", "start", "cmd.exe", "/k");
      List<String> newArgs = new ArrayList<>(windowArgs);
      newArgs.addAll(this.args);
      this.args = newArgs;
```

MustaphaOuchen: Concatenating the passed arguments is currently required, since not the subprocess starts the application, rather the cmd.exe will start our background process independently.

hohwille:
Build still failing with:
> [ERROR] Failures: 
[ERROR]   ProcessContextImplTest.enablingBackgroundProcessShouldNotBeAwaitedAndShouldNotPassStreams:164 
Wanted but not invoked:
processBuilder.redirectOutput(
    <custom argument matcher>
);
-> at java.base/java.lang.ProcessBuilder.redirectOutput(ProcessBuilder.java:793)
However, there were exactly 2 interactions with this mock:
processBuilder.command([]);
-> at com.devonfw.tools.ide.process.ProcessContextImpl.run(ProcessContextImpl.java:129)
processBuilder.start();
-> at com.devonfw.tools.ide.process.ProcessContextImpl.run(ProcessContextImpl.java:131)
[INFO] 
[ERROR] Tests run: 143, Failures: 1, Errors: 0, Skipped: 0

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -272,4 +248,90 @@ private String findBashOnWindows() {
...
+    if (context.getSystemInfo().isWindows() && !isBashScript) {
+
+      this.context.warning(
+          "Currently starting the process as background process in windows will use 'more' command to redirect output to a new cmd window!");
+
+      this.processBuilder.redirectOutput(Redirect.PIPE).redirectError(Redirect.PIPE);
+      List<String> windowsArgs = List.of("cmd.exe", "/c", "start", "cmd.exe", "/k");
+
+      String currentCommandToRunInCmd = this.arguments.stream().map(Object::toString).collect(Collectors.joining(" "));
+      currentCommandToRunInCmd += " | more";
```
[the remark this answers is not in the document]
hohwille: https://www.baeldung.com/linux/detach-process-from-terminal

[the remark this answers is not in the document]
hohwille: Or as an alternative we can run the process without getting the output. In the simplest way we simply then add `> /dev/null`.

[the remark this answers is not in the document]
MustaphaOuchen: So, that means I should assume that all Windows users have a bash and discard the output as "fallback" if no bash installation can be found?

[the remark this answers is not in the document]
hohwille: It was always a per-requisit of devonfw-ide that the user has git installed. This is still the same for IDEasy since we are calling git to clone the settings and the urls as external process. On Windows if you install git, you always have git-bash as well. On Linux and MacOS every user has bash even in a vanilla installation (Except some exotic linux distributions for pocket calculators or C64). So having bash is a given on any supported OS.

Whether the output is suppressed or rushes through the shell of the user should be parameterized. Ideally we introduce an enum state for this:
* suppress output (`>/dev/null 2>&1`)
* feed output into current terminal (run `bash -c` and pass the command with an added `&` like for the option above)
* open a new window with the output (could be done in the way it is implemented currently with CMD on windows but on linux/mac we need to figure out how to do that as we need to find the according terminal app and on MacOS there is not only the default MacOS Terminal but als iTerm2). Feel free to leave this option out for the moment but I can see this quite valuable for the future as well. And maybe we will allow some user specific configuration to configure one of these 3 options as default and per individual tool in the future... Let us always thing further and beyond but also continue with small agile steps from one working state to the next...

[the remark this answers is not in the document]
hohwille: BTW: If we combine all this in an enum, we could also replace the `capture` and `background` flags what prevents inconsistent combinations like `background=true & capture=true` and also increases readability of code:
~~pc.run(false, true, true);~~
```
pc.run(ProcessMode.BACKGROUND_SILENT);
```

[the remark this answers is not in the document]
MustaphaOuchen: I've implemented a ProcessMode enum (Thx for the suggestion, that makes the implenetation way easier to handle than having all possibilities as boolean parameters...) and refactored the code. After some local testing in Linux and Windows bash I think our desired behavior can only be implemented using the unix "disown" (See explanation in added JavaDoc). Instead of redirecting to the null file using '>/dev/null 2>&1', I've just used the  ProcessBuilder.Redirec.DISCARD which does the same OS independently. I've removed the open new window possibilty for now and added a todo comment in the ProcessMode enum.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
hohwille: We need to return `ProcessResult` and it has to have some exit code.
Surely we could also return different magic number for this or change exitCode to `Integer` and have it `null`.
But IMHO this is a reasonable behavior.
In bash the behavior is the same and the script shall continue successfully.
AFAIK if the command itself could not be started (e.g. because the command does not exist) then in bash you get an error but once it could be launched in background it is considered as success.
We can even remove the TODO here.

hohwille (changes requested):
@MustaphaOuchen thanks for your rework. You really mastered this tricky problem and created an excellent solution. Thanks also for adding very detailed and helpful JavaDoc to `ProcessMode`, etc. 👍 
I wanted to resolve all my comments that seemed to be just nice to have so I can even address them after the merge but the last one about Eclipse needs to be addressed before we can merge. Please have a look. If you can resolve all comments even better...

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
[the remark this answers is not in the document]
hohwille: Excellent - you solved this perfectly. Thanks 👍

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
hohwille: I will resolve this conversation, but leave it open for today so you can remove the TODO.

MustaphaOuchen: Done

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -272,4 +245,119 @@ private String findBashOnWindows() {
...
+      String findBashOnWindowsResult = findBashOnWindows();
+      if (findBashOnWindowsResult != null) {
+
+        bash = findBashOnWindowsResult;
+
+        // windows path must be converted to unix format and executable
+        String executablePath = this.arguments.get(0);
+        executablePath = convertWindowsPathToUnixPath(executablePath);
+
+        commandToRunInBackground = this.arguments.subList(1, this.arguments.size()).stream().map(Object::toString)
```
MustaphaOuchen, line 333: Note before merging anything:
I've made the wrong assumption that windows paths can only be the binary. Since in eclipse we may give paths as arguments, these must be also be converted to a Unix path before being passed to the shell.

hohwille: Ah, OK, got it: Before the command and args we executed directly. Now we delegate this to Bash and hence we (might) need to convert paths to bash syntax on Windows.
From my tests bash is able to handle this but only backslashes cause trouble:
```
$ bash -c 'cat D:\\projects\\devonfw\\setup'
#!/bin/bash

cd "$(dirname "${BASH_SOURCE:-$0}")" || exit 255
echo "Setting up your devonfw-ide in ${PWD}"
source scripts/devon ide setup "${@}"
```

hohwille: OK. Does not work from CMD:
```
C:\Users\hohwille>"C:\Program Files\Git\bin\bash.exe" -c "cat D:\projects\devonfw\setup"
cat: 'D:projectsdevonfwsetup': No such file or directory

C:\Users\hohwille>"C:\Program Files\Git\bin\bash.exe" -c "cat D:\\projects\\devonfw\\setup"
cat: 'D:projectsdevonfwsetup': No such file or directory
```

So you are right we need a workaround here.
See here for such a conversion that I have already implemented:
https://github.com/devonfw/IDEasy/blob/fdf34479b0ad6107e3908686242ced463299dff2/cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java#L206-L222

hohwille: Feel free to make public and static if you want to reuse.

cli/src/main/java/com/devonfw/tools/ide/tool/eclipse/Eclipse.java
```diff
@@ -80,7 +81,12 @@ protected ProcessResult runEclipse(boolean log, String... args) {
     Path javaPath = getCommandlet(Java.class).getToolBinPath();
     pc.addArg("-vm").addArg(javaPath);
     pc.addArgs(args);
-    return pc.run(log, false);
+
+    if (log) {
+      return pc.run(ProcessMode.BACKGROUND);
+    } else {
+      return pc.run(ProcessMode.BACKGROUND_SILENT);
+    }
```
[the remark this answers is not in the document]
hohwille: So IMHO we can pass `ProcessMode` as argument to this method instead of `boolean log`.
Then from `installPlugin` we can exec in foreground and from `run` we exec in background.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
MustaphaOuchen: @hohwille  I've fixed my wrong assumption with commit https://github.com/devonfw/IDEasy/pull/200/commits/38951e71a0af93c0d827713627c2bdda1e6766a9 and reused. Bit I have still the problem that if IDEasy is run via cmd, even though bash is used, the output is not printed on the console. But the application starts. This problem does not occur when we run IDEasy from bash.

cli/src/main/java/com/devonfw/tools/ide/tool/eclipse/Eclipse.java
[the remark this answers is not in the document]
MustaphaOuchen: Has been refactored in commit https://github.com/devonfw/IDEasy/pull/200/commits/a1ca068b614e6f2fbae4f97e87bffab762f24981

hohwille (approved):
@MustaphaOuchen thanks for the rework. Great job. Now ready for merge. 👍

PR #202, opened 2024-02-13
add JasyptUrlUpdater in UpdateManager

mvomiero:
added missing JasyptUrlUpdater object in UpdateManager


hohwille (approved):
@mvomiero thanks for fixing. 👍

PR #205, opened 2024-02-15
#31: implement tool commandlet for node package manager

### LINKED ISSUE #31: Implement ToolCommandlet for Node Package Manager
Create `ToolCommandlet` for Node Package Manager
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/npm

ndemirca:
fixes #31 

Unit tests are missing till a test infrastructure will be created


cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -1,21 +1,12 @@
...
 import java.net.http.HttpClient;
 import java.net.http.HttpClient.Redirect;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
-import java.nio.file.FileSystemException;
-import java.nio.file.Files;
-import java.nio.file.LinkOption;
-import java.nio.file.NoSuchFileException;
-import java.nio.file.Path;
+import java.nio.file.*;
```
aBega2000, line 9: I think the reformatting has changed this, but I think we should keep only the imports we use, not java.io.*

aBega2000 (commented):
Everything seems fine, except the one change that seems like the reformatting did. Thank you for this PR.

Issue #209, opened 2024-02-16
CI Build fails in GitHub


aBega2000:
### Expected behavior

We want the CI build to run successfully in Github, so that everything can work properly

### Actual behavior

The CI build fails and there is an error in the tests, namely:

[ERROR]   InstallCommandletTest.setUp:34 » FatalStartup java.lang.RuntimeException: java.io.IOException: Failed to bind to /0.0.0.0:1111

### Comments/Hints:

It seems like there is an error with WireMock, when we try to bind to port 1111 in the URLs. 
First a Quick Fix will be implemented, then a generally more reliable solution will be found and implemented

hohwille:
Reopening: We have only applied a quickfix (hack) with PR #210 so our CI gets green again.
However, we need to find a real solution:
* [This commit](https://github.com/devonfw/IDEasy/pull/210/commits/dcbece25c443edc1b245fa1285e9a98358fe9fda) is a brute hack to make it work that needs to be fixed properly.
* We should not use wiremock on different fixed ports in different tests and then end up with already allocated ports if tests run in parallel, etc. Either we can reduce everything to a single wiremock test (I guess that will not be possible) or we need to prevent that wiremock tests can run in parallel during the build or we need to make wiremock chose a free random port and then somehow inject that random port-number into the test-case to derive the correct download URL using that port rather than hardcoding it.

hohwille:
IMHO this issue is already fixed with PR #238 etc.

PR #210, opened 2024-02-16
#209: Quick Fix for CI build failure

aBega2000:
Fixes: #209 

This is a quick fix for the failure of the CI build.  In this PR is just the port number for JMC installation changed, from 1111 to 1112.

Another generally better solution is in investigation and will be implemented soon.

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/jmc/Jmc.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/jmc/Jmc.java
@@ -12,6 +12,7 @@ import com.devonfw.tools.ide.context.IdeContext;
 import com.devonfw.tools.ide.io.FileAccess;
 import com.devonfw.tools.ide.tool.LocalToolCommandlet;
 import com.devonfw.tools.ide.tool.ToolCommandlet;
+import com.devonfw.tools.ide.tool.java.Java;
 
 /**
  * {@link ToolCommandlet} for <a href="https://www.oracle.com/java/technologies/jdk-mission-control.html">JDK Mission
@@ -32,8 +33,7 @@ public class Jmc extends LocalToolCommandlet {
   @Override
   public boolean doInstall(boolean silent) {
 
-    // TODO The Java dependency should be also implemented, while the support for dependencies is not yet implemented
-    // getCommandlet(Java.class).install();
+    getCommandlet(Java.class).install();
     return super.doInstall(silent);
   }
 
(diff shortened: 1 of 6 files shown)


cli/src/main/java/com/devonfw/tools/ide/tool/jmc/Jmc.java
```diff
@@ -33,7 +32,7 @@ public Jmc(IdeContext context) {
   @Override
   public boolean doInstall(boolean silent) {
 
-    getCommandlet(Java.class).install();
+    // getCommandlet(Java.class).install();
```
MustaphaOuchen, line 35: I would probably add here a todo comment, so that it won't get lost if we have [#131](https://github.com/devonfw/IDEasy/issues/131) implemented. Please also add a todo comment in the jmc test concerning solving the binding port problem. Thank you!

hohwille: This change is for #131 and does not belong into this PR. We should revert this.

cli/src/test/java/com/devonfw/tools/ide/tool/Jmc/JmcTest.java
```diff
@@ -1,4 +1,4 @@
-package com.devonfw.tools.ide.Jmc;
+package com.devonfw.tools.ide.tool.Jmc;
```
hohwille, line 1: Please fix the package statement to fit to the path the file is located in (Eclipse would give you an error for this inconsistency, it seems that IntelliJ is very lax in this regard and does not complain). Also please rename `Jmc` package folder to `jmc` as Java-Package segments need to be `lowercase` and should avoid any `UPPERCASE` letters.
```suggestion
package com.devonfw.tools.ide.tool.jmc.Jmc;
```

hohwille: I was able to solve this on github with just my browser now.

hohwille:
The build is now red:
I cannot download java since that is using wiremock on port `1111` but also needed due to dependencies.

https://github.com/devonfw/IDEasy/blob/716614ed70b493406ebcb1ba6ecc2cdf0f164f92/cli/src/test/resources/ide-projects/_ide/urls/java/java/17.0.6/linux_x64.urls

Now I got your change to comment out the dependency to java with the TODO command. I assumed this was a change from #131 but it seems I was wrong... Sorry, that I now broke it, but I wanted to get this done asap and could not reach you.

hohwille:
Should be fixed now and then I can finally merge.

Issue #207, opened 2024-02-16
Restructure testing folder structure so that for each commandlet/testcase a seperate _ide and project folder exist


MustaphaOuchen:
Simply alter the folder hierarchy so that you have:

YourProjectCase (e.g. basic)
YourProjectCase -> _ide
YourProjectCase -> project

Change tests / AbstractIdeContextTest class accordingly to use right directories.

PR #211, opened 2024-02-16
#207: refactor test folder structure

MustaphaOuchen:
Fixes: #207 

Changes the folder testing infrastructure so that each test project can have its own _ide and project folder.


cli/src/test/java/com/devonfw/tools/ide/context/AbstractIdeContextTest.java
```diff
@@ -51,25 +51,27 @@ protected static IdeTestContext newContext(String projectName, String projectPat
...
     if (copyForMutation) {
-      Path projectDir = PATH_PROJECTS_COPY.resolve(projectName);
+      Path projectDir = PATH_PROJECTS_COPY.resolve(projectTestCaseName).resolve("project");
       FileAccess fileAccess = new FileAccessImpl(IdeTestContextMock.get());
       fileAccess.delete(projectDir);
       fileAccess.mkdirs(PATH_PROJECTS_COPY);
       fileAccess.copy(sourceDir, projectDir, FileCopyMode.COPY_TREE_OVERRIDE_TREE);
-      fileAccess.copy(PATH_PROJECTS.resolve(IdeContext.FOLDER_IDE), PATH_PROJECTS_COPY.resolve(IdeContext.FOLDER_IDE),
+      fileAccess.copy(PATH_PROJECTS.resolve(projectTestCaseName).resolve(IdeContext.FOLDER_IDE),
+          PATH_PROJECTS_COPY.resolve(projectTestCaseName).resolve(IdeContext.FOLDER_IDE),
```
hohwille, line 74: It would now make more sense and also be easier to just copy `projectTestCaseName` instead of its subfolders `project` and `_ide`.

MustaphaOuchen: Yeah, that's right. I've refactored it

cli/src/test/java/com/devonfw/tools/ide/context/AbstractIdeContextTest.java
```diff
@@ -51,42 +51,47 @@ protected static IdeTestContext newContext(String projectName, String projectPat
...
+   *        the test project in {@link #PATH_PROJECTS}. E.g. "basic".
    * @param projectPath the relative path inside the test project where to create the context.
    * @param copyForMutation - {@code true} to create a copy of the project that can be modified by the test,
    *        {@code false} otherwise (only to save resources if you are 100% sure that your test never modifies anything
    *        in that project.
    * @return the {@link IdeTestContext} pointing to that project.
    */
-  protected static IdeTestContext newContext(String projectName, String projectPath, boolean copyForMutation) {
+  protected static IdeTestContext newContext(String projectTestCaseName, String projectPath, boolean copyForMutation,
+      boolean debug) {
```
hohwille, line 63: what is `debug` flag? No JavaDoc given.
Please always prefer to add a new method signature but keep the existing one to prevent merge conflicts with other feature branches, etc.
I will add this to our coding conventions as I am repeating this from one review to the next.

MustaphaOuchen: That was just for testing, since somehow locally in windows the tests are green and the pipeline fails. The moving of files seems to have caused problems.

hohwille (approved):
@MustaphaOuchen thanks for this PR. Another step forward for our test-infrastructure. 👍 
As a follow up, I would suggest that we offer the ability to pass something like "basic/other-project" as argument instead of "basic" so if just a test-root folder is specified it defaults to `project` (`basic` --> `basic/project`) but if the given project contains a slash then that is interpreted as the relative path to the project itself and the toplevel folder of that given relative is the IDE_ROOT.
However, I will now merge this as it is a good improvement and we can do the next improvement as a separate PR.

Issue #34, opened 2023-09-22
Implement ToolCommandlet for SonarQube


hohwille:
Create `ToolCommandlet` for SonarQube
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/sonar

PR #216, opened 2024-02-20
#34: Implemented ToolCommandlet for SonarQube 

KGoroll1:
closes https://github.com/devonfw/IDEasy/issues/34

The sonar commandlet has been implemented.

Addressed issue: https://github.com/devonfw/IDEasy/issues/34


cli/src/main/java/com/devonfw/tools/ide/tool/sonar/Sonar.java
```diff
@@ -0,0 +1,100 @@
...
+  protected void initProperties() {
+
+    // Empty on purpose
+  }
+
+  @Override
+  public boolean install(boolean silent) {
+
+    getCommandlet(Java.class).install();
+    return super.install(silent);
```
hohwille, line 42: you may need to override postinstall and set the executable flags for the linux/macOS scripts since the download is a ZIP file and ZIP cannot represent executable flags.
https://github.com/devonfw/ide-urls/blob/master/sonar/sonar/10.4.0.87286/urls

I just checked the download from the sonarqube website and it also only gives me a ZIP. Broken by design of sonarqube. So we need to set the X-flag manually on Linux / Mac.

hohwille (approved):
@KGoroll1 Thank your for your PR. Excellent work - looking very good 👍 

I left you a hint what I guess to fix the linux/mac error.

hohwille (approved):
@KGoroll1 thanks for your PR. Great work and looking good to me 👍 
Regarding the documentation of the commandlets we recently agreed that we stop to redundantly document commandlets in asciidoc as we now have a better help and that should be the source of truth so we can avoid the redundant overhead maintaining the asciidocs that do not add further value.
I will have to check our DoD. In IDEasy the DoD needs to be different to the one in devonfw-ide. I assume that we do not have a DoD in IDEasy so I will have to create a story to add it.

Issue #219, opened 2024-02-21
Wrong executable extension in linux


mvomiero:
### Expected behavior

the recognition of the executable extension should depend from the operating system. For example, in Windows, a `.cmd` extension is preferred, while in Unix systems there is generally no extension for executables. (see example)

### Actual behavior

In the class SystemPath.java, there is `private static final List<String> EXTENSION_PRIORITY = List.of(".exe", ".cmd", ".bat", ".msi", ".ps1", "");` 

This priority list creates problems in Linux because "" should have priority for example on ".cmd". 

### Steps to reproduce (bug)

**EXAMPLE:** Commandlet mvn

1. in `cli/`, execute in WINDOW and LINUX
`devon mvn exec:java -Dexec.mainClass="com.devonfw.tools.ide.cli.Ideasy" -Dexec.args="--debug mvn --help"`

2. The executable generally differs from the operating system: (output from --debug)

* WIN: `Running command 'C:\Projects\devonfw\software\mvn\bin\mvn.cmd' with arguments '--help' ...`
_works without problems_
* LINUX: `Running command '/home/mvomiero/Projects/devonfw/software/mvn/bin/mvn.cmd' with arguments '--help' ...`
program exits with exit code 2:
`Running command '/home/mvomiero/Projects/devonfw/software/mvn/bin/mvn.cmd' with arguments '--help' failed with exit code 2!`

**_!!! the executable is the same, even though in linux there is a cmd without extension that should be executed. Problem is, that in the EXTENSION_PRIORITY list, ".cmd" has priority._** 

### NOTE: there are cases where the extension recognition properly works

**EXAMPLE**: commandlet Java

1. in `cli/`, execute in WINDOW and LINUX
`devon mvn exec:java -Dexec.mainClass="com.devonfw.tools.ide.cli.Ideasy" -Dexec.args="--debug java --help"`

2. The executable generally differs from the operating system: (output from --debug)

* WIN: ` Running command 'C:\Projects\devonfw\software\java\bin\java.exe' with arguments '--help' ...`
* LINUX: `Running command '/home/mvomiero/Projects/devonfw/software/java/bin/java' with arguments '-h' '-e' '-l' '-p' ...`

### Comments:

Don't know if the problem is actually the extension priority list or how the executables are installed in `_ide/software/default/`

### Affected version:

* OS: Linux/ - I still have to test Mac

PR #220, opened 2024-02-21
#219: wrong executable extension linux

mvomiero:
Fixes: #219

**NOTE: work in progress**

#### Approach:
different executable extension priority lists depending from the os.

#### Changes:
Splits the list EXTENSION_PRIORITY in EXTENSION_PRIORITY_WIN and EXTENSION_PRIORITY_WIN to have different priorities for the choice of the executable extension depending from the OS.

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
@@ -28,7 +28,8 @@ public class SystemPath {
 
   private final IdeContext context;
 
-  private static final List<String> EXTENSION_PRIORITY = List.of(".exe", ".cmd", ".bat", ".msi", ".ps1", "");
+  private static final List<String> EXTENSION_PRIORITY_WIN = List.of(".exe", ".cmd", ".bat", ".msi", ".ps1", "");
+  private static final List<String> EXTENSION_PRIORITY_UNIX = List.of(".exe", "", ".cmd", ".bat", ".msi", ".ps1");
 
   /**
    * The constructor.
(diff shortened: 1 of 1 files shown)


cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
```diff
@@ -28,7 +28,9 @@ public class SystemPath {
 
   private final IdeContext context;
 
-  private static final List<String> EXTENSION_PRIORITY = List.of(".exe", ".cmd", ".bat", ".msi", ".ps1", "");
+  private static final List<String> EXTENSION_PRIORITY_WIN = List.of(".exe", ".cmd", ".bat", ".msi", ".ps1", "");
+
+  private static final List<String> EXTENSION_PRIORITY_UNIX = List.of(".exe", "", ".cmd", ".bat", ".msi", ".ps1");
```
aBega2000, line 33: I just wonder if "" should be placed in the first place, and also, are the extensions that are only for WINDOWS also relevant to the list for UNIX? Does that also affect the WIN list, in the same sense?

aBega2000 (commented):
Thank you for this PR and for finding this bug. I have left a comment which I think should be addressed and researched a bit as well.

hohwille:
I fixed this in https://github.com/devonfw/IDEasy/pull/238/commits/78151534ae74dedb3637f380ffc834a56ded499d in PR #238

Usually I very much prefer that we have small stories and PRs fixing exactly that story but my time is very limited and the pressure is high to get this test-infrastructure available that blocks the entire team, so I assume, my merge will come first and thereby replace some other PRs.

hohwille:
IMHO this PR can now be closed. @mvomiero thanks for creating the issue and this PR. I could fix this very fast based on your work 👍

PR #221, opened 2024-02-21
#63: Creating the ide wrapper script

### LINKED ISSUE #63: Create ide wrapper script
As a devonfw-ide user, I want to be able to use the new `ide` CLI seamlessly so that I can continue to work like I was used to with devonfw-ide.
That is if I just call `ide` this will set my environments according to the IDEasy project I am currently located in.

For this we have already implemented our Java IDE and build-process to create native image binaries. All this already supports an environment commandlet that prints out all the variables to set.
However, a process (our ide.exe) cannot set/change the environments of the current shell (e.g. CMD, bash, or powershell) the user is currently running. Therefore we need a wrapper script similar like what we had before:
* CMD: https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/devon.bat
* Bash: https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/devon

Instead that new script should be quite different:
* instead of `devon[.bat]` it should be named `ide[.bat]`
* it should not contain the logic to find DEVON_IDE_HOME or other things but be even more simplistic:
* first it should check if arguments are given. If that is the case, then it shall call the native `ideasy[.exe]` (our native image from GraalVM) passing exactly those arguments the user had given. This can log output and interact with the user via CLI as usual.
* second it should afterwards always run the native `ideasy[.exe]` again with `env` as argument to trigger the `EnvironmentCommandlet`. In case we are running from bash (and we are on windows), then a second parameter `--bash` shall be supplied (`ideasy env --bash`). The output of this should be evaluated in an analogue way as we did to load the `devon.properties` in Bash/CMD (https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/functions-core#L9). Maybe we can even do KISS and redirect the entire env output to a pipe calling the `eval` command.

To simplify things you may develop and provide `ide` (bash) and `ide.bat` in separate steps as individual PRs.
As those files need to be included into our release package. My suggestion would be:
* add the scripts to a folder `cli/src/main/package`
* adapt the `cli/pom.xml` with a `build/resources` section that copies (and filters) the files from that folder to something like `target/package`. Please properly use [maven variables](https://cwiki.apache.org/confluence/display/MAVEN/Maven+Properties+Guide) instead of hardcoded `target` folder (that may be redirected e.g. to `eclipse-target`).
* Adapt the assembly config [exec.xml](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/assembly/exec.xml) (also consider renaming to `package.xml`) to ensure that the `target/package` folder is finally archived to our tar.gz file for the release.
* Adapt the [release](https://github.com/devonfw/IDEasy/blob/main/.github/workflows/release.yml) and [nightly-build](https://github.com/devonfw/IDEasy/blob/main/.github/workflows/nightly-build.yml) process to properly copy the native binaries into the correct places (`target/package/bin/`). Also rename the native image to `ideasy[.exe]` as otherwise we have a name clashing between the `ide` wrapper script and the `ideasy` (native image binary).

Again if you want to simplify. You can first place the scripts into the new location as a PR (first bullet point above) and later create a PR to integrate it and improve the release package.

Clarification of potential questions:
* How can the `ide[.bat]` script find the native binary? The most simple approach that I can think of would be that we have a central installation of IDEasy, that is inside IDE_ROOT that could be set as a global environment variable on the initial installation and always be on the PATH (`${IDE_ROOT}/bin` so e.g. `D:\projects\_ide\bin`) so we do not have to look for it (we have already agreed on this approach). However, the implication would be that if you install an update, you will update the CLI globally on your machine. If there are (breaking?) changes, this will apply to all IDE installations. An alternative would be to keep the old logic to find IDE_HOME in the script and have the ide-cli in a `bin` (or `scripts`) folder in IDE_HOME. After using devonfw-ide for years and having many parallel IDE instances I learned that updating is kind of tedious as I have to run the `ide update scripts` more than 10 times until all my installations are up-to-date. Therefore a global installation would be my first thought. Also this avoids redundancy and waste of disk-space if the `ide-cli.exe` is already 55MB. IMHO we just need to learn more about caring for downward compatibility and then this approach should be fine.

aBega2000:
Fixes #63


cli/pom.xml
```diff
@@ -105,6 +105,28 @@
...
+        <version>3.3.1</version>
+        <executions>
+          <execution>
+            <id>copy-resources</id>
+              <phase>validate</phase>
+            <goals>
+              <goal>copy-resources</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${basedir}/target/package</outputDirectory>
```
hohwille, line 119:
```suggestion
              <outputDirectory>${project.build.directory}/package</outputDirectory>
```
This is more clean as we reconfigure `target` folder in a profile.
See https://cwiki.apache.org/confluence/display/MAVEN/Maven+Properties+Guide

cli/src/main/assembly/exec.xml
```diff
@@ -23,5 +23,9 @@
         <include>${imageName}</include>
       </includes>
     </fileSet>
+    <fileSet>
+      <directory>${project.build.directory}/package</directory>
+      <outputDirectory>./target/package</outputDirectory>
+    </fileSet>
```
hohwille, line 29:
```suggestion
    <fileSet>
      <directory>${project.build.directory}/package</directory>
      <outputDirectory>.</outputDirectory>
    </fileSet>
```
Also remove the above `fileSet` that should be replaced by your new one.

cli/src/main/package/ide
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+# Check if arguments are given
+if [ -n "${1}" ]; then
+  # Call native ideasy.exe with user-provided arguments
+  ./ideasy "$@"
```
hohwille, line 6: added error handling here:
```
rc=$?
if [ "${rc}" != 0 ]; then
  exit ${rc}
fi
```
As our script will be sourced, we might want to avoid the local variable. I have not tested if $? changes by `if` statement... But nice to have for further improvements after merge...

cli/src/main/package/ide
```diff
@@ -0,0 +1,18 @@
...
+
+UNAME=$(uname)
+
+if [ "$UNAME" == "Linux" ] ; then
+	./ideasy env
+elif [ "$UNAME" == "Darwin" ] ; then
+	./ideasy env
+elif [[ "$UNAME" == CYGWIN* || "$UNAME" == MINGW* ]] ; then
+	./ideasy env --bash
+fi
```
hohwille, line 17:
```suggestion
if [ "${OSTYPE}" = "cygwin" ] || [ "${OSTYPE}" = "msys" ]; then
  ./ideasy env --bash
else
  ./ideasy env
fi
```
See https://github.com/devonfw/ide/blob/cdf4af5de21489f3fdcbfdab41af858822e0d598/scripts/src/main/resources/scripts/functions#L1785

Please also avoid tabs and use editor compliant with our code-style (e.g. Intellij or configure your editor like Notepad++ accordingly).

hohwille (changes requested):
@aBega2000 thanks for your PR. Looks good - nice work👍 
Did you actually test the github actions or did you just "guess" these changes? I assume you did some tests to find that `natives` copy logic in two workflows :)

I left some review comments for rework. Should be easy to complete.

hohwille:
https://github.com/devonfw/IDEasy/issues/63#issuecomment-1963770443

aBega2000:
I have made the required changed, and the commits contain all the changes that we also discussed. I also managed to fix the diff war by setting the core.autocrlf to false. Therefore the file pom.xml now shows only the changes from the main branch file. I also adapted the Test for the IdeHomeRequired, so that the PR passes all the tests.

cli/src/main/package/ide
hohwille: I changed it using `set -e` to solve the problem.

cli/src/main/assembly/exec.xml
hohwille: I updated the file removing the obsolete fileSet's.

hohwille (approved):
looking good now. Ready for merge.

PR #222, opened 2024-02-22
PR for refactoring test infrastructure #208

### LINKED ISSUE #208: Mock out ToolRepository so that commandlets can be tested with unit tests
- We want to test commandlets without being forced to use wiremock for each test.
- We need to mock out the downloading process and provide test data with a specific syntax, e.g. VersionToolOs...
- Instead of downloading, we simply copy the files recursively
- also Implement an example test case for an existing commandlet, e.g. Jmc
- Use the new folder structure  (see [#207](https://github.com/devonfw/IDEasy/issues/207))

MustaphaOuchen:
Closes #208 

Consider that https://github.com/devonfw/IDEasy/pull/228 should be merged first.

JmcTest class contains example tests


cli/src/test/java/com/devonfw/tools/ide/tool/jmc/JmcTest.java
```diff
@@ -79,24 +99,123 @@ private void setupMockServerResponse(String testUrl, String contentType, String
...
+    Jmc commandlet = new Jmc(context);
+    commandlet.setCommandletFileExtractor(commandLetExtractorMock);
+
+    // act
+    commandlet.install();
+
+    // assert
+    performPostInstallAssertion(context);
+  }
+
```
MustaphaOuchen, line 139: Comment in order not to forget to enable this test.

cli/src/test/java/com/devonfw/tools/ide/tool/jmc/JmcTest.java
```diff
@@ -79,24 +99,123 @@ private void setupMockServerResponse(String testUrl, String contentType, String
...
+  // TODO: Enable test as soon https://github.com/devonfw/IDEasy/pull/228 is merged
+  @Test
+  @Disabled
```
MustaphaOuchen, line 142: Comment in order not to forget to enable this test.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CommandletFileExtractorImpl.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+        fileAccess.copy(appPath, tmpDir);
+        pc.addArgs("detach", "-force", mountPath);
+        pc.run();
+      } else if ("msi".equals(extension)) {
+        this.context.newProcess().executable("msiexec").addArgs("/a", file, "/qn", "TARGETDIR=" + tmpDir).run();
+        // msiexec also creates a copy of the MSI
+        Path msiCopy = tmpDir.resolve(file.getFileName());
+        fileAccess.delete(msiCopy);
+      } else if ("pkg".equals(extension)) {
```
jan-vcapgemini, line 71: What about moving each setup file type into a separate extract method e.g. extractWindowsMsi, extractLinuxPkg etc.?
Maybe create an enum for each file type too?

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -595,10 +600,26 @@ public DirectoryMerger getWorkspaceMerger() {
     return this.workspaceMerger;
   }
 
+  public Path getDefaultExecutionDirectory() {
```
jan-vcapgemini, line 603: Javadoc missing.

jan-vcapgemini (changes requested):
Looks really nice, can't await to put it to use in my tool tests.
I've added some small change requests and suggestions/questions.

cli/src/test/java/com/devonfw/tools/ide/tool/jmc/JmcTest.java
MustaphaOuchen: Done

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
MustaphaOuchen: Done

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -234,7 +239,7 @@ public AbstractIdeContext(IdeLogLevel minLogLevel, Function<IdeLogLevel, IdeSubL
     this.downloadPath = this.userHome.resolve("Downloads/ide");
     this.variables = createVariables();
     this.path = computeSystemPath();
-    this.defaultToolRepository = new DefaultToolRepository(this);
+    this.defaultToolRepository = toolRepository == null ? new DefaultToolRepository(this) : toolRepository;
```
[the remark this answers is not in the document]
MustaphaOuchen: Done

cli/src/test/java/com/devonfw/tools/ide/commandlet/CommandLetExtractorMock.java
```diff
@@ -0,0 +1,52 @@
...
+import com.devonfw.tools.ide.io.FileAccess;
+
+/**
+ * Implementation of {@link CommandletFileExtractor} for testing.
+ */
+public class CommandLetExtractorMock implements CommandletFileExtractor {
+
+  private final IdeContext context;
+
+  public CommandLetExtractorMock(IdeContext context) {
```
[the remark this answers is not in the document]
MustaphaOuchen: Done

cli/src/test/java/com/devonfw/tools/ide/tool/jmc/JmcTest.java
```diff
@@ -79,24 +99,123 @@ private void setupMockServerResponse(String testUrl, String contentType, String
...
+  public void jmcShouldRunExecuteableSuccessfully() throws IOException {
```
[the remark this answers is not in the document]
MustaphaOuchen: Done

cli/src/test/java/com/devonfw/tools/ide/tool/jmc/JmcTest.java
```diff
@@ -39,33 +50,42 @@ static void tearDown() throws IOException {
     server.shutdownServer();
   }
 
+  @BeforeEach
+  public void setUpStream() {
+
+    // System.setOut(new PrintStream(outputStreamCaptor));
```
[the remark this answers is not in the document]
MustaphaOuchen: Yeah, good catch- done.

cli/src/test/java/com/devonfw/tools/ide/tool/jmc/JmcTest.java
```diff
@@ -25,11 +33,14 @@ public class JmcTest extends AbstractIdeContextTest {
 
   private static WireMockServer server;
 
-  private static Path resourcePath = Path.of("src/test/resources");
+  private static final Path RESOURCE_PATH = Path.of("src/test/resources");
 
   @BeforeAll
   static void setUp() throws IOException {
+
+    // TODO use random port number and create url file dynamically in project
```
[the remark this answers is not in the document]
MustaphaOuchen: Done

cli/src/test/java/com/devonfw/tools/ide/repo/ToolRepositoryMock.java
```diff
@@ -0,0 +1,127 @@
...
+    ToolDataForMock toolData = toolToToolData.get(tool);
+    String windowsFolderName = toolData.windowsFolderName;
+    String linuxFolderName = toolData.linuxFolderName;
+    String macFolderName = toolData.macFolderName;
+
+    Path baseDownloadMockPath = PROJECTS_TARGET_PATH.resolve(projectTestCaseName).resolve(MOCK_DOWNLOAD_FOLDER);
+    String mockProgram = "";
+
+    if (context == null) {
+      throw new IllegalStateException("Please set a IdeContext!");
```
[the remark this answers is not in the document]
MustaphaOuchen: Done

cli/src/test/java/com/devonfw/tools/ide/repo/ToolRepositoryMock.java
```diff
@@ -0,0 +1,127 @@
...
+  private IdeContext context;
+
+  // Save data of multiple tools to provide different faking for different tools
+  private final Map<String, ToolDataForMock> toolToToolData;
+
+  private record ToolDataForMock(String version, String windowsFolderName, String linuxFolderName,
+      String macFolderName) {
+  }
+
+  ;
```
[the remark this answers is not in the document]
MustaphaOuchen: Done

cli/src/main/java/com/devonfw/tools/ide/commandlet/CommandletFileExtractorImpl.java
MustaphaOuchen: Done with https://github.com/devonfw/IDEasy/pull/222/commits/c0399aa79ca64f6e2cbc36bb9fc87ae5fe73f970

cli/src/main/java/com/devonfw/tools/ide/commandlet/FileExtractor/ExctractorFileType.java
```diff
@@ -0,0 +1,5 @@
+package com.devonfw.tools.ide.commandlet.FileExtractor;
+
+enum ExtractorFileType {
```
jan-vcapgemini, line 3: Thanks for adding these, some javadoc would be nice though.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CommandletFileExtractorImpl.java
jan-vcapgemini: Looks much cleaner now, thanks.

cli/src/test/java/com/devonfw/tools/ide/tool/jmc/JmcTest.java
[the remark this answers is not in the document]
jan-vcapgemini: I meant something different, but ok ^^

jan-vcapgemini (approved):
Looks all fine to me. Added some comments for very small changes but I think this can be moved into review now.

cli/src/test/java/com/devonfw/tools/ide/tool/jmc/JmcTest.java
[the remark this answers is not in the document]
MustaphaOuchen: Ah, spelling, my bad :)

cli/src/main/java/com/devonfw/tools/ide/commandlet/FileExtractor/ExctractorFileType.java
MustaphaOuchen: Done

cli/src/main/java/com/devonfw/tools/ide/commandlet/FileExtractor/CommandletFileExtractor.java
```diff
@@ -0,0 +1,31 @@
+package com.devonfw.tools.ide.commandlet.FileExtractor;
```
hohwille, line 1: this is discouraged/anti-pattern to have Capitalized package segment.
It took me some time to understand this as it looks like the qualified name of a class `FileExtract` but as it turns out this is a package segment.
Never do such things in Java. Package-segments should *always* be `lowercase`.
IMHO this code belongs to the `io` package where also `FileAccess` is located and where existing methods like untar already exist.

cli/src/main/java/com/devonfw/tools/ide/commandlet/FileExtractor/CommandletFileExtractorImpl.java
```diff
@@ -0,0 +1,157 @@
...
+  public void extract(Path file, Path targetDir, boolean isExtract) {
+
+    FileAccess fileAccess = this.context.getFileAccess();
+    if (isExtract) {
+      Path tmpDir = fileAccess.createTempDir("extract-" + file.getFileName());
+      this.context.trace("Trying to extract the downloaded file {} to {} and move it to {}.", file, tmpDir, targetDir);
+      String extension = FilenameUtil.getExtension(file.getFileName().toString());
+      this.context.trace("Determined file extension {}", extension);
+      TarCompression tarCompression = TarCompression.of(extension);
+      ExtractorFileType fileType = ExtractorFileType.valueOf(extension.toUpperCase());
```
hohwille, line 47: this will throw an exception at runtime if the extension is not know here.
At least then the following code can never be reached anymore:
```
} else {
        throw new IllegalStateException("Unknown archive format " + extension + ". Can not extract " + file);
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/FileExtractor/CommandletFileExtractor.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.nio.file.Path;
+
+import com.devonfw.tools.ide.tool.ToolCommandlet;
+
+/**
+ * {@link CommandletFileExtractor} class which handles the extraction of downloaded installations of a
+ * {@link ToolCommandlet}.
+ */
+public interface CommandletFileExtractor {
```
hohwille, line 11: Why is this named `CommandletFileExtractor`? It does not seem to be commandlet specific.
Please rename it to `FileExtractor`.

cli/src/main/java/com/devonfw/tools/ide/commandlet/FileExtractor/CommandletFileExtractor.java
hohwille: Also to me it makes sense to move all extraction into this interface and implementation.
If we introduce a `FileExtractor` then untar and unzip should not stay in `FileUtil` but be moved here IMHO.
You have methods like `extractPkg` or `extractWindowsMsi` so why not having all these methods together here?
Also it would be nice to align the names then `unzip` --> `extractZip`, etc.

cli/src/main/java/com/devonfw/tools/ide/commandlet/FileExtractor/CommandletFileExtractorImpl.java
```diff
@@ -0,0 +1,157 @@
...
+          return getProperInstallationSubDirOf(subFiles[0]);
+        }
+      }
+      return path;
+    } catch (IOException e) {
+      throw new IllegalStateException("Failed to get sub-files of " + path);
+    }
+  }
+
+  private void extractLinuxDmg(Path file, Path tmpDir) {
```
hohwille, line 114: `extractLinuxDmg` is the incorrect name as `DMG` files are specific for MacOS.
IMHO I would strip out `Windows` and `Linux` from these methods and instead add such hints to the JavaDoc of the methods in the interface making all these methods public.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -590,6 +601,25 @@ public DirectoryMerger getWorkspaceMerger() {
     return this.workspaceMerger;
   }
 
+  /**
+   *
+   * @return the {@link #defaultExecutionDirectory} the directory in which a command process is executed.
```
hohwille, line 606: If you generate the JavaDoc as HTML this link will not work as private members are not part of the API-Doc and therefore the link is broken. Only use `@link` for public and protected members.
```suggestion
   * @return the default execution directory in which a command process is executed.
```

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -599,7 +629,9 @@ public GitContext getGitContext() {
   @Override
   public ProcessContext newProcess() {
 
-    return new ProcessContextImpl(this);
+    ProcessContext processContext = new ProcessContextImpl(this);
+    processContext.directory(this.getDefaultExecutionDirectory());
```
hohwille, line 633: The default here is `null`. Don't you break the "productive" usage here?

hohwille: I would consider to change `getDefaultExecutionDirectory()` to return `Path.get(".")` (ideally from a constant) instead of `null`.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -5,29 +5,28 @@
...
 import com.devonfw.tools.ide.property.StringListProperty;
-import com.devonfw.tools.ide.util.FilenameUtil;
 import com.devonfw.tools.ide.version.VersionIdentifier;
 
 /**
  * {@link Commandlet} for a tool integrated into the IDE.
  */
 public abstract class ToolCommandlet extends Commandlet implements Tags {
 
+  public CommandletFileExtractor commandletFileExtractor;
```
hohwille, line 28: IMHO this should be part of `IdeContext` so we do not need this here.

cli/src/test/resources/ide-projects/jmc/_ide/urls/java/java/17.0.6/linux_x64.urls
```diff
@@ -0,0 +1 @@
+http://localhost:1112/installTest/linux
```
hohwille, line 1: It seems you copy & pasted the `basic` project to `jmc`.
However, now with your changes only very few of the files you copied are really needed.
IMHO we should avoid duplicating lots of files for every commandlet that are never needed but instead start from scratch with the minimum required resources for the test. WDYT?

cli/src/test/resources/ide-projects/jmc/downloadMockLocation/org.openjdk.jmc-8.3.0-linux.gtk.x86_64/InstallTest.txt
```diff
@@ -0,0 +1 @@
+This is a test file.
```
hohwille, line 1: IMHO something is wrong here. Where does the `org.openjdk.` prefix come from?
Where does `linux.gtk.x86_64` come from?
We should build the unique location from these properties:
* tool
* edition
* version
* os (optional)
* arch (optional)
Further, I would recommend to not use any OS differences for tests unless really needed (if we have OS dependent code in the commandlet that we need to test).
Also I would suggest to not combine all the attributes into a single folder name.
We already have the folder based structure in `urls` so why not using the same here?
So use path `jmc/jmc/8.3.0/` instead of `org.openjdk.jmc-8.3.0-linux.gtk.x86_64`

cli/src/test/resources/ide-projects/jmc/downloadMockLocation/org.openjdk.jmc-8.3.0-linux.gtk.x86_64/JDK Mission Control/jmc
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+echo "Dummy jmc 8.3.0 on linux" > jmcTestRestult.txt
```
hohwille, line 3: I guess that when we just echo, we cannot easily get the result that was echoed in the test and assert it so most probably this is the easiest working solution.
However, please always also echo `$@` in such dummy commandlets so we can test the passed parameters (including parameters potentially added by the commandlet - see all the magic in eclipse that we can then test in such way).

cli/src/test/java/com/devonfw/tools/ide/tool/jmc/JmcTest.java
```diff
@@ -25,47 +31,54 @@ public class JmcTest extends AbstractIdeContextTest {
...
-  private static Path resourcePath = Path.of("src/test/resources");
+  private static final Path RESOURCE_PATH = Path.of("src/test/resources");
 
   @BeforeAll
-  static void setUp() throws IOException {
+  static void setUp() {
+
+    // TODO use random port number and create url file dynamically in project
+    // TODO ISSUE:https://github.com/devonfw/IDEasy/issues/223
     server = new WireMockServer(WireMockConfiguration.wireMockConfig().port(1112));
```
hohwille, line 41: Why dont we remove wiremock now? I thought this is the entire point of the refactoring and the new test infrastructure.

hohwille (changes requested):
@MustaphaOuchen Thanks for your PR. You did a lot of nice refactoring and improvements - great work 👍 

I consider this PR as a blueprint for all future tests therefore I aim for some perfection.
Hence, I was picky in the review and added various review comments.
My main concern is that we are still using wiremock here and IMHO the real purpose of this refactoring is that we wanted to get rid of it for simple command tests.
Maybe my idea of what you called `downloadMockLocation` was not clear enough.

We need to clarify how we proceed now...
IMHO it is not doable to complete this PR without investing some more hours what you won't be able to complete today.
I am not sure who of the team could easily take over this PR.
Maybe I need to reserve some hours next week to take over...

cli/src/test/java/com/devonfw/tools/ide/tool/jmc/JmcTest.java
```diff
@@ -79,24 +93,124 @@ private void setupMockServerResponse(String testUrl, String contentType, String
...
+    commandlet.install();
+
+    // assert
+    performPostInstallAssertion(context);
+  }
+
+  // TODO: This test currently fails because since JMC runs as a background process
+  // the test does not wait for the started process to finish and assertion is performed too early!
+  @Test
+  @Disabled
```
MustaphaOuchen, line 137: When we run a process in the background, we cannot be sure that it finishes before the test continues to run. I did not consider this before merging the changes due to processMode usage.

hohwille:
Thanks again for this great PR 👍 I decided to take over and created PR #238 based on this work.
I already solved the problems with the background execution during tests and also the testing the output (echo) from dummy tools as well as getting rid of wiremock and the dummy urls stuff. Also I found various more bugs while testing that I could fix. Seems like a breakthrough that we invested all this into the test-infrastructure.

Issue #225, opened 2024-02-23
backup() method tries to move a deleted file in Unix


mvomiero:
### Expected behavior

In the backup() method of FileAccessImpl, if the fileOrFolder given as argument is a symlink, it is simply deleted. (as long as my understanding goes, we don't need a backup). 

### Actual behavior

The symlink is deleted, but since there is no return condition, the function tries to move the deleted file to the target location.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. from `/cli`, run `devon mvn exec:java -Dexec.mainClass="com.devonfw.tools.ide.cli.Ideasy" -Dexec.args="--debug install jasypt"`
2. output is 

<img width="878" alt="Capture" src="https://github.com/devonfw/IDEasy/assets/102261609/60518e22-8e3e-4076-9ec3-a7e8bfd78136">

### Comments/Hints:
* a return in the if statement would already resolve the problem?
* for now I got the issue just for jasypt on linux and mac

### Affected version:

* OS:Linux/Mac

MustaphaOuchen:
Thank you for finding this bug. It is comprehensible and traceable that an already deleted symbolic link cannot be moved/backed up anymore. I would probably also consider if we may backup the file location where the symbolic linc actually points to instead of just returning?

PR #226, opened 2024-02-23
#225: Return statement in backup() function in case of symlinks

mvomiero:
Fixes: #225 by just adding a `return;` if the file is a symlink (it is then deleted and doesen't need to be moved).


MustaphaOuchen (approved):
After daily discussion: LGTM ! We probably need to create an issue ticket for "Files.isSymbolicLink()" because of its problems with windows junction!

hohwille (approved):
@mvomiero excellent finding/fix. Thank you 👍

mvomiero:
> After daily discussion: LGTM ! We probably need to create an issue ticket for "Files.isSymbolicLink()" because of its problems with windows junction!

Issue is now created: #231 
😃

PR #228, opened 2024-02-23
#227: Fix paths problem (IDEasy could not set the tool path on first installation)

### LINKED ISSUE #227: First tool installation of commandlet does not add tool to paths of SystemPath
### Expected behavior

The binary path should be added to path.

### Actual behavior
It is added to the "tool2pathMap" but not to the used "path" which is used to get the binary.

Install a program with ID-Easy for the first time, it will not start.

MustaphaOuchen:
Closes #227 

Our problem was that we just fill the paths variable of the SystemPath only during construction with the binary/commandlet locations. I think we have to do it also when tool2pathMap is changed, since tool2pathMap behaves like a "subset" of paths as far as I did understand.

This should fix that on the first start of the application the installed binary could not be found.


ndemirca:
LGTM

cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
```diff
@@ -167,11 +168,12 @@ public Path getPath(String tool) {
 
   /**
    * @param tool the name of the tool.
-   * @param path the new {@link #getPath(String) tool bin path}.
+   * @param path the new {@link #getPath(String) tool bin path} and {@link #paths}.
    */
   public void setPath(String tool, Path path) {
 
     this.tool2pathMap.put(tool, path);
+    this.paths.add(0, path);
```
hohwille, line 176: Salim did almost the same "fix" here:
https://github.com/devonfw/IDEasy/pull/143/files#diff-594121d7c6efe86c0927ed36bc6e69da51a4ee92f8ed2ac0958b118e3fc3f3ff

I will get in touch to discuss how we can get to the best solution.

hohwille (changes requested):
@MustaphaOuchen Thanks for your PR. I will reach out and discuss with you online so we can find the perfect solution together.

cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
MustaphaOuchen: @hohwille I changed it to iterate also over the tool2pathMap as discussed.

hohwille (approved):
@MustaphaOuchen thanks for the PR & fix. 👍 
Ready for merge.

PR #229, opened 2024-02-26
#1309: Improve plugin managment

diiinesh:
Solves  https://github.com/devonfw/ide/issues/1309
Improves plugin managment by solving the error of installing duplicate plugins


cli/src/main/java/com/devonfw/tools/ide/tool/PluginBasedCommandlet.java
```diff
@@ -3,6 +3,7 @@
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
```
mvomiero, line 6: Import not used

cli/src/test/java/com/devonfw/tools/ide/tool/PluginBasedCommandletTest.java
```diff
@@ -40,6 +34,12 @@ void testGetPluginsMap() {
     assertThat(plugin2.getName()).isEqualTo("anyedit");
 
     // Check if anyedit plugin has value "false" --> value from user directory
-    assertThat(plugin2.isActive()).isFalse();
+    assertFalse(plugin2.isActive());
```
mvomiero, line 37: probably better to use assertThat()? :)

mvomiero (commented):
except the unused import and maybe the refactoring for the assert(), LGTM :)

cli/src/main/java/com/devonfw/tools/ide/tool/PluginBasedCommandlet.java
diiinesh: I have checked again and Paths is used and needed.

diiinesh:
After reviewing and analyzing issue #1309 (https://github.com/devonfw/ide/issues/1309), I found that the issue of duplicate plugins, caused by not properly removing the plugins folder after every tool update, does not exist in IDEasy. In IDEasy, the plugin folder is removed correctly, and the plugins are reinstalled automatically.

This behavior has been thoroughly tested for both Eclipse and IntelliJ.

Additionally, I have made modifications to PluginBasedCommandlet.java. These modifications introduce two PluginMaps: one with plugin IDs as keys and another with plugin names as keys. This change enhances the efficiency and organization of plugin handling within the codebase.

mvomiero (approved):
Thank you for your implementation, LGTM 🌹

cli/src/main/java/com/devonfw/tools/ide/tool/PluginMaps.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  /**
+   * Returns the map of plugin descriptors keyed by plugin name.
+   *
+   * @return The map of plugin descriptors keyed by plugin name.
+   */
+  public Map<String, PluginDescriptor> getByName() {
+    return mapByName;
+  }
```
hohwille, line 41: I always like if we create dedicated structures but then we should also follow encapsulation:
Create the maps internally and remove them from `PluginBasedCommandlet` and instead delegate registration.

```suggestion
  
  private final IdeContext context;

  /**
   * The constructor.
   * @param context the {@link IdeContext}.
   */
  public PluginMaps(IdeContext context) {
    super();
    this.context = context;
    this.mapById = new HashMap<>();
    this.mapByName = new HashMap<>();
  }

  /**
   * @return the {@link PluginDescriptor} for the given {@link PluginDescriptor#getId() ID} or {@code null} if not found.
   */
  public PluginDescriptor getById(String id) {
    return this.mapById.get(id);
  }

  /**
   * @return the {@link PluginDescriptor} for the given {@link PluginDescriptor#getName() name} or {@code null} if not found.
   */
  public PluginDescriptor getByName(String name) {
    return this.mapByName.get(name);
  }
  
  public Collection<PluginDescriptor> getPlugins() {
    
    Map<String, PluginDescriptor> map = this.mapById;
    if (map.isEmpty()) {
      map = this.mapByName; // potentially plugins for this tool have no ID
    }
    return Collections.unmodifyableCollection(map.values());
  }
  
  /**
   * Registers a {@link PluginDescriptor} to this map.
   */
  protected void add(PluginDescriptor descriptor) {
    put(decriptor.getName(), descriptor, this.mapByName);
    String id = descriptor.getId();
    if ((id != null) && !id.isEmpty()) {
      put(id, descriptor, this.mapById);
    }
  }
  
  private void put(String key, PluginDescriptor descriptor, Map<String, PluginDescriptor> map) {
    PluginDescriptor duplicate = map.put(key, descriptor);
    if (duplicate != null) {
      this.context.info("Plugin with key {} was {} but got overridden by {}", key, duplicate, descriptor);
    }
  }
```

cli/src/main/java/com/devonfw/tools/ide/tool/PluginBasedCommandlet.java
```diff
@@ -1,195 +1,233 @@
...
+  public PluginDescriptor getPlugin(String key) {
+
+    if (key == null) {
+      return null;
+    }
+    if (key.endsWith(IdeContext.EXT_PROPERTIES)) {
+      key = key.substring(0, key.length() - IdeContext.EXT_PROPERTIES.length());
+    }
+
+    PluginDescriptor pluginDescriptor = getPluginsMap().getById().get(key);
```
hohwille, line 173:
```suggestion
    PluginDescriptor pluginDescriptor = getPluginsMap().getById(key);
```

cli/src/test/java/com/devonfw/tools/ide/tool/PluginBasedCommandletTest.java
```diff
@@ -1,44 +1,72 @@
...
+
+import com.devonfw.tools.ide.cli.CliException;
+import com.devonfw.tools.ide.context.AbstractIdeContextTest;
+import com.devonfw.tools.ide.context.IdeTestContext;
+import com.devonfw.tools.ide.tool.ide.PluginDescriptor;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.*;
```
hohwille, line 11: Can you verify your IDE settings (call `devon ide update`)?
With #52 we fixed this log time ago and prevent discouraged start-imports automatically.
If you still generate this in IntelliJ then your settings are outdated.

cli/src/main/java/com/devonfw/tools/ide/tool/PluginBasedCommandlet.java
```diff
@@ -1,195 +1,233 @@
...
+      for (PluginDescriptor plugin : mapByName.values()) {
+        if (!mapById.containsKey(plugin.getId())) {
+          if (plugin.isActive()) {
+            installPlugin(plugin);
+          } else {
+            handleInstall4InactivePlugin(plugin);
+          }
+        }
+      }
+      for (PluginDescriptor plugin : mapById.values()) {
```
hohwille, line 214: should IMHO now be simplified and would be sufficient as:
```suggestion
      for (PluginDescriptor plugin : getPluginsMap().getPlugins()) {
          if (plugin.isActive()) {
            installPlugin(plugin);
          } else {
            handleInstall4InactivePlugin(plugin);
          }
        }
      }
```

hohwille (changes requested):
@diiinesh thanks for your rework, improvements and testing on this feature.
I took your approach further to encapsulate the plugin mapping into the new `PluginMaps` class. See my review comments with the suggested code-changes.
The JUnit would still need adoption that I can not do reliably in the browser.

After all I am thinking that we might have over-complicated things with adding the additional by-id map:
* some plugins may not even have an ID but that is covered in the code
* eclipse plugin ID is a comma separated list of feature IDs. In case you have more than one feature overriding can become problematic: If you want to override the plugin configuration but want to have a more or less features, the plugin ID will change and therefore you are not technically overriding things even if you pick the same filename `«plugin».properties`. Therefore it is impossible to reduce the number of features but only add more features.
Maybe always using the `mapByName` for `getPlugins()` would be a better solution...

hohwille:
@diiinesh there are test errors:
```
[ERROR] com.devonfw.tools.ide.commandlet.UpdateCommandletTest.testRunPullSettingsAndUpdateSoftware -- Time elapsed: 0.032 s <<< ERROR!
java.lang.NullPointerException: Cannot invoke "com.devonfw.tools.ide.tool.PluginMaps.getPlugins()" because "this.pluginsMap" is null
	at com.devonfw.tools.ide.tool.PluginBasedCommandlet.doInstall(PluginBasedCommandlet.java:158)
```
Please have a look. Thanks.

hohwille (approved):
@diiinesh thanks for the update. You finally nailed it 👍

Issue #195, opened 2024-02-06
Installation of Java and running applications which require Java fails on Mac OS


MustaphaOuchen:
### Expected behavior

As a user of IDEasy, I want to install Java so that applications which require Java, like JMC, will run successfully on Mac OS.

### Actual behavior
If one tries to install Java the application will throw an java.lang.llegalStateException and applications which require Java will not find the JRE.

### Steps to reproduce (bug)

1.  Create a project environment with devonfw on a Mac
2.  Clone IDEasy in workspaces/main
3.  Navigate via Terminal to the IDEasy/cli project folder
4.  Install Java via: "devon mvn exec:java -Dexec.mainClass=com.devonfw.tools.ide.cli.Ideasy -Dexec.args=java"
5. Run an application which requires Java, JMC for example: "devon mvn exec:java -Dexec.mainClass=com.devonfw.tools.ide.cli.Ideasy -Dexec.args=jmc"

### Comments/Hints:

Screenshot of step 4.)

![Image](https://github.com/devonfw/IDEasy/assets/98693422/051ed7be-11dc-48e1-91c7-02480a251ef0)

Screenshot of step 5.)

![Image](https://github.com/devonfw/IDEasy/assets/98693422/7c705169-1db3-4227-98cf-e58d016b5e87)

### Affected version:

* OS: Mac

hohwille:
Already fixed with PR #238

PR #232, opened 2024-02-27
#195: java installation in macos

mvomiero:
Fixes: #195 

## CAUSE of the issue:
* copying `.ide.software.version` to the target Path was failing.
the problem was to my understanding and after my researches, that the source is a specific file path but the target is not recognized as a directory as it is, so when it tries to replace the target directory with the source file it fails.

## SOLUTION:
* adding .ide.software.version to the target Path solves the problem.
* the file has to be copied just if it is a new installation, so it is possible to enclose it in a if statement that checks it's a new installation

## COMMENTS:
* works fine for me,  but
* still not super experienced about all the installation of commandlets. Would be nice if someone else could have a look.


hohwille:
@mvomiero thanks for creating the bug issue, analyzing the problem and fixing it. Your PR is great. 👍 
However, I merged PR #238 that also fixed the same bug so we can now close this PR.

Issue #231, opened 2024-02-26
backup() method doesn't delete junctions in windows


mvomiero:
### Expected behavior

The FileAccessImpl.backup() method should delete junctions, as it does for symlinks, instead of copying them  to `devonfw/updates/backup/`

### Actual behavior

 junctions in `devonfw/software` are copied  to `devonfw/updates/backup/`

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

_**example: jmc**_

From `cli`:

1.  check if you have jmc alteady installed, otherwise install it 
`devon mvn exec:java -Dexec.mainClass="com.devonfw.tools.ide.cli.Ideasy" -Dexec.args="--debug install jmc" 
`
2. install another version of jmc to trigger the backup (e.g. 8.0.0)
`devon mvn exec:java -Dexec.mainClass="com.devonfw.tools.ide.cli.Ideasy" -Dexec.args="--debug install jmc 8.0.0" `

3. the junction in  `devonfw/software` will be now in  `devonfw/updates/backup/`
Console output:

`Creating backup by moving C:\Projects\devonfw\software\jmc to C:\Projects\devonfw\updates\backups\2024-02-26\jmc_17-22-12`

### Related/Dependent Issues

#225 was about the same method, but for a different case (symlinks, throwing exceptions).

### Affected version:

* OS: Windows

PR #233, opened 2024-02-27
#231: improve backup() method to delete junctions in win

mvomiero:
Fixes: #231 by adding a method isJunction() and deleting the file if it is a Junction in the backup() method.


VinceHeu (approved):

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -220,22 +219,37 @@ public String checksum(Path file) {
...
+      return false;
+    }
+
+    try {
+      BasicFileAttributes attr = Files.readAttributes(path, BasicFileAttributes.class, LinkOption.NOFOLLOW_LINKS);
+      return attr.isOther() && attr.isDirectory();
+    } catch (IOException e) {
+      throw new IllegalStateException(
+          "An unexpected error occurred whilst checking if the file: " + path + " is a junction", e);
+    }
```
hohwille, line 234: Didn't you get this code from `deleteLinkIfExists`?
You could improve by either reusing `deleteLinkIfExists` in the `backup` method or by now using `isJunction` inside `deleteLinkIfExists` function to avoid code duplication.

A small problem with the second approach is that as it seems the current implementation to detect a junction also checks for the existence in the same go:
```
    boolean exists = false;
    boolean isJunction = false;
      try { // since broken junctions are not detected by Files.exists(brokenJunction)
        BasicFileAttributes attr = Files.readAttributes(path, BasicFileAttributes.class, LinkOption.NOFOLLOW_LINKS);
        exists = true;
        isJunction = attr.isOther() && attr.isDirectory();
      } catch (NoSuchFileException e) {
        // ignore, since there is no previous file at the location, so nothing to delete
        return;
      }
    }
    exists = exists || Files.exists(path);
```
Also note that the existing code implementing this contains some very valuable comments that should not be lost.

hohwille (approved):
@mvomiero Thank you for your PR. Great that you found, analyzed and fixed the problem. 👍 
While your PR is fixing the problem it introduces some code duplication. I added a comment for rework. But apart from that all is fine and then it can be merged.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
mvomiero: **_I went for the second approach: using `isJunction `in `deleteLinkIfExists`:_**
* to make `isJunction `reusable the `NoSuchFileException `is catched to return false in case the file is not existing
* clearer structure of the `deleteLinkIfExists `function:
   * bools store if the link is a symlink or a path, checking if the file exists as well
   * if one of them is true, the file is deleted (in this way we don't even need to throw an exception as before, since we already know that it exists and is a link). If there is a problem with delete, `IOExecption `will be thrown as before.
 
 This approach improves readability and modularity of the code and reduces the `deleteLinkIfExists `from 28 to 11 lines 🙂
 
**_the first approach would have been in my opinion quite complicated and difficult to read since the moving of a directory would be inside a catch block and another catch block should be there for the IOExceptions. For example something like that:_**
```
  @Override
  public void backup(Path fileOrFolder) {

    try {
      // Delete the file or folder if it is a symbolic link or a junction
      deleteLinkIfExists(fileOrFolder);
    } catch (IllegalStateException e) {
      // Exception if the file is not deleted (it's not a link), backup has to be created.
      Path backupPath = this.context.getIdeHome().resolve(IdeContext.FOLDER_UPDATES).resolve(IdeContext.FOLDER_BACKUPS);
      LocalDateTime now = LocalDateTime.now();
      String date = DateTimeUtil.formatDate(now);
      String time = DateTimeUtil.formatTime(now);
      Path backupDatePath = backupPath.resolve(date);
      mkdirs(backupDatePath);
      Path target = backupDatePath.resolve(fileOrFolder.getFileName().toString() + "_" + time);
      this.context.info("Creating backup by moving {} to {}", fileOrFolder, target);
      move(fileOrFolder, target);
    } catch (IOException e) {
      throw new IllegalStateException("Failed to delete previous symlink or Windows junction at " + fileOrFolder, e);
    }
  }
```

* valuable comments are now there
* let me know if I oversaw something or if I should go for another type of refactoring, I'm at the beginning of my collaboration at IDEasy so happy to learn and get constructive feedbacks 🙂

hohwille (approved):
@mvomiero thanks for your update. You perfectly resolved the redundancies now. Great work 👍 
Ready for merge.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
hohwille: I fully agree and thank you for the excellent work and all the detailed explanations.

PR #238, opened 2024-03-03
#208: improve test infrastructure

### LINKED ISSUE #208: Mock out ToolRepository so that commandlets can be tested with unit tests

hohwille:
Fixes #208
Based on and replacement of the PR #222 work from @MustaphaOuchen 

* automatic test repository support in test context
* auto-mock process mode and print std out/error to logger in tests
* fixes some bugs with process context and env variables
* improves/fixes IDE_HOME detection
* improve generic extract support
* align test-cases and simplify test context creation


hohwille:
Usually I prefer small PRs for individual small stories. Here many different things have collected into a single monster PR what is usually undesired. However, as we all noticed, the test-infrastructure is required for the team to process as otherwise most other PRs are blocked. Also a team-review of this PR would most likely postpone this from a merge for at least a week.
I will therefore merge this now. Everybody is still invited to review and give feedback for improvements or to learn what has changed even after the merge...

PR #240, opened 2024-03-05
#238: fix process argument logging

hohwille:
fixes process argument logging that I broke (1st arg omitted) in PR #238.


ndemirca (approved):
LGTM

PR #242, opened 2024-03-05
#94: improve documentation and logging

### LINKED ISSUE #94: Ability to define plugins in user config

hohwille:
closes #94 that was already implemented by PR #141

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
@@ -1,6 +1,7 @@
 package com.devonfw.tools.ide.io;
 
 import java.nio.file.Path;
+import java.util.List;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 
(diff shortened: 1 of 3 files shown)


hohwille:
As there was no review feedback, I will simply merge now.

PR #243, opened 2024-03-06
#26: Implement ToolCommandlet for Jasypt

### LINKED ISSUE #26: Implement ToolCommandlet for Jasypt
Create `ToolCommandlet` for Jasypt
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/jasypt

mvomiero:
Closes #26


cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/Jasypt.java
```diff
@@ -0,0 +1,104 @@
...
+public class Jasypt extends LocalToolCommandlet {
+
+  private static final String USAGE_INFO = """
+      Jasypt encryption tool
+      Usage:
+       encrypt  <masterpassword>  <secret>             encrypt a secret with a master-password
+       decrypt  <masterpassword>  <secret>             decrypt an encrypted secret with a master-password
+      """;
+
+  public static final String CLASS_NAME_ENCRYPTION = "org.jasypt.intf.cli.JasyptPBEStringEncryptionCLI";
```
VinceHeu, line 27: Could be private

VinceHeu (changes requested):

cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/Jasypt.java
mvomiero: thank you for the tip 👍

cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/Jasypt.java
```diff
@@ -0,0 +1,104 @@
...
+    if (args.length == 0) {
+      this.context.info(USAGE_INFO);
+    } else if (args.length == 3 && args[0].equals("encrypt")) {
+      runJasypt(CLASS_NAME_ENCRYPTION, args);
+    } else if (args.length == 3 && args[0].equals("decrypt")) {
+      runJasypt(CLASS_NAME_DECRYPTION, args);
+    } else {
+      this.context.warning("Unknown arguments");
+      this.context.info(USAGE_INFO);
+    }
```
hohwille, line 59: If you use an enum property the CLI will already handle the case if an "unknow argument" was provided and therefore you could also omit the `else` block and even convert the `if ... else` to a `swtich` like done in Sonar commandlet.

cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/Jasypt.java
```diff
@@ -0,0 +1,104 @@
...
+
+  private void installJasyptArtifact() {
+
+    Mvn mvn = getCommandlet(Mvn.class);
+    mvn.install();
+    this.context.debug("installing jasypt artifact in: " + resolveJasyptJarPath());
+    mvn.runTool(null, "org.apache.maven.plugins:maven-dependency-plugin:3.1.2:get",
+        "-Dartifact=org.jasypt:jasypt:" + getInstalledVersion().toString());
+  }
+
```
hohwille, line 88: You correctly converted this from the bash code.
However, this does not make sense anymore. We have Jasypt here:
https://github.com/devonfw/ide-urls/tree/master/jasypt/jasypt

And `super..doInstall(silent)` will already install `jasypt` and also not in a hard-coded version (`3.1.2`) but in the version configurable via `JASYPT_VERSION`. So we actually do not need maven anymore here and you can kick this out (remove `installJasyptArtifact` method entirely).

IMHO you also need to override `isExtract()` to `false` as otherwise the `jasypt-*.jar` will be extracted.

```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/Jasypt.java
```diff
@@ -0,0 +1,104 @@
...
+    java.runTool(null, "-cp", resolveJasyptJarPath().toString(), className, ALGORITHM, GENERATOR_CLASS_NAME,
+        "password=" + args[1], "input=" + args[2]);
+  }
+
+  private Path resolveJasyptJarPath() {
+
+    Path m2Repo = context.getVariables().getPath("M2_REPO");
+    String installedVersion = getInstalledVersion().toString();
+    return m2Repo.resolve("org").resolve("jasypt").resolve("jasypt").resolve(installedVersion)
+        .resolve("jasypt-" + installedVersion + ".jar");
```
hohwille, line 101: See comments above. Using maven does not make sense anymore. We already get `jasypt-*.jar` installed into `software/jasypt`.
```suggestion
    return this.context.getFileAccess().findFirst(getToolPath(), f -> f.getFileName().toString().endsWith(".jar"), false);
```

cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/Jasypt.java
```diff
@@ -0,0 +1,104 @@
...
+  private void installJasyptArtifact() {
+
+    Mvn mvn = getCommandlet(Mvn.class);
+    mvn.install();
+    this.context.debug("installing jasypt artifact in: " + resolveJasyptJarPath());
+    mvn.runTool(null, "org.apache.maven.plugins:maven-dependency-plugin:3.1.2:get",
+        "-Dartifact=org.jasypt:jasypt:" + getInstalledVersion().toString());
+  }
+
+  private void runJasypt(String className, String[] args) {
```
hohwille, line 89: it would make more sense to define the signature as `runJasypt(String className, String masterPassword, String secret)`.
However, even better would be to define the arguments as individual properties:
```
public final StringProperty masterPassword;
public final StringProperty secret;

...
  this.masterPassword=add(new StringProperty("", true, "masterPassword"));
  this.secret=add(new StringProperty("", true, "secret"));
```

For perfection, you could create a class `PasswordProperty` that extends `StringProperty` and use that instead and then we could consider special handling of PasswordProperty later e.g. to prevent writing its value to log files or whatever may come in the future. But this is nice to have for now.

cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/Jasypt.java
```diff
@@ -0,0 +1,104 @@
...
+
+    Java java = getCommandlet(Java.class);
+    java.runTool(null, "-cp", resolveJasyptJarPath().toString(), className, ALGORITHM, GENERATOR_CLASS_NAME,
+        "password=" + args[1], "input=" + args[2]);
```
hohwille, line 93:
```suggestion
    java.runTool(null, "-cp", resolveJasyptJarPath().toString(), className, ALGORITHM, GENERATOR_CLASS_NAME,
        "password=" + this.masterPassword.get(), "input=" + this.secret.get());
```
After rethinking this I question ourselves more:
What if the end-user wants to change the algorithm? What if another algorithm does not take an initialization vector generator? Would it make more sense to define a `JASYPT_OPTS` property that we define in `IdeVariables` and that as the default value `algorithm=PBEWITHHMACSHA512ANDAES_256 ivGeneratorClassName=org.jasypt.iv.RandomIvGenerator` but can be overridden by projects or users in their `ide.properties`? IMHO that would make much more sense.

cli/src/test/java/com/devonfw/tools/ide/tool/jasypt/JasyptTest.java
```diff
@@ -0,0 +1,80 @@
...
+    String expectedMessage = """
+        Jasypt encryption tool
+        Usage:
+         encrypt  <masterpassword>  <secret>             encrypt a secret with a master-password
+         decrypt  <masterpassword>  <secret>             decrypt an encrypted secret with a master-password
+        """;
+    commandlet.arguments.setValue(arguments);
+    commandlet.run();
+    assertLogMessage(context, IdeLogLevel.INFO, expectedMessage);
+  }
```
hohwille, line 48: Sorry, you have implemented all the tests and now things have to change slightly rendering that test work void...
BTW: It is much better to write multiple `@Test` methods for individual aspects instead of writing one big `testJasypt()` method that then tests various other things. Have a look at other existing tests and how they do it.

cli/src/test/resources/ide-projects/jasypt/_ide/urls/jasypt/jasypt/1.9.3/status.json
```diff
@@ -0,0 +1,10 @@
+{
```
hohwille, line 1: none of these files in `_ide/urls` is actually used by your tests. Hence, you should remove them to avoid pointless clutter.

cli/src/test/resources/ide-projects/jasypt/project/settings/ide.properties
```diff
@@ -0,0 +1,6 @@
+JAVA_VERSION=17.0.10_7
+JASYPT_VERSION=1.9.3
+JMC_VERSION=8.3.0
+MVN_VERSION=3.9.6
+
+M2_REPO=~/.m2/repository
```
hohwille, line 6: Should all be removed now
```suggestion
```

cli/src/test/resources/ide-projects/jasypt/repository/jasypt/jasypt/default/InstallTest.txt
```diff
@@ -0,0 +1 @@
+This is a test file.
```
hohwille, line 1: Here you would actually need to place a `jasypt-default.jar` file. It can have the same content and the faked/mocked `java` could read and echo the content of the JAR file provided after `-cp` so you can test that all is working.

cli/src/test/resources/ide-projects/jasypt/repository/mvn/mvn/default/bin/mvn
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
hohwille, line 1: `mvn` and `mvn.bat` can be removed now.

cli/src/test/resources/ide-projects/jasypt/repository/java/java/default/bin/java.cmd
```diff
@@ -0,0 +1,3 @@
+@echo off
```
hohwille, line 1: if you do not want to test OS specific behavior you can remove the `java.cmd` so you only have to maintain `java` bash script.

documentation/jasypt.asciidoc
```diff
@@ -0,0 +1,76 @@
+:toc:
```
hohwille, line 1: Same as in other PRs. Sorry for the confusion. We discussed this in the team some time ago but nobody told this to the new people that joined the team recently:
We decided not to add redundant AsciiDoc documentation for commandlet usage since we have a very sophisticated help now that can give you all the information you need.
The redundant AsciiDoc will only outdate sooner or later when somebody changes a commandlet but forgets to update the AsciiDoc.
We already had this a lot in devonfw-ide.
See #247

hohwille (changes requested):
@mvomiero thank you for this PR. You migrated the bash code nicely to Java. 👍 
I did not see this coming before but after reviewing this PR I realized that the old behavior does not really make sense anymore.
I wrote tons of stories for migrating commandlets to IDEasy but here I forgot to think further so unfortunately this results in some rework for you.
Probably for the future we need to think about a quick handover checkpoint for such stories (after the developer has understood the story and started the very first things but before completing the code and writing any tests) to prevent this...

mvomiero:
@hohwille thank you for the review and the changes requested 💯 

I followed your suggestions for most of the changes, please note just some small details I had to adjust to make the code work:
* in `AbstractEnvironmentVariables` I changed a line to fix the bug we had about retrieving variables from `IdeVariables`
```diff
-     if (value != null) {
+     if (value == null) {
        value = var.getDefaultValueAsString(this.context);
      }
```
* for the **tests**, I had to hardcode the version of jasypt in the mocked .jar file name (`jasypt-1.9.3.jar` instead of `jasypt-default.jar`). The problem was, that when the commandlet is run, it looks for the artifact of the installed version, and since we input 1.9.3 as version in the properties, it will look for a file with the corresponding name. 
    * Since .jar files are included in the gitignore, i had to add a line to exclude `jasypt-1.9.3.jar`

Some other minor changes:
* I added a `Tag.ENCRYPTION` that has a parent `Tag.CRYPTO` for the commandlet.
* I changed the name of `masterPassword `and `secret `to `password `and `input`, which are actually the name of the arguments used when executing jasypt. e.g. `java -cp jasypt-1.9.3.jar ... password=xxx input=xxx`

Let me know if there are any problems or if something is not clear 😃

hohwille:
> @hohwille thank you for the review and the changes requested 💯
> 
> I followed your suggestions for most of the changes, please note just some small details I had to adjust to make the code work:
> 
>     * in `AbstractEnvironmentVariables` I changed a line to fix the bug we had about retrieving variables from `IdeVariables`
> 
> 
> ```diff
> -     if (value != null) {
> +     if (value == null) {
>         value = var.getDefaultValueAsString(this.context);
>       }
> ```

perfect. Thanks for finding and fixing.

>     * for the **tests**, I had to hardcode the version of jasypt in the mocked .jar file name (`jasypt-1.9.3.jar` instead of `jasypt-default.jar`). The problem was, that when the commandlet is run, it looks for the artifact of the installed version, and since we input 1.9.3 as version in the properties, it will look for a file with the corresponding name.

Correct.

>       
>       * Since .jar files are included in the gitignore, i had to add a line to exclude `jasypt-1.9.3.jar`

In such situations it is better not to do that. Simply use `git add -f jasypt-1.9.3.jar` instead.

> Some other minor changes:
> 
>     * I added a `Tag.ENCRYPTION` that has a parent `Tag.CRYPTO` for the commandlet.

Fully makes sense.

>     * I changed the name of `masterPassword `and `secret `to `password `and `input`, which are actually the name of the arguments used when executing jasypt. e.g. `java -cp jasypt-1.9.3.jar ... password=xxx input=xxx`

That is the only issue I have. The parameters from jasypt are very generic and this makes it much more confusing for end-users:
We want to encrypt passwords with a master-password. If a parameter is just called `password` and both parameters are passwords which of those is it? I personally would prefer here to stay with devonfw-ide compatbility and the from my PoV better understandable parameters `masterPassword` (the master password for encryption/decryption) and `secret` (the secret value to encrypt - e.g. a password).

> Let me know if there are any problems or if something is not clear 😃

Everything else is perfect - I just have the parameter name concern....

cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/Jasypt.java
```diff
@@ -0,0 +1,103 @@
...
+   *
+   * @param context the {@link IdeContext}.
+   */
+  public Jasypt(IdeContext context) {
+
+    super(context, "jasypt", Set.of(Tag.JAVA, Tag.ENCRYPTION));
+
+    this.command = add(new EnumProperty<>("", true, "command", JasyptCommand.class));
+    this.password = add(new PasswordProperty("", true, "password"));
+    this.input = add(new PasswordProperty("", true, "input"));
```
hohwille, line 41: The parameters from jasypt are very generic and this makes it much more confusing for end-users:
We want to encrypt passwords with a master-password. If a parameter is just called password and both parameters are passwords which of those is it? I personally would prefer here to stay with devonfw-ide compatbility and the from my PoV better understandable parameters:
* `masterPassword` - the master password for encryption/decryption
* `secret` - the secret value to encrypt - e.g. a password

hohwille (approved):
@mvomiero thanks for your PR and also for finding and fixing bugs. Great work 👍 
I left one comment on the parameter naming. Do you understand my PoV and agree?
If yes, you can simply update and I can merge.
Otherwise lets talk and you can convince me :)

cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/Jasypt.java
mvomiero: Yess now that I better understand that normally is used to encrypt passwords, I totally agree 😄

mvomiero:
Thank you @hohwille, I totally understand your PoV about the parameters naming. I totally agree with the `.gitignore `as well, I was not liking the solution either, so I added the file with with `git add -f `and removed it from the `.gitignore`
I merge then 🚀

PR #246, opened 2024-03-11
encoding error with graalvm

### LINKED ISSUE #215: Encoding bug with i18n of help
The resource bundles for i18n of help are processed with an invalid encoding:
image

> az             Werkzeug Kommando f�r die Azure Kommandoschnittstelle.

Should be `für` instead of `f�r`.
IMHO the resource bundles are encoded with UTF-8 what is correct but for some reason our GraalVM build native image treats them as something like `ISO-8859-1`.

diiinesh:
Fixes #215


hohwille (approved):
@diiinesh thanks for your PR. Your approach seems to make sense since we are using an old version of graalvm and the encoding bug was fixed already. 👍 
As this PR is still in draft mode, I assume you are still testing.
I already approve so once you take out of draft, we can merge and test this.

PR #248, opened 2024-03-15
Added Documentation for building an executable with graalvm

KGoroll1:
Adds documentation to build a native executable with graalvm. Currently the community edition of graalvm is used as it comes with a pre-installed native image. However, this version is outdated. Therefore, the next step should be to update the guide with a new version of graalvm. 

With a current version of graalvm you can build the native image yourself with `gu install native-image`. However, if you then run mvn package, the build will fail. This is probably just a version problem, but needs to be investigated further.


hohwille (approved):
@KGoroll1 thank you so much for this PR. 👍

PR #249, opened 2024-03-15
Feature/implement tool commandlet docker

### LINKED ISSUE #16: Implement ToolCommandlet for Docker
Create `ToolCommandlet` for Docker
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/docker
ATTENTION: Blocked by issue #47 that should be implemented first or together with this story.

KGoroll1:
closes #16 

Implemented the Docker GlobalToolCommandlet


cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -24,13 +26,70 @@ public abstract class GlobalToolCommandlet extends ToolCommandlet {
...
+
+    Path bashPath = this.context.getPath().findBinary(Path.of("bash"));
+    if (bashPath == null || !Files.exists(bashPath)) {
+      context.warning("Bash was not found on this machine. Not Proceeding with installation of tool " + this.tool);
+      return false;
+    }
+
+    PackageManager foundPackageManager = null;
+    for (PackageManager pm : commands.keySet()) {
+      if (Files.exists(this.context.getPath().findBinary(Path.of(pm.toString().toLowerCase())))) {
```
hohwille, line 62: It would be better to define a method in `PackageManager` enum for this instead of calling `toString().toLowerCase()`.
But nice to have - surely works...

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -24,13 +26,70 @@ public abstract class GlobalToolCommandlet extends ToolCommandlet {
...
+
+    Path binaryPath = this.context.getPath().findBinary(Path.of(getBinaryName()));
+
+    if (binaryPath != null && Files.exists(binaryPath) && !this.context.isForceMode()) {
+      IdeLogLevel level = silent ? IdeLogLevel.DEBUG : IdeLogLevel.INFO;
+      this.context.level(level).log("{} is already installed at {}", this.tool, binaryPath);
+      return false;
+    }
+
+    Path bashPath = this.context.getPath().findBinary(Path.of("bash"));
```
hohwille, line 54: maybe it makes sense to create a method `findBash()` reusing or replacing `findBashOnWindows()` as I assume we will have multiple places where we need this and can therefore centralize it including the error message if not found.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -24,13 +26,70 @@ public abstract class GlobalToolCommandlet extends ToolCommandlet {
...
+    int finalExitCode = 0;
+    if (foundPackageManager == null) {
+      context.warning("No supported Package Manager found for installation");
+      return false;
+    } else {
+      List<String> commandList = commands.get(foundPackageManager);
+      if (commandList != null) {
+        for (String command : commandList) {
+          ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.WARNING).executable(bashPath)
+              .addArgs("-c", command);
```
hohwille, line 77: why do we need `bash -c ` to run the command?
OK, later down I found out that we are passing pipes (`echo ... | dd ...`) what only works this way so this is fully correct.
I still would like to simplify the list to a single command that can then be a "script" potentially calling multiple commands (`foo && bar`).

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -0,0 +1,79 @@
...
+
+    commands.put(PackageManager.ZYPPER, Arrays.asList(
+        "sudo zypper addrepo https://download.opensuse.org/repositories/isv:/Rancher:/stable/rpm/isv:Rancher:stable.repo",
+        String.format("sudo zypper --no-gpg-checks install rancher-desktop=%s*", resolvedVersion)));
+
+    commands.put(PackageManager.APT, Arrays.asList(
+        "curl -s https://download.opensuse.org/repositories/isv:/Rancher:/stable/deb/Release.key | gpg --dearmor | sudo dd status=none of=/usr/share/keyrings/isv-rancher-stable-archive-keyring.gpg",
+        "echo 'deb [signed-by=/usr/share/keyrings/isv-rancher-stable-archive-keyring.gpg] https://download.opensuse.org/repositories/isv:/Rancher:/stable/deb/ ./' | sudo dd status=none of=/etc/apt/sources.list.d/isv-rancher-stable.list",
+        "sudo apt update",
+        String.format("sudo apt install -y --allow-downgrades rancher-desktop=%s*", resolvedVersion)));
```
hohwille, line 65: OK, now I at least get why you are using `Map<PackageManager, List<String>>`.
However, can't we still keep that simple as a String vararg?
I see that when there is `curl - s ... && echo ... && sudo apt update` it seems tricky to get the package manager.
So far we had simpler situations where we just called the package-manager and that was the first argument.

My conclusion would be:
1. we are running an installation so all logic can run with a single `sudo` command so that could be omitted here and done in `installWithPackageManger` method.
2. IMHO we can combine all the individual string arguments to a single script.
3. The remaining challenge is to also provide the package manager or OS so that `installWithPackageManger` can choose which of the installation options to take.

How about a simple API for the easy 90% case:
`installWithPackageManager("apt-get install unzip", "yum install unzip", "pacman -S unzip", "emerge install unzip", "zypp install unzip", "apk add unzip", "brew install unzip");`
Here we would extract the package manager directly from the command and convert it to our `Map` structure.
We can even make it smart to search in the string for any of the known package-managers if not already the first word is a package manager.
However, to avoid too much magic in really complex cases, we can still call a more complex method giving structured data instead of just Strings.
But still I think that `Map` is not the best option. What we want to provide is a "list" of "installation variants" per OS/Distribution/Package-Manager. For this a vararg of `PackageManagerCommand` would make more sense.
The class `PackageManagerCommand` can have the `PackageManager` and again a list of commands to keep the existing flexibility. A simple `PackageManagerCommand.of("apt install unzip")` would set `PackageManager` to `apt` and the commands list to the single string `apt install unzip`.

hohwille (approved):
@KGoroll1 thank you for this great PR. You solved a lot of problems here and it seems to work nicely. Well done. 👍 

I left some comments for improvements.
As you already had to leave our team, I think it is best if I create a new issue for addressing the review comments that I have left. I can then merge as is and someone else from the team can take the issue to aim for perfection :)

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
hohwille: Maybe from security PoV and zero trust we should question my point 1.

PR #250, opened 2024-03-17
#36: Tool Commandlet for tomcat

### LINKED ISSUE #36: Implement ToolCommandlet for Tomcat
Create `ToolCommandlet` for Tomcat
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/tomcat

### LINKED ISSUE #131: Support for tool dependencies

aBega2000:
fixes #36, fixes #131 

- This is a pull request where Tomcat is implemented and tested. 

- Additionally the support for tool dependencies is implemented and tested.


cli/src/main/java/com/devonfw/tools/ide/tool/tomcat/Tomcat.java
```diff
@@ -0,0 +1,114 @@
...
+
+    switch (command) {
+      case START:
+        // printTomcatPort();
+        arguments.setValueAsString("start", context);
+        super.run();
+        break;
+      case STOP:
+        arguments.setValueAsString("stop", context);
+        this.context.warning("TEST 2");
```
ndemirca, line 66: was this line for testing purposes?

cli/src/test/resources/ide-projects/tomcat/project/conf/ide.properties
ndemirca, line 1: do you need these empty files?

ndemirca (changes requested):

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -203,4 +220,119 @@ private ToolInstallation createToolInstallation(Path rootDir, VersionIdentifier
...
   }
 
+  /**
+   * Method to get the Path of the dependencies Json file
+   * 
+   * @return the {@link Path} of the dependencies file for the tool
+   */
+  public Path getDependencyJsonPath() {
+
+    Path urlsPath = this.context.getUrlsPath();
```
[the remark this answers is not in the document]
aBega2000: Done

cli/src/main/java/com/devonfw/tools/ide/tool/tomcat/Tomcat.java
```diff
@@ -0,0 +1,114 @@
...
+   */
+  public Tomcat(IdeContext context) {
+
+    super(context, "tomcat", Set.of(Tag.JAVA));
+    this.command = add(new EnumProperty<>("", true, "command", TomcatCommand.class));
+    add(this.arguments);
+  }
+
+  @Override
+  public boolean install(boolean silent) {
```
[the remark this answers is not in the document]
aBega2000: Yes, this method is necessary for the installation of the tool

cli/src/main/java/com/devonfw/tools/ide/tool/tomcat/Tomcat.java
aBega2000: Yes, now removed

cli/src/test/resources/ide-projects/tomcat/project/conf/ide.properties
aBega2000: Yes, without this file the tests fail

cli/src/main/java/com/devonfw/tools/ide/tool/tomcat/Tomcat.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  @Override
+  public void run() {
+
+    TomcatCommand command = this.command.getValue();
+
+    switch (command) {
+      case START:
+        // printTomcatPort();
```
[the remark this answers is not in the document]
aBega2000: This call of the function is necessary, but for the moment I am still trying to implement it

cli/src/main/java/com/devonfw/tools/ide/tool/tomcat/Tomcat.java
```diff
@@ -0,0 +1,114 @@
...
+    EnvironmentVariables variables = this.context.getVariables();
+    EnvironmentVariables typeVariables = variables.getByType(EnvironmentVariablesType.CONF);
+
+    typeVariables.set("CATALINA_HOME", getToolPath().toString(), true);
+    typeVariables.save();
+  }
+
+  @Override
+  protected void initProperties() {
+
```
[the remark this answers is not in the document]
hohwille: The reason is that the property `command` needs to be added before the property `arguments` that is typically auto-added by this `initProperties` method and here done in line 34 in the constructor.
@aBega2000 However, instead of resolving with with "it is neccessary" you could have changed the comment `Empty on purpose` to a reasonable explanation to better address such review feedback.

cli/src/test/java/com/devonfw/tools/ide/tool/tomcat/TomcatTest.java
```diff
@@ -0,0 +1,74 @@
...
+    tomcatCommandlet.run();
+
+    // assert
+    checkRunningTomcat(context);
+  }
+
+  private void checkDependencyInstallation(IdeTestContext context) {
+
+    assertLogMessage(context, IdeLogLevel.INFO, "The version 17.0.10_7 of the dependency java is being installed");
+    assertLogMessage(context, IdeLogLevel.INFO, "The version 17.0.10_7 of the dependency java was successfully installed");
```
hohwille, line 51: nice, but that is the easy case.
What if your `JAVA_VERSION` from `ide.properties` is not compatible with the `TOMCAT_VERSION` you want to run?
Then an additional `java` version that is compatible needs to be installed and set as `JAVA_HOME` before the tomcat is started. This is where it gets interesting and this is not tested. I assume it might not be working.

cli/src/test/resources/ide-projects/tomcat/project/software/tomcat/conf/server.xml
```diff
@@ -0,0 +1 @@
+<Connector port="49152"/>
```
hohwille, line 1: After reading the test, I understand why this file is there and what this magic number is for.
However, what confused me is that this is in a preinstalled `software/tomcat` folder.
First of all in IDEasy the `software` folder should only contain symlinks to the actual software installed to the software repository.
Second, a valid tool folder in `software` needs to have a `.ide.software.version` file (or `.devon.software.version`).
As this is not present, it should actually not be able to determine the current installed version of `tomcat` and therefore install the proper one, delete this `software/tomcat` directory and then create a symlink here pointing to the proper installation of tomcat. If that is not the case, did you find a bug with your test if it is green?
I would expect this file to be in `repository/tomcat/tomcat/default/conf/server.xml` instead of in the `software` folder.

cli/src/test/resources/ide-projects/tomcat/repository/java/java/default/readme
```diff
@@ -0,0 +1 @@
+Readme file
```
hohwille, line 1: The software installation should contain something more reasonable that you also assert in your test to see that the proper thing was installed. Here are multiple `readme` files that contain the wise words `Readme file` so in case the wrong `readme` file was copied, your test could not even detect this.
Due to my concern about the more complex case to be covered by the test, I would recommend to have two `java` versions like some 11 and some 21 version and have one configured as `JAVA_VERSION` that is incompatible with the `TOMCAT_VERSION` that would require the other and then test that starting tomcat did use the second one.

hohwille (changes requested):
@aBega2000 Thanks for your work on this story and continuously improving 👍 
Very nice that now via the `dependency.json` file tomcat can get Java automatically installed without the need to have this hard-coded and with the flexibility to choose such things according to the version to be installed.

As I warned from the start this was a complex story not easy to implement.
I left several review comments for improvement.
My major concern is that the advanced scenarios like a version conflict are not yet tested and therefore most probably not working. Feel free to schedule a meeting if you need some help or have further questions.

cli/src/test/resources/ide-projects/tomcat/repository/tomcat/tomcat/default/linux/bin/tomcat
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+echo "tomcat $*"
```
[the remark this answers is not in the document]
aBega2000: Fixed. Now there is the same test for linux, mac and windows

cli/src/test/resources/ide-projects/tomcat/repository/tomcat/tomcat/default/linux/InstallTest.txt
```diff
@@ -0,0 +1 @@
+This is a test file.
```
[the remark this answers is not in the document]
aBega2000: Fixed. The test is now same for the three Operating Systems, and I have tried keeping only the necessary files required for the tests

cli/src/test/java/com/devonfw/tools/ide/tool/tomcat/TomcatTest.java
aBega2000: In the properites file I have inserted the version 8u402b06, which is actually not compatible with the version 10.1.14 of tomcat (this one requires version 11 or later of java) and even thought these versions are not compatible, the right version of java is being found from the dependency file, namely 17.0.10_7, and installed as dependency. Is this the correct behavior?

cli/src/test/resources/ide-projects/tomcat/project/software/tomcat/conf/server.xml
aBega2000: I changed the location of the file accordingly, and also the port number to a more understandable one, namely 8080. This is also the port in which tomcat runs by default.

cli/src/test/resources/ide-projects/tomcat/repository/java/java/default/readme
aBega2000: I have removed the unnecessary readme files and for the folders I want to have included in github, I have added a couple of .gitkeep files

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -196,4 +215,130 @@ private ToolInstallation createToolInstallation(Path rootDir, VersionIdentifier
...
+          this.context.info("Necessary version of the dependency {} is already installed in repository", dependencyName);
+        }
+      }
+    }
+  }
+
+  private void installDependencyInRepo(String dependencyName, ToolCommandlet dependencyTool, VersionIdentifier dependencyVersionToInstall) {
+
+    this.context.info("The version {} of the dependency {} is being installed", dependencyVersionToInstall, dependencyName);
+    LocalToolCommandlet dependencyLocal = (LocalToolCommandlet) dependencyTool;
```
hohwille, line 278: This cast does not seem to make sense. If this method is defined in `LocalToolCommandlet` and also requires a `LocalToolCommandlet` as argument, then do not take `ToolCommandlet` as argument.
However, if you already want to think further to potentially also support dependencies to global tools in the future, this is great. Exactly for this purpose I created this code-spot for the feature you are building:
https://github.com/devonfw/IDEasy/blob/6b87d34f85ca03f286cd0b8b90d7b485b8835f49/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java#L107

hohwille (changes requested):
@aBega2000 thanks for taking this complex story further. You already made some good progress 👍 
However, I still think that this story is too complex to get it solved via review comment feedback ping-pong.
IMHO we need to arrange a screen-sharing session and get this solved together to come to an end in a more efficient way.

cli/src/main/java/com/devonfw/tools/ide/tool/Dependency.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Stream;
+
+/**
+ * Class to represent the functionality of installing the dependencies when a tool is being installed.
+ */
+public class Dependency {
```
hohwille, line 24: Should be part of the URL model (`com.devonfw.tools.ide.url.model`):
```
public class UrlDependenciesFile extends AbstractUrlFile<UrlEdition> {
 ...
}
```
Does not need to depend on `IdeContext`.
In `UrlEdition` you can integrate this easily just like the `securityFile`.

hohwille (approved):
@aBega2000 I reviewed and tested your latest code of this PR and tomcat is working. Great work👍 
For the lots of review comments, that we discussed in a personal meeting, I created issue #439 for the review rework and therefore resolve the review comments and merge this as is.

PR #257, opened 2024-03-20
#254: Part 2: add missing quotation marks

### LINKED ISSUE #254: ide wrapper bash script failing:
### Expected behavior

As a IDEasy user, I want to set my environment variables using `ide` command so that I can work as simple as with `devon` CLI.

### Actual behavior

```
$ ide
/d/projects/IDEasy/test/bin/ide: eval: line 22: syntax error near unexpected token `('
```

1. run `ide` wrapper script

### Related/Dependent Issues

* issue #63 
* PR #221 

### Comments/Hints:

According to my analysis the error seems to come from path entries with `Program Files (x86)`.
My assumption is that we need to double quote the variable values at least for bash.

### Affected version:

* OS: Windows
* latest SNAPSHOT of IDEasy.

VinceHeu:
Closes #254


hohwille (approved):
@VinceHeu thanks for this PR and the fast fix. Solution is perfect 👍 
Ready for merge.

Issue #260, opened 2024-03-22
ide wrapper script ends terminal on error


hohwille:
### Expected behavior

As a IDEasy user, I want to use `ide` command seamless like `devon` so that I can switch from devonfw-ide to IDEasy.

### Actual behavior

In order to change environment variables the `ide` wrapper script has to be sourced as otherwise it is called as a sub-process with its own inherited environment variables and on exit you return to the parent process what is the users shell (e.g. `git-bash`) and the environment variable changes are lost and have no effect. (see #63):
```
echo 'alias ide="source $(which ide)"' > ~/.bashrc
```
Now, I caused a bug when I tried to simplify the error handling in PR #221 when I decided to use `set -e`.
This is causing to exit the process as soon as an error happens what seemed to make sense at the first glance.
However, now that the `ide` script is sourced, we need to understand that process is our shell (e.g. `git-bash`) and the sourcing behaves as if the code is pasted into my terminal.
So as a result now on any error the process - what is the shell itself - exits.
Therefore, if you are not inside a devonfw-ide/IDEasy project, you will get the message `You are not inside an IDE installation: D:\projects` and an error (exit code not `0`) what will exit your shell.
Or to make it even more obvious:
Using `set -e` in the shell itself is a no-go as any non successful command (`grep 'nonexistent-word' file` or `[ 0 = 1 ]`) will exit your shell.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. install IDEasy
2. add the alias to your `.bashrc` (see above)
3. open a shell in a non devonfw-ide/IDEasy project (e.g. `C:\Windows`)
4. call `ide`

Your shell exits and your terminal closes (or your tab if you are using Windows-Terminal and have other tabs open).

### Related/Dependent Issues

* #63 
* #5 

### Comments/Hints:

We have to reintroduce the error handling (with `return_code=$?`).
Now that we have working SNAPSHOT releases and the major blockers resolved, it is very easy to locally modify the `ide` wrapper script and do a lot of testing of all the variants in order to create a PR fixing the problem.

### Affected version:

* OS: any
* Shell: Bash (but also zsh or other GNU shells)

mvomiero:
## Adding an alias for ide:
⚠️ please note that this line of code doesen't work because of three reasons:
```bash
echo 'alias ide="source $(which ide)"' > ~/.bashrc
```
1. the ` ' ` prevent the interpretation of the variable `$(which ide)`
[...]

PR #270, opened 2024-03-27
#260: Implement exit code in wrapper script

mvomiero:
Fixes: #260 

Note:
* the if statement to check if arguments are given is removed, since Ideasy should execute even in this case (printing the help usage for the user)


jan-vcapgemini (approved):
Looks fine to me.

Issue #204, opened 2024-02-14
Wrong commandlet version as CLI argument during installation causes problems


mvomiero:
### Expected behavior

we have to choose how to handle it. I suppose a warning or an exception. 

### Actual behavior

when running `install <commandlet> <version>` , the JAVA_VERSION variable in ide.properties is set even if the version is wrong, and the commandlet is not executed because the version is not found (see example down).

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

Example:

*  run with CLI arguments: `install java blabla` (instead of a proper version)
*  output:
    ` JAVA_VERSION=blabla has been set in SETTINGS@C:\Projects\devonfw\settings\ide.properties`
    ` java.lang.IllegalArgumentException: Version blabla for tool java does not exist in edition java.`

And since the variable is stored, the problem persists for the next executions, or for other commandlets if they are for example based in java.

PR #271, opened 2024-03-27
#204: Non existing version installation

mvomiero:
Fixes: #204


VinceHeu (approved):

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -366,6 +370,17 @@ public void setVersion(VersionIdentifier version, boolean hint) {
...
 
+  private void checkVersionExistence(String tool, String edition, VersionIdentifier version) {
+
+    try {
+      UrlMetadata metadata = this.context.getUrls();
+      metadata.getVersionFolder(tool, edition, version);
+    } catch (IllegalArgumentException e) {
+      throw new CliException(
+          "Version " + version + " for tool " + tool + " does not exist in edition " + edition + ".");
+    }
```
hohwille, line 381: The method `getVersion` already throws a `CliException` if a version pattern is given but no matching version could be found.
IMHO it would make sense to change `getVersionFolder` to throw `CliException` instead of `IllegalArgumentException` and avoid this catch block. And be aware that such "exception mapping" is typically kind of unsafe as some other developer might add another `if` check also throwing `IllegalArgumentException` for a different purpose and then we would cause a lot of confusion with this. Also you are not passing the catched exception `e` to the new exception.

hohwille (approved):
@mvomiero thanks for your PR and the nice improvement. 👍 
While your code is correct, I left a review comment for improvement.
Please have a look and consider as small update. Apart from that ready for merge.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
mvomiero: Perfect, makes totally sense. Then the check could be done directly in one line in `setVersion `so having a function `checkVersionExistence` doesent make sense anymore.

PR #273, opened 2024-03-28
#22: implement tool commandlet graalvm

### LINKED ISSUE #22: Implement ToolCommandlet for GraalVM
Create `ToolCommandlet` for GraalVM
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/graalvm

VinceHeu:
fixes #22


cli/src/test/java/com/devonfw/tools/ide/tool/graalvm/GraalVmTest.java
```diff
@@ -0,0 +1,87 @@
...
+    // arrange
+    IdeTestContext context = newContext(PROJECT_GRAALVM);
+    GraalVm commandlet = new GraalVm(context);
+
+    // act
+    commandlet.run();
+
+    // assert
+    String graalvmExport = "export GRAALVM_HOME=" + context.getSoftwarePath().resolve("extra/graalvm");
+    Path file = context.getConfPath().resolve("devon.properties");
```
mvomiero, line 50:
```suggestion
    Path file = context.getConfPath().resolve("ide.properties");
```

since now the variables are stored in `ide.properties`

jan-vcapgemini:
Thanks for your addition.
I've checked the graalvm process and think that we need to check if graalvm is already installed.
Currently the whole install process (copying of files, symlink creation etc.) gets repeated each time I start graalvm.
Maybe add a check for ide.software.version before calling install again.

mvomiero (changes requested):
Good job in developing this commandlet!🌴 
some implementations and it is good to go!

cli/src/main/java/com/devonfw/tools/ide/tool/graalvm/GraalVm.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  @Override
+  public void run() {
+
+    Path toolPath = getToolPath();
+    try {
+      if (!Files.exists(toolPath)) {
+        Files.createDirectories(toolPath);
+      }
```
mvomiero, line 50:
```suggestion
    try {
      if (!Files.exists(toolPath)) {
        Files.createDirectories(toolPath);
        super.install(true);
      }
```

mvomiero: Probably you could nest the installation here, so it is triggered just if the tool is not installed I suppose... otherwise the installation process gets started every time the commandlet is run 😺

cli/src/main/java/com/devonfw/tools/ide/tool/graalvm/GraalVm.java
```diff
@@ -0,0 +1,65 @@
...
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to create new directory:" + e);
+    }
+    super.install(true);
```
mvomiero, line 54: then remove it from here
```suggestion
```

mvomiero (commented):
Except a super small suggestion, LGTM 🌞

cli/src/main/java/com/devonfw/tools/ide/tool/graalvm/GraalVm.java
VinceHeu: I think it has to be outside of this if-condition. Otherwise it's not possible to install another update/version after the first installation.

cli/src/main/java/com/devonfw/tools/ide/tool/graalvm/GraalVm.java
```diff
@@ -0,0 +1,66 @@
...
+
+    Path toolPath = getToolPath();
+    try {
+      if (!Files.exists(toolPath)) {
+        Files.createDirectories(toolPath);
+
+      }
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to create new directory:" + e);
+    }
```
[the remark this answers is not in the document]
hohwille: However, I do not understand the entire purpose of this.
If that folder is not present did you face a error if you do not create it?
IMHO then that "error" should be fixed in `LocalToolCommandlet`.
From what I can see the folder is created here:
https://github.com/devonfw/IDEasy/blob/bd09ce4b82e387cc323a2cb996eada7c0dd86e6a/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java#L84

But maybe the logic to determine the already existing version fails if the path does not exist or why did you do this?

cli/src/main/java/com/devonfw/tools/ide/tool/graalvm/GraalVm.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @Override
+  public void postInstall() {
+
+    EnvironmentVariables envVars = this.context.getVariables().getByType(EnvironmentVariablesType.CONF);
+    envVars.set("GRAALVM_HOME", getToolPath().toString(), true);
+    envVars.save();
+    super.postInstall();
+  }
```
hohwille, line 65: This is how we solved such problems in the bash code of devonfw-ide.
In IDEasy we have a much smarter solution by simply defining `GRAALVM_HOME` in `IdeVariables`.
WDYT?
p.s.: This is nice to have. We can also merge and later add an improvement for this. Besides there are already other occurrences of this "pattern" already in IDEasy so we could better create a new issue to do a cleanup/refactoring of this pattern with `IdeVariables` for all occurrences and resolve this conversation. Also this is something to be announced and shared in the team so every developer gets aware of this learning.

cli/src/main/java/com/devonfw/tools/ide/tool/graalvm/GraalVm.java
```diff
@@ -0,0 +1,66 @@
...
+   */
+  public GraalVm(IdeContext context) {
+
+    super(context, "graalvm", Set.of(Tag.JAVA));
+  }
+
+  @Override
+  public Path getToolPath() {
+
+    return this.context.getSoftwarePath().resolve("extra").resolve(getName());
```
hohwille, line 33: Could you add a constant for this folder in `IdeContext` as we had more than one tool using this folder and with string literals we will cause redundancy then.
The general purpose of the `extra` folder is to separate tools that should not be added to the `PATH` variable automatically.

cli/src/test/resources/ide-projects/graalvm/repository/graalvm/graalvm/default/windows/bin/testFile.cmd
```diff
@@ -0,0 +1 @@
+echo graalvm windows %*
```
hohwille, line 1: this file is IMHO never used at all. Best practice is to avoid pointless clutter and only add test resources that are actually needed. You have a 2nd copy of it in the `default` edition. Also I wonder if we need the `community` edition at all. IMHO also the entire `repository/graalvm/community` folder is pointless and can be deleted.

hohwille:
I also just found this one:
```
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.023 s <<< FAILURE! -- in com.devonfw.tools.ide.tool.graalvm.GraalVmTest
[ERROR] com.devonfw.tools.ide.tool.graalvm.GraalVmTest.testGraalVmInstallInDirectoryExtra -- Time elapsed: 0.015 s <<< FAILURE!
java.lang.AssertionError: Step Install graalvm already ended with true and cannot be ended again with false
	at com.devonfw.tools.ide.step.StepImpl.end(StepImpl.java:137)
	at com.devonfw.tools.ide.step.StepImpl.error(StepImpl.java:123)
	at com.devonfw.tools.ide.step.Step.error(Step.java:142)
	at com.devonfw.tools.ide.tool.LocalToolCommandlet.doInstall(LocalToolCommandlet.java:96)
```
This does not look like an error cause by you but an existing bug in the code (most probably from me).
What confuses me is that the CI build for main is green so maybe there is a bug but your test is the first constellation revealing it. I will investigate this for you so you do not have to bother with this...

hohwille (changes requested):
added ide.properties to fix build

hohwille:
OK, my bad: I broke your green build with [this commit](https://github.com/devonfw/IDEasy/pull/273/commits/49e5b6bd3d34d45a2966538929791871f387f15a).
But here is the good news:
* It revealed the bug with the `Step` described above
* It revealed another bug in our EnvironmentVariable infrastructure: If the `ide.properties` or `devon.properties` does not exist at all in the location (e.g. `conf`) then it is `null` and `save()` will fail.
* It revealed what the purpose of the "empty" properties file in `conf` was about and so I re-added it with a comment explaining its purpose.

cli/src/main/java/com/devonfw/tools/ide/tool/graalvm/GraalVm.java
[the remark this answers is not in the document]
VinceHeu: I was thinking too complicated and I didn't have in mind that the folders are also created there. I changed it

cli/src/main/java/com/devonfw/tools/ide/variable/IdeVariables.java
```diff
@@ -49,7 +48,7 @@ public interface IdeVariables {
   VariableDefinitionString DOCKER_EDITION = new VariableDefinitionString("DOCKER_EDITION", null, c -> "rancher");
 
   /** {@link VariableDefinition} for {@link com.devonfw.tools.ide.context.IdeContext#getWorkspaceName() WORKSPACE}. */
-  VariableDefinitionString GRAALVM_EDITION = new VariableDefinitionString("GRAALVM_EDITION", null, c -> "community");
+  VariableDefinitionString GRAALVM_EDITION = new VariableDefinitionString("GRAALVM_EDITION", null, c -> "graalvm");
```
hohwille, line 51: This is the default anyway so we can delete it now.
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/tool/graalvm/GraalVm.java
```diff
@@ -0,0 +1,54 @@
...
+  public VersionIdentifier getInstalledVersion() {
+
+    return super.getInstalledVersion(getToolPath());
+  }
+
+  @Override
+  public void run() {
+
+    super.install(true);
+  }
```
hohwille, line 44: I guess we can stay with the default now. However, we need to verify that the proper toolname is set (you might need to override `getBinaryName()`) as there is no `graalvm` binary but instead `java` (or `javac` or `gu`, etc.).
Also you most probably need to change the `PATH` to prepend with GRAALVM_HOME/bin or you need to make the `Path` see variable `binaryPath` in "super.run()" absolute instead of relative...
```suggestion
```

hohwille (approved):
@VinceHeu thanks for the updates. You are approaching the finishing line. 👍 
I left comments for the final completion.
After that we can merge. I will also create a new story for plugin support via `gu install`.

PR #274, opened 2024-04-02
#256: improve package manager API 

### LINKED ISSUE #256: Improve package manager installation from PR 249
In PR #249 I left several review comments for improvement.
To avoid blockage, I decided to merge PR #249 but create this issue so someone else from the team can later address my review comments and implement the suggested improvements:

* [ ] centralize a `findBash()` method
* [ ] improve API of `installWithPackageManger` method.

salimbouch:
closes #256


cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -855,4 +857,81 @@ public boolean apply(CliArguments arguments, Commandlet cmd, CompletionCandidate
     return true;
   }
 
+  public Path findBashPath() {
+
+    Path bashPath;
+    if (systemInfo.isLinux() || systemInfo.isMac()) {
+      bashPath = Path.of("/bin/bash");
```
hohwille, line 864: The existing behavior was to find `bash` on the `PATH` as not all unix variants have it in `/bin/`:
```suggestion
      bashPath = Path.of("bash");
```

As this will currently raise an invalid warning, we would need to either only do the warning in the Windows case or we actually should consider to find it from the `PATH` here.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -855,4 +857,81 @@ public boolean apply(CliArguments arguments, Commandlet cmd, CompletionCandidate
     return true;
   }
 
+  public Path findBashPath() {
+
+    Path bashPath;
+    if (systemInfo.isLinux() || systemInfo.isMac()) {
```
hohwille, line 863: Invert the logic  and use `if (SystemInfoImpl.INSTANCE.isWindows()) { ... } else { ... }`.
This was actually inconsistent before your PR change (see `ProcessContextImpl`).

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -855,4 +857,81 @@ public boolean apply(CliArguments arguments, Commandlet cmd, CompletionCandidate
...
+    return null;
+  }
+
+  private String findBashOnWindows() {
+
+    // Check if Git Bash exists in the default location
+    Path defaultPath = Path.of("C:\\Program Files\\Git\\bin\\bash.exe");
+    if (Files.exists(defaultPath)) {
+      return defaultPath.toString();
+    }
```
hohwille, line 892: Why do we actually have these methods returning the `Path` as `String`?
IMHO we should only have methods returning it as `Path`.
As you can see in `findBashOnWindows()` we are creating it as `Path` and then converting it to string in order to convert it back to `Path` and if called from `findBash()` again converting it to `String`. This does not make sense to me.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -356,7 +305,7 @@ private void modifyArgumentsOnBackgroundProcess(ProcessMode processMode) {
     // try to use bash in windows to start the process
     if (context.getSystemInfo().isWindows()) {
 
-      String findBashOnWindowsResult = findBashOnWindows();
+      String findBashOnWindowsResult = this.context.findBash();
       if (findBashOnWindowsResult != null) {
 
         bash = findBashOnWindowsResult;
```
hohwille, line 311: here we can now replace all the code with just `String bash = this.context.findBash()` independent of the OS.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
salimbouch: The bash variable will be added to the ProcessBuilder, which requires the path of bash as a string. Can't we just provide the path as a String then?

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
salimbouch: only one method `findBash()` now. Bash is always available in linux and mac returning "bash" is enough, and in windows() we decided that it is a prerequisite for using IDEasy (findBashOnWindows() throws Exception when no Bash is found). So something like should be enough?

```
  public String findBash() {

    String bash = "bash";
    if (SystemInfoImpl.INSTANCE.isWindows()) {
      bash = findBashOnWindows();
    }

    return bash;
  }
```

hohwille: I guess the idea was to get a reasonable error if `bash` is not found because it is not on the `PATH` (maybe IDEasy had a bug and broke the `PATH`).
But in general you are correct that every reasonable linux and also MacOS should have `bash` installed and have it on `PATH`.
Also the question is what Java `Process` and `ProcessBuilder` are doing when we invoke `bash` and it is not on `PATH`. This should also raise an error telling that `bash` was to be executed but could not be found.
So maybe your KISS suggestion is valid.
However, I would still argue that the method should return `Path` instead of String as already commented.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
hohwille: You are right. If you change these `findBash*` methods to return `String` it is also fine.
It was just confusing to have both and convert back and forth.
So applying your simple code snippet suggested above should do it and we can resolve both review conversations.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
hohwille: Sorry, if I was not clear enough.
I was suggesting to remove the entire code block
```
String bash = "bash";

// try to use bash in windows to start the process
if (context.getSystemInfo().isWindows()) {

  String findBashOnWindowsResult = this.context.findBash();
  if (findBashOnWindowsResult != null) {

    bash = findBashOnWindowsResult;
  } else {
     context.warning(
            "Cannot start background process in windows! No bash installation found, output will be discarded.");
     this.processBuilder.redirectOutput(Redirect.DISCARD).redirectError(Redirect.DISCARD);
     return;
   }
 }
 ```
 With just this:
 ```
   String bash = this.context.findBash();
 ```
 And maybe keep this:
 ```
  if (bash == null) {
     context.warning("Cannot start background process in windows! No bash installation found, output will be discarded.");
     this.processBuilder.redirectOutput(Redirect.DISCARD).redirectError(Redirect.DISCARD);
     return;
   }
```

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -306,7 +255,7 @@ private String addExecutable(String executable, List<String> args) {
       // here we want to have native OS behavior even if OS is mocked during tests...
       // if (this.context.getSystemInfo().isWindows()) {
       if (SystemInfoImpl.INSTANCE.isWindows()) {
-        String findBashOnWindowsResult = findBashOnWindows();
+        String findBashOnWindowsResult = this.context.findBash();
```
hohwille: Same here. This code does not make sense any more.
We can simplify the entire block
```
String bash = "bash";
      interpreter = bash;
      // here we want to have native OS behavior even if OS is mocked during tests...
      // if (this.context.getSystemInfo().isWindows()) {
      if (SystemInfoImpl.INSTANCE.isWindows()) {
        String findBashOnWindowsResult = findBashOnWindows();
        String findBashOnWindowsResult = this.context.findBash();
Comment
        if (findBashOnWindowsResult != null) {
          bash = findBashOnWindowsResult;
        }
      }
      args.add(bash);
```

With just:
```
  interpreter = "bash";
  args.add(this.context.findBash());
```

salimbouch:
@hohwille FYI: I still haven't completely tested the changes to Docker and GlobalToolCommandlet on linux.

PR #276, opened 2024-04-02
#63: check if inside ide installation for wrapper script

mvomiero:
Fixes: #63 

I haven't fully understood this part of the issue:
_"Please note that you cannot output the message to std out from ideasy env --bash. So either we need to print it as error or we have to think of a different approach"_

With this PR, the following is output:
* `IDE environment variables have been set for ...` to stdout if inside an IDE installation
* `You are not inside an IDE installation: ...` to stderr if outside an IDE installation

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
@@ -126,11 +126,10 @@ public abstract class AbstractIdeContext implements IdeContext {
    * @param minLogLevel the minimum {@link IdeLogLevel} to enable. Should be {@link IdeLogLevel#INFO} by default.
    * @param factory the {@link Function} to create {@link IdeSubLogger} per {@link IdeLogLevel}.
    * @param userDir the optional {@link Path} to current working directory.
-   * @param toolRepository @param toolRepository the {@link ToolRepository} of the context. If it is set to {@code null}
-   * {@link DefaultToolRepository} will be used.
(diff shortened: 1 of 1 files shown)


cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -717,25 +711,34 @@ public void setLogLevel(IdeLogLevel logLevel) {
...
     CliArgument current = arguments.current();
     if (!current.isEnd()) {
       String keyword = current.get();
       Commandlet firstCandidate = this.commandletManager.getCommandletByFirstKeyword(keyword);
       boolean matches;
-      if (firstCandidate != null) {
+      if (firstCandidate == null) {  // no arguments
+        if (current.getArgs().isEmpty()) {
+          info(getMessageIdeHomeFound());
+          return ProcessResult.SUCCESS;
```
jan-vcapgemini, line 733: Better remove the return, otherwise we can't show the help if no arguments were provided.
```suggestion

```

jan-vcapgemini (changes requested):
Thanks for your addition. I've just added one small CR.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
hohwille: I do not get this change.
If the `firstCandidate` was not found, the first argument is not the keyword of a `Commandlet`.
This can happen e.g. if a add an option to a commandlet (see e.g. #279).
In such case the second `for` loop below handles the case to find the proper commandlet.
This however has nothing to do with the fact if `IDE_HOME` was found or not.

tobka777:
already fixed in #282

PR #286, opened 2024-04-12
#130: Add default content length to progress bar

### LINKED ISSUE #130: ProgressBar AssertionError, if content-length not set
### Expected behavior

If a download does not provide a content length it should still be downloaded using the workaround show [here](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java#L104).
### Actual behavior

An assertion error will be thrown if the content length is not set and/or provided with the download. This error is throw in the IdeProgressBarTestImpl as it asserts that the total of the progress bar equals its max before closing, as can be seen [here](https://github.com/devonfw/IDEasy/blob/main/cli/src/test/java/com/devonfw/tools/ide/io/IdeProgressBarTestImpl.java#L57).

1.  Mock a server, which contains/downloads a File without specifing the Content length in the header
2.  Run the InstallCommandlet with the URL from the mocked server.

### Affected version:

* OS: Windows
* OS : probably Linux/Mac aswell

ndemirca:
Closes #130


cli/pom.xml
```diff
@@ -90,6 +90,12 @@
       <artifactId>jansi</artifactId>
       <version>${jansi.version}</version>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.13.2</version>
```
hohwille, line 96: we already use JUnit 5 for our tests.
Why do you want to add JUnit 4 that is outdated?

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -131,9 +123,14 @@ private void downloadFileWithProgressBar(String url, Path target, HttpResponse<I
           fileComplete = true;
         } else {
           bufferedOut.write(data, 0, count);
-          pb.stepBy(count);
+          if (contentLength > 0) {
+            pb.stepBy(count);
+          } else {
+            break;
```
hohwille, line 129: this `break` will cause that the download is aborted without writing the file to the end.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -147,24 +144,44 @@ private void downloadFileWithProgressBar(String url, Path target, HttpResponse<I
...
       int readBytes;
 
       try (IdeProgressBar pb = this.context.prepareProgressBar("Copying", size)) {
         while ((readBytes = in.read(buf)) > 0) {
           out.write(buf, 0, readBytes);
-          pb.stepByOne();
+          if (size > 0) {
+            pb.stepByOne();
+          } else {
+            break;
```
hohwille, line 160: same pattern here

cli/pom.xml
ndemirca: Import added this one automatically. I will change it to Junit5

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -147,24 +144,44 @@ private void downloadFileWithProgressBar(String url, Path target, HttpResponse<I
...
+
+    if (contentLength == 0) {
+      progressBar.stepBy(10000000);
+    }
+  }
+
+  private void informAboutSettingDefaultContentLength(long contentLength) {
+
+    if (contentLength == 0) {
+      this.context.warning("Content-Length was not provided by download/copy source. Using fallback: Content-Length for the progress bar is set to 10000000.");
```
hohwille, line 181: The default value `10000000` is more or less randomly chosen but there is not big rationale behind it.
If you copy it to multiple places, it would be better to define a constant.

BTW: Before the log statement was also including the URL. IMHO it is helpful to have this information to see where the problem is coming from e.g. if we receive logs via github issues or support.

cli/src/test/java/com/devonfw/tools/ide/io/FileAccessImplTest.java
```diff
@@ -267,6 +270,52 @@ public void testSymlinkShortcutPaths(@TempDir Path tempDir) {
...
+  /**
+   * Test of {@link FileAccessImpl#download(String, Path)} with default value for missing content length if the url starts with http.
+   */
+  @Test
+  public void testDownloadWithDefaultValueForMissingContentLength() {
+
+    // arrange
+    IdeTestContext context = newContext("basic");
+    FileAccess fileAccess = new FileAccessImpl(context);
+    String url = "https://getsamplefiles.com/download/zip/sample-1.zip";
```
hohwille, line 282: Aint it a bug of this external service to have no `Content-Length` header?
Assuming they will "fix" this "bug" then this test will break? Or not break anymore even in case the code is not working without `Content-Length`?
If we need to test something like this explicitly then probably wiremock is the better choice and for the progress bar we already have tests using wiremock...

cli/pom.xml
hohwille: IntelliJ is smart but can also cause harm with too much magic ;)
Thanks for the update 👍

jan-vcapgemini (changes requested):
This fix looks quite solid, thanks. I've added some small CR's. After these, the PR should be ready for review.

cli/src/test/java/com/devonfw/tools/ide/io/IdeProgressBarTest.java
```diff
@@ -1,42 +1,106 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.any;
+import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlMatching;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Test of {@link IdeProgressBar}.
+ */
+@WireMockTest(httpPort = 8080)
```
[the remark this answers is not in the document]
ndemirca: This can be a new investigation task. I did not change this one, instead I added a constant at this place but the formatting is displaying this as a change.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
```diff
@@ -14,10 +14,9 @@ public interface FileAccess {
    * Downloads a file from an arbitrary location.
    *
    * @param url the location of the binary file to download. May also be a local or remote path to copy from.
-   * @param targetFile the {@link Path} to the target file to download to. Should not already exists. Missing parent
-   *        directories will be created automatically.
+   * @param targetFile the {@link Path} to the target file to download to. Should not already exists. Missing parent directories will be created automatically.
    */
-  void download(String url, Path targetFile);
+  void download(String url, Path targetFile, boolean test);
```
jan-vcapgemini, line 19: Please add `test` param to javadoc.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -145,26 +143,55 @@ private void downloadFileWithProgressBar(String url, Path target, HttpResponse<I
    * @param source Path of file to copy
    * @param target Path of target directory
    */
-  private void copyFileWithProgressBar(Path source, Path target) throws IOException {
+  private void copyFileWithProgressBar(Path source, Path target, boolean test) throws IOException {
```
jan-vcapgemini, line 146: Please add `test` param to javadoc.

cli/src/test/java/com/devonfw/tools/ide/io/IdeProgressBarTest.java
```diff
@@ -1,42 +1,100 @@
...
+   * Tests if {@link FileAccessImpl#download(String, Path, boolean)} with default value for missing content length is working properly.
+   *
+   * @param tempDir temporary directory to use.
+   */
+  @Test
+  public void testProgressBarCopyWithDefaultValueForMissingContentLength(@TempDir Path tempDir) {
+
+    //arrange
+    String taskName = "Copying";
+    stubFor(any(urlMatching("/os/.*")).willReturn(aResponse().withStatus(200).withBody(new byte[MAX_LENGTH])));
```
jan-vcapgemini, line 86: Can be removed.
```suggestion
```

jan-vcapgemini (changes requested):
Thank you for your last adjustments. I've added some very small CR's.

jan-vcapgemini (approved):

cli/src/main/java/com/devonfw/tools/ide/io/AbstractIdeProgressBar.java
```diff
@@ -0,0 +1,68 @@
...
+    if (this.maxLength > 0) {
+      // check if maximum overflow
+      if (this.currentProgress > this.maxLength) {
+        this.currentProgress = this.maxLength;
+        doStepTo(this.maxLength);
+        return;
+      }
+    }
+
+    doStepBy(stepSize, this.currentProgress);
```
hohwille, line 57:
```suggestion
    doStepBy(stepSize, stepSize);
```

cli/src/main/java/com/devonfw/tools/ide/io/AbstractIdeProgressBar.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  @Override
+  public void close() {
+    if (this.currentProgress < this.maxLength) {
+      // TODO: Check if doStepTo should be used instead.
+      doStepBy(this.maxLength - this.currentProgress, this.currentProgress);
```
hohwille, line 64: Agreed:
```suggestion
      doStepTo(this.maxLength);
```

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -161,26 +159,49 @@ private void downloadFileWithProgressBar(String url, Path target, HttpResponse<I
...
 
     try (InputStream in = new FileInputStream(source.toFile()); OutputStream out = new FileOutputStream(target.toFile())) {
+      long size;
+      if (test) {
+        size = 0L;
+      } else {
+        size = source.toFile().length();
+      }
```
hohwille, line 172: if you need to test this somehow, extract to protected method `long getFileSize(Path)` create a subclass of `FileAccessImpl` for test that overrides and returns `0` instead. Then in your JUnit replace the real `FileAccess` with the test instance... Seems a little bit more complicated but adding this `test` flag to our official API seems odd to me.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
```diff
@@ -24,8 +24,9 @@ public interface FileAccess {
    * @param url the location of the binary file to download. May also be a local or remote path to copy from.
    * @param targetFile the {@link Path} to the target file to download to. Should not already exists. Missing parent directories will be created
    *     automatically.
+   * @param test boolean indicating if this is used in a test or not.
    */
-  void download(String url, Path targetFile);
+  void download(String url, Path targetFile, boolean test);
```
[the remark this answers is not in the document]
ndemirca: this is a left over from the first draft and should be removed

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -161,26 +159,49 @@ private void downloadFileWithProgressBar(String url, Path target, HttpResponse<I
...
       } catch (Exception e) {
         throw new RuntimeException(e);
       }
     }
   }
 
+  private void informAboutSettingDefaultContentLength(long contentLength, String url, Path path) {
+
+    String source;
+    if (contentLength == 0) {
```
hohwille, line 194: why are you using `0` for undefined?
Files can be `0` bytes large and then the user should not get this warning for every file that was empty but gets copied.
Also if the file size is know but smaller than e.g. 1k we should not give a progress bar.
What is good is that we have a dedicated method `copyFileWithProgressBar` for this use-case so we do not do this for every file. However, I assume this is used when we get a tool release package from a custom tool repository using a file URL and that can also be a network mount that is slower so already copying 1MB could take some seconds...

cli/src/main/java/com/devonfw/tools/ide/io/IdeProgressBarConsole.java
```diff
@@ -1,33 +1,88 @@
...
+    ProgressBarBuilder pbb = new ProgressBarBuilder();
+    // default (COLORFUL_UNICODE_BLOCK)
+    pbb.setStyle(ProgressBarStyle.builder().refreshPrompt("\r").leftBracket("\u001b[33m│").delimitingSequence("")
+        .rightBracket("│\u001b[0m").block('█').space(' ').fractionSymbols(" ▏▎▍▌▋▊▉").rightSideFractionSymbol(' ')
+        .build());
+    // set different style for Windows systems (ASCII)
+    if (this.systemInfo.isWindows()) {
+      pbb.setStyle(ProgressBarStyle.builder().refreshPrompt("\r").leftBracket("[").delimitingSequence("")
+          .rightBracket("]").block('=').space(' ').fractionSymbols(">").rightSideFractionSymbol(' ').build());
+    }
```
hohwille, line 50: Now that we have this code in review again:
It would be better to set the default stype in an `else` block. You are first stetting the style and then on windows setting it again overriding it.
Or even better since both variants share the same code fragment:
```suggestion
    String leftBracket, rightBracket, fractionSymbols;
    char block;
    if (this.systemInfo.isWindows()) {
      // set different style for Windows systems (ASCII)
      leftBracket = "[";
      rightBracket = "]";
      fractionSymbols = ">";
      block = '=';
    } else {
      // default (COLORFUL_UNICODE_BLOCK)
      leftBracket = "\u001b[33m│";
      rightBracket = "│\u001b[0m";
      fractionSymbols = " ▏▎▍▌▋▊▉";
      block = '█';
    }
    pbb.setStyle(ProgressBarStyle.builder().refreshPrompt("\r").leftBracket(leftBracket).delimitingSequence("")
        .rightBracket(rightBracket).block(block).space(' ').fractionSymbols(fractionSymbols).rightSideFractionSymbol(' ')
        .build()); 
```

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
jan-vcapgemini: I've set the default for undefined to -1 now.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
jan-vcapgemini: I've implemented this as you've suggested, thanks.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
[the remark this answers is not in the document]
jan-vcapgemini: Removed now.

cli/src/main/java/com/devonfw/tools/ide/io/AbstractIdeProgressBar.java
```diff
@@ -0,0 +1,68 @@
...
+        return;
+      }
+    }
+
+    doStepBy(stepSize, this.currentProgress);
+  }
+
+  @Override
+  public void close() {
+    if (this.currentProgress < this.maxLength) {
```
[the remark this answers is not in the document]
jan-vcapgemini: I've added a check for `maxLength < 0` which will simply return and not change the progress bar anymore.

jan-vcapgemini:
I'm closing this PR and continuing it in #575 for a fresh team review and less cluttering.

Issue #264, opened 2024-03-22
Prevent windows file lock error when installing tool update


hohwille:
### Expected behavior

As a IDEasy user, I want to install a new tool version seamlessly so that I can stay up-to-date and have no hustle.

### Actual behavior

```
$ ideasy install java 17.0.6_10
JAVA_VERSION=17.0.6_10 has been set in SETTINGS@D:\projects\IDEasy\settings\ide.properties
Creating backup by moving D:\projects\IDEasy\software\java to D:\projects\IDEasy\updates\backups\2024-03-22\java_17-13-29
An unexpected error occurred!
We are sorry for the inconvenience.
Please check the error below, resolve it and try again.
If the error is not on your end (network connectivity, lack of permissions, etc.) please file a bug:
https://github.com/devonfw/ide/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=IllegalStateException%3A+Failed+to+move+D%3A%5Cprojects%5CIDEasy%5Csoftware%5Cjava+to+D%3A%5Cprojects%5CIDEasy%5Cupdates%5Cbackups%5C2024-03-22%5Cjava_17-13-29
java.lang.IllegalStateException: Failed to move D:\projects\IDEasy\software\java to D:\projects\IDEasy\updates\backups\2024-03-22\java_17-13-29
        at com.devonfw.tools.ide.io.FileAccessImpl.move(FileAccessImpl.java:272)
        at com.devonfw.tools.ide.io.FileAccessImpl.backup(FileAccessImpl.java:261)
        at com.devonfw.tools.ide.tool.LocalToolCommandlet.doInstall(LocalToolCommandlet.java:79)
        at com.devonfw.tools.ide.tool.ToolCommandlet.install(ToolCommandlet.java:189)
        at com.devonfw.tools.ide.commandlet.InstallCommandlet.run(InstallCommandlet.java:49)
        at com.devonfw.tools.ide.context.AbstractIdeContext.applyAndRun(AbstractIdeContext.java:770)
        at com.devonfw.tools.ide.context.AbstractIdeContext.run(AbstractIdeContext.java:734)
        at com.devonfw.tools.ide.cli.Ideasy.runOrThrow(Ideasy.java:89)
        at com.devonfw.tools.ide.cli.Ideasy.run(Ideasy.java:52)
        at com.devonfw.tools.ide.cli.Ideasy.main(Ideasy.java:28)
[...]

Issue #230, opened 2024-02-26
backup() method fails to move directories


mvomiero:
### Expected behavior

if I have already a version of a tool installed in `devonfw/software/` and I want to install another one, IDEasy should move the current tool installation directory to `devonfw/updates/backup/` trough the method `FileAccessImpl.backup()`

### Actual behavior

Works good for symlinks but not for directories like ` java/` or `mvn/`: an exception is thrown, `java.lang.IllegalStateException: Failed to move`

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

Example: java

From `cli/`:

1.  make sure you have java installed or install it: 
    `devon mvn exec:java -Dexec.mainClass="com.devonfw.tools.ide.cli.Ideasy" -Dexec.args="--debug install java"
`

2.  install now another version (e.g. 17.0.2_8):
     `devon mvn exec:java -Dexec.mainClass="com.devonfw.tools.ide.cli.Ideasy" -Dexec.args="--debug install java 17.0.2_8"
`

3. Output:

  
<img width="878" alt="Capture" src="https://github.com/devonfw/IDEasy/assets/102261609/ee434b1a-be69-424d-b9c7-af151d0e2714">

**_NOTE: you will have to delete JAVA_VERSION from `devonfw/settings/ide.properties` if you want to run java again, otherwise you will keep getting the Failed to move exception_**

    

## Related/Dependent Issues

#225 was caused by the same method but here is actually another situation, since it's a directory and not a symlink.

### Affected version:

* OS: Windows and Linux, still not tested on Mac

mvomiero:
Probably the problem is just because ideasy is using java itself so can't move the installation folder

hohwille:
What you are seeing here is the famous Windows File Locking "bug".
Windows has the habit to lock files if they are in used. So if you start java in the `software/java/bin/java.exe` (e.g. you launch your IntelliJ or Eclipse or have a long maven process running) that folder gets locked by Windows and cannot be deleted until the lock is returned what by default only happens when the process holding the lock ends.
https://github.com/devonfw/ide/blob/master/documentation/advanced-tooling-windows.asciidoc#cope-with-file-locks

As the new IDEasy will only use symlinks/junctions the problem you are reporting is a known bug that already exists in devonfw-ide. So I would consider to close this as wontfix. However, before doing so we should improve the error handling.
[...]

PR #288, opened 2024-04-15
#264: prevent Windows file lock

mvomiero:
Fixes: #264
Fixes: #230 

After extensive research and several attempts, it's evident that the reported issue isn't a bug but rather a result of **using devon instead of ideasy** while installing new tools. For instance:

The problem we were getting was that the moving of the installed software directory fails because of the windows file lock on the directory, e.g.` java.lang.IllegalStateException: Failed to move D:\projects\IDEasy\software\java to D:\projects\IDEasy\updates\backups\2024-03-22\java_17-13-29`.

This issue specifically arises when the tools are installed using 'devon' (**_as a directory_**) instead of 'ideasy' (**_as a junction_**).
When installed with 'ideasy', the file lock automatically applies to the target path of the junction `IDE_ROOT/_ide/software/default/...`, allowing the deletion of the junction itself in `IDE_HOME/software/`. Consequently, updating a tool version becomes seamless.

The only fix it had to be implemented is for a another exception that was arising in the `delete `method in `FileaccessImpl`, that wasn't properly recognizing and deleting junctions.

```diff
-      if (Files.isSymbolicLink(path)) {
+      if (Files.isSymbolicLink(path) || isJunction(path)) {
        Files.delete(path);
```

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
@@ -1,5 +1,19 @@
 package com.devonfw.tools.ide.io;
 
+import com.devonfw.tools.ide.cli.CliException;
+import com.devonfw.tools.ide.context.IdeContext;
+import com.devonfw.tools.ide.os.SystemInfoImpl;
+import com.devonfw.tools.ide.process.ProcessContext;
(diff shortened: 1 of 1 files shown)


hohwille (approved):
@mvomiero thanks for this PR and the nice and simple fix to the problem. 👍 
We need to test that again in the native image but I will merge so we can test it after next nightly SNAPSHOT deploy.

Issue #154, opened 2023-12-11
Show project name in header of intellij


jan-vcapgemini:
As an IDEasy user, I want intellij to display my project name in the header like we currently do for eclipse.
Currently intellij shows the workspace name only.

hohwille:
currently unclear if possible at all. IF we can pass a title via CLI this will be easy to implement.

PR #290, opened 2024-04-15
#154 Show project name in header of intellij

jan-vcapgemini:
Fixes: #154 

Implements:
* added TextMerger with `.name` extension to DirectoryMerger
* moved copy method from FallbackMerger to FileMerger
* added test to DirectoryMergerTest

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/merge/TextMerger.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/merge/TextMerger.java
@@ -26,14 +26,16 @@ public class TextMerger extends FileMerger {
   @Override
   public void merge(Path setup, Path update, EnvironmentVariables variables, Path workspace) {
 
-    if (Files.exists(update)) {
-      copy(update, workspace);
-    } else if (Files.exists(setup) && !Files.exists(workspace)) {
-      copy(setup, workspace);
+    Path template = update;
+    if (!Files.exists(template)) {
+      template = setup;
+      assert Files.exists(template);
+      if (Files.exists(workspace)) {
+        return; // setup is only applied for initial setup if workspace file does not yet exist
+      }
     }
-
     StringBuilder inputBuffer = new StringBuilder();
-    try (BufferedReader reader = Files.newBufferedReader(update)) {
+    try (BufferedReader reader = Files.newBufferedReader(template)) {
       String line;
       String resolvedValue;
       while ((line = reader.readLine()) != null) {
--- a/cli/src/main/java/com/devonfw/tools/ide/merge/TextMerger.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/merge/TextMerger.java
@@ -1,0 +1,59 @@
+package com.devonfw.tools.ide.merge;
+
+import com.devonfw.tools.ide.context.IdeContext;
+import com.devonfw.tools.ide.environment.EnvironmentVariables;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * Implementation of {@link FileMerger} for Text files.
+ */
+public class TextMerger extends FileMerger {
+
+  /**
+   * The constructor.
+   *
+   * @param context the {@link #context}.
+   */
+  public TextMerger(IdeContext context) {
+
+    super(context);
+  }
+
+  @Override
+  public void merge(Path setup, Path update, EnvironmentVariables variables, Path workspace) {
+
+    if (Files.exists(update)) {
+      copy(update, workspace);
+    } else if (Files.exists(setup) && !Files.exists(workspace)) {
+      copy(setup, workspace);
+    }
+
+    StringBuilder inputBuffer = new StringBuilder();
+    try (BufferedReader reader = Files.newBufferedReader(update)) {
+      String line;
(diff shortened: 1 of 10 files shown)


cli/src/main/java/com/devonfw/tools/ide/merge/TextMerger.java
```diff
@@ -0,0 +1,59 @@
...
+      while ((line = reader.readLine()) != null) {
+        resolvedValue = variables.resolve(line, workspace.getFileName());
+        inputBuffer.append(resolvedValue);
+        inputBuffer.append('\n');
+      }
+    } catch (IOException e) {
+      throw new IllegalStateException("Could not read text file: " + workspace, e);
+    }
+    try {
+      Files.write(workspace, inputBuffer.toString().getBytes());
```
hohwille, line 48: to me this implementation seems inconsistent with the `copy` methods first and then overriding the copied file anyways with this code - so why copying the file first anyway if it is overwritten?

I suggested an according change so if you accept, we can resolve this one.

Side-note: In general it is best-practice to always use streaming where possible. That is not reading entire files into the heap (memory) and then writing them but read and write in chunks with both an open reader and writer in the same loop so you do not need to allocate too much memory and keep the garbage collector busy. However, here your approach is perfectly fine:
* We know that the template files are small so it does not matter (different story if you implement a server application that may process arbitrary files that can be gigabytes large, also possible allowing an attacker to force a server crash via OutOfMemoryError)
* Here this approach is even better assuming that during resolving of variables at the end of the template an exception would be raised, your approach will not write anything into the workspace where with streaming you might have written some chunks into a partial file until the exception occurred.
I am just writing this note, as I also see this a lot in other projects where in my reviews I have to educate devs on this topic.

cli/src/main/java/com/devonfw/tools/ide/merge/TextMerger.java
```diff
@@ -0,0 +1,59 @@
...
+  public void merge(Path setup, Path update, EnvironmentVariables variables, Path workspace) {
+
+    if (Files.exists(update)) {
+      copy(update, workspace);
+    } else if (Files.exists(setup) && !Files.exists(workspace)) {
+      copy(setup, workspace);
+    }
+
+    StringBuilder inputBuffer = new StringBuilder();
+    try (BufferedReader reader = Files.newBufferedReader(update)) {
```
hohwille, line 36:
```suggestion
    Path template = update;
    if (!Files.exists(template)) {
      template = setup;
      assert Files.exists(template);
      if (Files.exists(workspace)) {
        return; // setup is only applied for initial setup if workspace file does not yet exist
      }
    }

    StringBuilder inputBuffer = new StringBuilder();
    try (BufferedReader reader = Files.newBufferedReader(template)) {
```

cli/src/main/java/com/devonfw/tools/ide/variable/IdeVariables.java
```diff
@@ -55,9 +54,12 @@ public interface IdeVariables {
   VariableDefinitionString JASYPT_OPTS = new VariableDefinitionString("JASYPT_OPTS", null,
       c -> "algorithm=PBEWITHHMACSHA512ANDAES_256 ivGeneratorClassName=org.jasypt.iv.RandomIvGenerator");
 
+  /** {@link VariableDefinition} for {@link com.devonfw.tools.ide.context.IdeContext#getIdeHome() PROJECT_NAME}. */
+  VariableDefinitionPath PROJECT_NAME = new VariableDefinitionPath("PROJECT_NAME", null, c -> c.getIdeHome().getFileName());
```
hohwille, line 58: 1. IdeHome may be null causing an NPE here.
2. If the variable is already called `PROJECT_NAME`, I would rather suggest to use `VariableDefinitionString` instead.

cli/src/test/java/com/devonfw/tools/ide/merge/DirectoryMergerTest.java
```diff
@@ -114,6 +116,8 @@ public void testConfigurator(@TempDir Path workspaceDir) throws Exception {
     // assert
     mainPrefs = PropertiesMerger.load(mainPrefsFile);
     assertThat(mainPrefs).containsOnly(JAVA_VERSION, JAVA_HOME, THEME_HACKED, UI_HACKED, EDITOR, INDENTATION_HACKED);
+
+    assertThat(namePath).hasContent("project" + " - main\ntest");
```
hohwille, line 120:
```suggestion
    assertThat(namePath).hasContent("project - main\ntest");
```

hohwille (approved):
@jan-vcapgemini thanks for your PR. Nice work looking good. 👍 
I left a few suggestions for improvement that you can apply/commit.
Then we can merge.
p.s.: The test data does not seem to reflect the real use-case. 1. if we have a "merger" that cannot merge, it does not make sense to have both `setup` and `update` since the `update` template always wins. Further, in reality, I assume the template will only contain `${PROJECT_NAME}`. However, it is good that you are testing a "more complex" scenario here.

cli/src/main/java/com/devonfw/tools/ide/merge/TextMerger.java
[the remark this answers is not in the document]
hohwille: is now fixed as well and cannot happen anymore.

PR #297, opened 2024-04-17
#25: implement tool commandlet for intellij

### LINKED ISSUE #25: Implement ToolCommandlet for Intellij
Create `ToolCommandlet` for Intellij
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/intellij

ATTENTION: Blocked by issue #10 that needs to be implemented first.

ndemirca:
Closes #25


cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -33,8 +35,21 @@ public boolean install(boolean silent) {
   @Override
   public void installPlugin(PluginDescriptor plugin) {
 
-    // TODO Auto-generated method stub
-
+    // no override needed
   }
```
salimbouch, line 39: maybe you could just remove the method if it's not needed.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -33,8 +35,21 @@ public boolean install(boolean silent) {
...
 
+  @Override
+  public void postInstall() {
+
+    super.postInstall();
+
+    Path binPath = getToolPath().resolve("bin");
+    if (this.context.getSystemInfo().isLinux()) {
```
salimbouch, line 47: use if `(SystemInfoImpl.INSTANCE.isLinux())`. the other one is inconsistent I heard.

salimbouch (commented):
Looks great to me 👍

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
ndemirca: This method must be declared.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
ndemirca: the method is already deleted

cli/src/main/java/com/devonfw/tools/ide/log/IdeSubLogger.java
```diff
@@ -43,7 +43,7 @@ default String log(Throwable error, String message) {
 
   /**
    * @return {@code true} if this logger is enabled, {@code false} otherwise (this logger does nothing and all {@link #log(String) logged messages} with be
-   * ignored).
+   * ignored). >>>>>>> main
```
hohwille, line 46: Leftover from git merge conflict...
```suggestion
   * ignored).
```

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -23,6 +36,61 @@ public Intellij(IdeContext context) {
     super(context, "intellij", Set.of(Tag.INTELLIJ));
   }
 
+  @Override
+  protected void runIde(String... args) {
```
hohwille, line 40: dont you want to override `runTool` here if that method by default does not behave correct for Intellij?

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -23,6 +36,61 @@ public Intellij(IdeContext context) {
...
+   * @return the {@link ProcessResult}.
+   */
+  protected ProcessResult runIntelliJ(ProcessMode processMode, String... args) {
+
+    Path toolPath;
+    if (this.context.getSystemInfo().isWindows()) {
+      toolPath = getToolBinPath().resolve(IDEA64_EXE);
+    } else if (this.context.getSystemInfo().isLinux()) {
+      toolPath = getToolBinPath().resolve(IDEA_BASH_SCRIPT);
+    } else {
```
hohwille, line 80: If the binary is OS dependent, the correct way would be to override the `getToolBinary()` method.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -23,6 +36,61 @@ public Intellij(IdeContext context) {
...
+      toolPath = Path.of("open");
```
hohwille, line 81: this is more a MacOS workaround and should not be a result of `getToolBinary()`. However, the `open` command needs to get the application as argument what would be the tool binary. We could even implement a global macos workaround to use `open` automatically when the ProcessContext tries to execute a MacOS `*.app`.
The current approach is rather hard to understand from the code as some parts are in `runIde` method adding magic arguments in case of MacOS and here in this method we see that `open` is used in case of MacOS. So only after reading the entire code and puzzling things together I can see that on MacOS `open -na ...somepath... --args «workspacepath» «args»` is called.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -23,6 +36,61 @@ public Intellij(IdeContext context) {
...
+  @Override
+  protected void runIde(String... args) {
+
+    install(true);
+
+    Step stepRun = this.context.newStep("Running IntelliJ");
+    try {
+      ProcessResult result;
+      if (this.context.getSystemInfo().isMac()) {
+        Path ideaPath = getToolPath().resolve("Contents").resolve("MacOS").resolve(IDEA);
```
hohwille, line 48: This looks odd to me. IMHO tool path should point to `software/intellij` if that contains `Contents/MacOS/idea` then already something went wrong before because our MacOS workaround did not work out.
It is sometimes helpful in such cases to first use `devon «tool»` on the according OS to manually test what is happening and what structure is created and then try to understand and make IDEasy creating the analog result (assuming that devonfw-ide is behaving correct, what is not always the case, as I faced today in some tests).

cli/src/test/resources/ide-projects/intellij/repository/intellij/intellij/default/mac/bin/idea.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+echo "intellij mac $*"
```
hohwille, line 2: The structure of the MacOS IntelliJ IMHO is very different. Also your code that is trying to create a Path pointing to `Contents/MacOS/idea` but that does not exist at all.
Therefore the test is not realistic and could be improved.
I am very aware that we also need a manual test or automated e2e test to really verify that this code is able to install and run IntelliJ on MacOS.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
hohwille: So I just opened my MacBookPro and had a look for you:
```
➜  software ls -la intellij/
total 1872
drwxr-xr-x   4 hohwille  staff     128 Mar 19 09:15 .
drwxr-xr-x  14 hohwille  staff     448 Mar 12 06:03 ..
-rw-r--r--   1 hohwille  staff       9 Mar 19 09:15 .devon.software.version
-rwxr-xr-x   1 hohwille  staff  951888 Mar 12 06:03 idea
➜  software ls -la intellij
lrwxr-xr-x  1 hohwille  staff  50 Mar 19 09:15 intellij -> macos/intellij/IntelliJ IDEA CE.app/Contents/MacOS
```
So IMHO this is not yet correct and most probably will not work on MacOS unless I am missing something.

hohwille (changes requested):
@ndemirca thanks for your PR and adding support for IntelliJ what is a very crucial feature for IDEasy. 👍 
I left some review comments for rework. Please have a look.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -23,6 +38,27 @@ public Intellij(IdeContext context) {
...
+  @Override
+  protected String getBinaryName() {
+
+    Path toolBinPath = getToolBinPath();
+    if (this.context.getSystemInfo().isWindows()) {
+      return toolBinPath.resolve(IDEA64_EXE).toString();
+    } else if (this.context.getSystemInfo().isLinux()) {
+      return toolBinPath.resolve(IDEA_BASH_SCRIPT).toString();
+    } else {
+      return getToolPath().resolve("IntelliJ IDEA CE.app").resolve("Contents").resolve("MacOS").resolve(IDEA).toString();
```
hohwille, line 58: IMHO this is still wrong as we discussed in our meeting.
I already merged the PR #359 and resolved the conflicts into this PR.
If we did everything correct there is no such thing anymore than `IntelliJ IDEA CA.app` inside the `software` folder.
Also please be aware the different editions of intelliJ will have a different `*.app` name so your code would never work with the ultimate edition used by many projects.
I hope it was already clarified meanwhile why we need the generic MacOS workaround and avoid such ugly OS-switches here and handling of MacOS apps in individual commandlets.
Ideally like in devonfw-ide after installation we can just call `idea` on any plattform to launch IntelliJ.
This PR therefore still needs some reworking...

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -31,10 +67,31 @@ public boolean install(boolean silent) {
   }
 
   @Override
-  public void installPlugin(PluginDescriptor plugin) {
+  protected void postInstall() {
+
+    super.postInstall();
+    if (this.context.getSystemInfo().isMac()) {
+      setMacOsFilePermissions(getToolPath().resolve("IntelliJ IDEA CE.app").resolve("Contents").resolve("MacOS").resolve(IDEA));
+    }
```
hohwille, line 75: same here...

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -31,10 +67,31 @@ public boolean install(boolean silent) {
...
+  }
 
-    // TODO Auto-generated method stub
+  private static void setMacOsFilePermissions(Path binaryFile) {
 
+    if (Files.exists(binaryFile)) {
+      String permissionStr = FileAccessImpl.generatePermissionString(493);
```
hohwille, line 81: Seems I missed to discuss on this API.
IMHO this is very bad design:
* numeric file permissions must always be given in octal notation or nobody can understand what `493` actually is. If you write it octal as `0755` at least some bash/unix experienced user will understand what this.
* We should avoid direct usage of `FileAccessImpl` in commandlets. That is why we have `FileAccess` interface.
* Why do we provide a method in FileAccessImpl that converts a number to a String and then we again convert the String to a `Set` of `PosixFilePermissions` in order to then natively set this via `Files.setPosixFilePermissions`? The idea of `FileAccess` is to create a higher level API and abstraction to such lower-level Java NIO operations and make our live easier. So if we believe that using numeric (octal) notation is the most readable way why dont we create a method in `FileAccess` to set file permissions to a given file with a given numeric value? However, it may be even simpler and more readable to simply define constants for the combinations that we need like this:
```
/** {@link PosixFilePermission}s for "rwxr-xr-x" or 0755. */
public static Set<PosixFilePermission> RWX_RX_RX = Set.of(PosixFilePermission.OWNER_READ, PosixFilePermission.OWNER_WRITE, PosixFilePermission.OWNER_EXECUTE, PosixFilePermission.GROUP_READ, PosixFilePermission.GROUP_EXECUTE, PosixFilePermission.OTHERS_READ, PosixFilePermission.OTHERS_EXECUTE);
```
* My real suggestion would rather be to add a method `FileAccess.makeExecutable(Path)` that will simply implement what `chmod a+x «path»` will do and could also be implemented to work on Windows (see also [here](https://stackoverflow.com/questions/664432/how-do-i-programmatically-change-file-permissions)).

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -31,10 +67,31 @@ public boolean install(boolean silent) {
...
+      Set<PosixFilePermission> permissions = PosixFilePermissions.fromString(permissionStr);
+      try {
+        Files.setPosixFilePermissions(binaryFile, permissions);
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      } finally {
+        return;
```
hohwille, line 88: This does not seem to add any value and is IMHO only causing confusion.

```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -31,10 +67,31 @@ public boolean install(boolean silent) {
...
+      }
+    }
   }
 
-}
\ No newline at end of file
+  @Override
+  public void installPlugin(PluginDescriptor plugin) {
+
```
hohwille, line 95: This is not implemented so this PR does not entirely close #25. If you do something like this on purpose, then create a new issue for IntelliJ plugin support and leave a comment here like `TODO https://github.com/devonfw/IDEasy/issues/«id»`. Otherwise we still end up in the same situation that we have currently that some comandlets seem to be present but are incomplete and if we have no issue about the problem, we might forget about it and currently merging this PR would automatically close #25.

hohwille (changes requested):
@ndemirca Thanks for your improvements and all the great work in this PR. 👍 
However, we still have some last mile to go to complete this story.
See my review comments for details.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
jan-vcapgemini: I've added a very simple workaround for mac `ultimate` and `community/intellij` edition.
The dmg extractor still keeps the IntelliJ IDEA CE.app/Contents/... in the software folder, so this way should suffice for now. For another method to ideally call it with `idea` on each platform, we need to do some more research and discussions.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
jan-vcapgemini: As described above.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
jan-vcapgemini: I've implemented your suggestions now.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
jan-vcapgemini: Adjusted.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
jan-vcapgemini: Adjusted with new issue.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
jan-vcapgemini: For further improvements I've added an issue: https://github.com/devonfw/IDEasy/issues/440

hohwille:
Thanks @jan-vcapgemini for resolving the merge problems that broke the DMG extraction.
I replaced this PR by #447 so I will close this now.

PR #300, opened 2024-04-18
added .gitattributes

jan-vcapgemini:
added .gitattributes from ide


hohwille (approved):
@jan-vcapgemini thank you so much for adding this. 👍 
I will merge asap as it is great to have our config from devonfw-ide also in IDEasy.
Sorry, that I missed this out when populating the IDEasy repo.
p.s.: the +x flag issue is still missing here so we need to add that after the merge.

Issue #319, opened 2024-05-08
Omit "Already up to date" from git pull


jan-vcapgemini:
### Expected behavior

As an IDEasy user, I want to update my urls in the background automatically without being notified that everything is up to date.

### Actual behavior
Nearly every process which needs urls will report an "Already up to date" message.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. run `ide install jmc`

### Comments/Hints:
An easy fix could be to add the `--quiet` option to the git pull.
https://github.com/devonfw/IDEasy/blob/0e8f59f1a68cceb2fa9f4f5f3593d0d09bcc5ae8/cli/src/main/java/com/devonfw/tools/ide/context/GitContextImpl.java#L187

### Affected version:

* IDEasy 2024.05.001-alpha-SNAPSHOT

PR #322, opened 2024-05-08
#319: omit already up to date

jan-vcapgemini:
Fixes: #319 

Implements:
* added `--quiet` option to git pull


cli/src/main/java/com/devonfw/tools/ide/context/GitContextImpl.java
```diff
@@ -179,7 +179,7 @@ public void pull(Path targetRepository) {
     this.processContext.directory(targetRepository);
     ProcessResult result;
     // pull from remote
-    result = this.processContext.addArg("--no-pager").addArg("pull").run(PROCESS_MODE);
+    result = this.processContext.addArg("--no-pager").addArg("pull").addArg("--quiet").run(PROCESS_MODE);
```
hohwille, line 182: this will make every "pull" quiet.
However, a nice and pragmatic KISS solution.
If we later want to have situations where we do not want pull to be quiet, we can still add a boolean flag to this method. So I will already resolve this and merge the PR.

hohwille (approved):
@jan-vcapgemini thanks for your PR and the KISS style solution. 👍 
Ready for merge.

hohwille:
As this is still in status `in proress` I will wait and clarify on Tuesday if we should merge. Maybe you have already local addon changes that you want to push.
However, if you want we can merge as is...

jan-vcapgemini:
I've tested the -q option while being offline using `git pull -q`.
The result looks good:
`fatal: unable to access 'https://github.com/devonfw/ide-urls.git/': Could not resolve host: github.com`
This PR can merged unless we need a more complex solution e.g. `quiet` flag in GitContext.

Issue #331, opened 2024-05-10
CobiGen installation is not working properly


jan-vcapgemini:
### Expected behavior

As an IDEasy user I expect the installer to install CobiGen properly so that` ide cobigen` works.

### Actual behavior
CobiGen is not installed properly and `ide cobigen` command throws an exception about the cobigen file not being found.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. run: `ide cobigen`

### Comments/Hints:
It looks like the wrong file was added to ide-urls. We should add the tgz file instead of the jar.
See: https://repo.maven.apache.org/maven2/com/devonfw/cobigen/cli/2021.12.006/
If the tgz gets unpacked properly, the bin folder and cobigen script will be found.

### Affected version:

* IDEasy 2024.05.001-alpha-SNAPSHOT

PR #332, opened 2024-05-14
#331: fixed CobiGen installation

jan-vcapgemini:
Fixes: #331 

Implements:
* replaced jar with tar.gz file (taken from devonfw-ide implementation)
* added simple install and run tests


mvomiero (approved):
Thank you for your implementation, LGTM 🛩️

hohwille (approved):
@jan-vcapgemini thanks for fixing. Even though we consider dropping cobigen from IDEasy  (see #335) it is still supported at least in devonfw-ide so already fixing the updater fully makes sense.

Issue #236, opened 2024-02-28
Eclipse Plugin installation completely broken


hohwille:
When I install Eclipse during the plugin-installation very strange things are happening:
![image](https://github.com/devonfw/IDEasy/assets/1923119/cfbf7738-ddc6-4bbc-b694-2d0d12a30024)

Even when I select the workspace from my current IDE it opens the Eclipse GUI and then it closes again automatically.
On the console I get things like this:
```
$ ./ideasy.exe shell
ide> install eclipse 2023-09
ECLIPSE_VERSION=2023-09 has been set in SETTINGS@D:\projects\IDEasy\settings\ide.properties
Creating backup by moving D:\projects\IDEasy\software\eclipse to D:\projects\IDEasy\updates\backups\2024-02-28\eclipse_09-44-12
Successfully installed eclipse in version 2023-09 replacing previous version 2023-12
Failed to install plugin startexplorer (de.bastiankrol.startexplorer.feature.feature.group): exit code was 0
!SESSION 2024-02-28 09:44:13.075 -----------------------------------------------
eclipse.buildId=4.29.0.20230907-1200
java.version=17.0.10
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en_US
Framework arguments:  -product org.eclipse.epp.package.java.product org.eclipse.equinox.p2.director -repository https://basti1302.github.io/startexplorer/update/ -installIU de.bastiankrol.startexplorer.feature.feature.group
Command-line arguments:  -os win32 -ws win32 -arch x86_64 -product org.eclipse.epp.package.java.product -consoleLog org.eclipse.equinox.p2.director -repository https://basti1302.github.io/startexplorer/update/ -installIU de.bastiankrol.startexplorer.feature.feature.group

!ENTRY ch.qos.logback.classic 1 0 2024-02-28 09:44:15.639
Activated before the state location was initialized. Retry after the state location is initialized.

!ENTRY org.eclipse.core.resources 2 10035 2024-02-28 09:45:22.560
The workspace exited with unsaved changes in the previous session; refreshing workspace to recover changes.

!ENTRY ch.qos.logback.classic 1 0 2024-02-28 09:45:23.014
Logback config file: D:\projects\IDEasy\workspaces\main\.metadata\.plugins\org.eclipse.m2e.logback\logback.2.2.0.20230625-0847.xml

[...]

PR #333, opened 2024-05-14
#236: fixed Eclipse plugin installation

jan-vcapgemini:
Fixes: #236 

Implements
* added missing arguments
* applied code reformat


mvomiero (approved):
LGTM!

hohwille (approved):
@jan-vcapgemini thanks for your analysis of the problem and the fixing PR.
Looks good to me and ready for merge. Great work 👍 
Probably `runEclipse` will still be used for lower-level checks where the additional arguments may partially get in the way but let use wait and see. If that ever happens, we can still create a new issue...

cli/src/main/java/com/devonfw/tools/ide/tool/eclipse/Eclipse.java
```diff
@@ -68,14 +67,14 @@ protected ProcessResult runEclipse(ProcessMode processMode, String... args) {
     pc.executable(toolPath);
     Path configurationPath = getPluginsInstallationPath().resolve("configuration");
     this.context.getFileAccess().mkdirs(configurationPath);
+    pc.addArg("-application " + toolPath);
```
hohwille, line 70: mhm, did you ever test this (manually)?
The application argument takes a classname like `org.eclipse.equinox.p2.director` as parameter.
Passing `toolPath` here does not seem to make any sense to me and IMHO cannot work.
Also this argument must not be always present - if you launch the regular Eclipse GUI this should be omitted.
See the old  code to understand what needs to happen:
https://github.com/devonfw/ide/blob/cdf4af5de21489f3fdcbfdab41af858822e0d598/scripts/src/main/resources/scripts/command/eclipse#L187

jan-vcapgemini: I've tested only the start of eclipse and didn't realize that the application param had no effect because the plugins were already installed.
I've adjusted the installPlugin method now and tested the install process after deleting the whole plugins folder in my test project.

hohwille (approved):
@jan-vcapgemini thanks for rework. Ready for merge now 👍

PR #349, opened 2024-05-23
#80 GUI prototype

### LINKED ISSUE #80: Create a first prototype of a IDEasy GUI with JavaFx
As an IDEasy user, I want to have GUI in addition to the CLI so that I can do some things easier and also get interactive support.
For this story, we just want to build a first protoype of the GUI that has only a minimum functionality:
* create a new maven module `gui` next to `cli` and add it to the top-level `pom.xml` as module.
* add a dependency to the `cli` as well as to `openjfx`.
* Create a JavaFx application with a `main` method inside the `gui` project.
* Create a `Scene` with some root container like a `VBox` and attach a `Label` with the text "welcome to IDEasy" to it and attach the `scene` to the `primaryStage` that you then show.
* Test that when launching the main program a window appears with the "welchome to IDEasy" message.

This will be our first JavaFx "hello world" step. Until this is done, I will create an issue with the next step to improve the GUI.

slskiba:
Closes #80 
Shows a welcome message in the window. 
Display the cli project's version in the window title as poc of module dependency.


gui/src/main/java/com/devonfw/ide/gui/AppLauncher.java
```diff
@@ -0,0 +1,31 @@
+package com.devonfw.ide.gui;
+
+public class AppLauncher {
```
hohwille, line 3: What is the purpose of this class? IMHO you should just put a regular `main` method (without creating your own `Thread`) into `App` like suggested in every JavaFx tutorial.
Maybe the CobiGen example that Jan showed you was rather misleading than helpful...

gui/src/test/java/com/devonfw/ide/gui/AppBaseTest.java
```diff
@@ -0,0 +1,100 @@
...
+   */
+  @Override
+  public void start(Stage stage) throws Exception {
+
+    this.mainStage = stage;
+    FXMLLoader loader = new FXMLLoader(getClass().getResource("main-view.fxml"));
+    this.mainRoot = loader.load();
+    stage.setScene(new Scene(this.mainRoot));
+    stage.show();
+    stage.toFront();
```
hohwille, line 63: This looks really wired to me.
You are redundantly repeating code from `App` like `new FXMLLoader(getClass().getResource("main-view.fxml"))` in this test instead of testing `App` itself.
Maybe I have no experience with automated UI testing of OpenJFX apps but I cannot believe that this is the "solution"...

hohwille (commented):
@slskiba thanks for this PR. You really have quickly jump started into the project and already implemented a working solution for the story 👍 
As always I am picky and give you lots of review comments for improvement.
Also I am aware that you marked this PR as draft so it is WIP and maybe some of my points are already on your todo list...
I hope my comments help you to continue and improve.

gui/src/main/java/com/devonfw/ide/gui/AppLauncher.java
slskiba: Running the main within App does not work for me, it throws an Error saying JavaFX runtime components are missing. Calling the App from the Launcher however works. 

Seems to be a common solution for the problem: https://stackoverflow.com/questions/56894627/how-to-fix-error-javafx-runtime-components-are-missing-and-are-required-to-ru

gui/src/main/java/com/devonfw/ide/gui/App.java
```diff
@@ -0,0 +1,58 @@
...
+      e.printStackTrace();
+    }
+    return app;
+  }
+
+  @Override
+  public void start(Stage stage) throws IOException {
+
+    Parent root = FXMLLoader.load(App.class.getResource("main-view.fxml"));
+    App.scene = new Scene(root, 320, 240);
```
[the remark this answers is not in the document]
slskiba: Knowing reasonable width/height values at this point seems challenging. Setting the window to half the screen size for now.

gui/src/main/java/com/devonfw/ide/gui/AppLauncher.java
hohwille: I am totally puzzled that another `main` method delegating to the actual `App.main` method solves such error but I pulled your feature-branch and can confirm this strange effect.
Since I was using JPMS with my other project using JavaFx I never have seen this bug since the actual root of the bug is the java module system and using `add opens` directives the error can also be worked around.
Most probably with Eclipse and EclipseFx Plugin it would also work but since we are using IntelliJ let us keep this workaround unless we may find an even better solution.
Since you nicely reworked and simplified this the extra `AppLauncher` class does not hurt or seem like any maintenance overhead so thanks for your solution.
What you however should do is always adding JavaDoc and here from the JavaDoc you could link the stack-overflow discussion so also in the future when this PR is history and forgotten developers can still understand what this is all about.

gui/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.devonfw.tools.IDEasy.dev</groupId>
+    <artifactId>ide</artifactId>
+    <version>dev-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+  <groupId>com.devonfw.ide</groupId>
```
hohwille, line 12: wrong `groupId`. Please use this:
```suggestion
  <groupId>com.devonfw.tools.IDEasy</groupId>
```

hohwille (approved):
@slskiba thanks for your rework. Looks nice now.
I added the `java.version` to the top-level POM. Could you remove it from the `cli/pom.xml` since it is redundant otherwise (could not do that easily in the browser).

gui/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+  <properties>
+    <mainClass>com.devonfw.ide.gui.AppLauncher</mainClass>
+    <javafx.version>21</javafx.version>
+    <testfx.version>4.0.18</testfx.version>
+    <javafx.maven.plugin.version>0.0.8</javafx.maven.plugin.version>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.devonfw.tools.IDEasy</groupId>
```
hohwille, line 26: nicer this way:
```suggestion
      <groupId>${project.groupId}</groupId>
```

hohwille (approved):
@slskiba thanks for perfectly addressing all my review comments.
Now ready for merge.

Issue #316, opened 2024-05-07
Jasypt encryption and decryption fails with versions other than 1.9.3


mvomiero:
### Expected behavior

As a IDEasy user, I want to properly be able to use all jasypt versions for encryption and decryption

### Actual behavior

If the version of Jasypt is different then `1.9.3`, the command fails since the encryption class is not found.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. install jasypt in a different version then `1.9.3`: `ide install jasypt 1.0`
2. this is the output:
``` bash
$ ide jasypt encrypt master password
Already up to date.
Error: Could not find or load main class org.jasypt.intf.cli.JasyptPBEStringEncryptionCLI
Caused by: java.lang.ClassNotFoundException: org.jasypt.intf.cli.JasyptPBEStringEncryptionCLI
Running command 'C:\TRY\Projects\devonfw\software\java\bin\java.exe' with arguments '-cp' 'C:\TRY\Projects\devonfw\software\jasypt\jasypt-1.0.jar' 'org.jasypt.intf.cli.JasyptPBEStringEncryptionCLI' 'algorithm=PBEWITHHMACSHA512ANDAES_256' 'ivGeneratorClassName=org.jasypt.iv.RandomIvGenerator' 'password=master' 'input=password' failed with exit code 1!
```

### Comments/Hints:
* the problem is that the artifact classes used for encryption and decryption are implemented just for version `1.9.3`

PR #354, opened 2024-05-24
#316: fix Jasypt version

mvomiero:
Fixes: #316 by checking if the version is equals or greater then `1.9.3 `before adding it to the `urls `with `JasyptUrlUpdater`


cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/JasyptUrlUpdater.java
```diff
@@ -1,11 +1,19 @@
...
 
 /**
  * {@link MavenBasedUrlUpdater} for jasypt
  */
 public class JasyptUrlUpdater extends MavenBasedUrlUpdater {
+
+  public static final DefaultArtifactVersion MIN_VERSION = new DefaultArtifactVersion("1.9.3");
+
+  private static final Pattern VERSION_PATTERN = Pattern.compile("(\\d+)\\.(\\d+)\\.(\\d+)");
+
```
hohwille, line 16: Not used anywhere.
```suggestion
```

cli/pom.xml
```diff
@@ -90,6 +90,12 @@
       <artifactId>jansi</artifactId>
       <version>${jansi.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-artifact</artifactId>
+      <version>3.6.3</version>
+    </dependency>
+
```
hohwille, line 98: I do not see the need for this dependency. Can you simply remove it?

cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/JasyptUrlUpdater.java
```diff
@@ -24,4 +32,11 @@ protected String getMavenArtifcatId() {
     return "jasypt";
   }
 
+  @Override
+  public boolean isValidVersion(String version) {
+
+    DefaultArtifactVersion artifactVersion = new DefaultArtifactVersion(version);
+    return artifactVersion.compareTo(MIN_VERSION) >= 0;
```
hohwille, line 39: Could you just use `VersionIdentifier` instead of `DefaultArtifactVersion` to avoid the overhead of maven artifact dependency?
`VersionIdentifier` also has Methods like `isGreaterOrEqual` that will make the code more readable.

hohwille (changes requested):
@mvomiero Thanks for your PR. Functionally working fine but I would love to propose a better solution.
See my review comments...

cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/JasyptUrlUpdater.java
mvomiero: implemented!

hohwille (approved):
@mvomiero thanks for the improvements 👍 
Will merge now.

PR #355, opened 2024-05-24
#344: add generic list support to property

### LINKED ISSUE #326: Uninstall multiple tools at the same time
Enhance the UninstallCommandlet implementation to support uninstalling multiple tools at the same time.

### LINKED ISSUE #344: Add generic list support on Property
Currently every `Property` is designed to take exactly one value.
As a special feature, we have implemented `StringListProperty` that can take any number of `String` values.
With PR #258 we have now the need to also support list of other property types.
The idea is that instead of adding support for `ToolListProperty` and maybe many other `*ListProperty` classes on the long run causing quite some duplication and redundancy, it would be better to extend `Property` class itself with the ability to take multiple values.

The idea is that instead of 
https://github.com/devonfw/IDEasy/blob/b2754e933b89db43d9a661f451691b469c12672b/cli/src/main/java/com/devonfw/tools/ide/property/Property.java#L43
We will maintain a potential list of values:
```java
protected List<V> valueList;
```
The implementer may decide if it is easier to additionally have this `valueList` or to change `value` from `V` to `List<V>`.

To avoid breaking existing code the methods `getValue()` and `setValue(V)` should keep its API compatible.
So e.g. if we really have a `List` of values then `getValue()` returns the first value from the list if available or otherwise `null`.
Methods will be added like `getValue(int i)` and `getValueCount()` to access the values from within the `List`.

At construction time every property should have the capability to be configured to hold a single value and the existing constructors should stay compatible providing this behavior while optional additional arguments can set `multiValued` flag to `true` (and can be added later if ever needed potentially cardinalities like `minValueCount` and `maxValueCount` so the CLI could e.g. take only a number of values within this range).

The PR implementing this new feature should also remove `StringListProperty` and refactor code using it to use `StringProperty` with `multiValued` flag instead.
To keep CLI untouched the old behavior of calling `setValueAsString` multiple times actually adding values to `valueList` should be kept. The implementer may decide if `setValue` will also behave the same way. However, if `multiValued` is `false` the implementation must not add multiple values into the underlying list if `setValue[AsString]` is called multiple times.

ndemirca:
Closes #344 
Closes #326

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -24,7 +24,7 @@ jobs:
     if: ${{ needs.verify_commit.outputs.RUN_BUILD == 'true' }}
     strategy:
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
+        os: [windows-latest, ubuntu-latest, macos-latest, macos-13]
 
     steps:
       - uses: actions/checkout@v3
--- a/cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
@@ -78,8 +78,8 @@ public abstract class AbstractUpdateCommandlet extends Commandlet {
           this.context.debug("Configuration {} already exists - skipping to copy from {}", confPath, child);
         } else {
           if (!basename.equals("settings.xml")) {
-            this.context.info("Copying template {} to {}.", child, confPath);
-            this.context.getFileAccess().copy(child, confPath);
+            this.context.info("Copying template {} to {}.", child, conf);
+            this.context.getFileAccess().copy(child, conf);
           }
         }
       }
--- a/cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
@@ -282,18 +282,20 @@ public class FileAccessImpl implements FileAccess {
   @Override
   public void copy(Path source, Path target, FileCopyMode mode) {
 
+    if (mode != FileCopyMode.COPY_TREE_CONTENT) {
+      // if we want to copy the file or folder "source" to the existing folder "target" in a shell this will copy 
+      // source into that folder so that we as a result have a copy in "target/source".
+      // With Java NIO the raw copy method will fail as we cannot copy "source" to the path of the "target" folder.
+      // For folders we want the same behavior as the linux "cp -r" command so that the "source" folder is copied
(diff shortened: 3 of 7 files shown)


hohwille (approved):
@ndemirca thanks for your PR. You have solved all the complex changes. Well done 👍 
I left quite some review comments for improvement.
You could simplify and reduce changes in existing code following our [coding-convetions](https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc#refactorings).
I tried to make constructive changes and keep it still simple so even though it might look like a lot of comments, it should be possible to resolve quickly so we can merge.

cli/src/main/java/com/devonfw/tools/ide/property/StringProperty.java
```diff
@@ -18,7 +18,20 @@ public class StringProperty extends Property<String> {
...
+    this(name, required, alias, false, null);
+  }
+
+  /**
+   * The constructor.
+   *
+   * @param name the {@link #getName() property name}.
+   * @param required the {@link #isRequired() required flag}.
+   * @param alias the {@link #getAlias() property alias}.
+   * @param multivalued
```
jan-vcapgemini, line 30:
```suggestion
   * @param multivalued the boolean flag about multiple arguments.
```

cli/src/main/java/com/devonfw/tools/ide/property/StringProperty.java
```diff
@@ -28,10 +41,11 @@ public StringProperty(String name, boolean required, String alias) {
    * @param required the {@link #isRequired() required flag}.
    * @param alias the {@link #getAlias() property alias}.
    * @param validator the {@link Consumer} used to {@link #validate() validate} the {@link #getValue() value}.
+   * @param multivalued
```
jan-vcapgemini, line 44:
```suggestion
   * @param multivalued the boolean flag about multiple arguments
```

cli/src/main/java/com/devonfw/tools/ide/property/StringProperty.java
```diff
@@ -46,4 +60,9 @@ public String parse(String valueAsString, IdeContext context) {
     return valueAsString;
   }
 
+  public String[] asArray() {
```
jan-vcapgemini, line 63:
```suggestion
  @Override
  public String[] asArray() {
```

hohwille (approved):
@ndemirca Great rework 👍 Thanks!
Are you still working on the constructors without the `boolean multiValued` parameter to keep signature compatibility and reduce the changes in commandlets?
Then I can wait a moment. Apart from that ready for merge.

ndemirca:
@hohwille the changes about duplications in constructors is in the last commit. You can merge it, thanks!

jan-vcapgemini (approved):

Issue #345, opened 2024-05-22
Improve Jasypt ToolCommandlet


mvomiero:
* Add missing JavaDocs
* More detaild bundle commandlet description
* Consistency of the parsing of the variable `JASYPT_OPTS` if user-defined

PR #358, opened 2024-05-27
#345: Improve Jasypt ToolCommandlet

mvomiero:
Fixes: #345


cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/Jasypt.java
```diff
@@ -9,19 +9,22 @@
...
 /**
- * {@link ToolCommandlet} for <a href="http://www.jasypt.org/">Jasypt</a>, The java library which allows to add basic
- * encryption capabilities with minimum effort.
+ * {@link ToolCommandlet} for <a href="http://www.jasypt.org/">Jasypt</a>, The java library which allows to add basic encryption capabilities with minimum
+ * effort.
  */
 public class Jasypt extends LocalToolCommandlet {
 
-  public final EnumProperty<JasyptCommand> command;
+  final EnumProperty<JasyptCommand> command;
```
hohwille, line 23: We use `public` here since this is the API to the parameters/properties of commendlets.
If one commandlet want to use another commandlet from a different package, you cannot set properties unless they are public (or via `getProperty(String)` method what is not type-safe and would require casting).

cli/src/test/resources/ide-projects/jasypt/repository/java/java/default/bin/java
```diff
@@ -1,3 +1,7 @@
 #!/bin/bash
 echo "executing java:"
-cat $2 # .jar file
\ No newline at end of file
+cat $2 # .jar file
+
+if [ ! -z "$6" ]; then
+    echo "$6"
+fi
```
hohwille, line 7: I missed this on the review of the original PR creating Jasypt commandlet but why do we cat a JAR file?
Isn't that crazy causing pointless spam in our logs? Or am I missing something?
For tests it is often the simplest solution to just to `echo $*` that will print out all arguments and then write the tests accordingly. Then you do not have to change the script if a test wants to add an additional parameter.
I do not understand what `$6` should be reading this script without digging deep.

hohwille (approved):
@mvomiero thanks for your PR and for improving the code and help messages. Very nice 👍 
I left some comments with my findings for potential improvement.
Please have a look.
Otherwise ready for merge.

cli/src/test/resources/ide-projects/jasypt/repository/java/java/default/bin/java
mvomiero: I have modified it like that:
```bash
#!/bin/bash
echo "${@:6}"
```
so just the optional arguments stored in JASPT_OPTS are printed out, the assertion is quite simple then:
```java
assertLogMessage(context, IdeLogLevel.INFO, context.getVariables().get("JASYPT_OPTS"));
```
We just want to check that the java executable is properly executed with the right arguments!

The problem with printing out all the arguments is that they are a lot and computed dynamically so for example the path would be different for the different os. 
https://github.com/devonfw/IDEasy/blob/81bb8bbe4262645b83cec11c0490afb5991a34b1/cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/Jasypt.java#L93-L94
But if you think it is better @hohwille I can do the assertion checking all the arguments 😃

mvomiero: I have modfied the mocked java executable as follows:

```bash
#!/bin/bash
echo "${@:6}" # optional arguments
```

that allows to log just the optional arguments stored in the variable JASYPT_OPTS, and accordingly quite easily test it with 

```java
assertLogMessage(context, IdeLogLevel.INFO, context.getVariables().get("JASYPT_OPTS"));
```


I thought it is the best and cleaner approach since otherwise it would be quite complicated and long in the test to compute the different arguments that are passed to java to execute jasypt (some of them are computed at runtime too):
https://github.com/devonfw/IDEasy/blob/81bb8bbe4262645b83cec11c0490afb5991a34b1/cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/Jasypt.java#L93-L94

But @hohwille if you still think it is better for the java executable to print out all the arguments I can modify it and compute them accordingly in the assertion in the test.

hohwille: IMHO better but nice to have. Lets first merge. We can still create a new PR for further enhancement.

Issue #364, opened 2024-05-28
Add intel MacOS to nightly build


jan-vcapgemini:
We should add another OS to our nightly build pipeline.
As @tobka777 already found out, we can just add `macos-13`.

PR #365, opened 2024-05-28
#364: added intel support for MacOS

jan-vcapgemini:
Fixes: #364 

Implements:
* added intel support for MacOS (macos-13)

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -24,7 +24,7 @@ jobs:
     if: ${{ needs.verify_commit.outputs.RUN_BUILD == 'true' }}
     strategy:
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
+        os: [windows-latest, ubuntu-latest, macos-latest, macos-13]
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-build.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     if: ${{ needs.verify_commit.outputs.RUN_BUILD == 'true' }}
     strategy:
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
+        os: [windows-latest, ubuntu-latest, macos-latest, macos-13]
```
hohwille, line 27: Awesome, thanks. Is that really all we need to do?
At least you need to add the same here:
https://github.com/devonfw/IDEasy/blob/998ef92c86551576eaca1fc8e2f005a67fb8c18d/.github/workflows/release.yml#L13

hohwille (approved):
@jan-vcapgemini Thanks for the PR 👍 
Please have a look at my comment, but should then be ready for merge.

PR #367, opened 2024-05-28
#361: fixed nightly build pipeline

### LINKED ISSUE #361: The nightly build shows red even though the pipeline ran successfully
If there are no new code changes, the nightly build currently aborts, turning the pipeline red. 

To fix this, the termination process will be removed so the nightly pipeline always runs. As long as the build is successful, the pipeline will remain green.

ndemirca:
Closes #361


.github/workflows/nightly-build.yml
```diff
@@ -9,14 +9,34 @@ jobs:
     runs-on: ubuntu-latest
     name: Verify latest commit
     outputs:
-      RUN_BUILD: ${{ steps.verify_commit.outputs.RUN_BUILD }}
+      RUN_BUILD: ${{ steps.set_output.outputs.RUN_BUILD }}
```
hohwille, line 12: I am not getting this change. There is no such step `set_output` in this workflow.
Did you test the changed workflow on your fork?
There once was a KT session in our team about testing workflow changes on fork.
You can ask others, at least @jan-vcapgemini will know...

hohwille (approved):
@ndemirca Thanks for your PR. I now understand your idea and think this is a very smart solution that you found. Great 👍 
I left a comment to assure we do not break things by merging without testing, since we want to avoid that our actions break and we then cannot get any nightly releases for a while until fixed again.

.github/workflows/nightly-build.yml
hohwille: Maybe this helps: https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs

PR #372, opened 2024-05-29
#361: fixed nightly build

### LINKED ISSUE #361: The nightly build shows red even though the pipeline ran successfully

ndemirca:
Closes #361


hohwille:
I am always happy about simple solutions... However, wont the removing `cancel` let the build fail if nothing has changed and therefore the nightly build would again be red even though nothing is wrong? If you have tested this and can confirm this is not the case, I am happy to merge but I doubt this since I had all these discussions with @alfeilex and he came up with the current approach to cancel as a neutral result that is neither red nor green and for some time it was working such that the final badge was showing the last non-neutral result.

PR #374, opened 2024-06-03
#294: added conf folder if it does not already exist

### LINKED ISSUE #294: EnvironmentVariables.save() failes if the underlying properties file does not exist
### Expected behavior

As a developer, I want to save my changes to `EnvironmentVariables` with the `save()` method so that it works as expected.

### Actual behavior

If the underlying properties file does not exist (neither `ide.properties` nor `devon.properties`) then the `save()` method fails with an exception.

1. Create a test project with no `conf` folder
2. Create a JUnit test using `newContext` to create a test context based on the test project from 1.
3. ensure this JUnit invokes code like the following:
```
    EnvironmentVariables envVars = this.context.getVariables().getByType(EnvironmentVariablesType.CONF);
    envVars.set("VARIABLE_NAME", "VARIABLE_VALUE", true);
    envVars.save();
```

### Related/Dependent Issues

#273

ndemirca:
Closes #294


cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
```diff
@@ -105,7 +103,15 @@ public void save() {
...
-    try (BufferedReader reader = Files.newBufferedReader(this.propertiesFilePath)) {
+    
+    if (!Files.exists(newPropertiesFilePath)) {
+      try {
+        Files.createFile(newPropertiesFilePath);
+      } catch (IOException e) {
+        throw new IllegalStateException("Failed to create properties file with" + newPropertiesFilePath, e);
+      }
+    }
+    try (BufferedReader reader = Files.newBufferedReader(newPropertiesFilePath)) {
```
hohwille, line 114: IMHO it would be smarter to skip the reading of the file instead of creating an empty file and then reading the empty file.
Further, we should also consider what happens if the parent folder does not exist...
Unfortunately Java's file operations have never been designed robust and idempotent so we have to do all manually (neither `createFile` nor `newBufferedWriter` will work of the parent folder such as `conf` or `~/.ide` does not exist).
For robustness, I would suggest that we add an `mkdirs` on the parent folder.

hohwille (approved):
@ndemirca thanks for your PR. Will technically fix the reported bug 👍  
But since we are on this I have some suggestion to make it robust also for the case that the parent folder does not exist. See my comment.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
hohwille: added as TODO comment to the code so we do not forget it but can now merge.

PR #375, opened 2024-06-04
#296 Replace InetAddress.isReachable with alternative

### LINKED ISSUE #296: inOnline() Method of AbstractIdeContext does not work properly
inOnline() Method of AbstractIdeContext returns on Ubuntu in WSL2 false, even though you are online. Therefore the installation of a tool will be interrupted before it is complete.

Steps to reproduce: execute any tool on Ubuntu (e.g. jmc).
Happens on MacOS too.

Relevant code that is of question:
https://github.com/devonfw/IDEasy/blob/b2754e933b89db43d9a661f451691b469c12672b/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L600

slskiba:
Fixes #296. 

InetAddress.isReachable() seems to require root privileges to run properly on unix. 

Solution not tested on MacOS.

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
@@ -40,7 +40,8 @@ import com.devonfw.tools.ide.url.model.UrlMetadata;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
-import java.net.InetAddress;
+import java.net.URL;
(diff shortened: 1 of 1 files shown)


mvomiero (approved):
Thanks for your implementation 🚀 

Tested  on mac and works properly!

hohwille (approved):
@slskiba Thanks for your PR. Reading the HTML seems to be little bit more of waste but thanks to @mvomiero your approach is confirmed to work so we have a good solution. 👍 
Ready for merge.

Issue #337, opened 2024-05-16
IDEasy Mac intel architecture support for native image builds


jan-vcapgemini:
We need MacOs intel support for our build pipeline as we currently use mac-latest which appears to be the ARM architecture.
@tobka777 already found a fitting os tag: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners e.g. `macos-13`

hohwille:
As PR #365 did not fix this issue, I will reopen.
Things to consider:
* https://github.com/devonfw/IDEasy/blob/e2ab777405570d8968785571f06688a1336f7acc/cli/pom.xml#L155-L169
* https://github.com/devonfw/IDEasy/blob/e2ab777405570d8968785571f06688a1336f7acc/cli/pom.xml#L290
* https://github.com/devonfw/IDEasy/blob/998ef92c86551576eaca1fc8e2f005a67fb8c18d/.github/workflows/release.yml#L13

so we have to add the architecture also to the classifier and add a fourth release to our pom, etc.

PR #376, opened 2024-06-04
#337 add mac architectures

jan-vcapgemini:
Fixes: #337

Implements:
* os-maven-plugin
* automatic detection of OS and architectures


hohwille (changes requested):
@jan-vcapgemini thanks for this PR. If I understood this approach correctly, there are issues that Java/Maven reports the architecture not properly normalized (for system property `os.arch`) so the same thing can get different values depending on the platform.
Therefore insted of profiles triggered by OS activation setting things like this:
```
       <os.detected.classifier>windows</os.detected.classifier>
       <os.detected.arch>x64</os.detected.arch>
```
You now use [os-maven-plugin](https://github.com/trustin/os-maven-plugin/) to generate such properties in a normalized way. While this seems like a smart and valid approach, you need to be aware that the previous properties where somehow made up by some developer in our team. Even though os-maven-plugin uses the same property names, it does use them not with the same semantics.
IMHO you need to change at least
```
<releaseName>${project.artifactId}-${os.detected.classifier}-${os.detected.arch}</releaseName>
```
to 
```
<releaseName>${project.artifactId}-${os.detected.classifier}</releaseName>
```
or to
```
<releaseName>${project.artifactId}-${os.detected.name}-${os.detected.arch}</releaseName>
```
if I read the documentation of that plugin correctly.

Further, we have standardized our architectures differently and use `x64` instead of `x86_64` so we still might need to find a way how to remap these but this is some nice to have that can also be addressed after merging this PR.

jan-vcapgemini:
Can be closed because #471 was already merged.

PR #382, opened 2024-06-05
#321 Fix PATH formatting of 'env' command on Windows

### LINKED ISSUE #321: Paths not formatted correctly when running ide env --bash
### Expected behavior

As a user, I want the Paths to be in Linux syntax when I run "ide env --bash" in terminal

### Actual behavior
 Some Paths are correctly set to Linux syntax, but there are others who are still in Windows syntax, namely "export MAVEN_ARGS" and "WORKSPACE_PATH"

1.  Open git bash terminal
2.  Run ide env --bash

image

slskiba:
Fixes #321.

Converting back to forward slashes in VariableDefinition's toString method seems pointless, since conversion to unix-style paths happens in https://github.com/devonfw/IDEasy/blob/81bb8bbe4262645b83cec11c0490afb5991a34b1/cli/src/main/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandlet.java#L77

The removed line caused forward/backward slash mismatch in `env` output on windows, and caused paths to not be converted to unix-style when running `env --bash`


mvomiero (approved):
Great job 👍 
LGTM

hohwille (commented):
@slskiba thanks for your PR. This change does make sense. However, please be aware that this does not fix bug #321 as further changes need to be done to format path's in git-bash syntax on windows.
As maybe therefore we again need a special handling for `Path` here (e.g. after we added an (optional) `boolean bash` attribute to this `toString` method), I am unsure if it makes more sense to come up with the complete fix for #321 before merging...

mvomiero:
I have tested the solution of @slskiba and it seems to work and fix issue #321  pretty good to me. 🚀 
* This was before the bugfix, executing `ide env --bash`:
env before change  

* and this is after:
env after change

the only problems are:
* the `M2_REPO` variable, that is wrongly formatted, because it is defined as `VariableDefinitionString` instead of `VariableDefinitionPath`. Issue #390 is already created for that.
* `MAVEN_ARGS` default value is two arguments, the second one a path, but it isn't a path itself...so it has to be formatted like that to properly work in windows no?

hohwille:
@mvomiero thanks for your testing and transparent feedback.
What I wanted to say is that we should still add some special handling for the type `Path` when formatting variables to String (for bash).
As it seems we have this already working on String level somewhere else:
https://github.com/devonfw/IDEasy/blob/c5bdc9b43da13cb737cfbbafaa101fe7e99232aa/cli/src/main/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandlet.java#L77-L94

A smarter way would still be to consider the value type and to pass the `bash` flag here:
```
this.context.getVariables().collectVariables(bash);
```
and there handle `Path` specially.
However, I agree with the fix of the issue and maybe it is also a good feature that we can convert windows vs. MSYS/bash path on a String basis since end-users may put anything like this into their `ide.properties`:
```
MY_CUSTOM_PATH=C:\users\mylogin\foo\bar
```
So lets agree to merge this and also close the bug.
However, after debugging what is going on (also for `PATH`), I will later create another PR to do some further nice-to-have improvements (as suggested above) since the current behavior is a little strange as it partially converts/repairs `PATH` with an incorrect result and then throws this result away by recomputing it again properly.

hohwille (approved):
So with @mvomiero explanation I now understand that we later convert the Windows path syntax with backslashes on a string level and this does not work if we replace these backslashes with slashes.
So @slskiba thanks for this quickfix of #321 that resolves the issue since I already merged #391.

hohwille:
For the record: even though the build was green on github now on Windows `DirectoryMergerTest` fails because the backslashes are no longer normalized. I will fix this in my refactoring of all this path conversion stuff.

slskiba:
I had noticed this too, but wasn't fast enough to comment to prevent this merge.
A quick note on the failing test:
The `DirectoryMergerTest.testConfigurator` uses the test file resources/templates/update/main.prefs, which contains the testdata line of `java.home=${IDE_HOME}/software/java`. Previously the backslashes in `IDE_HOME` were replaced with forward slashes by the removed code. Now, on windows, `java.home` resolves in a path containing part backslash, part forward slashes. Replacing the forward slashes of `java.home` to backslashes, makes the test pass on windows; however creates the opposite problem on unix systems.

Issue #362, opened 2024-05-27
Improve logging for errors related to windows file lock of directories


mvomiero:
As already reported in the issues #264 and #230, we know that the `backup` method of `FileAccessImpl` doesen't work on directories that are locked with windows file lock. 
This problem was fixed by #288 and it's the result of using `devon` instead of `ideasy` while installing new tools _(installed as directories instead of symlinks or junctions)_. 

Working in a hybrid environment (devon + ideasy) causes still problems (see below) . We could therefore develop a better logging system (e.g. giving more information about if the given path is a symlink, a junction or a directory). 

```bash
$ ide install eclipse
Creating backup by moving D:\projects\IDEasy\software\java to D:\projects\IDEasy\updates\backups\2024-05-24\java_19-48-18
Step 'ide (install,eclipse)' failed: java.lang.IllegalStateException: Failed to move D:\projects\IDEasy\software\java to D:\projects\IDEasy\updates\backups\2024-05-24\java_19-48-18
Step 'Install java (21*)' failed: java.lang.IllegalStateException: Failed to move D:\projects\IDEasy\software\java to D:\projects\IDEasy\updates\backups\2024-05-24\java_19-48-18
2 step(s) failed out of 2 steps.
An unexpected error occurred!
We are sorry for the inconvenience.
Please check the error below, resolve it and try again.
If the error is not on your end (network connectivity, lack of permissions, etc.) please file a bug:
https://github.com/devonfw/ide/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=IllegalStateException%3A+Failed+to+move+D%3A%5Cprojects%5CIDEasy%5Csoftware%5Cjava+to+D%3A%5Cprojects%5CIDEasy%5Cupdates%5Cbackups%5C2024-05-24%5Cjava_19-48-18
java.lang.IllegalStateException: Failed to move D:\projects\IDEasy\software\java to D:\projects\IDEasy\updates\backups\2024-05-24\java_19-48-18
        at com.devonfw.tools.ide.io.FileAccessImpl.move(FileAccessImpl.java:278)
        at com.devonfw.tools.ide.io.FileAccessImpl.backup(FileAccessImpl.java:267)
        at com.devonfw.tools.ide.tool.LocalToolCommandlet.doInstall(LocalToolCommandlet.java:82)
        at com.devonfw.tools.ide.tool.ToolCommandlet.install(ToolCommandlet.java:180)
[...]

PR #385, opened 2024-06-06
#362: improve logging with file type and special log for windows file lock

mvomiero:
Fixes: #362


cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -275,7 +275,11 @@ public void move(Path source, Path targetDir) {
     try {
       Files.move(source, targetDir);
     } catch (IOException e) {
-      throw new IllegalStateException("Failed to move " + source + " to " + targetDir, e);
+      String fileType = Files.isSymbolicLink(source) ? "symlink" : isJunction(source) ? "junction" : Files.isDirectory(source) ? "directory" : "file";
+      if (this.context.getSystemInfo().isWindows() && fileType.equals("directory")) {
+        this.context.warning("On Windows, file operations could fail due to file locks. Please ensure the files in the moved directory are not in use.");
```
hohwille, line 280: I would rather include this to the exception message instead of an extra log.
Our plan is also to include some feature into IDEasy to support the user reporting an error as bug on github with the exception stacktrace included.
Further, it would be even more helpful to include this link as well:
https://github.com/devonfw/IDEasy/blob/main/documentation/how-to-unlock-files.adoc (should we rename the documentation page to `windows-file-lock.adoc`?)

hohwille (approved):
@mvomiero thanks for your PR. Looks good 👍 
I have left a comment for further improvement. Please have a look and see if you can also address this to make it perfect :)

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
mvomiero: Done, the rename as well 🛩️

PR #393, opened 2024-06-14
#321: refactoring to improve path conversion logic

### LINKED ISSUE #321: Paths not formatted correctly when running ide env --bash

hohwille:
Fixes #321

* remove path conversion logic from `EnvironmentCommandlet` and moved it to `WindowsPathSyntax`
* instead just determine proper `WindowsPathSyntax` according to OS and `--bash` flag
* pass this `WindowsPathSyntax` to `collectVariables` method of `EnvironmentVariables` that was extended with a variant taking this parameter.
* This passes it further to `getValue` method that has been extended in the same way.
* Again it is passed further to `getDefaultValueAsString` method of `VariableDefinition` that has been extended accordingly.
* Finally, `VariableDefinitionPath` and `VariableDefinitionSystemPath` now honor this new parameter to do the actual path conversion.


hohwille:
Seems I did something wrong. My code/test behaves differently on Linux and therefore fails:
```
[ERROR] Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.015 s <<< FAILURE! -- in com.devonfw.tools.ide.os.WindowsPathSyntaxTest
[ERROR] com.devonfw.tools.ide.os.WindowsPathSyntaxTest.testMsys -- Time elapsed: 0.007 s <<< FAILURE!
org.opentest4j.AssertionFailedError: 

expected: "/c/Windows/system32/drivers/etc/hosts"
 but was: "C:\Windows\system32\drivers\etc\hosts"
```
Seems that `Path` on Linux represents the exact path given as string differently than on Windows...

hohwille:
Ok the Linux/Mac error is a nice one:
If we do `Path.of("C:\Windows\system32\drivers\etc\hosts")` on Windows, we get the proper absolute path while on Linux or Mac, we get a relative `Path` with just one segment (file or folder) named `C:\Windows\system32\drivers\etc\hosts`.
OS abstraction is usually great in Java and you get reproducible behavior on any OS but not work here since the FileSystem implementation is intentionally OS specific. So either we have to ignore such tests if we are not on Windows or we have to build in ugly workarounds/hacks for it...

slskiba:
With the missing change you pushed, output looks good now for `env` on Windows. However for `env --bash`, some variables are still not converted correctly: 
```
ide> env --bash
export PATH="/c/projects/IDEasy/software/node:/ [...]
[...]
export AWS_SHARED_CREDENTIALS_FILE="C:\projects\IDEasy\conf\aws\credentials"
IDE_HOME="/c/projects/IDEasy"
export MAVEN_ARGS="-s C:\projects\IDEasy\conf\mvn\settings.xml"
[...]
```

hohwille:
I have updated, tested, and debugged my code with more care and detail.
```
export AWS_SHARED_CREDENTIALS_FILE="C:\projects\IDEasy\conf\aws\credentials"
```
This is fixed on my end now.
```
export MAVEN_ARGS="-s C:\projects\IDEasy\conf\mvn\settings.xml"
```
I added a change to "fix" this but commented it out since it is most probably correct as is and "fixing" it to git-bash syntax might even break maven.
We can test this and if it works with maven (even without IDEasy involved) properly with git-bash syntax, we can comment in the out-commented code to avoid potential confusion.
However, in general on Windows the Windows Paths are typically correct and even required for native processes. Only to make things working in git-bash itself, we need all this messy path conversion stuff.
I also have some doubts that converting `M2_REPO` to git-bash syntax will be correct and not cause any maven bug.

cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
```diff
@@ -2,17 +2,14 @@
...
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
slskiba, line 12: Avoid star import

slskiba (commented):
Now works as I would expect it to 👍

cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
hohwille: Thanks for the notice. I manually disabled this already multiple times. As I wrote in #52 we have a serious problem with these settings.
I will manually fix this somehow to complete the PR.

PR #394, opened 2024-06-17
#360 validation of multivalued properties

### LINKED ISSUE #360: Add validation of multi-valued properties into Property class
Currently, the validation of a single value is supported. Extend the code to also support multivalues.

slskiba:
Fixes #360.
Calling validate on all values, not just the first, of the property. As discussed, leaving the Consumer and its validation handling untouched for now, sending individual values instead of validating the list as a whole.


hohwille (approved):
@slskiba thanks for your PR. Nice fix 👍 
Ready for merge.

PR #397, opened 2024-06-18
#395 Uninstall graalvm successfully 

### LINKED ISSUE #395: Uninstall graalvm is not successful, fails to find installed version
### Expected behavior

As a end user, I want to uninstall graalvm using the `uninstall graalvm` command.

### Actual behavior

graalvm is still present and fully functional after running `uninstall graalvm`. 
image

### Steps to reproduce (windows)

1. `install graalvm`
2. `uninstall graalvm`

slskiba:
Fixes #395. 

- Moved uninstall method into ToolCommandlet, to allow for tool-specific uninstallation processes
- Moved current uninstall logic to LocalToolCommandlet, as it only is viable for local tools
- Implemented placeholder uninstall method for GlobalToolCommandlet
- use getToolPath to locate the installed tool, thus fixing the graalvm uninstall error


cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -143,4 +143,10 @@ protected boolean doInstall(boolean silent) {
     postInstall();
     return true;
   }
+
+  @Override
+  public void uninstall() {
+    //TODO: handle "uninstall <globaltool>"
```
hohwille, line 149: For the moment I also do not see a better quickfix.
However, could you be so nice and create a new issue about implementing uninstall for global tools and link it in this comment?

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -176,7 +176,28 @@ protected void postExtract(Path extractedDir) {
 
   }
 
-  private ToolInstallation createToolInstallation(Path rootDir, VersionIdentifier resolvedVersion, Path toolVersionFile, boolean newInstallation) {
+  public void uninstall() {
+
+    try {
+      String commandletName = this.getName();
```
hohwille, line 182: you could even just use `this.tool` instead of `commandletName` :)

And for the record: qualifying variables with `this.` makes sense so you directly see if the source-code (even without IDE so e.g. in the browser on github) that this is a member variable and not a local variable.
For methods the `this.` prefix does not make so much sense since there are no "local methods".
BTW: Instead if you want to make the call of a static method explicit you can use the syntax `«ClassName».«methodName»(«args»)`.

hohwille (approved):
@slskiba thanks for your PR. You are rocket fast madam :) Excellent solution.
Nice since with using `getToolPath()` it will now automatically work even for `GraalVm` without overriding uninstall logic. Nice job 👍 
I left some minimal comments but can be merged.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
hohwille: And easy alternative would be that we never support uninstalling global tools automatically and change the `UninstallCommandlet` to only accept `LocalToolCommandlet`s by crafting an according property for that.

slskiba: I think it would be best to keep UninstallCommandlet to accept GlobalToolCommandlets for the following two reasons:
For cases where uninstalling global tools by IDEasy is not be supported, this method can be utilized to provide reasonable feedback to the user (i.e. instructions on how to manually uninstall), before stating the error "Couldn't uninstall [..]".  
For global tools installed by package managers (see e.g. docker for Linux), uninstalling global tools can be supported by IDEasy.

slskiba: > For the moment I also do not see a better quickfix. However, could you be so nice and create a new issue about implementing uninstall for global tools and link it in this comment?

Created #399

hohwille: > I think it would be best to keep UninstallCommandlet to accept GlobalToolCommandlets for the following two reasons:

OK

> For cases where uninstalling global tools by IDEasy is not be supported, this method can be utilized to provide reasonable feedback to the user (i.e. instructions on how to manually uninstall), before stating the error "Couldn't uninstall [..]".

The user could then not assign "docker" or "pgadmin" as argument to the `uninstall` commandlet and tab-completion would not suggest global tools. If the user manually types these as arguments the CLI would raise something like "invalid argument 'docker' for local tool".
Still I agree that an average user may not even know the difference between local and global tools.

> For global tools installed by package managers (see e.g. docker for Linux), uninstalling global tools can be supported by IDEasy.

On windows this may be quite difficult to automate and MacOS is missing the concept to uninstall an application. Apple just forgot about this idea and still argue that you can simply go to your `Application` folder and manually delete the app. What they forget to see is that such Application can do several tweaks to your computer (e.g. writing changes to `.bashrc`, registering workflows, etc.) that are not undone when you simply delete the app folder.

However, let us just agree that we also want to support uninstall for global tools in #399 as good as we can do.
I therefore resolve this review comment and merge this PR. Thanks @slskiba for your great work. 👍

Issue #323, opened 2024-05-08
PATHS showing more than once when running ide env


aBega2000:
### Expected behavior

As a user, I want the PATHS to be shown only once when I run ide env

### Actual behavior

Some Paths are shown more than once

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. open git bash terminal
2. run "ide env"

![image](https://github.com/devonfw/IDEasy/assets/145266540/8b0c1b3a-df1c-4a11-b61f-ca790e28810b)
This Image shows some of the repeated PATHs

hohwille:
I had already analyzed this bug and found the root problem and the solution that I posted in our chat but so it wont get lost, I will add it here:
https://github.com/devonfw/IDEasy/blob/0e8f59f1a68cceb2fa9f4f5f3593d0d09bcc5ae8/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L264
Method `setCwd` is called from here:
https://github.com/devonfw/IDEasy/blob/0e8f59f1a68cceb2fa9f4f5f3593d0d09bcc5ae8/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L176
But `ideRoot` is set only after that:
https://github.com/devonfw/IDEasy/blob/0e8f59f1a68cceb2fa9f4f5f3593d0d09bcc5ae8/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L200
Hence `ideRoot` is `null` and therefore the `getTool` method cannot detect that something is a tool path:
https://github.com/devonfw/IDEasy/blob/0e8f59f1a68cceb2fa9f4f5f3593d0d09bcc5ae8/cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java#L99-L102

hohwille:
BTW: Another question would be what should happen now if I go outside of an IDEasy project and just type `ide`.
It currently gives me this
```
$ ide
IDE_ROOT is not set or not a valid directory.
```
Expected would be something like this:
```
$ ide
You are not inside an IDEasy installation: /d/projects
```
(please create a new bug issue for that)
But the actual question is if after calling `devon` the `PATH` should be "resetted" so that all IDEasy tools should now be removed from the PATH again or not?

With `devon` it does not do this... So to remove the tools from the `PATH` I would simply have to open a new shell?

```
hohwille@CE49454 MSYS /d/projects
$ echo $PATH
[...]

PR #401, opened 2024-06-19
#323: Remove paths duplicates

mvomiero:
Fixes: #323


cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -177,14 +177,14 @@ public AbstractIdeContext(IdeLogLevel minLogLevel, Function<IdeLogLevel, IdeSubL
...
+
     // detection completed, initializing variables
-    setCwd(userDir, workspace, currentDir);
     Path ideRootPath = null;
     if (currentDir == null) {
       info(getMessageIdeHomeNotFound());
     } else {
       debug(getMessageIdeHomeFound());
-      ideRootPath = this.ideHome.getParent();
+      ideRootPath = currentDir.getParent();
```
hohwille, line 187: can be `null` here
```suggestion
      if (currentDir != null) {
        ideRootPath = currentDir.getParent();
      }
```

hohwille (approved):
@mvomiero thanks for your PR. Nice that you have been able to understand the bug and create a solution 👍 
Please avoid NPE then we can merge.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
slskiba: The check for currentDir being null already happens in line 183 just above

mvomiero: https://github.com/devonfw/IDEasy/blob/88385bc05d214fe5c166238ae86eba48bb18be8f/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L183-L187

there is already a check for that a couple of lines before 😉

slskiba (approved):

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
hohwille: Oh, sorry for the spam. That review was directly before our daily and I completed it in a hurry.
My mistake...

Issue #400, opened 2024-06-19
ide output outside of an IDEasy installation


mvomiero:
### Expected behavior

As a IDEasy user, if I type `ide` outside of an IDEasy installation, I should get an informative output like:

```bash
$ ide
You are not inside an IDEasy installation: /d/projects
```

### Actual behavior

This is the current output:

```bash
$ ide
IDE_ROOT is not set or not a valid directory.
```

### Steps to reproduce the bug

run `ide` outside of an IDEasy installation

### Related bugs
#323

PR #405, opened 2024-06-21
#400: ide output outside IDEasy installation

mvomiero:
Fixes: #400 

* error log properly updated
* removed following block of code in order to print the same log just once:
https://github.com/mvomiero/IDEasy/blob/f5b0765bd3099ca647db9430f35a4b4454f69c3f/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L184

```diff
-    if (currentDir == null) {
-      info(getMessageIdeHomeNotFound());
-    } else {
+   if (currentDir != null) {
```


cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -200,7 +198,7 @@ public AbstractIdeContext(IdeLogLevel minLogLevel, Function<IdeLogLevel, IdeSubL
       }
     }
     if (ideRootPath == null || !Files.isDirectory(ideRootPath)) {
-      error("IDE_ROOT is not set or not a valid directory.");
+      error("You are not inside an IDEasy installation: " + System.getProperty("user.dir"));
```
hohwille, line 201: Actually this is a different error/warning. Even if IDE_HOME is not found because you are outside of an IDEasy project, the installer should set a environment variable IDE_ROOT on your OS (e.g. in Windows environment variables for user).
What you are actually looking for is this one:
https://github.com/devonfw/IDEasy/blob/ee596dadce0cf902cdce23abe346c1c2ac5d3015/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L283
It is already printed here:
https://github.com/devonfw/IDEasy/blob/ee596dadce0cf902cdce23abe346c1c2ac5d3015/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L183-L184

IMHO we need to investigate this further, because you removed the actual code that should print the expected message.

hohwille: I rather think that the only problem is that `info` needs to be changed to `error` since the first is writing to standard out while the latter is writing to standard error.
Standard out is written into the variable for the variables to set using `eval` command.
If the exit code is not 0 that output is discarded by wrapper script.
You can test that is works properly in IntelliJ but the output is hidden by the wrapper script.

mvomiero: changed info to error, the log was printed anyways two times. 

I went for this new solution that avoids redundant code as well (calling `getMessageIdeHomeNotFound()` two times) :
```diff
-    if (ideRootPath == null || !Files.isDirectory(ideRootPath)) {
-      error("IDE_ROOT is not set or not a valid directory.");
-    }
+    if (ideRootPath == null || !Files.isDirectory(ideRootPath) || currentDir != null) {
+      error(getMessageIdeHomeNotFound());
+    }
```

* Note: `getMessageIdeHomeNotFound()` hat to be changed because in some cases `this.cwd` was equal to null

hohwille:
Seems with my PR #424 I ran into so many small problems that I fixed, that we overlapped.
I did the same fix as this PR in #424.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
hohwille: I already fixed this in PR #424 so that would replace this PR.
However, there is a strange bug happening on github that cannot be reproduced locally on any OS including linux preventing the merge of PR #424. Since I do not have time currently to investigate this further, I wanted to merge this PR and later rebase PR #424.
However, this is still incorrect: There are two more or less independent warnings:
* IDE_ROOT is not defined
* IDE_HOME is required but not defined

I do not see a duplication of messages and still think you are removing an error log message that was still intended.
```
$ ideasy --env
You are not inside an IDE installation: null
IDE_ROOT is not set or not a valid directory.
```

The actual problem is the `null` and the fact that our wrapper script is kind of buggy or the log-level of the IdeHome message is info but should be error.
Maybe I can find the time to reduce one of my PRs and get the actual bash-completion stuff out until it gets green and I can reuse it to fix #400. That would most probably be the best way to go to get this all sorted out quickly...

hohwille:
@mvomiero thanks for creating this PR and to fix bug #400. 👍 
Sorry, that I meanwhile did the same fix. Due to the review discussions, it seemed easier to apply my fix with PR #476. Hence closing this PR.

PR #409, opened 2024-06-25
Implement uninstall for docker on linux

### LINKED ISSUE #399: Handle uninstall of global tool commandlets
Currently the uninstall commandlet only implements a solution for local tool commandlets. Global commandlets, such as docker, are not successfully uninstalled after running `ide uninstall <tool>`. 

As per the change implemented in #397, GlobalToolCommandlet implements its own uninstall method [(see here)](https://github.com/slskiba/IDEasy/blob/a6e3512ffdb91ac87d77ae7e1008af01eba57ea8/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java#L149), simply providing a generic error that the tool could not be uninstalled:

```
  @Override
  public void uninstall() {
    //TODO: handle "uninstall <globaltool>"
    this.context.error("Couldn't uninstall " + this.getName());
  }
```

Global tools should implement uninstall procedures if possible. However in cases where uninstallation by IDEasy is not possible, we should provide the user with productive output, for example manual uninstall instructions. Currently, only the generic error is being displayed. 

# Additional hint

We should try to implement this per OS in a rather generic way instead of making every global tool commandlet having the burden to write custom implementation of uninstall for all OS.
For Linux we should try to find a way to improve our package manager infrastructure so we might be able to derive the uninstall from the install configuration. E.g.
`sodo apt install package1` --> `sudo apt remove package1`
IMHO Apple on Mac OS has entirely forgotten to address an uninstall concept. It simply does not exist. There simply is an `/Application` folder and to install an app you copy it there and to uninstall you remove it from there.
In order to also cleanup files and settings created by some app on uninstall there is no support from Apple, there are 3rd party apps like AppCleaner but nothing IDEasy can rely on.

slskiba:
Implements #399 for docker on linux. Does **not** close #399! 

Generic handling of uninstall for global tools still needs to be implemented, however this PR enables uninstall for docker on linux.


hohwille (approved):
@slskiba thanks for your PR. Changes look good to me. 👍 
What I think we need to consider for `runWithPackageManager` is that we need to add logging of the executed command in `ProcessContext` before actually execution the process.
By default this should happen on `trace` level but it should be possible to configure this as parameter to `run` method or we can alternatively create a method for it.
Then `runWithPackageManager` should change that log-level to `info`.
A linux user cannot be blindly prompted for its root password from `sudo` without knowing what is actually happening and what command is to be executed.

PR #411, opened 2024-06-25
#410: Refactor get version and edition

### LINKED ISSUE #410: Inconsistent behavior of get-version and get-edition
### Actual behavior
Inconsistent behaviour:
`get-edition` of a local tool that is not installed provides the configured edition and instructions on how to install the tool, however
`get-version` of a local tool that is not installed results in a `CliException: Tool <tool> is not installed`
`get-version` of a global tool fails in  `CliException: Tool <tool> is not installed`: Tool not installed, whether it is installed or not. 

### Related Issues
#399

### Comments/Hints:
The current implementation for get-version and get-edition is only applicable to local tools.

slskiba:
Closes #410. 

The problem described in issue #399 is of similar nature to the commands `get-version` and `get-edition`. I suggest adjusting #399 to also find a general solution for global tools for these commands in addition to `uninstall`. 

Changes with this PR:
- moved logic to get installed version and edition of a tool to LocalToolCommandlet, as the current implementation is only applicable to those.
- added placeholder methods to get installed version and edition of a global tool (to be adressed by (adjusted) #399). User now receives an error message for tools that do not support these commands as of now, instead of an exception. 
- changed `get-version` of tool that is not installed to match the behavior of `get-edition`: display configured versions and instructions to install.


cli/src/main/java/com/devonfw/tools/ide/commandlet/VersionGetCommandlet.java
```diff
@@ -41,7 +39,11 @@ public void run() {
...
     VersionIdentifier installedVersion = commandlet.getInstalledVersion();
     if (installedVersion == null) {
-      throw new CliException("Tool " + commandlet.getName() + " is not installed!", ProcessResult.TOOL_NOT_INSTALLED);
+      this.context.info("The configured version for tool {} is {}", commandlet.getName(),
+          commandlet.getConfiguredVersion());
+      this.context.info("To install that version call the following command:");
+      this.context.info("ide install {}", commandlet.getName());
+      return;
     }
     this.context.info(installedVersion.toString());
```
hohwille, line 48: See my comment in the issue. I would rather do something like this:
```suggestion
      this.context.debug("The configured version for tool {} is:", commandlet.getName());
      this.context.info(commandlet.getConfiguredVersion());
      this.context.debug("To install that version call the following command:");
      this.context.debug("ide install {}", commandlet.getName());
    } else {
      this.context.info(installedVersion.toString());
    }
```

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -176,6 +176,73 @@ protected void postExtract(Path extractedDir) {
...
+   * @return the installed edition of this tool or {@code null} if not installed.
+   */
+  public String getInstalledEdition(Path toolPath) {
+
+    if (!Files.isDirectory(toolPath)) {
+      this.context.debug("Tool {} not installed in {}", getName(), toolPath);
+      return null;
+    }
+    try {
+      String edition = toolPath.toRealPath().getParent().getFileName().toString();
```
hohwille, line 231: First of all if the software was installed by `devonfw-ide` it will not be a symlink or junction and therefore this will cause an odd result.
Second, I am not sure if `toRealPath()` will resolve the link both for symlinks and junctions. The JavaDoc is unclear regarding to OS specific things like junctions.
However, I get the expectation that `toRealPath()` will then resolve to something like `D:\projects\_ide\software\default\java\java\17.0.6_10` and the filename of the parent is `java` what is indeed the installed edition.
However if it was not installed via symlink/junction, the result would IMHO be `software`.
It is unlikely that a tool will ever have an edition named `software` but this looks a little fragile to me.
Can we also get the version and edition of a custom tool?
IMHO this is not yet implemented but there it may again be different...

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -176,6 +176,73 @@ protected void postExtract(Path extractedDir) {
...
+  /**
+   * @param toolPath the installation {@link Path} where to find currently installed tool. The name of the parent
+   *     directory of the real path corresponding to the passed {@link Path path} must be the name of the edition.
+   * @return the installed edition of this tool or {@code null} if not installed.
+   */
+  public String getInstalledEdition(Path toolPath) {
+
+    if (!Files.isDirectory(toolPath)) {
+      this.context.debug("Tool {} not installed in {}", getName(), toolPath);
+      return null;
```
hohwille, line 228: I would still expect to get the configured edition if not installed.
The behavior of this code snipped can be considered as OK but then the `EditionGetCommandlet` also needs to be changed to handle `null` value and then print the configured edition.

hohwille (commented):
@slskiba thanks for your PR. You properly analyzed and understood the problem. Your change also looks good to me. 👍 
However, it turns out this code was created rather by developers than from a story with an initial design and expectation of what should happen.
I now also understood properly that we have to distinguish between installed and configured version/edition. Since we seem to request both with the same commandlet (if the tool is installed, then print the installed value otherwise the configured value)...

Maybe I also jumped in to early on a final review. IMHO it will still be good to discuss the concept with the team before a final decision. However, most probably we will do the changes I proposed and can then merge...

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
slskiba: The EditionGetCommandlet already handles null values. When no installation is found, it prints the configured edition. https://github.com/devonfw/IDEasy/blob/79ff36f4f03126d465c164f72a52231e5f0c2931/cli/src/main/java/com/devonfw/tools/ide/commandlet/EditionGetCommandlet.java#L41

hohwille (approved):
@slskiba I created #428 for my concerns and therefore resolved my review comments and approve your PR. You addressed something that was broken and made it better so merging your PR makes sense and is an improvement to the current state.
For #428 we will create a separate PR later.

Issue #406, opened 2024-06-21
Nightly Build fails because unable to find ide-doc Snapshot Artifact


mvomiero:
### Expected behavior

Build should work.

### Actual behavior

Build fails: 

https://github.com/devonfw/IDEasy/actions/runs/9607252940/job/26498103697#step:4:90

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:2.8:copy (copy-documentation) on project ide-cli: Unable to find artifact. Could not find artifact com.devonfw.tools.IDEasy:ide-doc:pdf:2024.06.002-alpha-SNAPSHOT in ossrh (https://s01.oss.sonatype.org/content/repositories/snapshots/)
```

### Steps to reproduce bug

1. Run nightly build on ubuntu latest

### Hints
Related maven config is here:
https://github.com/devonfw/IDEasy/blob/ee596dadce0cf902cdce23abe346c1c2ac5d3015/cli/pom.xml#L227-L250

Problems:
* we now only build the cli for native-image without other modules (since gui failed there and nothing else is/should be required for native-image)
* therefore IDEasy documentation is not build before
* however, maven build for CLI seems to require documentation artifact to be build before.

### Suggestion
We may move the maven logic to include the documentation into the CLI package into a profile that only gets triggered by deployment or that is active by default but can be disabled in nightly build maven call for native image.

### Note
When working on the workflow we should also align it with the Release workflow:
Release workflow is simply doing
```
cd cli
mvn «args»
```
while currently NightlyBuild is doing
```
mvn «args» -pl cli
```
More or less both approaches work, but we should better align our workflows and IMHO the `cd cli` approach is simpler to understand.

hohwille:
@mvomiero thanks for creating this issue! 👍 
Sorry, I created a duplicate with #408.
Will copy some additional information over to this one and then close mine.

PR #416, opened 2024-06-25
Fix nightly build

mvomiero:
Fixes: #406


slskiba (approved):

hohwille (commented):
@mvomiero thanks for your PR. Your changes to the workflow seem perfectly correct.
However, I am questioning the changes to the POM.
As it seems this was already prepared via profiles so maybe nothing needs to be changed in POM.
You did move the `assembly` to the native profile. However, for the matrix OS native build we only want the binaries. If we there already create the result package, then we would also need to include the documentation since it is required in the final package.

However we also do this:
```
- uses: actions/upload-artifact@v3
        with:
          name: natives
          path: cli/target/*.tar.gz
```
So we transfer a *.tar.gz from the native build to the actual release build.
IMHO we should discuss this in a meeting and maybe need to redesign both nightly and release workflow. I still wonder why the release workflow is successful and does not run into the error the nightly build is causing...

hohwille (approved):
@mvomiero after rethinking this:
Currently we are blocked because Nightly Build is not working. So lets merge this and create a new story for future improvements and redesign. A SNAPSHOT release without PDF documentation is still better than none at all...

PR #431, opened 2024-07-01
#30 tool commandlet for lazydocker

### LINKED ISSUE #30: Implement ToolCommandlet for LazyDocker
Create `ToolCommandlet` for LazyDocker
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/lazydocker

slskiba:
Closes #30 

- Installs lazydocker only if a sufficient version of docker is preinstalled.
- Warns if the version of docker-compose is not sufficient for lazydocker.


cli/src/main/java/com/devonfw/tools/ide/tool/lazydocker/LazyDocker.java
```diff
@@ -0,0 +1,70 @@
...
+
+    if (!result.getOut().isEmpty()) {
+      VersionIdentifier dockercomposeversion = VersionIdentifier.of(result.getOut().get(0).toString());
+
+      if (dockercomposeversion.compareVersion(VersionIdentifier.of("1.23.2")).isLess()) {
+        this.context.info("Found docker-compose version:" + dockercomposeversion);
+        this.context.info(
+            "If you want to use docker-compose with lazydocker, then you'll need at least version 1.23.2 of docker-compose");
+      }
+    }
```
hohwille, line 66: should we invent a `preInstall` method like the `postInstall`?

hohwille: Another problem I see with the current approach is that `install` method is called to ensure that a tool is installed even if it is already installed. In case we would have a transitive dependency you would spam the logs with all these messages about lazydocker even if it is already installed for the installation of another tool what could cause confusion. Maybe not a very significant scenario but from that PoV we could even decide to remove this logic at all since docker `< 1.25` is not really relevant anymore.
Another side-node:
If docker is running, I get this:
```
$ docker version --format '{{.Client.APIVersion}}'
1.44
```
But if docker is not running, I get this:
```
$ docker version --format '{{.Client.APIVersion}}'
1.45
error during connect: this error may indicate that the docker daemon is not running: Get "http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.45/version": open //./pipe/docker_engine: The system cannot find the file specified.
```
I would rather consider all this a design flaw of docker, but I am wondering how this IDEasy could would behave here. At least to my tests it seems that the error message is written to standard error and looking at our code, it would be ignored just like the rest of the result like the return code (which is `1` if docker daemon is not running).

cli/src/main/java/com/devonfw/tools/ide/tool/lazydocker/LazyDocker.java
```diff
@@ -0,0 +1,70 @@
...
+
+    return super.doInstall(silent);
+  }
+}
```
hohwille, line 70: If I remember correctly, we had some fun to properly run lazydocker on Windows and had to invent some workarounds.
IMHO this is missing here:
https://github.com/devonfw/ide/blob/cdf4af5de21489f3fdcbfdab41af858822e0d598/scripts/src/main/resources/scripts/command/lazydocker#L22-L27

hohwille (approved):
@slskiba thanks for your PR. Nice job 👍 
I found some small gaps. Please have a look and address my review comments.
Then we should be able to merge this PR.

cli/src/main/java/com/devonfw/tools/ide/tool/lazydocker/LazyDocker.java
slskiba: I think there is no irrelevant log spam if `install lazydocker` would be called when it is already installed imho. It would only produce info's if docker had been removed after installing lazydocker, which I think is very relevant, since it renders lazydocker useless. The warning about docker-compose being out of date also seems relevant to me when checking the lazydocker installation.  

I intentionally ignored the return code and the error message of the version check, since if docker is running or not does not affect the installation of lazydocker. Once lazydocker is installed and started, it will itself notify the user about docker not running. 

But please let me know if you preferred the version checks to be removed. If not, I will be moving them to a `preInstall `method.

cli/src/main/java/com/devonfw/tools/ide/tool/lazydocker/LazyDocker.java
slskiba: Is the intention to start lazydocker in a seperate powershell window? I had tested running lazydocker on Windows, and it worked just fine for me.

hohwille: Then maybe something changed meanwhile. There was some problem on Windows that it locked or ended the bash terminal or so. I do not really remember. We can tomorrow see if someone in the team remembers. Otherwise, we can also merge as is if you have already tested this on Windows it should be fine and we do not need such powershell workaround...

hohwille: Since we could not really dig out something more here in our daily, we can merge as is.
If there is a bug, we can fix later...

PR #432, opened 2024-07-01
#418 Fix autocomplete of version

### LINKED ISSUE #418: auto-completion of version still buggy
### Expected behavior

As a IDEasy user, I want to use auto-completion so that I can type faster and avoid typos.
Therefore I expect to get all matching versions of a tool properly suggested e.g. when completing `ide install aws 2.1`.

### Actual behavior

```
$ ide complete install aws 2.
2.17.1
2.*
...
$ ide complete install aws 2.1
2.1.39
2.1*
2.1.38
2.1.37
2.1.36
2.1.35
2.1.34
2.1.33
2.1.32
2.1.31
2.1.30
2.1.29
2.1.28
2.1.27
2.1.26
2.1.25
2.1.24
2.1.23
2.1.22
2.1.21
2.1.20
2.1.19
2.1.18
2.1.17
2.1.16
2.1.15
2.1.14
2.1.13
2.1.12
2.1.11
2.1.10
2.1.9
2.1.8
2.1.7
2.1.6
2.1.5
2.1.4
2.1.3
2.1.2
2.1.1
2.1.0
2.1*!
```

As we can see completing `2.1` only suggests versions that match `2.1*` (starting with `2.1.`) but this is not what the user wants.
Instead it should also match versions such as `2.17.1` or `2.16.12` just to give a few missing examples.

1. see above, you can also run `ide shell` and use embedded auto-completion.

### Related/Dependent Issues

* PR #371
* PR #384

### Comments/Hints:

I have the feeling that this is related to the semantic sorting of versions (2.17.0, 2.16.0, ..., 2.1.0, 2.1.1) and the binary search in combination with the lexicographic matching (and `startsWith`) in completion (2.1 should match 2.17 and 2.16 but those are in the sorted list before 2.1 so only the versions from 2.1 in the sort order are considered)

### Affected version:

* current version (2024.06.002-alpha-SNAPSHOT)

slskiba:
Fixes #418

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/completion/CompletionCandidateCollector.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/completion/CompletionCandidateCollector.java
@@ -1,10 +1,11 @@
 package com.devonfw.tools.ide.completion;
 
-import java.util.Arrays;
-import java.util.List;
-
 import com.devonfw.tools.ide.commandlet.Commandlet;
 import com.devonfw.tools.ide.property.Property;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * Collects the {@link CompletionCandidate}s for auto-completion.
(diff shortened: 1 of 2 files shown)


hohwille (approved):
@slskiba thanks for your fix. Your solution is functionally correct. 👍 
After the merge, we can still check if we have lists of versions so large that we need some performance tuning (what was intended by the previous implementation). However, simplicity is key and Java should IMHO be fast enough to match hundreds of versions this way.

PR #424, opened 2024-06-25
#417: added completion script for bash

hohwille:
fixes #417

## Current problems

~~When your are outside of an IDEasy project (no IDE_HOME), you get the error message and that also breaks the completion:~~
```
$ ide complete in
You are not inside an IDE installation: D:\projects
IDE_ROOT is not set or not a valid directory.

Error: IDEasy failed with exit code 1
```

~~Next our completion needs to be changed to be IDE_HOME aware. That is commandlets shall not be suggested by completion if they require ide home but the user is not in IDE_HOME. Otherwise I get completion support for e.g. `ide install mvn` but then when I press return I will get the error above. However, `ide help` or `ide create` or `ide --version` can be called from anywhere and should always be offered by completion.~~

Issue #417, opened 2024-06-25
Provide bash auto-completion


hohwille:
As a IDEasy user, I want to have tab completion for `ide` command in bash so that I can type faster and avoid typos.

Old solution in devonfw-ide was:
* https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/autocomplete
* https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/devon-autocomplete
* https://github.com/devonfw/ide/blob/cdf4af5de21489f3fdcbfdab41af858822e0d598/scripts/src/main/resources/scripts/command/ide#L197-L203

New solution can be simplified significantly since we only need to delegate to `ideasy complete -- «completion arguments»`.

PR #444, opened 2024-07-04
#417: bash completion and processable output

hohwille:
implements #417 
retry / replace PR #424 that has strange build errors on github.


hohwille:
OK. At least it is now confirmed that the build error is reproducible and does not seem to relate to some git history bug.
However, when I run the tests locally on Windows or on Linux (Ubunutu) then all tests succeed.
So there seems to be some magic happening on github causing the difference.

hohwille:
closing as duplicate of PR #424

PR #445, opened 2024-07-05
#437 Refactor npmurlupdater

### LINKED ISSUE #437: Refactor NpmUrlUpdater
As discussed in #435 we want to refactor `NpmUrlUpdater`.
It is currently a `WebsiteUrlUpdater` what is designed for grepping versions out of HTML content and should only be used as last resort solution if there is no better way.
However, here the configured website is serving JSON and not HTML.
Therefore, it would make sense to migrate to `JsonUrlUpdater`.
Then we also do not need to "grep" the versions via regex pattern but find them from the JSON structure without the need for a regex.

Since the npmjs registry is a global concept just like the maven central repository, we should consider creating an abstract base class `NpmBasedUrlUpdater` that contains the logic to parse the JSON from the NPM registry so sub-classes of it like `NpmUrlUpdater` will only need minimal code/configuration.

slskiba:
Fixes #437.
- Implements `NpmBasedUrlUpdater` as an extension of `JsonUrlUpdater`
- `NpmUrlUpdater` is now a `NpmBasedUrlUpdater` instead of a `WebsiteUrlUpdater`
- Added Tests for `NpmUrlUpdater`

There is a rather unclean workaround to fix versions being filtered properly for npm, so that alpha/beta versions are not being made available. These versions were not available with the prior used `WebsiteURLUpdater`. This problematic behaviour however is not specific to npm: Currently, e.g. python too makes alpha, beta and rc versions available to the user. 
This problem arises due to these tools implementing their own `update` method, in which `doAddVersion` instead of `addVersion` of the `AbstractUrlUpdater` is being utilized, and by that, circumventing the `mapVersion` call in which these versions would have been filtered.

I am open to suggestions for a prettier solution for this workaround, however it seems to me it requires some refactoring on `AbstractUrlUpdater`, and might be best to do this in a seperate issue, and consider the refactoring of NpmUrlUpdater completed.

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/url/updater/NpmBasedUrlUpdater.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/url/updater/NpmBasedUrlUpdater.java
@@ -1,6 +1,7 @@
-package com.devonfw.tools.ide.tool.npm;
+package com.devonfw.tools.ide.url.updater;
 
-import com.devonfw.tools.ide.url.updater.JsonUrlUpdater;
+import com.devonfw.tools.ide.npm.NpmJsonObject;
+import com.devonfw.tools.ide.npm.NpmJsonVersion;
 
 import java.util.Collection;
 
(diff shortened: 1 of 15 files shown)


cli/src/main/java/com/devonfw/tools/ide/url/updater/NpmBasedUrlUpdater.java
```diff
@@ -0,0 +1,102 @@
...
+      String response = doGetResponseBodyAsString(doGetVersionUrl());
+      NpmJsonObject jsonObj = MAPPER.readValue(response, getJsonObjectType());
+
+      for (NpmJsonVersion item : jsonObj.getVersions().getVersionMap().values()) {
+        String version = item.getVersion();
+        //TODO: this is not the right place to filter versions
+        // Also missing the logging of AbstractUrlUpdater's addVersion on which versions were filtered
+        if (mapVersion(version) == null) {
+          continue;
+        }
```
slskiba, line 62: This is the workaround mentioned above.

PR #447, opened 2024-07-05
#25: implement tool commandlet for intellij

hohwille:
implements #25
replacement for PR #297


hohwille:
We still have tons of bugs on MacOS.
However, this way it is working (without security limitations of new MacOS) and was also tested successfully by @jan-vcapgemini 
We will merge this now and then create bug issues for future improvements.

hohwille:
We still have tons of bugs on MacOS.
However, this way it is working (without security limitations of new MacOS) and was also tested successfully by @jan-vcapgemini 
We will merge this now and then create bug issues for future improvements.

PR #454, opened 2024-07-08
#429 writing properties to conf

### LINKED ISSUE #429: Option to configure writing properties changes to conf instead of settings
Currently when I do things like `ide set-version mvn 3.9.0` or `ide set-edition docker docker`, I will actually do changes in my `$IDE_HOME/settings/ide.properties` (or `devon.properties` in case of legacy support).
I would propose to add an option `--cfg=«env»` (to `ContextCommandlet`) that can switch this to use `conf` directory instead of `settings` (`--cfg=conf`) or even to `~/.ide/ide.properties` (`--cgf=home`).
Then I could change my docker edition only for personal use instead of in settings what only makes sense when you want to share the change with your team.

slskiba:
Resolves #429 

- Adds the `--conf`, `--workspace`, and `--home` options to both `set-edition` and `set-version` to allow writing the version properties to these corresponding locations. 
- Settings location used previously for `set-edition` and `set-version` remains the default


hohwille (commented):
@slskiba thanks for your PR. You nicely added the ability to provide the `EnvironmentVariablesType` as optional argument 👍 

With the approach of the new flags we could run `ide set-edition --conf --home --workspace intellij ultimate` and would even get support for this via auto-completion. However this command is inconsistent and does not make sense. That is why I was suggesting to use `--cfg=conf` or `--cfg=home` or `--cfg=workspace` to avoid such inconsistency.
What do you think?
Otherwise we IMHO would need to add an ability to express dependencies between `Properties` and honor that in auto-completion. That seems very complicated and could be easily avoided with the single property that takes an enum value instead of just a flag.

cli/src/main/java/com/devonfw/tools/ide/commandlet/EditionSetCommandlet.java
```diff
@@ -35,9 +31,7 @@ public EditionSetCommandlet(IdeContext context) {
     addKeyword(getName());
     this.tool = add(new ToolProperty("", true, "tool"));
     this.edition = add(new EditionProperty("", true, "edition"));
-    this.conf = add(new FlagProperty("--conf", false, null));
-    this.home = add(new FlagProperty("--home", false, null));
-    this.workspace = add(new FlagProperty("--workspace", false, null));
+    this.cfg = add(new EnumProperty<>("cfg", false, null, EnvironmentVariablesType.class));
```
hohwille, line 34: IMHO you need to make this an option as otherwise it will later be tricky to maintain the CLI if other args may be added.
Now, I could provide `conf` or `user` as last argument to make it optional. But only if we go for `--cfg=conf` or `--cfg=user` we will have the ability also in the future to have multiple such optional CLI aspects since assuming we would later add an additional optional `StringProperty` how could our CLI parser know if `conf` should be the value for `cfg` or for the new `StringProperty` argument?

```suggestion
    this.cfg = add(new EnumProperty<>("--cfg", false, null, EnvironmentVariablesType.class));
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/EditionSetCommandlet.java
```diff
@@ -52,13 +46,14 @@ public void run() {
...
-    } else if (this.home.isTrue()) {
-      commandlet.setEdition(edition, true, EnvironmentVariablesType.USER);
-    } else if (this.workspace.isTrue()) {
-      commandlet.setEdition(edition, true, EnvironmentVariablesType.WORKSPACE);
+    EnvironmentVariablesType env = this.cfg.getValue();
+    if (env == EnvironmentVariablesType.SETTINGS || env == EnvironmentVariablesType.CONF || env == EnvironmentVariablesType.USER
+        || env == EnvironmentVariablesType.WORKSPACE) {
+      commandlet.setEdition(edition, true, env);
+    } else if (env == EnvironmentVariablesType.RESOLVED || env == EnvironmentVariablesType.SYSTEM) {
+      context.error("Invalid option for --cfg: " + env);
```
hohwille, line 54: ah, I did not think about this one. In general it was smart to recycle the existing enum. However, in this case, you should create a new enum with a `EnvironmentVariablesType toType()` method so that auto-completion will not suggest options that do not make any sense.

hohwille (approved):
@slskiba thanks also for the fast and excellent rework. Everything perfect now 👍 
Ready for merge.

PR #455, opened 2024-07-08
#428 Improve get version and edition

### LINKED ISSUE #428: Redesign get/set-version/edition and add get-installed-version/edition
Thoughts on the current state:

Shouldn't `get-version` be symmetric to `set-version`? Currently when I do `ide set-version mvn 3.9.6` I will configure `MVN_VERSION=3.9.6`. Now what would a new IDEasy expect to get if he after that runs `ide get-version mvn` what should get the `mvn` version? Without knowing details about IDEasy, I would expect to get `3.9.6` as output. However, if maven was already installed before, I then will get the installed version that might be different so maybe I will get `3.9.7`.

Therefore I am thinking about adding new commandlets `get-installed-version` and `get-installed-edition` that will give me the version/edition that is actually installed. The commandlets `get-version` and `get-edition` can then be simplified to only print the configured values and the UX would be IMHO more intuitive.

slskiba:
Resolves #428 

- Renamed `getEdition` of `ToolCommandlet` to `getConfiguredEdition`
- `get-version` and `get-edition` now behave identically w.r.t. edition/version specifics 
- Adds flags `--installed` and `--configured` to both `get-version` and `get-edition`, to get the corresponding output
- Using no flag (or both), will provide both the configured and installed version(/edition):
  - if an installed tool's version is equal to the configured version, that version will be provided
  - if a tool is not installed, the configured versions along with install instructions are provided
  - if a tool is installed, but the version does not match the configured version, both versions are provided along with install instructions

Here is an example of the implemented behaviour:

```
ide> get-version aws
2.2.0

ide> set-version aws 2.1.0
AWS_VERSION=2.1.0 has been set in SETTINGS@C:\projects\IDEasy\settings\ide.properties
To install that version call the following command:
ide install aws

ide> get-version aws
The installed version for tool aws is 2.2.0
The configured version for tool aws is 2.1.0
To install the configured version call the following command:
ide install aws

ide> get-version aws --configured
2.1.0

ide> get-version aws --installed
2.2.0

ide> install aws
Start: Install aws
[..]
Successfully installed aws in version 2.1.0 replacing previous version 2.2.0
Successfully completed ide (install,aws)

ide> get-version aws
2.1.0
```


hohwille (approved):
@slskiba thanks for your PR and nice that you brought in your own idea to come up with a better solution than what I suggested. 👍 
First I had the fear that `ide get-version --installed --configured` would then result in an inconsistent request but you solved this nicely.
What could be improved is that you could avoid some redundancy and simplify the if/else flow if you would simply cover the case if both flags are false. If I get it correct this default should behave identical to if both flags are true. Then you could simply have two proper booleans for both flags and then can just handle two if blocks sequentially.
However, this is nice-to-have and can also be done after merge.

PR #467, opened 2024-07-09
#161 Adds feature of notifying the user when ide-settings has updates

### LINKED ISSUE #161: Feature to add notification when ide-settings has updates
From our survey of experiences with devonfw-ide, there is a feature request to add notifications when the `ide-settings` are updated. This way the user can decide whether or not to update the ide. 

One idea would be to get the latest status/commits of the `ide-settings` repo after running `devon`. Then the user gets the information and can decide to update the ide with the latest `ide-settings`.

There is also the idea in the survey to have a health check or report of the actual installation. This report could also include information about updates of any kind, such as new releases or new settings.

The original feature ideas of the survey are: 
- Push notification that an updated ide-settings is available, or even required updates (or at least an active contradiction that one does not want to update yet)
- Some kind of "health check" or report on the actual installation (vs. ide-settings) for support/debugging purposes

diiinesh:
closes #161 

- Added functions `fetchIfNeeded` and `isRepositoryUpdateAvailable` to `GitContextImpl`.
- Added test cases for the new functions.

These changes improve usability by informing the user if there are updates available in the settings repository.


cli/src/main/java/com/devonfw/tools/ide/context/GitContextImpl.java
```diff
@@ -69,6 +71,84 @@ public void pullOrCloneIfNeeded(String repoUrl, String branch, Path targetReposi
...
+    Path gitDirectory = targetRepository.resolve(".git");
+    // Check if the .git directory exists
+    if (!this.context.isForceMode() && Files.isDirectory(gitDirectory)) {
+      Path fetchHeadPath = gitDirectory.resolve("FETCH_HEAD");
+      long currentTime = System.currentTimeMillis();
+      // Get the modification time of the FETCH_HEAD file
+      try {
+        long fileModifiedTime = Files.getLastModifiedTime(fetchHeadPath).toMillis();
+        // Check if the file modification time is older than the delta threshold
+        if ((currentTime - fileModifiedTime > GIT_FETCH_CACHE_DELAY_MILLIS.toMillis())) {
```
slskiba, line 88: ..MILLIS.toMillis() sounds rather unintuitive. I'd suggest removing the _MILLIS suffix from the constants
```suggestion
        if ((currentTime - fileModifiedTime > GIT_FETCH_CACHE_DELAY.toMillis())) {
```

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -894,6 +894,7 @@ private boolean applyAndRun(CliArguments arguments, Commandlet cmd) {
       if (cmd.isIdeHomeRequired() && (this.ideHome == null)) {
         throw new CliException(getMessageIdeHomeNotFound());
       }
+      getGitContext().fetchIfNeeded("origin","main", getSettingsPath());
       cmd.run();
```
slskiba, line 898: We should be aware that currently this adds somewhere around 0.5 seconds to every command run

cli/src/main/java/com/devonfw/tools/ide/context/GitContextImpl.java
```diff
@@ -69,6 +71,84 @@ public void pullOrCloneIfNeeded(String repoUrl, String branch, Path targetReposi
...
+          if (isRepositoryUpdateAvailable(targetRepository, remoteName, branch)) {
+            this.context.info("Updates are available for the settings repository. Please pull the latest changes.");
```
slskiba, line 90: Maybe we can provide more concrete infos on what actions the user can take in this case. I.e. refer to `ide update`

cli/src/test/java/com/devonfw/tools/ide/context/GitContextTest.java
```diff
@@ -161,4 +162,111 @@ public void testRunGitPullWithForceStartsCleanup(@TempDir Path tempDir) {
...
+    // Mocking repository update availability
+    outs.add("new updates available");
+
+    // act
+    gitContext.fetchIfNeeded(repoUrl, remoteName, tempDir);
+
+    // assert
+    // Since FETCH_HEAD is old and updates are available, fetch should occur
+    assertThat(outs.isEmpty()).isFalse();
+  }
```
slskiba, line 249: This test does not test anything but the add method of java's List

slskiba (changes requested):
Nice job 👍

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
hohwille: As commented in #161 you should follow the concept of #118 to only trigger expensive git operations after some delay (e.g. every 2h). Further as discussed on the chat do NOT fetch if you already have local changes.
Also we do not own the real settings repository but only our template repository.
Projects can change the default branch and the user can also be switched to a different branch than the default one.
Either determine which branch the current user is on and if there is a corresponding remote branch or fetch all branches. But hardcoding `main` is wrong.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -918,6 +918,10 @@ private boolean applyAndRun(CliArguments arguments, Commandlet cmd) {
           debug(getMessageIdeHomeFound());
         }
       }
+      if (getGitContext().isRepositoryUpdateAvailable(getSettingsPath()))
+        info("Updates are available for the settings repository. If you want to pull the latest changes, call ide update.");
+      else
```
hohwille, line 923: can you check your IDE setup?
You should call `ide update` to get your IntelliJ settings up-to-date so that code is nicely autoformatted, curly braces added, etc.

hohwille (commented):
@diiinesh Thanks for your updates. Code looks nice now and generally should work fine.
However, in the story #161 we discussed the workflow as following:

1. check if changes are available by comparing local and remote branch and if changes are available print according message to user ✔️ 
2. if no local changes are know, fetch from git remote ✔️ 
3. if changes had been fetched from the server, do the same step as 1. again ❌ 

So currently if changes have been pulled, then the user gets feedback about updates only on the next call of IDEasy.
For the last one, I would suggest that `fetchIfNeeded` would return a `boolean` giving the feedback if actual changes could be fetched from the server.
Then move your code that does the `isRepositoryUpdateAvailable` check and potentially prints the message to a private method returning `boolean`.
Then if `fetchIfNeeded` returned `true` call that private method again.

Last challenge seems to be that `git fetch` does not nicely give us feedback if we are already up-to-date or some real changes had been pushed. Seems we need to do a web-search to find the magic option to get this information.

```
$ git fetch
remote: Enumerating objects: 51, done.
remote: Counting objects: 100% (51/51), done.
remote: Compressing objects: 100% (27/27), done.
remote: Total 33 (delta 19), reused 0 (delta 0), pack-reused 0
Unpacking objects: 100% (33/33), 4.48 KiB | 97.00 KiB/s, done.
From git.host.com:project
 * [new branch]      feature/feature1 -> origin/feature/feature1
$ echo $?
0
```
So here we should get `true` as result.
```
$ git fetch
From git.host.com:project
 * [new branch]      feature/feature1 -> origin/feature/feature1
$ echo $?
0
```
While here nothing was fetched and we should get `false`.

UPDATE: Sorry, my observation is not exactly correct. The `fetchIfNeeded` method also does the check again but as discussed already, that check does not belong there since it has nothing to do with `git fetch` and is specific for the `settings` while `fetchIfNeeded` can take any git repo as argument.

hohwille (approved):
@diiinesh excellent now. Thanks for completing this PR to perfection. 👍 

Now ready for merge.

Issue #469, opened 2024-07-10
Intellij is not using custom idea.properties file


jan-vcapgemini:
### Expected behavior

As an IDEasy Intellij user I expect Intellij to detect my `idea.properties` file in my workspace.

### Actual behavior
Intellij does not detect the `idea.properties` file and won't apply our custom plugin etc. paths.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Run `ide intellij`
2. Change some settings f.e. code style
3. Check if `.intellij` or `.idea` folders in `workspace/main` were changed

### Comments/Hints:
It seems we are missing a crucial environment variable:
IDEA_PROPERTIES
See this example for our old implementation from devonfw-ide: https://github.com/devonfw/ide/blob/cdf4af5de21489f3fdcbfdab41af858822e0d598/scripts/src/main/resources/scripts/command/intellij#L148C3-L148C61

Also take a look at how the default idea.properties location can be changed: https://intellij-support.jetbrains.com/hc/en-us/articles/207240985-Changing-IDE-default-directories-used-for-config-plugins-and-caches-storage
### Affected version:

* IDEasy: 2024.07.003

Issue #339, opened 2024-05-17
Refactoring IntelliJ and android-studio


ndemirca:
Refactor the code by identifying the common implementations of the tool commandlets "IntelliJ" and "android-studio" and remove duplicates

hohwille:
PR #339 was merged but this is not really completed:
* [ ] Introduce an abstract class `IdeaBasedIdeToolCommandlet` between `IdeToolCommandlet` and its current implementations `Intellij` and `AndroidStudio`. There might come support for additional such IDEs to IDEasy in the future (PyCharm, WebStorm, etc.).
* [ ] Extract common functionalities to the new abstract base class. 
  * [ ] IMHO plugin installation is the same for all IDEA/JetBrains IDEs except for some parameterization of the download URL and the IDE folder name in the plugins folder.
  * [ ] `IDEA_PROPERTIES` mechanism is the same for all IDEA/JetBrains IDEs only the name of the environment variable differs. The Variable value differs per workspace. Therefore setting it once in postInstall with a fixed workspace name is buggy. We need to "export" this property via ProcessContext for each call of that IDE. Also it seems an existing flaw that we use the same-filename for different IDE products: See [android-studio](https://github.com/devonfw/ide-settings/blob/main/android-studio/workspace/update/idea.properties) and [IntelliJ](https://github.com/devonfw/ide-settings/blob/main/intellij/workspace/update/idea.properties). As both files have different content, they also must have different filenames as otherwise one overrides the other.
  * [ ] The `getBinaryName()` problem is the same for all IDEA/JetBrains IDEs. The current solution is still not nice. I would prefer a postInstall hook that creates a simple wrapper bash script if the default command is not already present. So create bash script `studio` that only calls `studio64.exe $*`, etc.

PR #470, opened 2024-07-10
#469: added IDEA_PROPERTIES to variables

jan-vcapgemini:
Fixes: #469 

Implements:
* new IDEA_PROPERTIES environment variable pointing to workspaces/main/idea.properties files


hohwille (approved):
@jan-vcapgemini thanks for this PR and especially tracing down the root of the problem 👍 

We could think if we want to only export this in `AbstractIdeaIdeCommandlet` (see #339) so the user will not have this in his shell.
However, all has pros and cons. I will merge this as is and we can still think if we want to address this with #339 or not. An advantage of your solution in this PR is also that if we would ensure there is always an `idea` command for IntelliJ on the path, then our config would still apply if some user starts intellij without `ide` script as wrapper.

PR #446, opened 2024-07-05
#337: added macos-13 to release

jan-vcapgemini:
Fixes: #337 

Implements:
* adjusted tgz file names to latest pom release name

PR #471, opened 2024-07-11
#337: mac architectures combined

jan-vcapgemini:
Combines: #446 and #376
Fixes: #337


hohwille (approved):
@jan-vcapgemini thanks for the quick solution combining your PRs 👍 
Now we can merge this and should be able to get a consistent result on `main` branch.

Issue #468, opened 2024-07-09
Intellij not started in background


hohwille:
### Expected behavior

As a IDEasy and IntelliJ user, I want to launch intellij in background so that I can continue working in my shell while using IntelliJ.

### Actual behavior

The `ide intellij` command blocks until the IntelliJ IDE is terminated.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. `ide intellij`

### Related/Dependent Issues

#25

### Comments/Hints:

Do we run process with `BACKGROUND` as mode?

### Affected version:

* OS: Windows
* 2024.07.003-alpha-SNAPSHOT
* IntelliJ 2023.3.7

hohwille:
https://github.com/devonfw/IDEasy/blob/09abbeb67c9377e53279d715c8eaddc1caaee523/cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java#L211

PR #474, opened 2024-07-11
#468: start intellij in background

jan-vcapgemini:
Fixes: #468 

Implements:
* started intellij in background (replaced ProcessMode with ProcessMode.BACKGROUND)


hohwille (approved):
@jan-vcapgemini thanks for the quick fix. Awesome 👍

PR #475, opened 2024-07-11
#337: fixed os and architecture names

jan-vcapgemini:
Fixes: #337

Implements:
* fixed os and architecture names


hohwille (changes requested):
@jan-vcapgemini thanks for your PR.
However, this change does not make sense to me as it will surely break the build even more.
You are now attaching 4 times the same file instead of the 4 different OS-specific releases.
IMHO you can close this PR.

PR #481, opened 2024-07-16
#437 and #458 Refactor JsonUrlUpdaters

### LINKED ISSUE #437: Refactor NpmUrlUpdater

### LINKED ISSUE #458: Refactor JsonUrlUpdaters
It looks like the implementations of the JsonUrlUpdaters branched from the [IntellijUrlUpdater](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/tool/intellij/IntellijUrlUpdater.java) and the [JavaUrlUpdater](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/tool/java/JavaUrlUpdater.java).
The IntellijUrlUpdater is using a workaround for different editions and was re-used (copy&pasted) in various other JsonUrlUpdaters (Android Studio, Python). The problem with these implementation of the updaters is that the `mapVersion` (filtering of alpha and beta versions) is missing.

We now need to use the JavaUrlUpdater as base and adjust all other JsonUrlUpdaters to it.

More details about the `mapVersion` issue can be found here: https://github.com/devonfw/IDEasy/pull/445

List of JsonUrlUpdate subclasses
- [x] `GithubUrlUpdater`
- [ ] #483
- [ ] #484
- [ ] `AndroidStudioUrlUpdater`
- [x] `JavaUrlUpdater`

slskiba:
Fixes #437. Fixes #458. 

Refactoring of all JsonUrlUpdaters

- Npm now uses `JsonUrlUpdater` instead of `WebsiteUrlUpdater`, with a `NpmBasedUrlUpdater` in between. Introduced the necessary classes to map the json structure from the npm registry
- Unified the implementations of all other JsonUrlUpdaters:
   - common update method in `JsonUrlUpdater`
   - Introduced `JsonVersionItem` as a means to identify the json items holding version and download information to allow for common methods signatures in `JsonUrlUpdater`
   - Replaced `collectVersionsFromJson` with `collectVersionsWithDownloadsFromJson` and a common implementation for JsonUrlUpdaters to allow for simultaneous grabbing of version with download urls from `JsonVersionItem`

- Add tests to Java and Npm UrlUpdaters
- Fixed test of Python UrlUpdater (since we are now not making beta versions available anymore)


cli/src/test/java/com/devonfw/tools/ide/tool/java/JavaUrlUpdaterTest.java
```diff
@@ -0,0 +1,89 @@
...
+  private static final String EXPECTED_ABODY_CHECKSUM = "de08da1685e537e887fbbe1eb3278fed38aff9da5d112d96115150e8771a0f30";
+
+  /**
+   * Test of {@link JsonUrlUpdater} for the creation of {@link JavaUrlUpdater} download URLs and checksums.
+   *
+   * @param tempDir Path to a temporary directory
+   * @throws IOException test fails
+   */
+
+  @Test
```
jan-vcapgemini, line 43:
```suggestion
  /**
   * Test of {@link JsonUrlUpdater} for the creation of {@link JavaUrlUpdater} download URLs and checksums.
   *
   * @param tempDir Path to a temporary directory
   * @throws IOException test fails
   */
  @Test
```

cli/src/test/java/com/devonfw/tools/ide/tool/npm/NpmJsonUrlUpdaterTest.java
```diff
@@ -0,0 +1,151 @@
...
+  private static final String EXPECTED_ABODY_CHECKSUM = "de08da1685e537e887fbbe1eb3278fed38aff9da5d112d96115150e8771a0f30";
+
+  /**
+   * Test of {@link JsonUrlUpdater} for the creation of {@link NpmUrlUpdater} download URLs and checksums.
+   *
+   * @param tempDir Path to a temporary directory
+   * @throws IOException test fails
+   */
+
+  @Test
```
jan-vcapgemini, line 42:
```suggestion
  /**
   * Test of {@link JsonUrlUpdater} for the creation of {@link NpmUrlUpdater} download URLs and checksums.
   *
   * @param tempDir Path to a temporary directory
   * @throws IOException test fails
   */
  @Test
```

cli/src/main/java/com/devonfw/tools/ide/tool/python/PythonUrlUpdater.java
```diff
@@ -61,51 +80,37 @@ protected Class<PythonJsonObject> getJsonObjectType() {
...
 
-  @Override
-  protected void collectVersionsFromJson(PythonJsonObject jsonItem, Collection<String> versions) {
-
-    throw new IllegalStateException();
-
-  }
-
   @Override
   public void update(UrlRepository urlRepository) {
```
jan-vcapgemini, line 84: Do we still need to override the update method?
Won't we lose the isMissingOs check then?

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/IntellijUrlUpdater.java
```diff
@@ -20,24 +16,19 @@
...
   private static final String ULTIMATE_EDITION = "ultimate";
-
   private static final String COMMUNITY_EDITION = "intellij";
-
   private static final ObjectMapper MAPPER = JsonMapping.create();
 
-  private static final Logger logger = LoggerFactory.getLogger(IntellijUrlUpdater.class);
-
   @Override
   public void update(UrlRepository urlRepository) {
```
jan-vcapgemini, line 28: Do we still need to override the update method?
Won't we lose the isMissingOs check then?

jan-vcapgemini (changes requested):
Looks like a really nice refactor of the UrlUpdaters to me.
I've added some small questions and CR's, but I think this looks quite solid already.

cli/src/main/java/com/devonfw/tools/ide/common/JsonVersionItem.java
```diff
@@ -0,0 +1,8 @@
+package com.devonfw.tools.ide.common;
+
+/**
+ * Interface for a data object that can be read from and written to JSON.
+ */
+public interface JsonVersionItem {
+
+}
```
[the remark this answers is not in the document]
slskiba: I have used this to allow for a common implementation of `collectVersionsWithDownloadsFromJson` in `JsonUrlUpdater`, so that each subclass does not have to implement it themselves. 

The `JsonUrlUpdater` is now based on the subclasses `JsonObject` and `JsonVersionItem`, with the first representing the whole json to parse the versions from, and the latter the individual items that contain the versioning number and possibly download links. Each subclass implements their own method of the relation between `JsonVersionItem`s (`JVI`) and the `JsonObject` (`J`) by overriding 
`protected abstract Collection<JVI> getVersionItems(J jsonObject)`

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/IntellijUrlUpdater.java
slskiba: Currently, the `update` method of `AbstractUrlUpdater` (and the derived method in `JsonUrlUpdater`) do not support multiple editions. 
I tried to keep this PR to the refactoring of `JsonUrlUpdaters`, and only minimally touch the `AbstractUrlUpdater` where necessary. Therefore `IntellijUrlUpdater` yet has to have its own update method to support multiple editions. I would suggest to make a seperate PR to addess point 2 of #458

cli/src/main/java/com/devonfw/tools/ide/tool/python/PythonUrlUpdater.java
slskiba: The update method reads the json response into the corresponding `JsonObject`. 
However for python this generic implementation did not quite fit, since python's json is different from the others, and has to be read into a a `PythonRelease` array. This is then wrapped into the JsonObject to be send to the `collectVersionsWithDownloadsFromJson` method.

slskiba: The `isMissingOs` check is part of `collectVersionsWithDownloadsFromJson` implemented in the `JsonUrlUpdater`, and is not lost.

jan-vcapgemini: Hmm, this could be refactored a bit too I think, the update method of `AbstractUrlUpdater` looks quite similar to the `collectVersionsWithDownloadsFromJson` method, apart from the `addVersion(urlVersion, item);`. But this could be done in a separate PR of course.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/IntellijUrlUpdater.java
jan-vcapgemini: Added a new issue #483

cli/src/main/java/com/devonfw/tools/ide/npm/NpmJsonDist.java
```diff
@@ -0,0 +1,20 @@
+package com.devonfw.tools.ide.npm;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * JSON data object for a version of Npm. We map only properties that we are interested in and let jackson ignore all others.
+ *
+ * @see NpmJsonObject#getVersions()
+ */
+public class NpmJsonDist {
```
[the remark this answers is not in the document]
hohwille: Records are immutable and existing frameworks like hibernate, jackson, mapstruct, etc. are only slowly adding support for that since it is way complicated to unmarshall to a record and adding such support requires Java support for records what again requires "Java >= 14" and drops support for older Java versions.
While we might think that Java 17 or 21 should already be state of the art everywhere, there are still projects using Java8 in the wild and some of these libraries do not want to leave such projects left in the rain...
However, for jackson it seems that this should already work:
https://stackoverflow.com/questions/68394911/why-record-class-cant-be-properly-deserialized-with-jackson
So we could go for this even though surely a nice-to-have change...

cli/src/main/java/com/devonfw/tools/ide/common/JsonVersionItem.java
```diff
@@ -0,0 +1,8 @@
+package com.devonfw.tools.ide.common;
+
+/**
+ * Interface for a data object that can be read from and written to JSON.
+ */
+public interface JsonVersionItem {
+
```
hohwille, line 7: I would suggest that we even make it simpler by adding `getVersion()` method here:

```suggestion
  String getVersion();
```

If the version should have a different property name in JSON, you can use Jackson annotation.
Else if the version is transient and needs to be computed out of other attributes, we simply implement it as a getter and annotate with `@JsonIgnore` if needed.
If we later still have a totally unexpected and odd case, we can sill override `String getVersion(JVI jsonVersionItem)` and implement some nonsense for `getVersion()` like `throw new UnsupportedOperationException();` but I guess this will never happen and then we can even prevent to override this method for every JSON updater making it even more simple. WDYT?

slskiba: `getVersion` of the `JsonUrlUpdater` now simply delegates to the `jsonVersionItem.getVersion()` method, and have moved the `getVersion` logic of the `GithubUrlUpdater` and `JavaUrlUpdater` into the appropriate version items.

cli/src/main/java/com/devonfw/tools/ide/npm/NpmJsonDist.java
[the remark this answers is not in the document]
slskiba: I have changed almost all of the classes that represent JSON data objects to be records.

PR #482, opened 2024-07-17
#430 Add source of env variable for debugging

### LINKED ISSUE #310: Add a new property to env to display the file origin of each variable
As an IDEasy user, I want to know the source file of each environment variable.

### LINKED ISSUE #430: config debug support
We already have `ide env` (or `ide env --bash` for git-bash users on Windows).
However, sometimes users of IDEasy want to understand where a variable comes from.
Our [configuration](https://github.com/devonfw/IDEasy/blob/main/documentation/configuration.adoc) is quite flexible but can somehow also be complex since I have many `ide.properties` files to dig in finding the source of a (mis)configuration.

The aim of this story is to support debugging of the output if the log-level is at least on debug so via `ide --debug env` or `ide -d env`.
So instead of getting something like this:
```
DOTNET_VERSION="6*"
NODE_VERSION="v18*"
DOCKER_EDITION=docker
WORKSPACE="main"
WORKSPACE_PATH="D:\projects\IDEasy\workspaces\main"
JAVA_VERSION="17*"
```

I would get this:
```
from settings/ide.properties:
DOTNET_VERSION="6*"
NODE_VERSION="v18*"
JAVA_VERSION="17*"
from config/ide.properties:
DOCKER_EDITION=docker
from defaults:
WORKSPACE="main"
WORKSPACE_PATH="D:\projects\IDEasy\workspaces\main"
```

the lines starting with `from` would be debug log statements (in gray).
For this we would need to add the source of the property to the result object (e.g. in `VariableLine` as optional property).
Further, in debug mode, we would need to sort/group the results by the source to get the desired output.

slskiba:
Closes #430. Maybe also closes #310


cli/src/main/java/com/devonfw/tools/ide/environment/VariableLine.java
```diff
@@ -288,4 +304,9 @@ public static VariableLine of(boolean export, String name, String value) {
     return new Variable(export, name, value);
   }
 
+  public static VariableLine of(boolean export, String name, String value, Object source) {
```
jan-vcapgemini, line 307: Please add Javadoc.

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -132,7 +133,7 @@ private final Collection<VariableLine> collectVariables(boolean onlyExported) {
   /**
    * @param variables the {@link Set} where to add the names of the variables defined here.
    */
-  protected void collectVariables(Set<String> variables) {
+  protected void collectVariables(Map<String, String> variables) {
```
jan-vcapgemini, line 136: Why was this changed from a Set (prevents duplicates) to a Map?

jan-vcapgemini (changes requested):
Looks good to me, finally we can check where the variable came from.
I've added some small CR's and a question.
Maybe we should add some more tests for special cases and proper output validation too, what do you think?

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
slskiba: Since the variables are the keys of the map, duplicates are still prevented. A map now allows us to additionally keep track of the source of the variable, which is contained in the value part of the map.

jan-vcapgemini: Ah, ok. I was too focused on HashMaps allowing duplicate values but missed the key. At least the Javadoc got corrected now, missed this too...

jan-vcapgemini (approved):
Thanks for adding the test and solving my question. Ready for review now.

hohwille (commented):
@slskiba thanks for your PR. All seems functionally correct. 👍 
I am still wondering what would be the best solution here.
Actually the idea was that `collectVariables` would just give use all variables defined in `ide.properties` reps. `devon.properties` files and we wanted to avoid the overhead to evaluate their values in the first place so if someone just needs the names of the variables we can avoid the overhead and further redefined variables do not need to be evaluated multiple times of different levels...
Now after revisiting our current code and state, I come to the impression that it would make more sense to refactor `collectVariables` to return `Collection<VariableLine>` instead.
Internally in the recursive method we could then have a `Map<String, VariableLine>` and use the variable name as key. This way we could prevent further evaluation of variable values if the `VariableLine` is already present in the map. After the recursive invocation we can simply return `map.values()`. For me that would make more sense and lead to a better design.
WDYT?

cli/src/test/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandletTest.java
```diff
@@ -34,6 +38,11 @@ public void testRun() {
     assertLogMessage(context, IdeLogLevel.INFO, "JAVA_VERSION=\"17*\"");
     assertLogMessage(context, IdeLogLevel.INFO, "INTELLIJ_EDITION=\"ultimate\"");
     assertLogMessage(context, IdeLogLevel.INFO, "DOCKER_EDITION=\"docker\"");
+    //assert messages of debug level grouping of environment variable are present
+    assertLogMessage(context, IdeLogLevel.DEBUG, "from defaults:");
+    assertLogMessage(context, IdeLogLevel.DEBUG, "from " + settingsIdeProperties + ":");
+    assertLogMessage(context, IdeLogLevel.DEBUG, "from " + confIdeProperties + ":");
```
hohwille, line 44: Seems we should improve our log assertion support since we actually want to assert the order of the log statements here since otherwise the test will still be green if things are logged totally out of order...

hohwille: Created issue #507

hohwille (approved):
Since my suggestions seem to add quite some complexity, we decided to merge as is, and address that later.
I created #507 
Ready for merge.

PR #486, opened 2024-07-22
#303 Handle null return of findbash

### LINKED ISSUE #303: Make findBash method fully consistent
From #256 and before we still have a little inconsistency:
* it may be that bash could not be found (maybe it is even not installed or simply not on PATH)
* On Windows in some cases `findBash()` may return `null`
https://github.com/devonfw/IDEasy/blob/f23348b910a2f3e7dd3da2f1e98cc0fb38acb920/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L973
https://github.com/devonfw/IDEasy/blob/f23348b910a2f3e7dd3da2f1e98cc0fb38acb920/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L987
* On Windows in another case it will throw an exception
https://github.com/devonfw/IDEasy/blob/f23348b910a2f3e7dd3da2f1e98cc0fb38acb920/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L992
* On Linux or Mac `bash` could not be on the `PATH` and but `findBash()` just always returns `bash` without checking. Please note that in #274 this was suggested for KISS and I agreed since `bash` is a prerequisite on every sane Linux and MacOS installation.

Already this is somehow not consistent:
We should either return `null` if bash is not found or throw an exception but not do a mixture of both.

If it can return `null` then we have places where we need to handle this but currently do not do this:
* https://github.com/devonfw/IDEasy/blob/f23348b910a2f3e7dd3da2f1e98cc0fb38acb920/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java#L82C12-L85
* https://github.com/devonfw/IDEasy/blob/f23348b910a2f3e7dd3da2f1e98cc0fb38acb920/cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java#L270

And here we actually do a `null` check and have a fallback what might argue for returning `null` if bash is not available:
https://github.com/devonfw/IDEasy/blob/f23348b910a2f3e7dd3da2f1e98cc0fb38acb920/cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java#L310-L316

However, since the beginning of devonfw-ide we have `bash` as an absolute prerequisite of our product and without nothing is working. Since we also need git and therefore with Git-bash we also have `git` on Windows, we do not really have to implement strategies how to deal with the situation that `bash` is not available and can also simply raise an exception directly.
But we need to handle that error case properly since we manipulate the `PATH` variable and this could result that `bash` may not be found anymore if something goes wrong with the `PATH` variable. Therefore in such case the user should get an appropriate error message instead of a NullPointerException that is not helpful.

slskiba:
Closes #303


cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -1064,7 +1064,8 @@ private String findBashOnWindows() {
       }
     }
     // no bash found
-    throw new IllegalStateException("Could not find Bash. Please install Git for Windows and rerun.");
+    //throw new IllegalStateException("Could not find Bash. Please install Git for Windows and rerun.");
```
hohwille, line 1067: Lets better follow clean code
```suggestion
```

hohwille (changes requested):
@slskiba thanks for your PR and fixing the bash inconsistency. Your choice to let `findBash()` return `null` if no bash was found seems reasonable. 👍 
However, I am unsure if compensation of the problem is possible or makes sense in these 3 cases... Wouldn't it be smarter to create another function in `IdeContext` like this:
```java
default String findBashRequired() {
  String bash = findBash();
  if (bash == null) {
    String message = "Could not find bash what is a prerequisite of IDEasy.";
    if (getSystemInfo().isWindows()) {
      message = message + "\nPlease install Git for Windows and rerun.";
    }
    throw new IllegalStateException(message);
  }
}
```

IMHO if we want to run a `bash` script and we do not find bash, we cannot just return `null`  and continue. Also when we want to install a global tool and have to run some package manager installation in bash we cannot just return `false` - it seems that this boolean return of `executePackageManagerCommand` is saying if the installation was successful or not. However, global tool commandlets just seem to pass that boolean as a result of `doInstall` what is incorrect: 
```java
  /**
   * Installs or updates the managed {@link #getName() tool}.
   *
   * @param silent - {@code true} if called recursively to suppress verbose logging, {@code false} otherwise.
   * @return {@code true} if the tool was newly installed, {@code false} if the tool was already installed before and nothing has changed.
   */
  protected abstract boolean doInstall(boolean silent);
```

So IMHO we should simply use `findBashRequired()` in these 3 places and then can skip the `null` checks there...

hohwille (approved):
@slskiba thanks for addressing all my review feedback perfectly 🚀 
Ready for merge.

Issue #466, opened 2024-07-09
workspace not properly detected


hohwille:
If I go to a specific workspace, that workspace should be configured for `WORKSPACE` and `WORKSPACE_PATH`.
However, if I am not inside a specific workspace, this should fallback to `main`.
When I go to the `workspaces` directory itself, I am not inside a workspace and should get `WORKSPACE=main` and `WORKSPACE_PATH=$IDE_HOME/workspaces/$WORKSPACE` (resolved). But this is not the case:
```
hohwille@CE49454 MSYS /d/projects/IDEasy/workspaces
$ ide intellij
Configuring workspace workspaces for IDE intellij
Step 'ide (intellij)' failed: java.lang.IllegalStateException: Could not write to text file: D:\projects\IDEasy\workspaces\.idea\.name
1 step(s) failed out of 1 steps.
An unexpected error occurred!
We are sorry for the inconvenience.
Please check the error below, resolve it and try again.
If the error is not on your end (network connectivity, lack of permissions, etc.) please file a bug:
https://github.com/devonfw/ide/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=IllegalStateException%3A+Could+not+write+to+text+file%3A+D%3A%5Cprojects%5CIDEasy%5Cworkspaces%5C.idea%5C.name
java.lang.IllegalStateException: Could not write to text file: D:\projects\IDEasy\workspaces\.idea\.name
        at com.devonfw.tools.ide.merge.TextMerger.merge(TextMerger.java:52)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:64)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:68)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:68)
        at com.devonfw.tools.ide.tool.ide.IdeToolCommandlet.configureWorkspace(IdeToolCommandlet.java:235)
        at com.devonfw.tools.ide.tool.ide.IdeToolCommandlet.run(IdeToolCommandlet.java:200)
        at com.devonfw.tools.ide.context.AbstractIdeContext.applyAndRun(AbstractIdeContext.java:897)
        at com.devonfw.tools.ide.context.AbstractIdeContext.run(AbstractIdeContext.java:847)
        at com.devonfw.tools.ide.cli.Ideasy.runOrThrow(Ideasy.java:92)
[...]

PR #488, opened 2024-07-23
#466: fixed workspace fallback to main

jan-vcapgemini:
Fixes: #466 

Implements:
* added fallback to main if no workspace was found
* added several tests for mentioned bugs


slskiba (approved):
LGTM 👍

hohwille (approved):
@jan-vcapgemini thank you for your PR. Now reading your diff I finally understand the bug and see what went wrong. Thanks for fixing and nice that you also improved the test to cover the gap. Great work 👍

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -178,6 +178,11 @@ public AbstractIdeContext(IdeLogLevel minLogLevel, Function<IdeLogLevel, IdeSubL
       currentDir = getParentPath(currentDir);
     }
 
+    // fallback to main
+    if (workspace.isEmpty()) {
+      workspace = WORKSPACE_MAIN;
+    }
+
```
hohwille, line 185: BTW: I would be a little bit more clean to solve this in line 168 like this:
```java
        if (FOLDER_WORKSPACES.equals(name1) && !name2.isEmpty()) {
```

hohwille: I just edited this online on github. Seems cleaner to not set workspace to empty string and then reset to `main` afterwards. So we can merge and include in todays release.

Issue #433, opened 2024-07-01
Add plugin installation support to intellij


jan-vcapgemini:
As a IDEasy intellj user, I want to be able to install intellij plugins.

Currently the `installPlugin` method is not implemented for Intellij yet.

Needs to be implemented here:
https://github.com/devonfw/IDEasy/blob/eae8fd2796e382f6ff84def23cb6ca550d760930/cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java#L36

Old implementation can be found here:
https://github.com/devonfw/ide/blob/cdf4af5de21489f3fdcbfdab41af858822e0d598/scripts/src/main/resources/scripts/command/intellij#L74-L111

PR #487, opened 2024-07-23
#223 dynamic wiremock ports

slskiba:
Closes #223 and closes #193

PR #480, opened 2024-07-15
#339: refactor intellij and android-studio

jan-vcapgemini:
Fixes: #339

Implements:
* removed unnecessary and unused methods and variables
* synchronized both tool classes

PR #490, opened 2024-07-25
#433: Implement Intellij plugins installation

mvomiero:
Fixes: #433


cli/src/main/java/com/devonfw/tools/ide/tool/intellij/IntellijPluginManager.java
```diff
@@ -0,0 +1,140 @@
...
+
+    Step step = this.context.newStep("Install plugin: " + pluginId);
+    try {
+      String buildVersion = readBuildVersion();
+
+      Path installationPath = commandlet.getPluginsInstallationPath();
+      ensureInstallationPathExists(installationPath);
+
+      if (downloadUrl == null || downloadUrl.isEmpty()) {
+        downloadUrl = String.format("https://plugins.jetbrains.com/pluginManager?action=download&id=%s&build=%s", pluginId, buildVersion);
```
jan-vcapgemini, line 53: Maybe move this link to a constant too?

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/IntellijPluginManager.java
```diff
@@ -0,0 +1,140 @@
...
+      }
+      step.close();
+    }
+  }
+
+  private String readBuildVersion() throws IOException {
+    Path buildFile = commandlet.getToolPath().resolve(BUILD_FILE);
+    if (context.getSystemInfo().isMac()) {
+      buildFile = context.getSoftwareRepositoryPath().resolve("default").resolve("intellij/intellij").resolve(commandlet.getInstalledVersion().toString())
+          .resolve("IntelliJ IDEA" + commandlet.generateMacEditionString() + ".app").resolve("Contents/Resources").resolve(BUILD_FILE);
```
jan-vcapgemini, line 79: This can probably be optimized, the resolve of the path to the app looks a bit too specific to me and maybe other methods e.g. `MacOsHelper` can be useful here too.

jan-vcapgemini (changes requested):
Looks like a really nice and clean implementation. I've added some small CR's to resolve.

hohwille (approved):
@mvomiero Thanks for your PR and implementing this story. Looks good to me 👍 
I have a few questions:
* Why do we need `extractJar`? A `*.jar` is technically just a ZIP archive used for Java code. If we have the situation that directories are missing in JARs and then the extraction is failing because only the file entry is present and the parent directory was not created - then wouldn't it make sense to fix this in `extractZip` instead?
* Does it really make sense to extract `PluginManager` classes? Surely it is good to keep class files small and follow SoC. However, we have to implement the same for `AndroidStudio`, `Eclipse`, `VSCode` and maybe more IDEs in the future. For reuse it would make more sense to me to put the generic code to download and potentially unpack into `IdeToolCommandlet` for reuse. Then only few lines may remain for IntelliJ specific plugin stuff...

We can still merge as is and refactor later (by someone else when you have left). But maybe the questions could tackle some important information from you that can help us with this...

mvomiero:
Implemented the refactoring:
* with a general `PluginInstaller` connected to `IdeToolCommandlet` (for download and extraction)
*  and `IntellijPluginInstaller` that _extends_ `PluginInstaller`, specific for the `Intellij` commandlet (computing the right download URL or read the `build.txt` of intellij).


I ll try to work on the other changes tomorrow 🛩️

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -78,7 +81,7 @@ protected void postInstall() {
     }
   }
 
-  private String generateMacEditionString() {
+  String generateMacEditionString() {
```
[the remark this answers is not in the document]
mvomiero: moved to `IntellijPluginInstaller`, can then be removed when #480 is merged. Becuse is still used in `Intellij` right now.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/IntellijPluginManager.java
mvomiero: I think I would leave it like that to make it easier for the developer to see where the values or the arguments are gonna be replaced in the string replacing `%s`

cli/src/test/java/com/devonfw/tools/ide/tool/intellij/IntellijTest.java
```diff
@@ -10,26 +17,31 @@
 import com.devonfw.tools.ide.log.IdeLogLevel;
 import com.devonfw.tools.ide.os.SystemInfo;
 import com.devonfw.tools.ide.os.SystemInfoMock;
+import com.github.tomakehurst.wiremock.junit5.WireMockTest;
 
 /**
  * Integration test of {@link Intellij}.
  */
+@WireMockTest(httpPort = 8080)
```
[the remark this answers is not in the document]
mvomiero: Added the `// TODO` for now, then #487 will take care of it

hohwille:
> Why do we need extractJar? A *.jar is technically just a ZIP archive used for Java code. If we have the situation that directories are missing in JARs and then the extraction is failing because only the file entry is present and the parent directory was not created - then wouldn't it make sense to fix this in extractZip instead?

As discussed in our meeting it seems I was wrong. There are cases where JAR files are not 100% ZIP compatible and can cause errors on extraction. I still would love to see the stacktrace but it seems that IJ foldercompact plugin would be such a case so we can reproduce it this way and investigate further.

Issue #301, opened 2024-04-18
Mutation testing


ndemirca:
**Mutation testing** is a technique for assessing the quality of tests by creating artificially faulty versions of the source code, known as "mutants". The goal is to determine if your tests can detect these mutants.  It provides an indicate how effective your tests are if they can uncover potential flaws in the code. 

1. Research various tools and frameworks, such as PIT, for conducting mutation testing.
2. Evaluate and determine the most suitable option for our project.
3. Present the findings to the team.
4. Collaboratively decide on the preferred solution.
5. Implement the chosen approach.

slskiba:
[PIT](https://pitest.org/) seems to be the best adopted mutation testing system for java. 

Alternatives to PIT are often developed for academic purposes and no longer actively developed or supported, such as [muJava](http://cs.gmu.edu/offutt/mujava/), [Jester / Simple Jester](http://jester.sourceforge.net/), [Jumble](http://jumble.sourceforge.net/), or [Javalance](http://www.st.cs.uni-saarland.de/mutation/)

[The Major Mutation Framework](https://mutation-testing.org/downloads/) is in active development, with its latest release in May 29, 2024. However it does not yet support maven, and the standalone usage might be very tricky to implement.

PR #491, opened 2024-07-25
#301 mutation testing

slskiba:
Suggested realization of #301 using https://pitest.org/

In IDEasy project directory, run the following command to generate a Pit Test Coverage Report of all of the cli module:
```
mvn test-compile org.pitest:pitest-maven:mutationCoverage -DtargetClasses=com.devonfw.tools.ide* -DtargetTests=com.devonfw.tools.ide* -pl cli
```
To run the mutation test on all of the cli module takes a bit more than an hour to run. To test only a subset of classes, specify them accordingly as targetClasses/targetTests, where targetClasses refer to be classes whose code is to be mutated, and targetTests refers to the tests that the mutations are tested against. Ideally, these two should stand in relation to one another. 

As of 25.07.2024, the report yields the following overall result: 
image
where
**Line Coverage** denotes the amount lines covered by tests in proportion to the total number of executable lines in a class. 
**Mutation Coverage** denotes the amount of mutations killed in proportion to the total number of mutations created (including those not covered by tests). It reaches 100% iff both line coverage and test strength (see below) reach 100%.
**Test Strength** denotes the amount of mutations killed in proportion to the total number of mutations covered by tests

In the report, individual classes can be examined for their test and mutation coverage and the applied mutations. 

A mutation is said to survive if all tests pass with the modified code. If a mutation is killed, it means that at least one test did not pass when run on the mutated code.


jan-vcapgemini:
Please add: https://github.com/pitest/pitest-junit5-plugin/blob/master/LICENSE.txt to your `LICENSE.txt`

pom.xml
```diff
@@ -84,6 +84,18 @@
...
       </plugin>
+      <plugin>
+        <groupId>org.pitest</groupId>
+        <artifactId>pitest-maven</artifactId>
+        <version>1.16.1</version>
+        <dependencies>
+          <dependency>
+            <groupId>org.pitest</groupId>
+            <artifactId>pitest-junit5-plugin</artifactId>
+            <version>1.2.1</version>
```
jan-vcapgemini, line 95: Shouldn't we add the LICENSE to our documentation too?

jan-vcapgemini (approved):
Looks good to me. I've added a question about the "missing" LICENSE.

pom.xml
jan-vcapgemini: License is not relevant to users, so this is fixed.

.github/workflows/pitest.yml
```diff
@@ -0,0 +1,28 @@
+name: Pitest
```
hohwille, line 1: Workflow name and filename is a little bit cryptic.
`mutation-test` would be more self explanatory.

hohwille (approved):
@slskiba thanks for analyzing this innovative topic and providing this solution 👍 
I left a nice to have improvement comment.
Otherwise ready for merge.

hohwille (approved):
@slskiba thanks for making this perfect. 👍 
Ready for merge now.

PR #493, opened 2024-07-26
#489: Fix separator comparison

### LINKED ISSUE #489: VersionRange not working correctly
### Expected behavior

As a user, I want the VersionRange to work correctly with "Exclusive and Inclusive". So if the VersionRange is [11,22), it should include every version from 11 (including) to 22 (excluding).

### Actual behavior

While tested with tomcat as tool and java as dependency, this VersionRange [11,22) includes versions like 22_36, which is not supposed to do.

### Steps to reproduce (bug)

In VersionRangeTet class, a test can be added to get this behavior. The following code can be added in the testOf() Method: 

String v4String = "[11,22)";
VersionRange v4 = VersionRange.of(v4String);
assertThat(v4.contains(VersionIdentifier.of("22_36"))).isEqualTo(false);

This assert is supposed to be true, so the VersionRange [11,22) is not supposed to include version 22_36, but it actually does.

### Comments/Hints:

Here is a picture of the test case: 

image

slskiba:
Closes #489. 

Fixes the issue, now ensures that "22_1 > 22", while still "22_1 < 22.1".

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/version/VersionSegment.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/version/VersionSegment.java
@@ -216,11 +216,18 @@ public class VersionSegment implements VersionObject<VersionSegment> {
     if (!lettersResult.isEqual()) {
       return lettersResult;
     }
-    //a version with a separator is greater than a version without, a version with a "." separator is greater than a version with a "_" separator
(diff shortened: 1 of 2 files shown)


jan-vcapgemini (approved):
Looks good to me. Thanks for this fix.

hohwille (approved):
@slskiba thanks for extending the tests and fixing the bug in version comparison. 👍

Issue #427, opened 2024-06-27
Improve deployment workflows (nightly and release)


hohwille:
Both nightly build and release workflow build the native image (via matrix OS for each involved target platform OS/arch).
There are several things to improve:
* we want to support more target platforms (see e.g. #337)
* on the according platform builds (matrix os steps) we should IMHO only build the native image and upload it but do nothing else
* on the main build machine we should do the full build except for the native-image and just collect the native images build before. This includes building the documentation, building the `gui`, and for the `cli` finally build the assembly with everything together (documentation + native images + scripts and other package resources).

What I also do not get is where do we do the "loop" over all OS combinations in the main build? Actually I would expect that either in the workflow build or in our CLI POM or in the assembly we define that we want to have a release for win-x64, max-x64, mac-arm, linux-x64 (and potentially more). In the POM I see this:
https://github.com/devonfw/IDEasy/blob/dd966b4a603d68f0693b07c8a76d63dd626421f9/cli/pom.xml#L18C5

So for the native-image build this somehow makes sense to me. But how does the main build finally create multiple OS/arch specific releases? My only explanation is that he does not at all but just downloads all that was uploaded from the matrix OS builds and maven just attaches them as artifacts and therefore includes them as-is into the `deploy`. Therefore, I see that we will need some rework. I still also do not get why the nightly build failed lacking the documentation to include while the release build did not fail. Why is that? To me this does not make sense.

BTW: I do not want to cause extra complexity. We can also stay with the approach that the native matrix OS build is already building the full package and target release. However, then before we would need to build things like the documentation first and then include that single documentation into all the matrix os builds (via maven repo? or via GHA uploads/downloads). I just have the impression that this will all make it much more in-transparent and therefore complex.
Currently, I do not fully understand what is going on and I would consider myself as a maven expert. So who will then be able to understand it and is capable to improve and apply fixes in the future? So we should find a simple and understandable solution and for what remains complex we should add documentation (e.g. comments or even asciidoc file explaining how all works together with GHA, maven, assembly, etc.). 

[...]

PR #494, opened 2024-07-26
#427: optimize nightly builds

jan-vcapgemini:
Fixes: #427 

Implements:
* new job management with shared results (documentation and native images)
* dynamic profile builds with automatic archive and upload/download of github artifacts
* update of `actions/upload/download-artifact` to `v4`
* some documentation
* update of `native-maven-plugin` to `0.10.2`
* still some redundancies in the `cli/pom.xml`


.github/workflows/release.yml
```diff
@@ -21,23 +47,26 @@ jobs:
...
+      - name: Build native image
         shell: bash
         run: |
-          maven_config="$(cat .mvn/maven.config)"
-          current_version="${maven_config/#*-Drevision=}"
-          current_version="${current_version/ */}"
           cd cli
-          mvn -B -ntp -Pnative -DskipTests=true package
-      - uses: actions/upload-artifact@v3
+          mvn -B -ntp -P${{ matrix.os }} '-P!build-documentation' -DskipTests=true package
```
hohwille, line 59: I would not create a profile for each matrix OS. Your 4 profiles for this in the POM are 98% copy&paste.
Simply keep a single `native` profile and pass the assembly XML file name part as variable:
```suggestion
          mvn -B -ntp -Dideasy.assembly.id=${{ matrix.os }} -DskipTests=true package
```

cli/pom.xml
```diff
@@ -140,54 +140,141 @@
...
-                      <file>${project.build.directory}/${project.artifactId}-osx-aarch_64.tar.gz</file>
-                      <type>tar.gz</type>
-                      <classifier>mac-arm</classifier>
-                    </artifact>
-                  </artifacts>
+                  <appendAssemblyId>true</appendAssemblyId>
+                  <descriptors>
+                    <descriptor>/src/main/assembly/release-windows-x86_64.xml</descriptor>
+                  </descriptors>
+                </configuration>
```
hohwille, line 203: As described above. Just keep one of these 4 profiles as `native` and simply reference descriptor as
```
<descriptor>/src/main/assembly/release-${ideasy.assembly.id}.xml</descriptor>
```
Rename the descriptors to match the github OS image names and we can avoid all the copy & paste.

hohwille (approved):
@jan-vcapgemini thanks for this PR. You did a great job finding a good solution for the build chaining. 👍 
However, I left comments for further improvement.
To build the documentation, we can simply build in `documentation` folder.
To build the package, we already build in `cli` folder - hence, no profile to include or exclude build of documentation is required.
We can avoid all the copy&paste of profiles and simplify as commented.

cli/pom.xml
jan-vcapgemini: I've implemented all of your suggestions now and tested the release and nightly build workflows on my own fork.
Seems to be working as expected.

hohwille (commented):
@jan-vcapgemini thanks for the rework. 🚀 
I still have doubts that the documentation is just uploaded and downloaded but not installed into maven local repo and assume this will fail at least for release build but we will see today as I will build a release after merging and maybe this will also proof me wrong.
Ready for merge.

hohwille:
> I still have doubts that the documentation is just uploaded and downloaded but not installed into maven local repo and assume this will fail at least for release build but we will see today as I will build a release after merging and maybe this will also proof me wrong.

In my last review now I finally got it. We do not use dependency-plugin anymore to get the doc PDF from local repo and copy it to `target/package` but this is directly done via upload and download of artifacts. Makes sense now.

PR #495, opened 2024-07-28
#352: implemented new design for tools like python and node

### LINKED ISSUE #352: Design for tools like node and python that hack their own installation
As a general best-practice I would expect a tool to allow a pristine installation. That is the tool gets installed in some installation folder and then nothing is changes for this installation (while the version of the installation remains the same).

Based on this concept we create the software repository in IDEasy. For every tool we create a folder with a sub-folder for every version of that tool that gets installed. Then we can simply set/change/remove a symbolic link to install, upgrade/downgrade or remove a tool from a project. As a benefit of this principle multiple IDEasy projects can share the installation of a tool if they also use the same version what allows to save disc-space. Also downloading, extracting, etc. does not have to be repeated and the installation and upgrade/downgrade is lightning fast if the desired version is also available in the software repository.

However, IDEasy wants to ensure the sandbox principle for projects. That is each project lives in isolation (like a docker container) and changes in project `A` do not effect project `B` (on the same machine). This implies that the installation of a software tool for a particular version remains pristine.

In general this is all given for most tools like `java`, `dotnet`, etc. and working smooth.
A challenge are tools that support plugins as these are typically "installed" inside the installation itself. However, if project `A` installs plugin `P1` for tool `T` then project `B` should not automatically get plugin `P1` if it also uses tool `T` (in the same version).
We solved this by a dedicated plugins folder inside the projects and relocation of the plugins for a tool into such dedicated project specific folder via some launch parameter or configuration. Beside some edge-cases (see #224) this is also solved and working.

However, we have two famous tools that completely violate this principle and cause severe issues:
1. node (aka node.js)
2. pyhton

# node
Inside a node installation all software that is based on `node` is installed on top manipulating the existing installation.
The most prominent aspect is already `npm` that comes with `node`. If the project `P1` and `P2` both want to use the latest version of `node` they share the installation in the software repository behind their same symbolic link. Now if `P1` and `P2` have different requirements for the version of `npm` we have a severe conflict as if `P1` installs `npm` version `V1` it will do this inside the installation that is also used by `P2` and therefore have such side effect. If `P2` instead needs a different `npm` version `V2` it will override it also for `P1` on installation and vice versa.
If even gets worse since `npm install -g «tool»` also installs any other tool globally (`-g`) into that installation causing the same side effects.

See also #324

# python
With `python` we more or less have the same story as with `node`. What is `npm` for `node` we have with `pip` for `python` and installing `pip` and other python based tools independent from the `python` installation itself trying to keep it pristine have failed.
Here we already have various tickets indicating the problem:
* https://github.com/devonfw/ide/issues/839
* https://github.com/devonfw/ide/issues/1170
* https://github.com/devonfw/ide/issues/1314

# comparison
A great design can be seen with `java` or `dotnet` that bring their own runtime and then independent of that other tools depending on it can be installed elsewhere and as long as they know where to find such runtime (via env variables like `PATH` or `JAVA_HOME, etc.) all is working fine.
However, since we are not the makers of `node` or `python` we cannot change this "flawed design" and it seems so fundamental that there is almost no chance that these projects are likely to consider our plea.

# proposed solution
As a solution, I would suggest to add some `protected boolean isIgnoreSoftwareRepo()` to `LocalToolCommandlet` that by default returns `false`. Then we can change the implementation in `LocalToolCommandlet` such that if the new method returns `true` we then do not delegate to `installInRepo` method on installation and also `installInRepo` method will fail. Instead the `install` method will then install the software directly in the software folder.
This will have the following limitations if that feature is activated (may be more, not complete):
* no workaround for bug #264
* no support for dependencies in conflicting versions to such tool (#131) possible (a project cannot have two tools that require different/incompatible versions of `node` or `python`).
* installation/upgrading is not lightning fast
* wast of disk-space if two projects use the same tool in the same version

We will first create a PR implementing this change/feature to `LocalToolCommandlet`.

However, then we can solve the problem:
* node commandlet is overriding `isIgnoreSoftwareRepo()` to `true`
* npm commandlet is a simple delegation to `npm install -g resolveVersion($NPM_VERSION)` (simple delegate similar to kubectl - see #81). Like in devonfw-ide we will have no `software/npm` folder but have `npm` inside `software/node`.
* python commandlet is overriding `isIgnoreSoftwareRepo()` to `true`
* pip commandlet works like `npm` or `kubectl` as a simple wrapper on top of python commandlet. We will have to install pip inside python and not in its own `software/pip` folder.

Here we will create a separate PR for node/npm (e.g. in the context of #324) based on the PR for `LocalToolCommandlet` and this story itself.
Also we can then also start implementing PR for #38 and another one for `pip` (story needs to be created).

salimbouch:
the method installTool (which is the old installInRepo) now performs the installation for both cases (in software repo or directly in software folder based on the return value of the method isIgnoreSoftwareRepo).

closes #352

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
@@ -176,7 +176,7 @@ public abstract class LocalToolCommandlet extends ToolCommandlet {
         if (this.context.isForceMode()) {
           fileAccess.delete(toolPath);
         } else {
-          if (resolvedVersion.equals(getInstalledVersion()) || !isIgnoreSoftwareRepo()) {
+          if (!isIgnoreSoftwareRepo() || resolvedVersion.equals(getInstalledVersion())) {
             this.context.debug("Version {} of tool {} is already installed at {}", resolvedVersion, getToolWithEdition(this.tool, edition), toolPath);
             return createToolInstallation(toolPath, resolvedVersion, toolVersionFile);
           }
@@ -187,7 +187,6 @@ public abstract class LocalToolCommandlet extends ToolCommandlet {
       }
     }
     Path target = toolRepository.download(this.tool, edition, resolvedVersion);
-    fileAccess.mkdirs(toolPath.getParent());
     boolean extract = isExtract();
     if (!extract) {
       this.context.trace("Extraction is disabled for '{}' hence just moving the downloaded file {}.", this.tool, target);
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
@@ -87,6 +87,10 @@ public abstract class LocalToolCommandlet extends ToolCommandlet {
         // we need to link the version or update the link.
         Path toolPath = getToolPath();
         FileAccess fileAccess = this.context.getFileAccess();
+        if (Files.exists(toolPath)) {
+          fileAccess.backup(toolPath);
+        }
+        fileAccess.mkdirs(toolPath.getParent());
         fileAccess.symlink(installation.linkDir(), toolPath);
       }
       this.context.getPath().setPath(this.tool, installation.binDir());
@@ -172,7 +176,7 @@ public abstract class LocalToolCommandlet extends ToolCommandlet {
         if (this.context.isForceMode()) {
           fileAccess.delete(toolPath);
         } else {
-          if (resolvedVersion.equals(getInstalledVersion())) {
+          if (resolvedVersion.equals(getInstalledVersion()) || !isIgnoreSoftwareRepo()) {
             this.context.debug("Version {} of tool {} is already installed at {}", resolvedVersion, getToolWithEdition(this.tool, edition), toolPath);
             return createToolInstallation(toolPath, resolvedVersion, toolVersionFile);
           }
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
@@ -74,7 +74,7 @@ public abstract class LocalToolCommandlet extends ToolCommandlet {
     Step step = this.context.newStep(silent, "Install " + this.tool, configuredVersion);
     try {
       // install configured version of our tool in the software repository if not already installed
-      ToolInstallation installation = installInRepo(configuredVersion);
+      ToolInstallation installation = installTool(configuredVersion);
       // check if we already have this version installed (linked) locally in IDE_HOME/software
       VersionIdentifier resolvedVersion = installation.resolvedVersion();
       if (resolvedVersion.equals(installedVersion) && !installation.newInstallation()) {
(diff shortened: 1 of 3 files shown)


cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -83,14 +83,16 @@ protected boolean doInstall(boolean silent) {
...
-      if (Files.exists(toolPath)) {
-        fileAccess.backup(toolPath);
+      if (!isIgnoreSoftwareRepo()) {
+        // we need to link the version or update the link.
+        Path toolPath = getToolPath();
+        FileAccess fileAccess = this.context.getFileAccess();
+        if (Files.exists(toolPath)) {
+          fileAccess.backup(toolPath);
+        }
+        fileAccess.mkdirs(toolPath.getParent());
```
hohwille, line 93: Isn't that duplicated now? See lines 195-198.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -152,17 +162,24 @@ public ToolInstallation installInRepo(VersionIdentifier version, String edition,
...
     Path toolVersionFile = toolPath.resolve(IdeContext.FILE_SOFTWARE_VERSION);
     FileAccess fileAccess = this.context.getFileAccess();
     if (Files.isDirectory(toolPath)) {
       if (Files.exists(toolVersionFile)) {
         if (this.context.isForceMode()) {
           fileAccess.delete(toolPath);
         } else {
-          this.context.debug("Version {} of tool {} is already installed at {}", resolvedVersion, getToolWithEdition(this.tool, edition), toolPath);
-          return createToolInstallation(toolPath, resolvedVersion, toolVersionFile);
+          if (resolvedVersion.equals(getInstalledVersion()) || !isIgnoreSoftwareRepo()) {
```
hohwille, line 179: Nice to have but Java is evaluating boolean expression from left to right. `isIgnoreSoftwareRepo()` will most probably do nothing than returning a boolean literal. In case it returns `true` then `getInstalledVersion()` will not be called avoiding minimal waste to determine the actually installed version.

```suggestion
          if (!isIgnoreSoftwareRepo() || resolvedVersion.equals(getInstalledVersion())) {
```

hohwille: BTW: Why do we not skip the installation if the installed version is already correct in case we do not install in software repo? So `ide install npm` will always reinstall even if the correct version is already installed?

hohwille (approved):
@salimbouch thanks for implementing this story. Your solution looks elegant to me. Also renaming the `installInRepo` method makes sense to me. Great job 👍 
I left some minor comments/questions but seems functionally be complete and ready for merge. If you can look at the comments and maybe consider small improvement, would be nice but we can also merge and later do such minor nice-to-have improvements.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
salimbouch: In case the installed version is already correct in case of installing directly in the software folder, `resolvedVersion.equals(getInstalledVersion())` would be true and the installation is skipped.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
salimbouch: `ToolPath` doesnt have the same value in both methods, in `doInstall` it's the software folder, and in `installTool` it's either the software folder or software repo. So in case we want to change the version of a tool we need to back up the tool in software folder, I couldn't find a better way to avoid this redundancy.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
salimbouch: I reversed the order inside the if.

PR #510, opened 2024-08-03
#507: improve collect variables

### LINKED ISSUE #507: Review improvements to collectVariables and test env debug from #482
From #482 we left some review concerns open to avoid blocking the merge:
* [x] Actually the idea was that collectVariables would just give use all variables defined in ide.properties reps. devon.properties files and we wanted to avoid the overhead to evaluate their values in the first place so if someone just needs the names of the variables we can avoid the overhead and further redefined variables do not need to be evaluated multiple times of different levels...
Now after revisiting our current code and state, I come to the impression that it would make more sense to refactor collectVariables to return Collection<VariableLine> instead.
Internally in the recursive method we could then have a Map<String, VariableLine> and use the variable name as key. This way we could prevent further evaluation of variable values if the VariableLine is already present in the map. After the recursive invocation we can simply return map.values(). For me that would make more sense and lead to a better design.
* [x] Seems we should improve our log assertion support since we actually want to assert the order of the log statements here since otherwise the test will still be green if things are logged totally out of order... (in `EnvironmentCommandletTest` we actually want to assert that log statements across different log levels are in a specific order).

hohwille:
fixes #507 with following improvements:
* fixed bug in log assertion `hasEntries` implementation
* sort variables in output of `EnvironmentCommandlet` for deterministic results and reliable testing
* allow tests to define log level so we can test behavior dependent on log-level
* extend `EnvironmentCommandletTest` to cover all variables with DEBUG and normal logging (INFO)
* rework `collectVariables` for better design.
* allow test project to define `home/system.properties` to make tests reliable and independent from `System.getenv`
* added `VariableSource` and used it in `EnvironmentVariables` and `VariableLine`
* improved maven folder and config logic


hohwille:
@slskiba I now understood the problems you had trying to address my review suggestion. Indeed it did not work directly as I suggested:

The recursive method currently collects the variable names in the map. Their values always have to be resolved on the highest level of the hierarchy (`EnvironmentVariablesResolved`). So in order to create the `VariableLine` that value needs to be resolved what requires the `EnvironmentVariablesResolved` instance. I decided to pass it as argument to the recursive `collectVariables` method as `resolver` to make it possible.
As you can see I did some more refactoring here and there where I came along and saw room for improvement.

hohwille:
Tests fixed, all green now. I am happy with the PR.

cli/src/test/java/com/devonfw/tools/ide/log/IdeTestLoggerAssertion.java
```diff
@@ -89,23 +89,38 @@ private void hasEntries(boolean nothingElseInBetween, IdeLogEntry... exprectedEn
 
     assert (exprectedEntries.length > 0);
     int i = 0;
-    for (IdeLogEntry entry : exprectedEntries) {
-      if (entry.equals(exprectedEntries[i])) {
+    int max = 0;
+    for (IdeLogEntry entry : this.entries) {
+      if (exprectedEntries[i].matches(entry)) {
```
slskiba, line 94: nit: typo in `exprectedEntries`

cli/src/test/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandletTest.java
```diff
@@ -8,33 +8,119 @@
...
+        IdeLogEntry.ofInfo("TEST_ARGS10=\"user10 workspace10\""), //
+        IdeLogEntry.ofInfo("TEST_ARGS2=\" user2 conf2\""), //
+        IdeLogEntry.ofInfo("TEST_ARGS3=\" user3 workspace3\""), //
+        IdeLogEntry.ofInfo("TEST_ARGS4=\" settings4\""), //
+        IdeLogEntry.ofInfo("TEST_ARGS5=\" settings5 conf5\""), //
+        IdeLogEntry.ofInfo("TEST_ARGS6=\" settings6 workspace6 conf6\""), //
+        IdeLogEntry.ofInfo("TEST_ARGS7=\"user7 settings7 workspace7 conf7\""), //
+        IdeLogEntry.ofInfo("TEST_ARGS8=\"settings8 workspace8 conf8\""), //
+        IdeLogEntry.ofInfo("TEST_ARGS9=\"settings9 workspace9\""), //
+        IdeLogEntry.ofInfo("TEST_ARGSa=\" user1 settings1 workspace1 conf1  user3 workspace3 confa\""), //
```
slskiba, line 118: Why is the spacing so odd in the test arguments?

cli/src/test/java/com/devonfw/tools/ide/log/IdeTestLoggerAssertion.java
```diff
@@ -89,23 +89,38 @@ private void hasEntries(boolean nothingElseInBetween, IdeLogEntry... exprectedEn
...
+        IdeLogEntry entry = exprectedEntries[i];
+        error.append(entry.level());
+        error.append(":");
+        error.append(entry.message());
+        error.append('\n');
+      }
+    }
+    error.append("\nBut could not find expected log entries:\n");
+    for (i = max; i < exprectedEntries.length; i++) {
+      IdeLogEntry entry = exprectedEntries[i];
```
slskiba, line 123: I think this might only work fully as intended for when `nothingElseInBetween` is set to true. 
Consider that some (faulty) code produces the following Info log:
```
INFO - SomeLogger - A
INFO - SomeLogger - C
```
And consider the test calls `assertThat(context).logAtInfo().hasEntries("A", "B", "C")` to check for all 3 messages to be present as intendedn. 
I would expect the test to fail and state the missing "B" message.

However currently the test will fail with the following info to the developer:
```
Found expected log entries:
INFO:A

But could not find expected log entries:
INFO:B
INFO:C
```
even though the "C" message had been present.

However, this does not affect the correctness of the test, only the information provided upon failure.

slskiba (commented):
Some really nice improvements to the log testing 👍

cli/src/test/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandletTest.java
hohwille: What exactly do you mean? IntelliJ formatter? Or do you mean the values of these `TEST_ARGS*` variables?
For the first IMHO all is working as expected and for the latter you need to ask @MattesMrzik who created these variables and their values. IMHO he tried to cover many edge-cases of variable resolution and therefore also added leading or duplicated whitespace, but I am just guessing...

cli/src/test/java/com/devonfw/tools/ide/log/IdeTestLoggerAssertion.java
hohwille: This is just a matter of how we report the error in cases where the assertion is failing.
My intention was just to be helpful and show the first line that did not match so the developer can faster find the problem. This feature already helped me in various situations.
Your expectation would be extraordinary complex to implement. Also consider that I may expect `"A", "A", "B", "A"` and then what should I get reported if I however have `"A", "B", "B", "A"` in the log?
To me it is important that the assertion is properly verified. There error message should be helpful for the developer but must not be perfect. If you want to dig deeper here, you could argue this block
```
} else if (expectedEntries[0].matches(entry)) {
  i = 1;
```
is somehow not perfectly accurate for every edge-case but I think we can still fix it whenever this problem should hit us (IMHO not likely to happen in the next decade, but we will see).

cli/src/test/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandletTest.java
hohwille: user home:
```
TEST_ARGS1=${TEST_ARGS1} user1
TEST_ARGS3=${TEST_ARGS3} user3
```
settings:
```
TEST_ARGS1=${TEST_ARGS1} settings1
```
workspace:
```
TEST_ARGS1=${TEST_ARGS1} workspace1
TEST_ARGS3=${TEST_ARGS3} workspace3
```
conf:
```
TEST_ARGS1=${TEST_ARGS1} conf1
TEST_ARGSa=${TEST_ARGS1} ${TEST_ARGS3} confa
```

So the expected result is correct but rather hard to understand...

cli/src/test/java/com/devonfw/tools/ide/log/IdeTestLoggerAssertion.java
hohwille: OK. I improved the error reporting to avoid confusion with a simple approach:
I only report the first exepcted log entry that is not matching.
Developer will see his test and all the expected log entries he is expecting so I will resolve this conversation.

Issue #512, opened 2024-08-05
Fix nightly build conclusion detection


jan-vcapgemini:
Nightly builds are currently running properly, but the last workflow status check does not.
It appears that github actions REST API calls and Github set-opt are outdated.

PR #513, opened 2024-08-05
#512: fix GitHub actions nightly build

jan-vcapgemini:
Fixes: #512 

Implements:
* updated set-output to new GITHUB_OUTPUT
* removed check_status step


.github/workflows/check-for-updates.yml
```diff
@@ -32,3 +34,30 @@ jobs:
...
+
+      - name: Print and handle the status
+        run: |
+          echo "The status of the last workflow run is: ${{ env.conclusion }}"
+          if [ "${{ env.conclusion }}" != "success" ]; then
+            echo "The last workflow did not succeed. Failing this workflow."
+            exit 1
+          else
+            echo "The last workflow succeeded. This workflow will succeed."
+          fi
```
hohwille, line 63: We do not need this complex stuff anymore:
The check-for-updates workflow will run every night and will always be green.
The nightly-build will run whenever there had been changes and show the proper build status.
The readme badge will show the status of the last nightly build what is exactly what we want to have.

```suggestion
```

hohwille (approved):
@jan-vcapgemini thanks for your PR and improving the nightly build. Thanks for also analyzing and fixing the issue in the `check-for-updates` workflow 👍 

As I am trying to say we should entirely remove this logic to dig in the job history and finally simplify the workflows as suggested.
If we can agree to apply my suggestion, we can merge asap.

Issue #439, opened 2024-07-02
Refactor and improve tool-dependencies and tomcat


hohwille:
From PR #250 we have to following review improvements left to do:
- [x] remove `TomcatCommand` and make `Tomcat` commandlet to simply delegate to `catalina.sh` passing any given arguments from `ide tomcat «args»`
* [x] extract `UrlDependencyFile` for URL-Model to support lazy loading and simpler access to `dependency.json`
* [x] create PR for `ide-urls` to add `dependencies.json` to `tomcat/tomcat`.
* [x] extract `withEnvVar` method from `ProcessContext` to own interface `EnvironmentContext`.
* [x] introduce new method to `LocalToolCommandlet` for setting environment variables `public void setEnvironment(EnvironmentContext context, Path toolPath)`.
* [x] refactor ~~installInRepo~~ `installTool` method so we can pass `VersionRage` (instead of `VersionIdentifier`) for a dependency (from the `DependencyInfo`).
* [x] when trying to resolve the dependency version and install the tool, first check if the resolved version of the configured tool version of the tool matches the dependency version range. If that is the case then use the tool installation preferred by the project (`software/«tool»` via `getToolPath()`). Only if that version does not match then install an additional version in repo if not already available.
* [x] If you install the tool version in the repo, try to reuse the existing code from `installInRepo` methods.
* [x] In `runTool` we can then first create the process context, pass it as `EnvironmentContext` to the installation method (needs some refactoring) so that there is only one single place in the code getting the dependencies and looping over them that does all together. That is the dependent tools are all installed and also their environment variables are set. For regular existing methods like `install()` and `install(boolean)` you could pass `null` as `EnvironmentContext` and in case that is `null` also omit to call the `setEnvironment(EnvironmentContext, Path)` method.

hohwille:
Issue was auto-closed. Not all tasks are completed/check. We have to double check if there is something left to be done.
Until that is clarified and all tasks are checked, I reopen it.

PR #515, opened 2024-08-05
#439: refactor tool dependencies and tomcat

aBega2000:
Fixes: #439 

With this PR I intend to solve the issue of tool-dependencies by refactoring and improving it as much as possible. 

As can be seen, many classes are changed and this has happened because of the change in the install method where I now pass EnvironmentContext as argument. Namely this came from the point: **"In runTool we can then first create the process context, pass it as EnvironmentContext to the installation method (needs some refactoring) so that there is only one single place in the code getting the dependencies and looping over them that does all together."**

The problem at the moment is that some tests are not running successfully anymore. If this approach that I am following is correct, then the last step for this issue would be to fix the mentioned tests so that everything runs smoothly.


cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
```diff
@@ -150,7 +150,7 @@ private static String getTool(Path path, Path ideRoot) {
 
   private Path findBinaryInOrder(Path path, String tool) {
 
-    List<String> extensionPriority = List.of("");
+    List<String> extensionPriority = List.of("", ".sh");
```
hohwille, line 153: I am not sure this is correct.
When I am in a shell and type `«cmd»` then I only execute this given command.
On windows in CMD or powershell, this will also execute `«cmd».exe`, `«cmd».bat`, `«cmd».msi`, etc. if available.
But in no shell on any OS that I am aware of this will run `«cmd».sh`.

I do not know why you really added this here but if you want to run `catalina.sh` then why not running it like that instead of just `catalina`. You can forget about `catalina.bat` on Windows if that was your concern. This exists only for convenience since Windows CMD is unable to directly execute `catalina.sh`.

cli/src/main/java/com/devonfw/tools/ide/tool/CustomToolCommandlet.java
```diff
@@ -16,15 +16,15 @@ public CustomToolCommandlet(IdeContext context, CustomTool customTool) {
   }
 
   @Override
-  public ToolInstallation installTool(VersionIdentifier version) {
+  public ToolInstallation installTool(EnvironmentContext environmentContext, VersionIdentifier version) {
```
hohwille, line 19: see first comment above and keep original method signature delegating to new one with `null` as fallback here.
As best-practice please add optional parameters at the end and not at the start.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -323,127 +318,66 @@ private ToolInstallation createToolInstallation(Path rootDir, VersionIdentifier
...
-    if (Files.exists(this.dependency.getDependencyJsonPath(getConfiguredEdition()))) {
-      setDependencyRepository(getInstalledVersion());
-    } else {
-      this.context.trace("No Dependencies file found");
-    }
-
-    ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.WARNING).executable(binaryPath).addArgs(args);
-
-    for (String key : this.dependenciesEnvVariablePaths.keySet()) {
+  public void setEnvironment(EnvironmentContext context, Path dependencyPath, String dependencyName) {
```
hohwille, line 321: Variable `context` collides with existing member variable `context` for `IdeContext`.

```suggestion
  public void setEnvironment(EnvironmentContext envContext, Path dependencyPath, String dependencyName) {
```

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -323,127 +318,66 @@ private ToolInstallation createToolInstallation(Path rootDir, VersionIdentifier
...
 
-      String dependencyPath = this.dependenciesEnvVariablePaths.get(key);
-      pc = pc.withEnvVar(key, dependencyPath);
-    }
-
-    pc.run(processMode);
+    String pathVariable = dependencyName.toUpperCase(Locale.ROOT) + "_HOME";
+    context.withEnvVar(pathVariable, dependencyPath.toString());
```
hohwille, line 324: IMHO you are missing the `null` check here since by default `null` is passed as `EnvironmentContext` but maybe it does not need to go here but in all places calling this method.

cli/src/test/java/com/devonfw/tools/ide/tool/tomcat/TomcatTest.java
```diff
@@ -16,36 +16,32 @@ public void testTomcat() {
...
-    checkRunningTomcat(context);
+    //    checkRunningTomcat(context);
   }
 
   private void checkDependencyInstallation(IdeTestContext context) {
 
-    assertThat(context).logAtInfo().hasEntries("The version 17.0.10_7 of the dependency java is being installed",
-        "The version 17.0.10_7 of the dependency java was successfully installed");
+    assertThat(context).logAtInfo().hasMessage("The version 17.0.10_7 of the dependency java is being installed");
+    assertThat(context).logAtInfo().hasMessage("The version 17.0.10_7 of the dependency java was successfully installed");
```
hohwille, line 31: Was this changed by accident? Before, I was checking also the order of the two log-statements. Why should that be wrong?

hohwille (commented):
@aBega2000 as requested I had a look at your PR. It looks already good to me and you are perfectly on the right track 👍 
Your changes to use the model with `UrlDependencyFile` are fully correct, and the way you integrate the environment variable settings is also looking good. I left some review comments for further directions and places to adopt. Please have a look.

hohwille (changes requested):
@aBega2000 thanks for your patient work on this complex topic.
You have already gathered very great improvements in this PR and addressed most of the aspects of the story #439.
I left some review comments for rework.
Please, have a look and consider some improvements.
You can also then take this PR out of draft mode.
IMHO we can approve if all review comments are addressed.
Feel free to get in touch if some review comments are unclear for you.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -323,127 +341,67 @@ private ToolInstallation createToolInstallation(Path rootDir, VersionIdentifier
...
-    if (Files.exists(this.dependency.getDependencyJsonPath(getConfiguredEdition()))) {
-      setDependencyRepository(getInstalledVersion());
-    } else {
-      this.context.trace("No Dependencies file found");
-    }
-
-    ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.WARNING).executable(binaryPath).addArgs(args);
-
-    for (String key : this.dependenciesEnvVariablePaths.keySet()) {
+  public void setEnvironment(EnvironmentContext envContext, Path dependencyPath, String dependencyName) {
```
jan-vcapgemini, line 344: Please add Javadoc.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -96,18 +98,36 @@ public void run() {
...
+
     if (toolVersion == null) {
       install(true);
-      binaryPath = toolPath;
     } else {
       throw new UnsupportedOperationException("Not yet implemented!");
     }
-    ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.WARNING).executable(binaryPath).addArgs(args);
 
+    ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.WARNING).executable(binaryPath).addArgs(args);
```
jan-vcapgemini, line 130: Should this be THROW too?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -175,6 +200,11 @@ public boolean install() {
    * @param silent - {@code true} if called recursively to suppress verbose logging, {@code false} otherwise.
    * @return {@code true} if the tool was newly installed, {@code false} if the tool was already installed before and nothing has changed.
    */
+  public boolean install(EnvironmentContext environmentContext, boolean silent) {
```
jan-vcapgemini, line 203: Please adjust Javadoc.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -160,10 +180,15 @@ public VersionIdentifier getConfiguredVersion() {
   }
 
   /**
-   * Method to be called for {@link #install(boolean)} from dependent {@link com.devonfw.tools.ide.commandlet.Commandlet}s.
+   * Method to be called for {@link #install(EnvironmentContext, boolean)} from dependent {@link com.devonfw.tools.ide.commandlet.Commandlet}s.
    *
    * @return {@code true} if the tool was newly installed, {@code false} if the tool was already installed before and nothing has changed.
    */
+  public boolean install(EnvironmentContext environmentContext) {
```
jan-vcapgemini, line 187: Please adjust Javadoc.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -175,6 +200,11 @@ public boolean install() {
    * @param silent - {@code true} if called recursively to suppress verbose logging, {@code false} otherwise.
    * @return {@code true} if the tool was newly installed, {@code false} if the tool was already installed before and nothing has changed.
    */
+  public boolean install(EnvironmentContext environmentContext, boolean silent) {
+
+    return doInstall(environmentContext, silent);
+  }
+
   public boolean install(boolean silent) {
```
jan-vcapgemini, line 208: Please add Javadoc.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -186,6 +216,9 @@ public boolean install(boolean silent) {
    * @param silent - {@code true} if called recursively to suppress verbose logging, {@code false} otherwise.
    * @return {@code true} if the tool was newly installed, {@code false} if the tool was already installed before and nothing has changed.
    */
+
+  protected abstract boolean doInstall(EnvironmentContext environmentContext, boolean silent);
```
jan-vcapgemini, line 220: Please adjust Javadoc.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -186,6 +216,9 @@ public boolean install(boolean silent) {
...
+
   protected abstract boolean doInstall(boolean silent);
```
jan-vcapgemini, line 222: Please add Javadoc.

jan-vcapgemini (changes requested):
The refactoring looks really nice, especially the runTool refactor. I've added quite some CR's for missing Javadocs only. Please check your IDE for yellow/orange hints on the right side of the code window to prevent this.

cli/src/main/java/com/devonfw/tools/ide/process/EnvironmentContext.java
```diff
@@ -1,5 +1,8 @@
 package com.devonfw.tools.ide.process;
 
+/**
+ * Interface for the environment context relatable in the case a tool need to be run with an environment variable
```
jan-vcapgemini, line 4:
```suggestion
 * Interface for the environment context relatable in the case a tool needs to be run with an environment variable.
```

cli/src/main/java/com/devonfw/tools/ide/url/model/file/UrlDependencyFile.java
```diff
@@ -32,11 +36,31 @@ public UrlDependencyFile(UrlEdition parent) {
...
 
+  /**
+   * @return the content of the dependency map of the dependency.json file
+   */
   public Map<VersionRange, List<DependencyInfo>> getDependencyMap() {
 
     return this.dependencyMap;
   }
 
+  public List<DependencyInfo> findDependenciesFromJson(Map<VersionRange, List<DependencyInfo>> dependencyMap, VersionIdentifier toolVersionToCheck) {
```
jan-vcapgemini, line 47: Please add Javadoc.

cli/src/main/java/com/devonfw/tools/ide/url/model/file/UrlDependencyFile.java
```diff
@@ -32,11 +36,31 @@ public UrlDependencyFile(UrlEdition parent) {
...
+      VersionRange foundToolVersionRange = map.getKey();
+
+      if (foundToolVersionRange.contains(toolVersionToCheck)) {
+        return map.getValue();
+      }
+    }
+    return null;
+  }
+
+  public boolean isDependencyMapNull() {
```
jan-vcapgemini, line 60: Please add Javadoc.

jan-vcapgemini (changes requested):
Some small CR's missing and then it's ready to merge.

jan-vcapgemini (approved):
Thanks for adding the missing Javadocs. Ready for review.

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -167,7 +167,7 @@ private void updateSoftware() {
 
       // update/install the toolCommandlets
       for (ToolCommandlet toolCommandlet : toolCommandlets) {
-        toolCommandlet.install(false);
+        toolCommandlet.install(null, false);
```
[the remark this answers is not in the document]
hohwille: Done perfectly. Thanks!

cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
hohwille: Was reverted so can be resolved.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
hohwille: Was done. Thanks.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
hohwille: Now `setEnvironment` does the `null` check so all perfect and this can be resolved.

cli/src/main/java/com/devonfw/tools/ide/tool/CustomToolCommandlet.java
hohwille: Done perfectly. Thanks.

cli/src/main/java/com/devonfw/tools/ide/tool/Dependency.java
```diff
@@ -49,10 +44,17 @@ public Dependency(IdeContext context, String tool) {
    * @param toolEdition the edition of the tool.
    * @return the {@link Path} of the dependency.json file
    */
-  public Path getDependencyJsonPath(String toolEdition) {
+  public boolean existsDependencyJsonPath(String toolEdition) {
 
-    Path toolPath = this.context.getUrlsPath().resolve(this.tool).resolve(toolEdition);
-    return toolPath.resolve(DEPENDENCY_FILENAME);
+    return Files.exists(getDependencyUrlEdition(toolEdition).getDependencyFile().getPath());
```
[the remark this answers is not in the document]
hohwille: Was done in `UrlDependencyFile` and `Dependency.java` was removed. Fully resolved. Thanks.

cli/src/main/java/com/devonfw/tools/ide/tool/Dependency.java
```diff
@@ -49,10 +44,17 @@ public Dependency(IdeContext context, String tool) {
...
+  }
+
+  private UrlEdition getDependencyUrlEdition(String toolEdition) {
+
+    UrlRepository urlRepository = new UrlRepository(this.context.getUrlsPath());
+    UrlTool urlTool = new UrlTool(urlRepository, tool);
+    UrlEdition urlEdition = new UrlEdition(urlTool, toolEdition);
+    return urlEdition;
```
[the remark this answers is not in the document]
hohwille: Was removed. Now `this.context.getDefaultToolRepository()` is used what is perfectly correct.

cli/src/main/java/com/devonfw/tools/ide/tool/Dependency.java
```diff
@@ -121,9 +116,9 @@ public Path versionExistsInRepository(Path dependencyRepositoryPath, VersionRang
     return Path.of("");
   }
 
-  private List<DependencyInfo> findDependenciesFromJson(Map<VersionRange, List<DependencyInfo>> dependencies, VersionIdentifier toolVersionToCheck) {
+  private List<DependencyInfo> findDependenciesFromJson(Map<VersionRange, List<DependencyInfo>> dependencyMap, VersionIdentifier toolVersionToCheck) {
 
-    for (Map.Entry<VersionRange, List<DependencyInfo>> map : dependencies.entrySet()) {
+    for (Map.Entry<VersionRange, List<DependencyInfo>> map : dependencyMap.entrySet()) {
```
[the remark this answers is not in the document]
hohwille: Was moved to `UrlDependencyFile`. Thanks.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -94,23 +95,42 @@ public void run() {
...
     if (toolVersion == null) {
-      install(true);
-      binaryPath = toolPath;
+      if (existsEnvironmentContext) {
+        pc = this.context.newProcess().errorHandling(ProcessErrorHandling.THROW).executable(binaryPath).addArgs(args);
+        install(pc, true);
+      } else {
+        install(true);
+        pc = this.context.newProcess().errorHandling(ProcessErrorHandling.THROW).executable(binaryPath).addArgs(args);
+      }
```
hohwille, line 115: Could you avoid the code duplication?
```suggestion
      pc = this.context.newProcess().errorHandling(ProcessErrorHandling.THROW).executable(binaryPath).addArgs(args);
      if (existsEnvironmentContext) {
        install(pc, true);
      } else {
        install(true);
      }
```

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -96,18 +98,36 @@ public void run() {
...
+    Path binaryPath;
+    binaryPath = Path.of(getBinaryName());
+
+    if (existsEnvironmentContext) {
+      ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.WARNING).executable(binaryPath).addArgs(args);
+
+      if (toolVersion == null) {
+        install(pc, true);
+      } else {
+        throw new UnsupportedOperationException("Not yet implemented!");
```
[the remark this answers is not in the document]
hohwille: Even though marked as outdated, this is still present.

hohwille (approved):
@aBega2000 thank you very much for your work and improvements. You are reaching the finishing line. I resolved all previous review comments that you solved with your rework. There are only very few review remarks to address (currently 5). Great work 👍 

Could you have a look and get the last review comments implemented then we can merge this PR and finally close the story.

cli/src/test/java/com/devonfw/tools/ide/tool/tomcat/TomcatTest.java
aBega2000: This was a mistake by my side and is now changed back to hasEntries()

cli/src/main/java/com/devonfw/tools/ide/process/EnvironmentContext.java
```diff
@@ -0,0 +1,7 @@
+package com.devonfw.tools.ide.process;
+
+public interface EnvironmentContext {
+
+  EnvironmentContext withEnvVar(String key, String value);
```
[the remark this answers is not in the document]
aBega2000: Done.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -323,127 +341,68 @@ private ToolInstallation createToolInstallation(Path rootDir, VersionIdentifier
...
-      throw new UnsupportedOperationException("Not yet implemented!");
-    }
+  public void setEnvironment(EnvironmentContext envContext, Path dependencyPath, String dependencyName) {
 
-    if (Files.exists(this.dependency.getDependencyJsonPath(getConfiguredEdition()))) {
-      setDependencyRepository(getInstalledVersion());
-    } else {
-      this.context.trace("No Dependencies file found");
+    if (envContext != null) {
+      String pathVariable = dependencyName.toUpperCase(Locale.ROOT) + "_HOME";
```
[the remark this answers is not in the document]
aBega2000: Actually the reason is that the actual tool that comes from this.tool is Tomcat. Because the way the flow of the program works, is that when tomcat is being installed, first java gets installed, and the dependency java gets added as the environment context. But it all happens in the Commandlet Tomcat. So I can't change this like it is suggested because than it would be: TOMCAT_HOME

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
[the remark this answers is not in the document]
aBega2000: I removed the check toolVersion==null, and as far as I tested everything runs smoothly. I also tested gcviewer manually with this change and it seems to work. It gets installed and it also runs

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
aBega2000: The problem here is that if I do the suggested change, many tests of tools stop running. As far as I've debugged, this happens because the pc must be initialized after the install method call for these tools. But for tomcat ( at least until now it is only tomcat ) the pc must be initialized before the call because it should be passed as an argument to the method call, namely install(pc, true); So I couldn't find another solution for this problem, but to duplicate that line of code

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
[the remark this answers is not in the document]
hohwille: To me that means that it is implemented wrong then.
We must be able to override this method for a specific tool. E.g. assume for Java, we want to additionally set `JRE_HOME` because some other tool might for some reason not honor `JAVA_HOME´.
All this will not work then. You need to call this method on the correct tool.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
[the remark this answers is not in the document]
hohwille: My suggestion was to entirely remove the `toolVersion` if it is not supported at all and is not used anywhere.
Since we later might need to reintroduce it, I set this to resolved.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
hohwille: > The problem here is that if I do the suggested change, many tests of tools stop running.

So what was the error then? We need to understand the problems to come up with a clean design.

cli/src/test/java/com/devonfw/tools/ide/tool/tomcat/TomcatTest.java
hohwille: Maybe on your local machine but not yet on github here in this PR.

hohwille: Some for `checkRunningTomcat` that was changed alike and then even commented out.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
aBega2000: As discussed, I created a private method for this case, but it seems we have found a bug in the flow of the program.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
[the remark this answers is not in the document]
aBega2000: I fixed this but to do so, I had to cast the dependencyTool to LocalToolCommandlet in order to call the setEnvironment method. Please let me know if this is right, or perhaps should I keep it as ToolCommandlet and move the setEnvironment method to the ToolCommandlet class?

cli/src/test/java/com/devonfw/tools/ide/tool/tomcat/TomcatTest.java
aBega2000: Fixed.

hohwille (approved):
@aBega2000 thanks for your excellent and patient rework. You resolved all review comments and brought this to the finishing line. Excellent job 👍 
Ready for merge.

PR #521, opened 2024-08-06
#465: improve editorconfig

### LINKED ISSUE #465: use .editorconfig to configure the code-style
In devonfw-ide and IDEasy we use a complex and generic configurator that reads configuration templates, resolves variables and then can merge this into the according configuration files of an IDE tool in the workspace before launching the IDE.

This approach is very generic and allows us to auto-configure things that our code never knew about.
However, this approach needs quite some internal knowledge about how that IDE tool stores its internal configuration on the disc and how the configuration file is structured. Therefore average users of devonfw-ide and IDEasy find it quite hard to make adjustments to such settings if they want to tailor the code-style to their project needs and demands.
Finally, sometimes IDE tools like IntelliJ change their configuration structure over night with a new release (see e.g. #464) and thereby breaking our efforts supporting the projects with a fully automated IDE setup.

The idea of this story is to investigate how the file `.editorconfig` can be used as an alternative approach to config files specific for particular IDE tools. Details can be found at https://editorconfig.org/
This has the following benefits:
* `.editorconfig` is a de-facto standard supported by almost every IDE tool (IntelliJ, Eclipse, VsCode, etc.) so with one file, we could manage the code-style for all IDEs what makes it much easier and more transparent for our end-users who want to customize these settings.
* Further, projects can even copy&paste the `.editorconfig` into their code git repository and maintain it there so it will also apply to users not using IDEasy for whatever reason. Also this would allow to have multiple code git repositories inside one IDEasy project that have the need for different code-styles. This is a demand we can currently not address. For governance this may look odd but with microserivces this scenario gets more and more relevant and we might loose users if we force them to have an IDEasy project for each microservice.
* By default `.editorconfig` only supports very basic and common settings like indentation size (`indent_size`)
* However, there are even ways to support proprietary settings of an IDE like Intellij. This may sound in-official from the [editorconfig spec](https://editorconfig.org/) PoV but it is much more likely to be a "stable API" than our current approach to directly tweak the internal XML config files managed by IntelliJ directly. 

KISS approach would be to add support for putting an `.editorconfig` file into `ide-settings`.
If that is present the configurator would simply copy it into your top-level workspace (e.g. `workspaces/main`).
We can later implement a real merger (the file-format of `.editorconfig` more or less the one used by windows ini files and also used by git config) but for the moment a simple override would already do.
My assumption would be that if the IDE is launched with that workspace folder as root directory, it would honor the `.editorconfig`.
This should be the first thing to test.
If not, try if it works if the `.editorconfig` is present in the code git repository. If only that works, we could copy the file if not yet present or we could introduce a feature to `ProjectCommandlet` somehow for this.

hohwille:
fixes #465 

* Improve editorconfig by continuation indent of JavaDoc
* applied changes to JavaDoc 
* added some missing JavaDoc

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,8 @@
+root = true
+
 [*]
 charset = utf-8
-end_of_line = crlf
+end_of_line = lf
 indent_size = 2
 indent_style = space
 insert_final_newline = true
@@ -95,7 +97,7 @@ ij_java_doc_align_param_comments = false
 ij_java_doc_do_not_wrap_if_one_line = true
 ij_java_doc_enable_formatting = true
 ij_java_doc_enable_leading_asterisks = true
-ij_java_doc_indent_on_continuation = false
+ij_java_doc_indent_on_continuation = true
 ij_java_doc_keep_empty_lines = true
 ij_java_doc_keep_empty_parameter_tag = true
 ij_java_doc_keep_empty_return_tag = true
--- a/cli/src/main/java/com/devonfw/tools/ide/cli/CliArgument.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/cli/CliArgument.java
@@ -224,7 +224,7 @@ public class CliArgument {
 
   /**
    * @return a {@link String} representing all arguments from this {@link CliArgument} recursively along is {@link #getNext(boolean) next} arguments to the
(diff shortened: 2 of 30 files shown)


jan-vcapgemini (approved):
Looks good to me, the indent on continuation fits too.

Issue #463, opened 2024-07-09
Set default Maven m2 repository location back to user home


jan-vcapgemini:
As an IDEasy user I want to prevent unnecessary downloads of maven dependencies.

Currently if no m2 home was set an ominous folder gets created and the maven m2 repository gets created several times resulting in long waiting times and waste of HDD space.

We need to make sure that the old location e.g. user home/m2 is set by default.

PR #522, opened 2024-08-07
#463: Set default Maven m2 repository location back to user home

jan-vcapgemini:
Fixes: #463 

Implements:
* added fallback to USER_HOME for mvn repository
* added new testMavenRepositoryPathFallsBackToUserHome to MvnTest
* added environment.properties to mvn test resources (required for proper simulation of USER_HOME)

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
@@ -475,7 +475,7 @@ public interface IdeContext extends IdeLogger {
   }
 
   /**
-   * @return the String value for the variable M2_REPO, or null if called outside an IDEasy installation.
+   * @return the String value for the variable M2_REPO, or falls back to the default USER_HOME/.m2 location if called outside an IDEasy installation.
    */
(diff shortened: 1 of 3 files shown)


hohwille (approved):
@jan-vcapgemini thanks for your PR. I agree it makes absolutely sense to have `M2_REPO` always configured instead of using `null` as fallback. Thanks for the observation and fix 👍 

~~One last thing for `getMavenRepository()`:
I tried to delegate such things to the `Mvn` commandlet and to avoid redundancies.
IMHO it should be done like this:~~
```
  default Path getMavenRepository() {

    if (getIdeHome() != null) {
      Mvn mvn = getCommandletManager().getCommandlet(Mvn.class);
      Path mavenConfFolder = mvn.getMavenConfFolder(true);
      return mavenConfFolder.resolve("repository");
    }
    return null;
  }
```

~~Then your fix should go to that `Mvn.getMavenConfFolder` method as otherwise it is inconsistent. That shows again a good example why such redundancies are evil:
You fix something but are not aware that the same code block was copy&pasted somewhere else where you then miss to apply the fix as well...~~

PR #529, opened 2024-08-09
#479 android studio plugin support

### LINKED ISSUE #479: Add plugin installation support to AndroidStudio
As an IDEasy intellj user, I want to be able to install android-studio plugins.

Currently the `installPlugin` method is not implemented for AndroidStudio yet.

Needs to be implemented here:
https://github.com/devonfw/IDEasy/blob/5a5c2abf0702cb5fbdbc14d7a2be36fe8eb00529/cli/src/main/java/com/devonfw/tools/ide/tool/androidstudio/AndroidStudio.java#L90

Old implementation can be found here: https://github.com/devonfw/ide/blob/cdf4af5de21489f3fdcbfdab41af858822e0d598/scripts/src/main/resources/scripts/command/android-studio#L64-L96

Hint:

We can probably use the same implementation from #433
See also #505

slskiba:
Closes #479. 
Adds plugin support for Android Studio. 

Refactored `IntellijBasedPluginInstaller` to `IdeaBasedPluginInstaller` to be used with both Intellij and AndroidStudio. At this point I see no reason to keep separate implementations for Intellij and Android Studio PluginInstallers derived from the IdeaBasedPluginInstaller.


cli/src/main/java/com/devonfw/tools/ide/tool/androidstudio/AndroidStudio.java
```diff
@@ -66,6 +67,14 @@ protected void postInstall() {
...
+
+    IdeaBasedPluginInstaller pluginInstaller = new IdeaBasedPluginInstaller(context, this);
+    String downloadUrl = pluginInstaller.getDownloadUrl(plugin);
+    pluginInstaller.installPlugin(plugin, downloadUrl);
+  }
+
+  @Override
+  public String getMacToolApp() {
+    return "Android Studio Preview.app";
   }
```
jan-vcapgemini, line 79: Couldn't this be automated with the `MacOsHelper/findLinkDir` code?
https://github.com/devonfw/IDEasy/blob/752f41ed424a0a74920aee6ecfaa44329e0a9f80/cli/src/main/java/com/devonfw/tools/ide/os/MacOsHelper.java#L91-L104
f.e. creating a new method in `MacOsHelper` to retrieve the `*.app` name.

jan-vcapgemini (approved):
Looks good now. Ready for review.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeaBasedPluginInstaller.java
```diff
@@ -47,23 +45,14 @@ private String readBuildVersion() {
     Path buildFile = commandlet.getToolPath().resolve(BUILD_FILE);
     if (context.getSystemInfo().isMac()) {
       MacOsHelper macOsHelper = new MacOsHelper(context);
-      Path rootToolPath = macOsHelper.findRootToolPath(this.commandlet, context);
-      buildFile = rootToolPath.resolve("IntelliJ IDEA" + generateMacEditionString() + ".app").resolve("Contents/Resources").resolve(BUILD_FILE);
+      Path appPath = macOsHelper.findAppDir(macOsHelper.findRootToolPath(this.commandlet, context));
```
hohwille, line 48: Nice that we now do not hardcode the app name here anymore.
However, IMHO the method `findRootToolPath` was already buggy before and does not honor the edition (always uses tool name also as edition what is incorrect here).
Also `"default"` is hardcoded there as redundant string literal instead of reusing `ToolRepository.ID_DEFAULT` so could you also please improve that alongside?

PR #532, opened 2024-08-09
#497 bugfix: match specific installed version with * 

### LINKED ISSUE #497: Get-version does not match an installed version with configured version "*"
When calling get-version on an installed tool that is configured with any version ("`*`"), any installed version should match this configuration, and thus only the installed version is to be printed, without instructions to install the tool.

### Example:
For lazydocker with configured version ("`*`"), and installed version 0.23.3:
#### Expected behavior
```
ide> get-version lazydocker
0.23.3
```
#### Actual behavior
```
ide> get-version lazydocker
The installed version for tool lazydocker is 0.23.3
The configured version for tool lazydocker is *
To install that version call the following command:
ide install lazydocker
```

### Comments/Hints:
Introduced by https://github.com/devonfw/IDEasy/pull/455

slskiba:
Fixes #497 

- Fixes a bug where get-version would still print install instructions even if the tool is installed and no specific version is configured
- Improves tests for get-version and set-version


cli/src/test/resources/ide-projects/settings/readme
```diff
@@ -0,0 +1 @@
+this is the IDE_ROOT directory
```
jan-vcapgemini, line 1: This file can be removed. Only required if we want to read from it in a test. Please use `.gitkeep` instead if a folder is needed.

cli/src/test/resources/ide-projects/settings/project/workspaces/main/readme
```diff
@@ -0,0 +1 @@
+this is the main workspace of basic
```
jan-vcapgemini, line 1: This file can be removed. Only required if we want to read from it in a test. Please use `.gitkeep` instead if a folder is needed.

cli/src/test/resources/ide-projects/settings/project/software/mvn/bin/readme
```diff
@@ -0,0 +1 @@
+this is the mvn bin folder
```
jan-vcapgemini, line 1: This file can be removed. Only required if we want to read from it in a test. Please use `.gitkeep` instead if a folder is needed.

cli/src/test/resources/ide-projects/settings/project/readme
```diff
@@ -0,0 +1 @@
+this is the IDE_HOME directory
```
jan-vcapgemini, line 1: This file can be removed. Only required if we want to read from it in a test. Please use `.gitkeep` instead if a folder is needed.

cli/src/test/resources/ide-projects/settings/project/home/readme
```diff
@@ -0,0 +1 @@
+this is the users HOME directory
```
jan-vcapgemini, line 1: This file can be removed. Only required if we want to read from it in a test. Please use `.gitkeep` instead if a folder is needed.

cli/src/test/resources/ide-projects/settings/project/home/Downloads/ide/readme
```diff
@@ -0,0 +1 @@
+this is the download cache
```
jan-vcapgemini, line 1: This file can be removed. Only required if we want to read from it in a test. Please use `.gitkeep` instead if a folder is needed.

cli/src/test/resources/ide-projects/settings/project/software/mvn/bin/mvn
jan-vcapgemini, line 1: Empty file needed?

jan-vcapgemini (changes requested):
Thanks for this bugfix. I've added some small CR's (mostly obsolete test files)

jan-vcapgemini (approved):
Nice addition to the project readme. Ready for review.

hohwille (approved):
@slskiba thanks for the actual fix and also all the effort to improve the tests and create a more dedicated test project for that to make the tests more maintainable. Great work 👍
Ready for merge.

Issue #398, opened 2024-06-18
Add help details for all existing commandlets


hohwille:
Story #320 was implemented with PR #389 but only the technical part.
All help details are now configured as TODO markers.
In order to make progress and allow the feature to be used in new commandlets already, we decided to merge with the TODOs and create this story to address the TODOs stepwise with PRs. This way multiple developers of our team can provide PRs that address the details for those commandelts they have knowledge about and they can therefore provide quickly.
The commandlets are:
* [x] cmd.--version.detail
* [x] cmd.aws.detail
* [x] cmd.az.detail
* [ ] ~~cmd.cobigen.detail~~
* [x] cmd.complete.detail
* [x] cmd.create.detail
* [x] cmd.docker.detail
* [x] cmd.dotnet.detail
* [x] cmd.eclipse.detail
* [x] cmd.env.detail
* [x] cmd.gcviewer.detail
* [x] cmd.get-edition.detail
* [x] cmd.get-version.detail
* [x] cmd.gh.detail
* [x] cmd.graalvm.detail
* [x] cmd.gradle.detail
* [x] cmd.helm.detail
* [x] cmd.help.detail
* [x] cmd.install.detail
* [x] cmd.intellij.detail
* [x] cmd.jasypt.detail
* [x] cmd.java.detail
* [x] cmd.jmc.detail
* [x] cmd.kotlinc.detail
* [x] cmd.kotlincnative.detail
* [x] cmd.list-editions.detail
* [x] cmd.list-versions.detail
* [x] cmd.mvn.detail
* [x] cmd.node.detail
* [x] cmd.npm.detail
* [x] cmd.oc.detail
* [x] cmd.quarkus.detail
* [x] cmd.repository.detail
* [x] cmd.set-edition.detail
* [x] cmd.set-version.detail
* [x] cmd.shell.detail
* [x] cmd.sonar.detail
* [x] cmd.terraform.detail
* [x] cmd.uninstall.detail
* [x] cmd.update.detail
* [x] cmd.vscode.detail

PR #535, opened 2024-08-13
#398: added more help details in en and de

jan-vcapgemini:
Addresses: #398 

Implements: 
* added java help details
* added version help details
* added complete help details
* added create help details
* added env help details
* added list-editions help details
* added list-versions help details
* added shell help details
* added uninstall help details
* added update help details
* added repostory help details


cli/src/main/resources/nls/Help.properties
```diff
@@ -10,18 +10,18 @@ cmd.build=Runs a build job with given arguments.
...
 cmd.build.val.args=Build arguments to be used when running a build job.
 cmd.complete=Internal commandlet for bash auto-completion.
-cmd.complete.detail=TODO complete
+cmd.complete.detail=Run 'ide complete <expression>' to activate the non-interactive autocompletion, replace <expression> with the expression you want to autocomplete.\nF.e type: 'ide complete in' to get 'install' and 'intellij' suggestions.
```
hohwille, line 13:
```
Usage: ide [option]* complete [<args>*]
Internal commandlet for bash auto-completion.
```
Why introducing `<expression>` instead of writing `<args>`?

Official British shortcut of "for example" is IMHO "e.g.". If you use in-official forms still terminate with dot "f.e."
https://dictionary.cambridge.org/dictionary/english/eg

cli/src/main/resources/nls/Help.properties
```diff
@@ -10,18 +10,18 @@ cmd.build=Runs a build job with given arguments.
...
 cmd.create=Create a new IDEasy project.
-cmd.create.detail=TODO create
+cmd.create.detail=Run 'ide create <project_name>' to create a new IDEasy project in the projects directory, replace <project_name> with the name of your project (it will become the directory name).
```
hohwille, line 15: DRY?
```
Usage: ide [option]* create <project> [<settingsRepository>]
Create a new IDEasy project.
```

cli/src/main/resources/nls/Help.properties
```diff
@@ -63,9 +63,9 @@ cmd.kotlincnative.detail=Kotlin/Native Compiler (kotlincnative) compiles Kotlin
 cmd.lazydocker=Tool commandlet for LazyDocker.
 cmd.lazydocker.detail=Lazydocker is a simple terminal UI for both docker and docker-compose. Detailed documentation can be found at https://github.com/jesseduffield/lazydocker
 cmd.list-editions=List the available editions of the selected tool.
-cmd.list-editions.detail=TODO list-editions
+cmd.list-editions.detail=To list all available editions of f.e. 'intellij' simply type: 'ide list-editions intellij'. IDEasy will then list 'ultimate' and 'intellij' (community edition).
```
hohwille, line 66: Never redundantly put such information into localized help messages. If we would add a new edition, to `intellij`, we would very likely forget to update this.
I very much like your idea to add web-links to further details like you have done for Java.
It would be lovely if we would do that for all our tools because this could really be helpful for the end-user.
In Git-Bash / WT the user can control-click the link to open in the browser...

cli/src/main/resources/nls/Help.properties
```diff
@@ -79,7 +79,7 @@ cmd.pgadmin.detail=pgAdmin is an open source management tool for PostgreSQL and
 cmd.quarkus=Tool commandlet for Quarkus (framework for cloud-native apps).
 cmd.quarkus.detail=Quarkus is a Kubernetes-native Java framework for building cloud-native applications. Detailed documentation can be found at https://quarkus.io/
 cmd.repository=Setup pre-configured git repositories (clone, build, import).
-cmd.repository.detail=TODO repository
+cmd.repository.detail=This will set up all pre-configured repositories. Simply call 'ide repository setup <your_project_name>', replace <your_project_name> with the name of your project config file located in 'settings/repository/your_project_name' and IDEasy will automatically clone, build and set up your project based on the existing property file.\nIf you omit the project name, all projects found in the repository directory will be pre-configured.
```
hohwille, line 82:
```
Usage: ide [option]* repository setup [<repository>]
Setup pre-configured git repositories (clone, build, import).
```
```suggestion
cmd.repository.detail=Without further arguments this will set up all pre-configured git repositories.\nAlso, you can provide an explicit git repo as `<repository>` argument and IDEasy will automatically clone, build and set up your project based on the existing property file.\nRepositories are configured in 'settings/repository/<repository>.properties' and can therefore be shared with your project team for automatic or optional setup.
```

cli/src/main/resources/nls/Help.properties
```diff
@@ -89,7 +89,7 @@ cmd.set-version.detail=This will set the according tool version variable in your
 cmd.set-version.opt.--cfg=Selection of the configuration file (settings | home | conf | workspace).
 cmd.set-version.val.version=The tool version to set.
 cmd.shell=Commandlet to start built-in shell with advanced auto-completion.
-cmd.shell.detail=TODO shell
+cmd.shell.detail=The interactive shell offers console users a new user experience with an advanced auto-completion. Most inputs can be autocompleted in this mode.\nIDEasy will offer several matching commands e.g. 'install' or options e.g. '-t' by simply hitting the 'Tab' key after the expression f.e. 'in' for 'install' and 'intellij'. You can use the Tab key to switch between both suggestions.\nTo exit this mode use 'CTRL+C', 'CTRL+D' or type 'exit'.
```
hohwille, line 92: Great explanation!

```
IDEasy will offer several matching commands e.g. 'install' or options e.g. '-t' by simply hitting the 'Tab' key after the expression f.e. 'in' for 'install' and 'intellij'. You can use the Tab key to switch between both suggestions.
```

Isn't that the same in git-bash completion?
Maybe here we should rather describe the unique features added on top like cursor navigation within the suggested options, etc.

hohwille (approved):
@jan-vcapgemini Thanks for your PR and providing the missing help texts. 👍 
I left some comments for improvement. We can even create a new issue (e.g. for adding URLs to all tools). I just want to tackle that we continuously question our status quo and consider how we can make IDEasy better every day...

cli/src/main/resources/nls/Help.properties
jan-vcapgemini: I've adjusted the explanation and added a git bash and jline3 reference to it now.

cli/src/main/resources/nls/Help.properties
jan-vcapgemini: I've removed the specific information about the intellij versions now.

cli/src/main/resources/nls/Help.properties
jan-vcapgemini: I've removed the repetition of the project_name now.

hohwille: I do not see this update.
Maybe instead of 
```
Run 'ide create <project_name> <settings_repository>' to create a new IDEasy project in the projects directory using your custom project settings...
```
just write
```
Create a new IDEasy project in the projects directory (IDE_ROOT) using your custom project settings...
```

Ah, and by the way:
> You can also omit the <settings_repository> to use the default project settings.

If you want to have the default settings, you need to provide `-` as `<settings_repository>`.
If you omit it, you will get an interactive question for the git settings URL.

jan-vcapgemini: I've adjusted it now as you've suggested.

PR #538, opened 2024-08-14
#37: Tool commandlet for visual studio code

### LINKED ISSUE #37: Implement ToolCommandlet for Visual Studio Code
Create `ToolCommandlet` for Visual Studio Code
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/vscode

ATTENTION: Blocked by issue #10 that needs to be implemented first.

Commandlet has already been added with PR #84 but is incomplete. This story may only be closed if also adding/removing plugings, configuring workspace, etc. is all implemented.

aBega2000:
Fixes #37


cli/src/test/java/com/devonfw/tools/ide/tool/vscode/VscodeTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+    // arrange
+    IdeTestContext context = newContext(PROJECT_VSCODE);
+    Vscode vscodeCommandlet = new Vscode(context);
+
+    // install
+    vscodeCommandlet.install();
+
+    // run
+    vscodeCommandlet.run();
```
slskiba, line 24: I would assume the `run` call is not necessary for the installation to complete successfully, please remove

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -28,4 +32,14 @@ protected String getBinaryName() {
...
   }
 
+  @Override
+  public void installPlugin(PluginDescriptor plugin) {
+    Path binaryPath = Path.of(getBinaryName());
+    ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.WARNING).executable(binaryPath);
+    pc.addArg("--install-extension");
+    pc.addArg(plugin.getId());
+    pc.run();
+  }
```
slskiba, line 42: Since you added plugin support for VSCode, it would be really nice to have a test covering this.

slskiba (changes requested):

hohwille (approved):
@aBega2000 thanks for your PR. 👍 
I have nothing to add to what @slskiba already left as review comments. Please address these review comments. Then we can resolve and merge.

cli/src/test/java/com/devonfw/tools/ide/tool/vscode/VscodeTest.java
```diff
@@ -34,5 +33,14 @@ private void checkInstallation(IdeTestContext context) {
 
     assertThat(context.getSoftwarePath().resolve("vscode/.ide.software.version")).exists().hasContent("1.92.1");
     assertThat(context).logAtSuccess().hasMessage("Successfully installed vscode in version 1.92.1");
+
+    Path executablePath = context.getSoftwareRepositoryPath().resolve("default/vscode/vscode/1.92.1/bin/");
+    if (context.getSystemInfo().isWindows()) {
+      executablePath = executablePath.resolve("code.cmd");
+    } else {
+      executablePath = executablePath.resolve("code");
+    }
```
hohwille: Isn't this an internal implementation detail? From an end-user perspective it is important that after installation the tool is present in `software` and can be executed and works.
We just changed some tools to not be installed via software repository. In such case assertions like this will break.
It is OK if that breaks and you get aware of such side-effect of your change but still I would not consider this as a best practice to add such assertions to all the tests of local tool commandlets. If we write a generic test the software repository or the `InstallCommandlet` such assertions do make sense but here I would not really recommand it.
Further, I do not understand why you distinguish by operating system here if in the lines above (31,32) you do not. There will always be `code.cmd` and `code` in the tests no matter what OS is tested. If you want to make your test OS dependent you need to do that consistently but here I do not really see the need. Maybe if you can add the MacOS app structure of VSCode this would be beneficial but that would require some more rework of the test and its test-data (test-project).

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -28,4 +32,14 @@ protected String getBinaryName() {
     return "code";
   }
 
+  @Override
+  public void installPlugin(PluginDescriptor plugin) {
```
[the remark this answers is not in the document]
hohwille: Also the active plugin installation is implemented differently:
In devonfw-ide we collect all active and all inactive plugins and then do one call of VSCode (`code`) for each list.
You trigger an invocation of VScode for each active plugin.
Assuming you have 10 active plugins configured this will launch VScode 10 times instead of only once what makes quite some difference in performance.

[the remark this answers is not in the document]
aBega2000: I have changed the implementation completely. I would like to know if this is the correct way to go about this issue, so that I can keep working on the tests, because the tests are still missing.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -28,4 +42,127 @@ protected String getBinaryName() {
...
+          this.context.error("Error copying directories: " + e);
+        }
+      } else {
+        try {
+          Files.createDirectories(vsCodeConf);
+        } catch (IOException e) {
+          this.context.error("Failed to create directories for vscode configuration: " + e);
+        }
+      }
+    }
```
hohwille, line 132: Correctly migrated. However, a very old legacy feature from devonfw-ide that we could also consider to remove.
Otherwise, I would also have review feedback like that you should use `FileAccess` instead of crafting your own `copyDirectory` method and you should also use `FileAccess.mkdirs`.
IMHO the easiest approach is to remove this block and the `copyDirectory` method.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -28,4 +42,127 @@ protected String getBinaryName() {
...
+
+    Path vsCodeExtensionFolder = this.context.getIdeHome().resolve("plugins/vscode");
+    try {
+      Files.createDirectories(vsCodeExtensionFolder);
+    } catch (IOException e) {
+      this.context.error("Failed to create directories for vscode extensions: " + e);
+    }
```
hohwille, line 139: use `FileAccess.mkdirs` instead and avoid using lowlevel Java IO directly.
Also the creation of this folder should IMHO be moved to `PluginBasedCommandlet` before new `installPlugins` method is called (line 155).
This will be always called on installation so the directory would already exist here.
Centralizing this makes sense since all plugin based tools otherwise need to implement this again and again so otherwise we only have it once in a central place.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -28,4 +42,127 @@ protected String getBinaryName() {
...
+
+    Path binaryPath;
+    binaryPath = Path.of(getBinaryName());
+
+    List<String> command = new ArrayList<>();
+    command.add("--new-window");
+    command.add("--user-data-dir=" + vsCodeConf);
+    command.add("--extensions-dir=" + vsCodeExtensionFolder);
+
+    String[] extensions = this.arguments.asArray();
```
hohwille, line 149: Why is this variable called `extensions`? This is very confusing to me since this is any argument that can also be passed from the end user via `ide vscode arg1 arg2`. You are currently implementing a feature about plugins/extensions but here you need to abstract from this since the `runTool` method is used for any situation where we start visual studio code.

However, there are further concerns here:
* Why converting a List to an Array if then you again convert it to a List again? That does not seem to make sense.
* Why are you using `this.arguments` here at all? The arguments for this `runTool` method are given as parameter `args`. There is a `runTool` method without arguments that will take the `this.arguments` and supply it to the `args` parameter of this method. But there are other cases like here where you need to install extensions. There the idea is that you can provide the arguments directly to the `args` parameter instead of modifying the state of the commandlet and changing the value of `this.arguments`. I will add a comment to the location where you are doing this to make it clear.

For the future we should even think how we can better extend additional features to the `runTool` method since you are completely overriding it copy&pasting quite some code from `LocalToolCommandlet.runTool` method. However, I will create a new story for this. It has nothing to do with your PR and your solution can be merged and that new story can be implemented later as a cleanup/improvement.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -28,4 +42,127 @@ protected String getBinaryName() {
...
+  }
+
+  private void doInstallPlugins(List<PluginDescriptor> pluginsToInstall) {
+    if (pluginsToInstall.isEmpty()) {
+      this.context.info("No plugins to be installed");
+    } else {
+      for (PluginDescriptor plugin : pluginsToInstall) {
+        this.arguments.addValue("--install-extension");
+        this.arguments.addValue(plugin.getId());
+      }
```
hohwille, line 75: You need to call `runTool` here with the arguments passed as `args` parameter to that method.
Otherwise you are causing a bug here:
* Assume the user will call `ide vscode arg1 arg2`.
* The `Vscode` commandlet gets `arg1` and `arg2` assigned to `this.arguments` property.
* Now if VSCode is not yet installed it will be installed and during that this method is called to install the plugins.
* Here you will add `--install-extensions`, etc. to `this.arguments` after `arg1` and `arg2`.

Instead we want to call `code` here via `runTool` to install the plugins now. If that fails, the installation fails and the actual run of `VSCode` intended by the end-user shall not even be triggered.
I hope it gets clear now.

hohwille (changes requested):
@aBega2000 Thanks for your rework and improvements. 👍 
Still it turns out some details are not yet clear and your implementation needs further rework.
I left several review comments for you to address.
As I cannot put a review comment on an empty file:
Please try to avoid inconsistent test-data except for cases where you explicitly want to test how the software behaves in case or corrupt data.
A Empty file is not a valid JSON file.
Simply put `{}` into `extensions.json` and you are safe.
Your current "hack" code migrated from bash would also in this case insert `"extensions": "plugin1,plugin". This may let your JUnit pass but this is still invalid JSON and the real VSCode application would complain about it.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
hohwille: Done by @aBega2000 since a while now. Hence. I am resolving this.

cli/src/test/java/com/devonfw/tools/ide/tool/vscode/VscodeTest.java
hohwille: Thanks for your update. Resolved.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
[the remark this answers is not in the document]
hohwille: Thanks for your update. See my new review. This is now resolved.

cli/src/main/java/com/devonfw/tools/ide/tool/plugin/PluginBasedCommandlet.java
```diff
@@ -151,19 +153,31 @@ protected boolean doInstall(boolean silent) {
...
+      Collection<PluginDescriptor> plugins = pluginMaps.getPlugins();
+      installPlugins(plugins);
     }
     return newlyInstalled;
   }
 
+  private void createExtensionFolder() {
+    Path extensionFolder = this.context.getIdeHome().resolve("plugins").resolve(this.tool);
+    FileAccess fileAccess = new FileAccessImpl(this.context);
+    fileAccess.mkdirs(extensionFolder);
```
hohwille, line 168:
```suggestion
    this.context.getFileAccess().mkdirs(extensionFolder);
```

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -28,4 +45,95 @@ protected String getBinaryName() {
...
+  private void doAddRecommendations(List<PluginDescriptor> recommendations) {
+    Path extensionsJsonPath = this.context.getWorkspacePath().resolve(".vscode/extensions.json");
+
+    ObjectMapper objectMapper = new ObjectMapper();
+    Map<String, Object> recommendationsMap;
+
+    try {
+      if (Files.exists(extensionsJsonPath) && Files.size(extensionsJsonPath) > 0) {
+        String content = Files.readString(extensionsJsonPath, StandardCharsets.UTF_8);
+        recommendationsMap = objectMapper.readValue(content, new TypeReference<Map<String, Object>>() {
```
hohwille, line 92: Use streaming instead of reading the entire `String` for efficiency:
```java
  try (BufferedReader reader = Files.newBufferedReader(extensionsJsonPath)) {
    recommendationsMap = objectMapper.readValue(reader, Map.class);
  }
```

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -28,4 +42,127 @@ protected String getBinaryName() {
...
+          Files.writeString(extensionsJsonPath, content, StandardCharsets.UTF_8);
+          this.context.success("Plugin recommendations have been updated in " + extensionsJsonPath);
+        } else {
+          this.context.success("All plugin recommendations are up to date.");
+        }
+
+      } else {
+        Files.writeString(extensionsJsonPath, "{" + recommendationsKey + "}", StandardCharsets.UTF_8);
+        this.context.success("Plugin recommendations have been created in " + extensionsJsonPath);
+      }
```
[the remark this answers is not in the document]
hohwille: Well done. 👍 Setting to resolved.

hohwille (approved):
@aBega2000 thanks for your rework. Looks very good already. 👍 

One review comment was still not completed and I added some small hints for final improvement. We can merge when these are resolved.

cli/src/main/java/com/devonfw/tools/ide/tool/plugin/PluginBasedCommandlet.java
```diff
@@ -151,19 +153,31 @@ protected boolean doInstall(boolean silent) {
...
+  }
+
+  protected void installPlugins(Collection<PluginDescriptor> plugins) {
```
jan-vcapgemini, line 171: Please add Javadoc.

cli/src/test/java/com/devonfw/tools/ide/tool/vscode/VscodeTest.java
```diff
@@ -0,0 +1,35 @@
...
+import com.devonfw.tools.ide.context.AbstractIdeContextTest;
+import com.devonfw.tools.ide.context.IdeTestContext;
+
+public class VscodeTest extends AbstractIdeContextTest {
+
+  private static final String PROJECT_VSCODE = "vscode";
+
+
+  @Test
+  public void testVscode() {
```
jan-vcapgemini, line 14:
```suggestion
  public void testVscodeInstall() {
```

jan-vcapgemini (changes requested):
Looks good so far. I've added some CRs to resolve.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
hohwille: @aBega2000 can you please fix this review comment? We already discussed that you missed this one and it is trivial to fix but currently this is causing a severe bug.

hohwille: Thanks for your update. Well done 👍

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -86,12 +85,16 @@ private void doAddRecommendations(List<PluginDescriptor> recommendations) {
     ObjectMapper objectMapper = new ObjectMapper();
     Map<String, Object> recommendationsMap;
 
+    try (BufferedReader reader = Files.newBufferedReader(extensionsJsonPath)) {
+      recommendationsMap = objectMapper.readValue(reader, Map.class);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
```
hohwille, line 96: You still need to check if the file exists as otherwise this will fail.
```suggestion
  if (Files.exists(extensionsJsonPath)) {
    try (BufferedReader reader = Files.newBufferedReader(extensionsJsonPath)) {
        recommendationsMap = objectMapper.readValue(reader, Map.class);
      } catch (IOException e) {
        throw new RuntimeException(e);
      }
    } else {
      recommendationsMap = new HashMap<>();
    }
```

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -100,13 +103,15 @@ private void doAddRecommendations(List<PluginDescriptor> recommendations) {
...
+      } else if (existingRecommendations.size() > recommendations.size()) {
         List<String> newRecommendations = new ArrayList<>();
         recommendationsMap = new HashMap<>();
         for (PluginDescriptor recommendation : recommendations) {
           newRecommendations.add(recommendation.getId());
         }
         recommendationsMap.put("recommendations", newRecommendations);
+      } else {
+        this.context.info("All recommendations are already set");
       }
```
hohwille, line 119: I do not see how these `if` Statements make sense. My suggestion:
```suggestion
     Set<String> existingRecommendationsSet = new HashSet<>(existingRecommendations);
     for (PluginDescriptor recommendation : recommendations) {
       String recommendationId = recommendation.getId();
       if (existingRecommendationsSet.add(recommendationId)) {
         existingRecommendations.add(recommendationId);
       }
     }
```

BTW: You can also count the added recommendations and if zero, skip the `objectMapper.writeValue` operation :)

slskiba:
CHANGELOG entry needs to be added for this PR still

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
jan-vcapgemini: I've Implemented your suggestion now.

jan-vcapgemini (approved):

CHANGELOG.adoc
```diff
@@ -7,7 +7,8 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 Release with new features and bugfixes:
 
 * https://github.com/devonfw/IDEasy/issues/628[#638]: Fixed update fails on first error
-* https://github.com/devonfw/IDEasy/issues/553[#554]: Missmatch of IDE_ROOT
+* https://github.com/devonfw/IDEasy/issues/37[#538]: Added Visual Studio Code (vscode) with plugin installation and plugin recommendation support
```
hohwille, line 10:
```suggestion
* https://github.com/devonfw/IDEasy/issues/37[#37]: Added Visual Studio Code (vscode) with plugin installation and plugin recommendation support
```

CHANGELOG.adoc
```diff
@@ -7,7 +7,8 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
...
+* https://github.com/devonfw/IDEasy/issues/553[#554]: Mismatch of IDE_ROOT
```
hohwille, line 11:
```suggestion
* https://github.com/devonfw/IDEasy/issues/553[#553]: Mismatch of IDE_ROOT
```

hohwille (approved):
@jan-vcapgemini thanks for taking over this PR and resolving the last review comments. 👍 

Still added some suggestions but will simply apply and then merge.

Issue #537, opened 2024-08-14
IDEasy repository setup command throws NPE


jan-vcapgemini:
### Expected behavior

As an IDEasy user I expect the repository commandlet to work properly when I call it with `ide repository setup` and omit the repository.

### Actual behavior
IDEasy throws a NullpointerException:

```
java.lang.NullPointerException: Cannot invoke "java.nio.file.Path.getFileSystem()" because "path" is null
	at java.base/java.nio.file.Files.provider(Files.java:104)
	at java.base/java.nio.file.Files.exists(Files.java:2514)
	at com.devonfw.tools.ide.property.PathProperty.validate(PathProperty.java:64)
	at com.devonfw.tools.ide.commandlet.Commandlet.validate(Commandlet.java:233)
	at com.devonfw.tools.ide.context.AbstractIdeContext.applyAndRun(AbstractIdeContext.java:912)
	at com.devonfw.tools.ide.context.AbstractIdeContext.run(AbstractIdeContext.java:864)
	at com.devonfw.tools.ide.cli.Ideasy.runOrThrow(Ideasy.java:88)
	at com.devonfw.tools.ide.cli.Ideasy.run(Ideasy.java:51)
	at com.devonfw.tools.ide.cli.Ideasy.main(Ideasy.java:28)

```
### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. run `ide repository setup`

### Affected version:

* IDEasy: 2024.08.001-beta-SNAPSHOT

PR #539, opened 2024-08-14
#537: fixed repository setup NPE

jan-vcapgemini:
Fixes: #537 

Implements:
* fixed NPE in PathProperty.validate method
* added a simple test for a missing but not required argument


cli/src/main/java/com/devonfw/tools/ide/property/PathProperty.java
```diff
@@ -60,8 +60,8 @@ public Path parse(String valueAsString, IdeContext context) {
   @Override
   public boolean validate() {
 
-    if (this.value != null) {
-      if (Files.exists(getValue())) {
+    if (!this.value.isEmpty()) {
+      if (getValue() != null && Files.exists(getValue())) {
```
hohwille, line 64: https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc#consider-extractig-local-variable-for-multiple-method-calls

hohwille (approved):
@jan-vcapgemini thanks for your PR. You properly fixed the bug and also before added a test to reproduce it. Great job 👍 
I left a nice-to-have improvement comment. Please have a look. We can then merge.

cli/src/main/java/com/devonfw/tools/ide/property/PathProperty.java
hohwille: Resolved this myself as constructive review. Also I validate all `Path` values in the `List` instead of just the first one to support multi-valued path property.

Issue #505, opened 2024-08-01
Refactor Jetbrains IDEs


jan-vcapgemini:
PR #480  was merged but this is not really completed:
- [x] Introduce an abstract class `IdeaBasedIdeToolCommandlet` between `IdeToolCommandlet` and its current implementations `Intellij` and `AndroidStudio`. There might come support for additional such IDEs to IDEasy in the future (PyCharm, WebStorm, etc.).
- [ ] Extract common functionalities to the new abstract base class. 
  - [x] IMHO plugin installation is the same for all IDEA/JetBrains IDEs except for some parameterization of the download URL and the IDE folder name in the plugins folder. UPDATE: This is more or less a duplicate of #479 - at least that would be the best way to resolve #479 instead of copy&paste approach.
  - [x] `IDEA_PROPERTIES` mechanism is the same for all IDEA/JetBrains IDEs only the name of the environment variable differs. The Variable value differs per workspace. Therefore setting it once in postInstall with a fixed workspace name is buggy since we write the resolved workspace name in there instead of `${WORKSPACE}`. We could better "export" this property via ProcessContext for each call of that IDE instead of writing to `ide.properties` files. ~~Also it seems an existing flaw that we use the same-filename for different IDE products: See [android-studio](https://github.com/devonfw/ide-settings/blob/main/android-studio/workspace/update/idea.properties) and [IntelliJ](https://github.com/devonfw/ide-settings/blob/main/intellij/workspace/update/idea.properties). As both files have different content, they also must have different filenames as otherwise one overrides the other.~~ - done with [this commit](https://github.com/devonfw/ide-settings/commit/8f6da8c195e2932976374913dd34ecdd344614df).
  - [ ] The `getBinaryName()` problem is the same for all IDEA/JetBrains IDEs. The current solution is still not nice. I would prefer a postInstall hook that creates a simple wrapper bash script if the default command is not already present. So create bash script `studio` that only calls `studio64.exe $*`, etc. UPDATE: This point is more or less a duplicate of #440 (but that issue is only about IntelliJ).

_Originally posted by @hohwille in https://github.com/devonfw/IDEasy/issues/339#issuecomment-2263000783_

PR #543, opened 2024-08-20
#519: install-plugin commandlet

### LINKED ISSUE #519: IdeToolCommandlet: add options and feature to install and uninstall plugins via CLI
As a IDEasy user, I want to be able to add or remove plugins to an IDE via CLI so that I can simplify or automate plugin installations.

In devonfw-ide we had such feature:
```
$ devon eclipse help
...
usage: devon eclipse [options] [args]

Manage Eclipse IDE and workspace.

Arguments:
 --all                           if provided as first arg then to command will be invoked for each workspace
 setup                           setup Eclipse (install or update)
 add-plugin «id» [«url»]         install an additional plugin
 run | start                     launch Eclipse IDE (default if no argument is given)
 ws-up[date]                     update eclipse workspace
 ws-re[verse]                    reverse merge changes from workspace into settings
 ws-reverse-add                  reverse merge adding new properties
 create-script                   create eclipse-main script if not already exists
 import «path» [«workingsets»]   import projects in «path» into current workspace. «workingsets» comma-separated list of workingsets (no spaces between ,), eg: ./myproject ws1,ws2
 -b | --batch         run in non-interactive mode (do not ask any questions).
 -q | --quiet         be quiet and avoid output.
 -d | --debug         be extra verbose and print more output for debugging.
 -f | --force         force proceeding even in case of warnings.
```

Please note that in IDEasy we so far only have proxy commands for tools so we create additional commandlets like `InstallCommandlet` or `VersionSetCommandlet` for tweaking additional things related to some tool instead of mixing this into the CLI of that according tool.

Therefore, it would be logical to create `InstallPluginCommandlet` (`ide install-plugin «tool» «plugin»`) as well as `UninstallPluginCommandlet`  (`ide uninstall-plugin «tool» «plugin»`) that both derive from an abstract `PluginCommandlet` taking an `PluginBasedCommandlet` as `«tool»` argument followed by a `«plugin»` descriptor.
The implementation should just delegate to according `installPlugin` or `uninstallPlugin` methods of `PluginBasedCommandlet`.

As always the story includes proper design so auto-completion is working as expected. Therefore a dedicated `PluginProperty` needs to be created (analogue to `RepositoryProperty`). Optionally instead of using `ToolProperty` a `PluginToolProperty` may be added so that only plugin based tools will be completed. (While writing this, I notice that auto-completion is still incomplete and missing e.g. in `RepositoryProperty` - so maybe we need to create another story for this and link it from here).

Feel free to create an extra issue for remove/uninstall plugin feature to keep PRs/stories smaller.
The actual plugin installation itself for concrete IDEs is handled by dedicated issues and not in scope of this story:
* #490 Intellij
* #479 Android-Studio
* #236 Eclipse
* #302 GraalVM

slskiba:
Closes #519
Also closes point 1 and 2.1 of #505  

- Adds `install-plugin` and `uninstall-plugin` commands with their respective functionalities and cli auto-completion; and tests.
- Refactoring: removed all `PluginInstaller` classes, and all plugin-related code from `IdeToolCommandlet`. Instead made `IdeToolCommandlet` extend `PluginBasedCommandlet`, and created `IdeaBasedIdeToolCommandlet` (currently used by Intellij and AndroidStudio) to handle the plugins. (I.e. 
  - before: `Intellij` >`IdeToolCommandlet` > `LocalToolCommandlet`
  - now: `Intellij` >`IdeaBasedIdeToolCommandlet` >`IdeToolCommandlet` > `PluginBasedCommandlet` >`LocalToolCommandlet` )
- Refactoring: moved plugin-related classes (that are also used by `mvn`) from both `tool` and `tool/ide` directories into their own package `tool/plugin`.


hohwille (approved):
@slskiba thanks for this PR. You make me very happy with this refactoring because this is what I wanted to have from the start but maybe it was too complex for some developers to understand and temporary lead to some copy&paste detours. Now we have a very nice design and do proper reuse without code-duplication for all this stuff. 🥇 

Some last remarks I noticed for potential later improvements:
This code was removed:
```
  public Path getPluginsInstallationPath() {

    // TODO add edition???
    return this.context.getPluginsPath().resolve(this.tool);
  }
```
The same still exists without the TODO in `PluginBasedCommandlet` but the `TODO` concern is still valid.

Classname `PluginMaps` seems confusing (also confirmed by the local variable you used for this ;) ). IMHO we should rename it to just `Plugins`. Also the getter is named `getPluginsMap()` and currently returns `PluginMaps` (notice the different places of the plural `s`). So that method should be renamed to `getPlugins()` and also should get some formal JavaDoc since it is now public.

All nice to have, so I will merge as-is now.

PR #546, opened 2024-08-20
Quick Fix: Check Java Installation Before Run GCViewer

alfeilex:
Add Java Installation check before running GCViewer.


hohwille (approved):
@alfeilex thanks for the lightning fast fix. 👍

PR #555, opened 2024-08-23
483: support multiple editions in update

### LINKED ISSUE #483: Refactor IntellijUrlUpdater to support multiple editions
Currently, the update method of `AbstractUrlUpdater` (and the derived method in `JsonUrlUpdater`) do not support multiple editions.

We should refactor `AbstractUrlUpdater` to support multiple editions in `update`. Probably set edition dynamically in `collectVersionsFromJson` in `IntellijUrlUpdater` on refactoring.

See: https://github.com/devonfw/IDEasy/pull/481#discussion_r1680909027

slskiba:
Closes #483 

- UrlUpdater base classes now support multiple editions
- The default implementation behaves as before: a single edition named after the tool
- Intellij therefore can be refactored, not having to implement its own update method anymore


cli/src/main/java/com/devonfw/tools/ide/url/updater/JsonUrlUpdater.java
```diff
@@ -100,11 +104,11 @@ protected void collectVersionsWithDownloadsFromJson(J jsonObj, UrlEdition editio
   /**
    * Gets the {@link JsonObject} from the response of the version URL.
    *
-   * @param response
+   * @param response String from the JSON API request
```
hohwille, line 107: Missing JavaDoc (you could also add that this method may now return null in case the edition is not present.
```suggestion
   * @param response String from the JSON API request
   * @param edition the edition data to get from the response.
```

hohwille (approved):
@slskiba thanks for your PR. You nicely added the ability to pass the edition as parameter instead of internally retrieving it from the getter. The way you add multiple editions as List without the need to refactor all other UrlUpdaters is smart and well designed. 👍 

The only little room for improvement that I can notice is that you are parsing the response as JSON for IntelliJ and Python multiple times only to get one item out of the JSON array. You could theoretically leave the `getJsonObjectFromResponse` without the `edition` and create a new parent container object for that and containing the editions. Then you could add a protected method `getJsonEdition(J rootObject, String edition)` to `JsonUrlUpdater` that gets the root JSON object and returns the actual edition JSON object from it for a given `edition` parameter. The you however, need 3 generic types: `J, JE, JVI` where `JE` is the JSON edition object. Other JSON UrlUpdaters would then declare `J = JE`. To avoid changes in other URL Updaters, you could extract `AbstractJsonUrlUpdater` that has these 3 generics while `JsonUrlUpdater` extends it and implement `getJsonEdition` to just return `rootObject`.
However, I do not see a problem with the current approach and we can also merge as is...

PR #558, opened 2024-08-26
helpful tweak for Win11

hohwille:
added registry files to enable or disable Windows Explorer context menu options vs. show more.
Source:
https://www.elevenforum.com/t/disable-show-more-options-context-menu-in-windows-11.1589/

Waiting before confirmation in chat below of the page before approval and merge.


hohwille:
@hohwille wrote
>Thank you for this helpful tweak. I would love to include this into an OSS product IDEasy licensed under ASL2.0 and therefore would like to kindly ask for your confirmation. I will leave your files as is including the headers with Author and Link to this page.
This way Windows users that install this product will already have the *.reg files on their disc and only need to merge them.

Brink (Administrator) responded:
> Hello, and welcome. :alien:
That'll be fine.

hohwille:
From review meeting and retest of this feature:
* `Enable_Show_more_options_context_menu.reg` immediately applies and brings the Win11 default back with `Show more options`
* `Disable_Show_more_options_context_menu.reg` does not immediately apply. When you right click on some file, you still have the default Win11 menu. You need to kill explorer.exe process and restart it with the task-manager. Then the change applies and you get the context menu with all the options immediately.

PR #559, opened 2024-08-26
#531: prevent logging from context constructor

### LINKED ISSUE #531: EnvironmentCommandlet can log warnings into the middle of environment variables
### Expected behavior

As a IDEasy user, I want to setup my project environment in any shell anywhere with just the command `ide` so that I can start working seamlessly from there.

### Actual behavior

Setting up the environment variables with `ide` commandlet can cause errors and even potentially security risks.
If variables cannot be evaluated, that should normally be logged as warning to notify the user about a problem.
However, the problem with `EnvironmentCommandlet` (`ide env`) is that here we are using the "log" as an "API" to the shell (bash or CMD) and send some kind of "code for execution".
Now, if some log message like `Undefined variable ${UNDEFINED} in 'SOME'` gets logged into this "code", that breaks the "API".
So most likely we have to control the log for this commandlet like we do in `CompleteCommandlet`.

### Steps to reproduce (bug)

1. add something like `FOO=${BAR} ${SOME} ${UNDEFINED}` to your `ide.properties`
2. run `ide` command in that project.

I then get errors like this in bash:
```
bash: $'\E[93mUndefined': command not found
bash: $'\E[93mUndefined': command not found
bash: $'\E[93mUndefined': command not found
```

### Comments/Hints:

See https://github.com/devonfw/IDEasy/blob/752f41ed424a0a74920aee6ecfaa44329e0a9f80/cli/src/main/java/com/devonfw/tools/ide/commandlet/CompleteCommandlet.java#L55

In [EnvironmentCommandletTest](https://github.com/devonfw/IDEasy/blob/main/cli/src/test/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandletTest.java) we also have a lot of redundancy now for windows and non-windows case in case of info-logging and also I had to comment out the assertion method like this: `log().hasEntries/*WithNothingElseInBetween*/(...)`.
To avoid the redundancies, we should manually build the expected `VariableLine` instances and then have an advanced assertion method that can take flags like `boolean quoteValue` and `boolean omitExport`.

### Affected version:

* OS: Windows (also Linux/Mac)
* IDEasy: 2024.08.001-beta-SNAPSHOT

hohwille:
fixes #531
* prevent logging from IdeContext constructor
* check if debug/trace logging was activated before disabling log-levels for processable output commandlets
* solve ContextCommandlet init issues


hohwille:
@hohwille wrote
>Thank you for this helpful tweak. I would love to include this into an OSS product IDEasy licensed under ASL2.0 and therefore would like to kindly ask for your confirmation. I will leave your files as is including the headers with Author and Link to this page.
This way Windows users that install this product will already have the *.reg files on their disc and only need to merge them.

Brink (Administrator) responded:
> Hello, and welcome. :alien:
That'll be fine.

cli/src/main/java/com/devonfw/tools/ide/context/IdeStartContext.java
```diff
@@ -0,0 +1,34 @@
+package com.devonfw.tools.ide.context;
+
+import java.util.Locale;
+
+import com.devonfw.tools.ide.log.IdeLogger;
+
+public interface IdeStartContext extends IdeLogger {
```
jan-vcapgemini, line 7: Please add Javadoc.

cli/src/main/java/com/devonfw/tools/ide/context/IdeStartContextmpl.java
```diff
@@ -0,0 +1,150 @@
...
+
+import com.devonfw.tools.ide.log.IdeLogLevel;
+import com.devonfw.tools.ide.log.IdeLogger;
+import com.devonfw.tools.ide.log.IdeSubLogger;
+import com.devonfw.tools.ide.log.IdeSubLoggerNone;
+
+/**
+ * Implementation of {@link IdeLogger}.
+ */
+public class IdeStartContextmpl implements IdeStartContext {
```
jan-vcapgemini, line 15: Please check Javadoc.

jan-vcapgemini (changes requested):
Thanks for your addition of the IdeStartContext. I've added some small CR's but I think overall this is ready to merge.

PR #561, opened 2024-08-26
#424: re-adding last relevant change

hohwille:
last relevant change from PR #424

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
@@ -930,20 +930,29 @@ public abstract class AbstractIdeContext implements IdeContext {
       Commandlet firstCandidate = this.commandletManager.getCommandletByFirstKeyword(keyword);
       boolean matches = false;
       if (firstCandidate != null) {
-        matches = apply(arguments.copy(), firstCandidate, collector);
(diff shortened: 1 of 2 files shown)


hohwille:
OK. Now at least the error has been isolated to the minimal diff.
Still confusing that locally all works correct.
Would be interesting to debug this on github somehow and see what goes wrong in this case.
Even though we create the context from the basic project we seem to end up with `ideHome` being `null`.
I am not getting how this is possible.
```java
  private boolean isIdeHome(Path dir) {

    if (!Files.isDirectory(dir.resolve("workspaces"))) {
      return false;
    } else if (!Files.isDirectory(dir.resolve("settings"))) {
      return false;
    }
    return true;
  }
```
this should be given for https://github.com/devonfw/IDEasy/tree/main/cli/src/test/resources/ide-projects/basic/project

hohwille:
OK, finally it gets obvious why this is happening. I was always looking in the wrong test.
There is `CompeleteTest` and `IdeCompleterTest`. The latter one contains the problem here:
https://github.com/devonfw/IDEasy/blob/44d5f0293d0f41d80508c65031cd17c59b4b8fda/cli/src/test/java/com/devonfw/tools/ide/completion/IdeCompleterTest.java#L131

hohwille:
Hurray, I was right and the build went green. Puzzle finally solved. Merging now.

Issue #251, opened 2024-03-18
Release workflows broken


hohwille:
### Expected behavior

As a developer of IDEasy, I want automated workflows that build releases so that I can download the latest SNAPSHOT or publish an official release.

### Actual behavior

Both the nightly-build workflow and the release workflow are broken.
Error from the log:
```
[ERROR] Failed to execute goal org.sonatype.plugins:nexus-staging-maven-plugin:1.6.13:deploy (injected-nexus-deploy) on project ide-cli:
Failed to install artifact com.devonfw.tools.IDEasy:ide-cli:tar.gz:linux:2024.03.001-alpha-SNAPSHOT: /home/runner/work/IDEasy/IDEasy/cli/target/ide-cli-linux-x64.tar.gz (No such file or directory) -> [Help 1]
```

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. See the nightly build workflow or run it manually

### Related/Dependent Issues

#221

### Comments/Hints:

This is where the release file for linux is defined and should be deployed:
https://github.com/devonfw/IDEasy/blob/8e44c02fffb25707b176042d53d4c564113fc0f0/cli/pom.xml#L155

As this used to work before and it now fails the question is:
Are all release packages not build anymore and therefore the file does not exist?
Or are they build but somehow now located in a different directory or with different name so they are not found?

### Affected version:

* OS: Linux, maybe all

hohwille:
Looking at the logs in more detail, I can find that the linux release is the first one trying to be installed what immediately fails:
```
[INFO] Installing /home/runner/work/IDEasy/IDEasy/cli/target/ide-cli-2024.03.001-alpha-SNAPSHOT.jar to /home/runner/work/IDEasy/IDEasy/documentation/target/nexus-staging/deferred/com/devonfw/tools/IDEasy/ide-cli/2024.03.001-alpha-SNAPSHOT/ide-cli-2024.03.001-alpha-SNAPSHOT.jar
[INFO] Installing /home/runner/work/IDEasy/IDEasy/cli/.flattened-pom.xml to /home/runner/work/IDEasy/IDEasy/documentation/target/nexus-staging/deferred/com/devonfw/tools/IDEasy/ide-cli/2024.03.001-alpha-SNAPSHOT/ide-cli-2024.03.001-alpha-SNAPSHOT.pom
[...]

PR #562, opened 2024-08-26
#523: exclude files from wrong OS

### LINKED ISSUE #523: Remove files specific for other OS from releases
We should exclude files specific for a particular OS from releases for another OS:
* Non-Windows (linux/mac): exclude `*.bat`
* All: exclude `package/system/«os»` for all OS except the current one.

### Affected version:

* IDEasy 2024.08.002-beta

hohwille:
fixes #523 
* exclude files from wrong OS in assembly files
* fix XSD link to https

Also fix OS and arch classifier to follow our own standard (see ide-urls).
Was also specified in #427:
> Actually I would expect that either in the workflow build or in our CLI POM or in the assembly we define that we want to have a release for win-x64, max-x64, mac-arm, linux-x64 (and potentially more)
See also #251

Fun fact is that even without this change, we got releases with `windows-x64`, `mac-arm`, etc. as specified so it worked.
The question is why did it work? At least the release artifact filenames in the target-folder are derived from these assembly IDs and have been somehow inconsistent, so for me this change makes sense.
UPDATE: I was blind. Each file is mapped in the POM. All non-sense because we never completed what I was trying to say in issues like #427
See also my comment on Teams:
> Ich hatte ja mal vorgeschlagen, dass wir in den matrix.os VMs nur das native image bauen (ideasy[.exe]) und das danach ins release einbauen, d.h. wir bauen 4 Assemblies die jeweils ein eigenes natives image nutzen, was im matrix.os build hochgeladen und danach im Gesamt-Build wieder runtergeladen wurde. Dann brauchen wir auch die Dokumentation nicht davor extra zu bauen und diese ganzen komischen Verrenkungen nicht machen und das Assembly-Plugin kann auch die Artefakte attachen und wir brauchen dafür kein extra build-helper-plugin. Dann wird es m.E. auch wieder konsistent.

Created new issue #563 for that.

For the record: I had a long long discussion with @jan-vcapgemini if our arch classifiers are OK or not.
We could easily agree that `arm` is correct and should be preferred over `aarch`.
We can also agree that 32-Bit is to be considered dead.
That we use `x64` what actually stands for `x86-64` on the one hand and `arm` without any `64` is kind of inconsistent.
My argument was however, that this was specified long time ago in `devonfw-ide` and the entire `ide-urls` repo is based on this convention. Changing this in IDEasy would require to change or rebuild the entire `ide-urls` repo and that would even break `devonfw-ide`. So consider this as given by legacy if you do not like it.
Might be more logical if we would instead have
* `x64` and `arm64`
* or `x86` and `arm`
* or `x86-64` and `arm-64`

However, we simply have `x64` and `arm` and that is somehow curved in stone so we better follow strictly our own conventions instead of doing something different in various places.


hohwille:
Our builds now fail with this error:
```
Run coveralls report        
coveralls: error while loading shared libraries: libevent-2.1.so.7: cannot open shared object file: No such file or directory
Error: Process completed with exit code 127.
```
Not related to the changes at all but some infrastructural problem of coveralls.
If that persists, we need to disable coveralls but first lets wait 24h or so.

cli/pom.xml
```diff
@@ -136,6 +136,13 @@
                     <exclude>system/**/*</exclude>
                   </excludes>
                 </resource>
+                <resource>
+                  <directory>src/main/package</directory>
```
hohwille, line 140: Please add an XML comment explaining why this is now duplicated.

hohwille: Done.

Issue #461, opened 2024-07-09
Add progress bar for extraction (unpacking)


jan-vcapgemini:
As an IDEasy user I want to see the status of my software installations.
Currently when a software is being installed the console does not update and show any progress apart from the download state.
Only if the user activates the `trace` mode, processes of the installation can be followed.

We need to add the progress bar functionality to FileUtil.extract/unpack functionality.

Issue #552, opened 2024-08-23
ZIP extraction buggy


slskiba:
I expect the plugins provided by <plugin>.properties files in the ide settings to be available for installation

### Actual behavior

```
> ide install intellij
Start: Install intellij
Due to lack of permissions, Microsoft's mklink with junction had to be used to create a Symlink. See https://github.com/devonfw/IDEasy/blob/main/documentation/symlinks.asciidoc for further details. Error was: C:\testprojects\dummyproject\software\intellij: A required privilege is not held by the client
You are on Windows and you do not have permissions to create symbolic links. Junctions are used as an alternative, however, these can not point to relative paths. So the source (..\..\_ide\software\default\intellij\intellij\2023.3.7) is interpreted as an absolute path.
Junction created for C:\testprojects\dummyproject\software\intellij <<===>> C:\testprojects\_ide\software\default\intellij\intellij\2023.3.7
Successfully installed intellij in version 2023.3.7
Start: Install plugin: EclipseCodeFormatter
Trying to download intellij-plugin-EclipseCodeFormatter.zip from https://plugins.jetbrains.com/pluginManager?action=download&id=EclipseCodeFormatter&build=IC-233.15325.26
Downloading 100% [========================] 10/10MiB (0:00:00 / 0:00:00) ?MiB/s
Step 'Install plugin: EclipseCodeFormatter' ended with failure.
Step 'ide (install,intellij)' failed: java.lang.IllegalStateException: Failed to extract C:\testprojects\_ide\tmp\EclipseCodeFormatter\intellij-plugin-EclipseCodeFormatter.zip to C:\testprojects\_ide\tmp\extract-intellij-plugin-EclipseCodeFormatter.zip
 Step 'Install plugin: EclipseCodeFormatter' failed: null
2 step(s) failed out of 4 steps.
An unexpected error occurred!
We are sorry for the inconvenience.
Please check the error below, resolve it and try again.
If the error is not on your end (network connectivity, lack of permissions, etc.) please file a bug:
https://github.com/devonfw/ide/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=IllegalStateException%3A+Failed+to+extract+C%3A%5Ctestprojects%5C_ide%5Ctmp%5CEclipseCodeFormatter%5Cintellij-plugin-EclipseCodeFormatter.zip+to+C%3A%5Ctestprojects%5C_ide%5Ctmp%5Cextract-intellij-plugin-EclipseCodeFormatter.zip
[...]

PR #567, opened 2024-08-27
#461: Add Progressbar for Unpacking Archives

alfeilex:
Fixes: #461 

For archives that use the `extractArchive()` function a progressbar is shown during unpacking:

Affected Archives are:
- ZIP
- TAR (transitive PKG)
- JAR


cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -613,8 +613,20 @@ public void extractTar(Path file, Path targetDir, TarCompression compression) {
   public void extractJar(Path file, Path targetDir) {
 
     this.context.trace("Unpacking JAR {} to {}", file, targetDir);
+    // calculate the size of archive for the progressbar max size
+    long sizeOfArchive = 0;
```
alfeilex, line 617: This workaround counting the size before init the progressbar works for ZIPs but JARs, as for example gcviewer.jar, have the size of -123

Probably that's still a bug in `extractJar()`

cli/src/test/java/com/devonfw/tools/ide/context/AbstractIdeTestContext.java
```diff
@@ -88,8 +88,8 @@ public IdeProgressBar prepareProgressBar(String taskName, long size) {
 
     IdeProgressBarTestImpl progressBar = new IdeProgressBarTestImpl(taskName, size);
     IdeProgressBarTestImpl duplicate = this.progressBarMap.put(taskName, progressBar);
-    // If we have multiple downloads, we may have an existing "Downloading" key
-    if (!taskName.equals("Downloading")) {
+    // If we have multiple downloads or unpacking, we may have an existing "Downloading" or "Unpacking" key
```
alfeilex, line 91: Not sure if that's wrong behavior, but if not put this condition then the tests for extracting tar files don't work. Probably because the tests run simultaniously? Could be `@BeforeEach` useful in this case?

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
alfeilex: progressbar for JAR extraction removed for now

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -659,8 +659,22 @@ public static String generatePermissionString(int permissions) {
...
     this.context.trace("Unpacking archive {} to {}", file, targetDir);
+    // calculate the size of archive for the progressbar max size
+    long sizeOfArchive = 0;
     try (InputStream is = Files.newInputStream(file); ArchiveInputStream ais = unpacker.apply(is)) {
       ArchiveEntry entry = ais.getNextEntry();
+      while (entry != null) {
+        sizeOfArchive += entry.getSize();
+        entry = ais.getNextEntry();
+      }
+      System.out.println("Size of Archive is: " + sizeOfArchive);
```
hohwille, line 670: please remove `System.out`. If you want to keep it use logging on debug level.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -659,8 +659,22 @@ public static String generatePermissionString(int permissions) {
...
 
     this.context.trace("Unpacking archive {} to {}", file, targetDir);
+    // calculate the size of archive for the progressbar max size
+    long sizeOfArchive = 0;
     try (InputStream is = Files.newInputStream(file); ArchiveInputStream ais = unpacker.apply(is)) {
       ArchiveEntry entry = ais.getNextEntry();
+      while (entry != null) {
+        sizeOfArchive += entry.getSize();
+        entry = ais.getNextEntry();
+      }
```
hohwille, line 669: did you test the overhead cause by this?
Does the read all data, unpack everything or will this just read some header/meta data and skip the rest of the payload?
As it seems the InputStream variants are slow in this regard.
We might need to implement this based on `ZipFile` instead:
https://stackoverflow.com/a/29331429/1506881

cli/src/test/java/com/devonfw/tools/ide/context/AbstractIdeTestContext.java
hohwille: Was added long time ago by @MustaphaOuchen 
I also would need to debug into this. Your change makes sense according to the existing workaround/hack.
However, IMHO the test progress bar should be improved but this can be done isolated from this PR so IMHO this change is fine and we can resolve this.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -684,6 +698,9 @@ private void extractArchive(Path file, Path targetDir, Function<InputStream, Arc
         if (isTar && !this.context.getSystemInfo().isWindows()) {
           Set<PosixFilePermission> permissions = PosixFilePermissions.fromString(permissionStr);
           Files.setPosixFilePermissions(entryPath, permissions);
+          pb.stepBy(((TarArchiveEntry) entry).getRealSize());
+        } else {
+          pb.stepBy(entry.getSize());
```
hohwille, line 703: oh, nice one - kind of odd API. Thanks for finding out such tricky things.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
hohwille: Also if you do that workaround later for `TarArchiveEntry`, then IMHO the same has to be done here as well.

hohwille (commented):
@alfeilex thanks for your PR. Great work. 👍 
On the long run, I would still suggest to create a generic abstraction of unpacking so we do not have to reimplement this redundantly for JAR, ZIP (when using JDK - see #552), etc.

My concern is still if the computation of the real size is causing a significant overhead and we should discuss this first. If that is clarified, we can approve but maybe the discussion will lead to some bigger changes/refactorings...

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
alfeilex: Sorry for that. I used that for testing and forgot to remove. 

done

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
alfeilex: Finally, there is not need for calculating the size of the archives. It's enough to take the file size. The issue before was probably that the ZIP extraction didn't work correctly and by increasing stepSize of the probgressbar the maximal size wasn't reached. Further, by using `ZipFile` we have the compressed size of entries and an overload iof the progressbar size doesn't occur anymore. 

Related issues: #575 and  #609

alfeilex:
Blocked by #609

After the changes are merged, I will adapt the code and we can merge that PR after reviewing.

hohwille (approved):
@alfeilex thanks for your improvements. Cool idea to use the filesize and the compressed size of the entries what is a huge simplification with no real disadvantage. 👍

PR #568, opened 2024-08-29
#294: Ensure parent directory exists and skip reading if file not exists

ndemirca:
Resolved TODO from PR #374


cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
```diff
@@ -104,27 +104,34 @@ public void save() {
...
-        line = reader.readLine();
-        if (line != null) {
-          VariableLine variableLine = VariableLine.of(line, this.context, getSource());
-          lines.add(variableLine);
-        }
-      } while (line != null);
     } catch (IOException e) {
-      throw new IllegalStateException("Failed to load existing properties from " + this.propertiesFilePath, e);
+      throw new IllegalStateException("Failed to create parent directories for " + newPropertiesFilePath, e);
+    }
```
hohwille, line 115: You have access to `context` and `FileAccess` here. We invented `FileAccess` to avoid repeating this low-level stuff all the time:
```suggestion
  Path parentFolder = newPropertiesFilePath.getParent();
  if (parentFolder != null) {
    this.context.getFileAccess().mkdirs(parentFolder);
  }
```

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
```diff
@@ -104,27 +104,34 @@ public void save() {
...
+          if (line != null) {
+            VariableLine variableLine = VariableLine.of(line, this.context, getSource());
+            lines.add(variableLine);
+          }
+        } while (line != null);
+      } catch (IOException e) {
+        throw new IllegalStateException("Failed to load existing properties from " + this.propertiesFilePath, e);
+      }
+    } else {
+      this.context.info("Properties file {} does not exist, skipping read.", newPropertiesFilePath);
```
hohwille, line 132: If we `save` but have nothing to read, DEBUG should be sufficient here to avoid spam.
```suggestion
      this.context.debug("Properties file {} does not exist, skipping read.", newPropertiesFilePath);
```

hohwille (approved):
@ndemirca thanks for addressing this leftover TODO via this PR 👍 

Please have a look at my suggestions. We can then merge.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
hohwille: Just for the record: Under which circumstances will `newPropertiesFilePath.getParent()` return `null`?
IMHO this can never happen in reality.
However, I just learned that NIO `Path` API is not so well designed since you cannot traverse up to the parent of a relative path in case you "leave" that relative path. I would not have expected this:
```
Path srcMain = Path.of("src/main");
Path src = srcMain.getParent(); // same as Path.of("src") what is fine
Path basedir = src.getParent(); // not pointing to the current working directory but giving null instead
```

Therefore you instead need to do
```
Path srcMain = Path.of("src/main").toAbsolutePath();
```
to make it work as expected...

So very safe that you check for `null`. One could somehow set it as `Path.of("ide.properties")` in case we once support `ide.properties` in `IDE_HOME` and that could cause an NPE.
Anyway I ensure that all this should really never happen here:
https://github.com/devonfw/IDEasy/blob/1e62bd2a4a949f9875cac16118bf5108940eb431/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L143

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
```diff
@@ -99,32 +99,34 @@ public void save() {
     }
     Path newPropertiesFilePath = this.propertiesFilePath;
     String propertiesFileName = this.propertiesFilePath.getFileName().toString();
+    Path propertiesParentPath = newPropertiesFilePath.getParent();
+
     if (LEGACY_PROPERTIES.equals(propertiesFileName)) {
-      newPropertiesFilePath = this.propertiesFilePath.getParent().resolve(DEFAULT_PROPERTIES);
+      newPropertiesFilePath = propertiesParentPath.resolve(DEFAULT_PROPERTIES);
```
hohwille, line 105: Oh, cool. You also found and fixed a critical bug here. Thanks 👍

cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
```diff
@@ -21,13 +21,16 @@ class EnvironmentVariablesPropertiesFileTest extends Assertions {
   /**
    * Test of {@link EnvironmentVariablesPropertiesFile} including legacy support.
    */
+
+  private final static Path envVarPath = Path.of("src/test/resources/com/devonfw/tools/ide/env/var/");
+  private final static EnvironmentVariablesType type = EnvironmentVariablesType.SETTINGS;
+  private final static AbstractEnvironmentVariables parent = null;
```
hohwille, line 27: https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc#constants
* modifier order for constants is `static final` not vice-versa
* constants in Java are always in UPPER_CASE syntax
* There is no sense in defining a constant for `null` here. Simply use `null` as literal or leave it as local variable.
```suggestion
  private static final Path ENV_VAR_PATH = Path.of("src/test/resources/com/devonfw/tools/ide/env/var/");
  private static final EnvironmentVariablesType TYPE = EnvironmentVariablesType.SETTINGS;
```

hohwille (approved):
@ndemirca thanks for your improvement and great that you added a test for saving of `EnvironmentVariables`. 👍 
Still I found small things in the review that we need to rework before merge.
Please have a look.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
hohwille: All solved. Now 👍

PR #569, opened 2024-08-30
#438: validationstate

### LINKED ISSUE #438: Introduce ValidationState
Currently our validation is not implemented well:
https://github.com/devonfw/IDEasy/blob/eae8fd2796e382f6ff84def23cb6ca550d760930/cli/src/main/java/com/devonfw/tools/ide/commandlet/Commandlet.java#L202-L216
https://github.com/devonfw/IDEasy/blob/eae8fd2796e382f6ff84def23cb6ca550d760930/cli/src/main/java/com/devonfw/tools/ide/property/Property.java#L456-L467

So we have validation result as `boolean` and also the concept to throw an exception if the validation fails in a properties validator.
Both approaches are not well designed.

Instead we should create an interface `ValidationResult`
```
public interface ValidationResult {
  boolean isValid();

  String getErrorMessage();
}
```

Then we can create an implementation `ValidationState`:
```
public class ValidationState implements ValidationResult {
  private final StringBuilder errorMessage;
  
  public boolean isValid() {
    return (this.errorMessage == null);
  }
  
  public String getErrorMessage() {
    if (this.errorMessage == null) {
      return null;
    }
    return this.errorMessage.toString();
  }

  public void addErrorMessage(String error) {
    if (this.errorMessage == null) {
      this.errorMessage = new StringBuilder(error.length());
    } else {
      this.errorMessage.append('\n');
    }
    this.errorMessage.append(error);
  }
}
```

This is just a design suggestion as starting point. We can also have a immutable implementation(s) of `ValidationResult` and have a `ValidationResult toResult()` method in `ValidationState`.

The final idea is that internal validation methods will take a `ValidationState` object as argument. This also applies for the validators of the `Property` classes. Here we need to change from `Consumer<V>` to a custom interface:
```
@FunctionalInterface
public interface PropertyValidator<V> {
  void validate(V value, ValidationState validationState);
}
```
The main API should take no argument and return `ValidationResult`.

An alternative approach is to skip the `ValidationState` and always return `ValidationState` and then add some aggregation logic that can compose multiple `ValidationState` objects into one. IMHO the latter approach is slightly more complex.

slskiba:
Closes #438. added exemplary usage to `InstallCommandlet`

TBD:
- Exemplary usage added in `InstallCommandlet` checks whether the given version is a valid one - this is a relevant check for a `VersionProperty` at any given place - should they now each implement an identical validator? 
- `PathProperty` solves this by overriding `validate()`, to implement general Path-related validation - or should this move into validator implementations?


cli/src/main/java/com/devonfw/tools/ide/commandlet/InstallCommandlet.java
```diff
@@ -29,7 +30,12 @@ public InstallCommandlet(IdeContext context) {
     super(context);
     addKeyword(getName());
     this.tool = add(new ToolProperty("", true, "tool"));
-    this.version = add(new VersionProperty("", false, "version"));
+    this.version = add(new VersionProperty("", false, "version",
+        (VersionIdentifier v, ValidationState state) -> {
```
hohwille, line 34: For lambdas it is convenient to use type inference:
```suggestion
        (v, state) -> {
```

cli/src/main/java/com/devonfw/tools/ide/property/Property.java
```diff
@@ -449,21 +451,24 @@ public boolean matches(String nameOrAlias) {
...
    * @return {@code true} if this {@link Property} is valid, {@code false} if it is {@link #isRequired() required} but no {@link #getValue() value} has been
-   * set.
-   * @throws RuntimeException if the {@link #getValue() value} is violating given constraints. This is checked by the optional {@link Consumer} function given
-   * at construction time.
+   *     set.
+   * @throws RuntimeException if the {@link #getValue() value} is violating given constraints. This is checked by the optional {@link Consumer} function
+   *     given at construction time.
    */
-  public boolean validate() {
+  public ValidationState validate() {
```
hohwille, line 458: From the outside we do not need to do any changes on the state.
To express this we should use the interface that was designed for this purpose.
```suggestion
  public ValidationResult validate() {
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/Commandlet.java
```diff
@@ -228,7 +229,9 @@ public boolean validate() {
       }
     }
     for (Property<?> property : this.propertiesList) {
-      if (!property.validate()) {
+      ValidationState state = property.validate();
+      if (!state.isValid()) {
+        context.debug(state.getErrorMessage());
```
hohwille, line 234: OK, so maybe we need to expose this concept further:
We should consider creating the `ValidateState` here and pass it to `property.validate(state)` as argument and return `void` instead.
Or as an alternative we could implement some mechanism to aggregate multiple validation states:
```
public class ValidationState {
...
  public void add(ValidationResult result) {
    if (!result.isValid()) {
      addErrorMessage(result.getErrorMessage()); // TODO: maybe this will cause duplicated newlines so this needs to be done smarter
    }
}
```

So in the end then the `validate()` method on `Commandlet` should return `ValidationResult` containing all violations of all properties. Also we should somehow add some magic to the above aggregation that tells the end-user which error message is about which property in a generic way.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -824,7 +824,7 @@ private boolean applyAndRun(CliArguments arguments, Commandlet cmd) {
 
     boolean matches = apply(arguments, cmd, null);
     if (matches) {
-      matches = cmd.validate();
+      matches = cmd.validate().isValid();
```
hohwille, line 827: IMHO this needs to go even further.
So also here we need to return `ValidationResult`.
Tricky thing is if `matches` is currently `false` but not from validation, we would even need a reason as `validationError`.
So most likely also `apply` needs to return `ValidationResult`.
Then `run(CliArguments)` can be changed to handle the `ValidationResult`
So instead of `boolean matches` declare the `ValidationResult` and keep the one from the `firstCandidate`.
Now instead at the
```
step.error("Invalid arguments: {}", current.getArgs());
```
We can now handle that `ValidationResult` if not valid and get the error message.
The error message could be appended to the error log message or IMHO even better, we could throw a CliException here from the ValidationResult. If we want to make that reusable, we can even derive a class `CliValidationException` that can be created from `ValidationResult`, `Commandlet`, and `args` to create a reasonable exception.
This makes it better for JUnit tests since we can catch such exception to assert validation error behavior instead of checking status codes and searching for log messages what is less stable.

hohwille (approved):
@slskiba thanks for your rework. You created a solid state and are almost there. 👍 
I left some comments to guide the way.
Once addressed, we can finally merge.

cli/src/main/java/com/devonfw/tools/ide/commandlet/Commandlet.java
```diff
@@ -219,20 +221,24 @@ public boolean isProcessableOutput() {
...
+        state.addErrorMessage("Required property has no value");
+        return state;
       }
     }
     for (Property<?> property : this.propertiesList) {
-      if (!property.validate()) {
-        return false;
+      ValidationResult state = property.validate();
+      if (!state.isValid()) {
+        context.error(state.getErrorMessage());
```
hohwille, line 237: For the error handling the caller needs to be responsible. Assuming we have two matching commandlets and the first one (e.g. `firstCandidate`) is invalid and we log an error but the second one is valid, then the logged error message will be confusing for the end-user since he did not do anything wrong.
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/property/Property.java
```diff
@@ -449,21 +452,24 @@ public boolean matches(String nameOrAlias) {
...
+   *     given at construction time.
    */
-  public boolean validate() {
+  public ValidationResult validate() {
+
+    ValidationState state = new ValidationState(this.getNameOrAlias());
 
     if (this.required && (getValue() == null)) {
-      return false;
+      state.addErrorMessage("Property " + this.name + " is required but no value has been set");
```
hohwille, line 464: The name of the property is now already included by the `ValidationState`:
```suggestion
      state.addErrorMessage("Value is required and cannot be empty.");
```

PR #575, opened 2024-09-02
#130: progressbar unknown file size

jan-vcapgemini:
Fixes: #130 
Continues: #286 

Implements:
* Unit tests for `ProgressBar`
* Integration tests for file download and file copy processes
* refactoring of `IdeProgressBar` into `AbstracIdeProgressBar`
* Handling of unknown file sizes (sets initial max size of ProgressBar to -1)


cli/src/main/java/com/devonfw/tools/ide/io/AbstractIdeProgressBar.java
```diff
@@ -0,0 +1,68 @@
...
+   * @param currentProgress current progress state.
+   */
+  protected abstract void doStepBy(long stepSize, long currentProgress);
+
+  /**
+   * Increases the progress bar by given step size.
+   *
+   * @param stepSize size to step by.
+   */
+  protected abstract void doStepBy(long stepSize);
```
hohwille, line 32: IMHO we can now remove this method or why do we have both `doStepBy` methods?
```suggestion
```

jan-vcapgemini: I've removed the method now.

cli/src/test/java/com/devonfw/tools/ide/io/IdeProgressBarTestImpl.java
```diff
@@ -34,27 +34,34 @@ public class IdeProgressBarTestImpl implements IdeProgressBar {
    * @param max maximum length of the bar.
    */
   public IdeProgressBarTestImpl(String name, long max) {
-
+    super(max);
     this.start = Instant.now();
     this.name = name;
     this.max = max;
```
slskiba, line 40: Why are we making another local field max, instead of using maxLength or AbstractIdeProgressBar? the super call seems completely unnecessary when not using maxLength.

slskiba (approved):
Looks great. :+1:
Just a minor remark/question, but ready to merge imo.

cli/src/test/java/com/devonfw/tools/ide/io/IdeProgressBarTestImpl.java
jan-vcapgemini: Thanks I've removed the duplicated field and added a getter for maxLength now.

slskiba: Resolved

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/Mvn.java
```diff
@@ -185,8 +185,9 @@ private Set<String> findVariables(String content) {
   @Override
   public void installPlugin(PluginDescriptor plugin) {
 
-    Path mavenPlugin = this.getToolPath().resolve("lib/ext/" + plugin.getName() + ".jar");
+    Path mavenPlugin = this.context.getSoftwarePath().resolve(this.tool).resolve("lib/ext/" + plugin.getName() + ".jar");
```
hohwille, line 188: Not wrong but just for my curiosity: Why did you had to change this? The method `getToolPath()` should give us the same result and makes it more compact and readable IMHO. I assume there is a reason why you changed this that I am missing.

hohwille (approved):
@jan-vcapgemini thanks for your PR. You created the perfect solution. Awesome that you found support for unknown length is already in the progress-bar library we are using so we could simplify this story a lot. 👍 
I left one questions for my curiosity. But this is ready for merge!

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/Mvn.java
jan-vcapgemini: This looks like an unintended change. Should be fixed with: #595

PR #578, opened 2024-09-03
#505: set IDEA_PROPERTIES in runTool instead of postInstall

### LINKED ISSUE #505: Refactor Jetbrains IDEs

slskiba:
Closes point 2.2 of #505


cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -357,16 +357,16 @@ private ToolInstallation createToolInstallation(Path rootDir, VersionIdentifier
...
-   * Method to set the environment variable for the dependency of the tool.
+   * Method to set an environment variable for the process context.
    *
    * @param envContext the {@link EnvironmentContext} of the tool.
-   * @param dependencyPath the {@link Path} of the dependency.
+   * @param pathVariable the pathVariable to set
+   * @param value the {@link Path value} of the pathVariable.
    */
-  private void setEnvironment(EnvironmentContext envContext, Path dependencyPath) {
+  protected void setEnvironment(EnvironmentContext envContext, String pathVariable, Path value) {
```
hohwille, line 366: This was a missunderstanding.
This method should be called only once per tool commandlet.
Its implementation can set multiple environment variables but the tool must be in control which variables are set to what value. If that information has to come from the outside, it cannot work properly.

Example:
* Install and run tomcat
* tomcat has dependecy on Java so we also need to install Java.
* According to the tomcat version only a dedicated version range of Java is compatible.
* We might notice that the project is using a Java version that does not match this version range.
* In that case we install an additional Java matching that version range if not already available in software repo
* Now we call this `setEnvironment` method on the Java commandlet with the `ProcessContext` created in the `Tomcat` commandlet that will later be used to start tomcat.
* The Java commandlet will set `JAVA_HOME` to the given `Path` pointing to the software installation (potentially in software repo or in the `software` folder of the project). It may even override this method to additionally set `JRE_HOME` to the same value. Maybe even the given `Path` needs to be added to the `PATH` variable of that process context. Things like this will all come in the future and are not yet implemented.
* Finally `Tomcat` is started via ProcessContext with `JAVA_HOME`, etc. set and therefore tomcat will use the compatible Java version instead of the incompatible one from the project.
* However, `Tomcat` commandlet shall not have hardcoded knowledge about its `Java` dependency and the fact that maybe some `JAVA_HOME` variable needs to be set to make it work. The idea is that for a tool a `dependencies.json` file can be added or extended to add a new dependency without the need to implement the same into our Java code.

Making this method `protected` is perfectly correct. Also making the JavaDoc more clear is very much appreciated.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -394,19 +394,20 @@ private void installDependencies(DependencyInfo dependencyInfo, EnvironmentConte
     VersionIdentifier dependencyVersionToInstall = this.dependency.findDependencyVersionToInstall(dependencyInfo);
 
     Path dependencyRepository = getDependencySoftwareRepository(dependencyName, dependencyTool.getConfiguredEdition());
+    String pathVariable = this.tool.toUpperCase(Locale.ROOT) + "_HOME";
```
hohwille, line 397: As stated above this is incorrect. Assuming here you are in the `Tomcat` commandlet and would decide that the variable needs to be set to `JAVA_HOME`. Maybe the `dependencyTool` which is `Java` in this example needs to decide something else instead of this default, this cannot be solved properly.
E.g. Maven might need `MAVEN_HOME` instead of `MVN_HOME` just to give another example.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -116,7 +116,7 @@ public void runTool(ProcessMode processMode, VersionIdentifier toolVersion, bool
     pc.run(processMode);
   }
 
-  private ProcessContext getProcessContext(Path binaryPath, String... args) {
+  protected ProcessContext getProcessContext(Path binaryPath, String... args) {
```
hohwille, line 119:
```suggestion
  protected ProcessContext createProcessContext(Path binaryPath, String... args) {
```

hohwille (changes requested):
@slskiba thanks for doing the setting of tool specific env variables in the right way. 👍 
Unfortunately there are still quite some left-overs from #131 and #439 that are a little in your way here and make things slightly more complicated.
As a learning I have to interfere really hard in the future if ever some developer tries to create such monster PRs anymore that grow so complex that the developer will loose track to understand all changes and keep everything in his head. Instead we must split into smaller stories before and have smaller PRs that only get merged if a high quality is given.
I hope my examples in the review comments make clear what the problem and the expectation is.
Feel free to invite me to a Teams meeting if you need further assistance how to reach that state and come to such a solution as this will also need some improvements leftover from the tool dependency stories...

cli/src/test/java/com/devonfw/tools/ide/tool/androidstudio/AndroidStudioTest.java
```diff
@@ -80,7 +80,6 @@ public void testAndroidStudioRun(String os, WireMockRuntimeInfo wmRuntimeInfo) t
   private void checkInstallation(IdeTestContext context) {
     // commandlet - android-studio
     assertThat(context.getSoftwarePath().resolve("android-studio/.ide.software.version")).exists().hasContent("2024.1.1.1");
-    assertThat(context.getVariables().get("STUDIO_PROPERTIES")).isEqualTo(context.getWorkspacePath().resolve("studio.properties").toString());
```
jan-vcapgemini, line 83: Why was this removed?

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
slskiba: I have reverted the setEnvironment method to its original state 👍

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
slskiba: I have reverted this change 👍

jan-vcapgemini (changes requested):
Looks good to me, can be moved to review after my CRs were resolved.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -357,16 +357,16 @@ private ToolInstallation createToolInstallation(Path rootDir, VersionIdentifier
...
-   * @param dependencyPath the {@link Path} of the dependency.
+   * @param value the {@link Path value} of the pathVariable.
    */
-  private void setEnvironment(EnvironmentContext envContext, Path dependencyPath) {
+  protected void setEnvironment(EnvironmentContext envContext, Path value) {
 
     if (envContext != null) {
       String pathVariable = this.tool.toUpperCase(Locale.ROOT) + "_HOME";
-      envContext.withEnvVar(pathVariable, dependencyPath.toString());
+      envContext.withEnvVar(pathVariable, value.toString());
```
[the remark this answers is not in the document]
slskiba: It is just a renaming to make it more general. This method is actually not used for this PR, but in the future might be used for things that are not related to dependencies. Therefore I suggest to keep this more general naming.

cli/src/test/java/com/devonfw/tools/ide/tool/intellij/IntellijTest.java
```diff
@@ -141,7 +141,6 @@ public void testIntellijRun(String os, WireMockRuntimeInfo wmRuntimeInfo) throws
   private void checkInstallation(IdeTestContext context) {
 
     assertThat(context.getSoftwarePath().resolve("intellij/.ide.software.version")).exists().hasContent("2023.3.3");
-    assertThat(context.getVariables().get("IDEA_PROPERTIES")).isEqualTo(context.getWorkspacePath().resolve("idea.properties").toString());
```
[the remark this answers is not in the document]
slskiba: Setting this variable is not part of the installation process anymore

cli/src/test/java/com/devonfw/tools/ide/tool/androidstudio/AndroidStudioTest.java
slskiba: Setting this variable is not part of the installation process anymore

hohwille: In order to test this, we would need to mock the `ProcessContext`, run AndroidStudio and then verify that this variable was set in the mocked `ProcessContext`.

hohwille: Feel free to create a new issue to improve our tests for this. Not needed to be implemented within this PR.

slskiba: Created #607

hohwille (approved):
@slskiba thanks for completing this PR. Nice work. 👍 
Ready for merge.
The proper usage of `setEnvironment` will be implemented later outside of this PR.

Issue #534, opened 2024-08-13
Smart git pull not working


hohwille:
### Expected behavior

Story #118 should be implemented with PR #163 

### Actual behavior

When I call autocompletion of a version (e.g. `ide -t complete set-version mvn 3.9.`) multiple times within a loop without any delay, I can observe that every such invocation triggers a `git pull`.
Therefore I conclude that #118 is not implemented correctly.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. `ide -t complete set-version mvn 3.9.`
2. `ide -t complete set-version mvn 3.9.`

### Related/Dependent Issues

#118
#163

### Comments/Hints:

Seems to need some detailed debugging session to understand why it is now working as planned.
I have chosen high priority since the auto-completion is very slow due to this bug.
For regular usage this does not cause harm, but I want to use auto-completion for versions.

### Affected version:

* OS: Windows
* IDEasy: 2024.08.001-beta-SNAPSHOT

PR #581, opened 2024-09-03
#534: fixed smart git pull

jan-vcapgemini:
Fixes: #534


hohwille (approved):
@jan-vcapgemini thanks for analysis and this fix as PR. Indeed you are perfectly correct and found the reason why the optimization never worked as planned. Great job 👍

PR #602, opened 2024-09-10
#553: fixed ide root in setup.bat script

### LINKED ISSUE #553: Missmatch of IDE_ROOT
I got this warning from IDEasy:
```
Variable IDE_ROOT is set to 'D:\projects\_ide' but for your project 'undefined' the path 'D:\projects' would have been expected.
```
This does not make sense:
- [ ] as documented in [variables](https://github.com/devonfw/IDEasy/blob/main/documentation/variables.adoc) the `IDE_ROOT` variable should point to the "projects" folder and not to the `_ide` sub-folder inside of it. That seems a bug in our `setup.bat` script.

https://github.com/devonfw/IDEasy/blob/043d854e5969b29435c7a66ded253e46a0f3dc5d/cli/src/main/package/setup.bat#L39

ndemirca:
Closes #553


cli/src/main/package/setup.bat
```diff
@@ -36,6 +36,7 @@ if "%USER_PATH%" == "" (
 ) else (
   setx PATH "%USER_PATH%;%CD%\bin"
 )
+cd ..
 setx IDE_ROOT "%CD%"
```
hohwille, line 40: You also need to return to the original directory (`_ide`) after the `setx` command since we later run the `setup` bash script that will otherwise not be found anymore and will fail.

hohwille (approved):
@ndemirca thanks for fixing this 👍

PR #603, opened 2024-09-10
Quick Fix: Revert UrlUpdaters

alfeilex:
For testing, I removed some updaters and these change is unfortunately was added in the PR.


hohwille (approved):
Thanks @alfeilex

Issue #604, opened 2024-09-10
Wrong Path Formating of MAVEN_ARGS variable on Windows


alfeilex:
### Expected behavior

Environment variables on Windows should be in windows path format:
`MAVEN_ARGS=-s C:\projects\dev\conf\mvn\settings.xml`

### Actual behavior
`MAVEN_ARGS=-s /c/projects/dev/conf/mvn/settings.xml` is not in windows path format on Windows.

### Steps to reproduce (bug)
1. Install `mvn` and try to run `mvn clean`
2. That error should occur:
```
ide mvn clean
[ERROR] Error executing Maven.
[ERROR] The specified user settings file does not exist: C:\c\projects\dev\conf\mvn\settings.xml
```
3. The reason for that might be the wrong format `MAVEN_ARGS` value as described in behaviors

### Affected version:
* OS: Windows 
* Shell: GitBash

ndemirca:
duplicate of https://github.com/devonfw/IDEasy/issues/592

hohwille:
@alfeilex 
Why do you think, this is a bug?
Maven is called via `mvn.bat` on Windows and need a Windows standard PATH rather than a MSYS path.
Is something going wrong with the current state?
IMHO it is a bug to change this...

hohwille:
> duplicate #592

I would disagree. This is a totally different issue.

hohwille:
For the record: IMHO this is related to this problem: https://issues.apache.org/jira/browse/MNG-8018

hohwille:
> Expected behavior
Environment variables on Windows should be in windows path format

This expectation is also wrong. In git-bash we need MSYS path syntax instead of Windows syntax to make it work.
This is the difference of
```
ide env
```
called in `CMD` vs.
```
ide env --bash
```
called in `git-bash`.

As we can see supporting Windows with CMD and git-bash is rather tricky.

PR #605, opened 2024-09-10
#604: Fix MAVEN_ARGS Path Format for Windows

alfeilex:
Fixes: #604 

Tested by creating a native image for Windows


cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
```diff
@@ -438,11 +438,11 @@ default String getMavenArgs() {
       return null;
     }
     String settingsPath;
-    WindowsPathSyntax pathSyntax = getPathSyntax();
-    if (pathSyntax == null) {
+    boolean isWindows = getSystemInfo().isWindows();
```
jan-vcapgemini, line 441: Do we need this boolean somewhere else too? If not, we can replace it in the if condition.

jan-vcapgemini (changes requested):
LGTM. The isWindows boolean can be removed I think.

hohwille (approved):
@alfeilex thanks for creating the bug issue and fixing PR. 👍 
I left a comment as you can simplify this to a single line since OS check is pointless. You do not need `WindowsPathSyntax` to convert from WINDOWS to WINDOWS since `Path` on Windows will always format in Windows syntax via `toString()`.

BTW: This is now resulting in the code how we wrote it before it was broken with #321

cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
alfeilex: Has been done with the removal of the OS checking

cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
```diff
@@ -438,11 +438,11 @@ default String getMavenArgs() {
...
+    if (!isWindows) {
       settingsPath = mvnSettingsFile.toString();
     } else {
-      settingsPath = pathSyntax.format(mvnSettingsFile);
+      settingsPath = WindowsPathSyntax.WINDOWS.format(mvnSettingsFile);
     }
```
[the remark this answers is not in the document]
alfeilex: Done

jan-vcapgemini (approved):
LGTM.

PR #606, opened 2024-09-10
#574: fixed getParentPath method

### LINKED ISSUE #574: mismatch of working directory
### Actual behavior

IDEasy incorrectly assumes that the current working directory would be `D:\projects\_ide\software\default\java\java\21.0.4_7` while it is actually  `/d/projects/IDEasy/software/java`.

image

### Steps to reproduce (bug) 

1.  execute $ ide help in /d/projects/IDEasy/software/java

ndemirca:
Closes #574


cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -313,17 +312,7 @@ private boolean isIdeHome(Path dir) {
...
-      if (!dir.equals(linkDir)) {
-        return linkDir;
-      } else {
-        return dir.getParent();
-      }
-    } catch (IOException e) {
-      throw new IllegalStateException(e);
-    }
-
+    return dir.getParent();
```
hohwille, line 315: Correct fix. However, having a private method that just wraps an existing method does not add any value.
IMHO we should now remove `getParentPath` method and call `getParent()` directly.

hohwille (approved):
@ndemirca thanks for your PR and fix. Looks good to me 👍 

IMHO this can be moved out of draft.

For the record: in devonfw-ide we resolved the link for the following use-case:
Assuming the end-user creates a symbolic link as a shortcut (e.g. of a repo inside a workspace into his home directory) then setting the environment via `devon`/`ide` would not work. However, if we resolve the link, it is still working. Now since we are using junctions and symbolic links in IDEasy, it caused bug #574. If later somebody complaints about this and wants to have this support back, we can still do the following: If IDE_HOME was not found this way, we do `toRealPath()` on the initial Path and rerun the code again that tries to find IDE_HOME from there. Not needed in this PR/Story but just wanted to explain this somewhere for general understanding.

PR #610, opened 2024-09-11
#525 Added online state check before downloading

### LINKED ISSUE #525: User can make downloads to install tools even when set to offline
### Expected behavior

When running "ide --offline install <tool>", I expect no downloads to be made. 

### Actual behavior

```
$ ide --offline install lazydocker 0.7.5
LAZYDOCKER_VERSION=0.7.5 has been set in SETTINGS@C:\projects\IDEasy\settings\ide.properties
Start: Install lazydocker
Trying to download lazydocker-0.7.5-windows-x64.zip from https://github.com/jesseduffield/lazydocker/releases/download/v0.7.5/lazydocker_0.7.5_Windows_x86_64.zip
Downloading 100% [==================================================================] 3/3MiB (0:00:00 / 0:00:00) ?MiB/s
Checksum 4a4e47a4a61a25cbc5e1175adf42c02b8957c1dbdb5e854d306151a08c51acad is correct.
Successfully installed lazydocker in version 0.7.5 replacing previous version 0.23.0
Successfully completed ide (install,lazydocker,0.7.5)

```

WorkingAmeise:
closes #525  

now Throws a CliOfflineException if the offline flag is set


cli/src/main/java/com/devonfw/tools/ide/repo/AbstractToolRepository.java
```diff
@@ -49,6 +50,9 @@ public Path download(String tool, String edition, VersionIdentifier version) {
     UrlDownloadFileMetadata metadata = getMetadata(tool, edition, version);
     VersionIdentifier resolvedVersion = metadata.getVersion();
     Set<String> urlCollection = metadata.getUrls();
+    if (context.isOffline()) {
+      throw new CliOfflineException("Not able to download " + tool + " because we are offline");
+    }
```
hohwille, line 55: This download method delegates to `FileAccessImpl.download` later (see below).
Why do we need to have this here?
Because we want to abort sooner and avoid waste (creating temporary path)? Or because we want to provide additional information?

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -95,6 +96,9 @@ public void download(String url, Path target) {
     this.context.info("Trying to download {} from {}", target.getFileName(), url);
     mkdirs(target.getParent());
     try {
+      if (context.isOffline()) {
+        throw new CliOfflineException("Not able to download because we are offline");
+      }
```
hohwille, line 101: The `CliOfflineException` already has a standard message in its default constructor.
If we want to have a custom message for downloads, I would suggest that you provide this also in `CliOfflineException` as an additional String constructor taking the `url` as argument and giving some context information so we can also see which tool this was and what version and edition since all is somehow in the URL. Ideally you make the constructor private and expose it as a static "factory" method with a speaking name `public static CliOfflineException ofDownload(String url) { return new CliOfflineException(url); }` with
```java
private CliOfflineException(String url) {
  super("You are offline and cannot download from URL " + url);
}
```

hohwille (approved):
@WorkingAmeise thanks for your PR. You found the relevant spots in the code where to implement this feature. Also throwing `CliOfflineException` is the perfect solution. 👍 
You PR is functionally correct. I just left some comments for further thoughts and improvements. Please have a look.

hohwille:
Can you please also check DoD:

CHANGELOG entry needs to be added.

cli/src/main/java/com/devonfw/tools/ide/repo/AbstractToolRepository.java
WorkingAmeise: We can download Tools via the implementation in AbstractToolRepository.java or we can download plugins via the implementation in FileAccessImpl.java in both cases we need to cancel since we are offline. Additionaly the download method of FileAccessImpl is called from other methods directly, so we need to have the check in both methods.

hohwille (approved):
@WorkingAmeise thanks for your PR and your rework.
You addressed my review comment well and solved the centralization of messages in `CliOfflineException` perfectly. 👍 

Please remove `CHANGELOG.adoc.orig` as commented and we can merge this PR.

Issue #612, opened 2024-09-11
Automatically generated issue URL is still pointing to ide instead of IDEasy


jan-vcapgemini:
### Expected behavior

As an IDEasy user I expect the automatically generated issue URL to be pointing to the IDEasy repository.

### Actual behavior
The automatically generated issue URL points to the ide repository instead.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Generate an error and click on the issue URL.

### Affected version:

* OS: Windows
* IDEasy: 2024.09.002-beta-SNAPSHOT

PR #613, opened 2024-09-11
#612: fixed issue URL

jan-vcapgemini:
Fixes: #612

Implements:
* fixed URL


hohwille (approved):
@jan-vcapgemini thanks for finding and fixing this bug 👍

PR #609, opened 2024-09-11
#552: Fix zip extraction

slskiba:
Closes #552. ~~Closes also: #550~~ (edit: PR #615 now handles #550)

- Separates extraction for zip and tar
- Use `ZipInputStream` instead of `ArchiveInputStream` for zip files
- Fixes download url containing spaces due to plugin names to be formatted correctly
- Fixes two buggy log statements

PR #615, opened 2024-09-12
#550: Fix download url generated from plugin id

### LINKED ISSUE #550: Installation of plugins for IDEs fails when the plugin-id contains a space character
As an IDEasy user, I want to install plugins for my IDEs as defined by the plugins properties files in the provided settings. 
In a plugin.properties file, the plugin_id may contain a space character (e.g. "PlantUML integration"), causing the download, and subsequently the installation of the plugin, to fail. 

Requirement: the PlantUMLIntegration.properties file needs to be present in settings/intellij/plugins, with the following content:
```
plugin_id=PlantUML integration
plugin_active=false
```
Then:

1. Set plugin_active in PlantUMLIntegration.properties to true
2. `install intellij`

Or alternatively:
1. `install-plugin intellij PlantUMLIntegration`

### Comments/Hints:
Replacing the space character of the plugin_id with "%20" makes the download succeed - however then the plugin installation fails at the extraction of the downloaded zip file. The field plugin_id in the properties file should keep the space character, to match the plugin_id provided from jetbrains.

slskiba:
Closes #550. 

Requires merging of #609 to be fully function for zip compressed plugins.

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeaBasedIdeToolCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeaBasedIdeToolCommandlet.java
@@ -3,6 +3,8 @@ package com.devonfw.tools.ide.tool.ide;
 import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


hohwille (approved):
@slskiba thanks for this PR and finding a proper solution. I do not know why `+` is not supported as that seems a common HTTP standard but for sure you did this on purpose.
Using `URLEncoder` from JDK is a good choice for the fix. 👍

jan-vcapgemini (approved):
LGTM.

PR #619, opened 2024-09-16
#618: improve IdeProgressBarConsoleTest

### LINKED ISSUE #618: IdeProgressBarConsoleTest slow in maven
On my machine during mvn install the test `IdeProgressBarConsoleTest.testProgressBarStyleSwitchOnNonWindows` takes more than 2 minutes what seems extremely long/slow.
In IntelliJ the entire test class with all test methods runs in less than 2 seconds.

From Maven surefire report:
```
  <testcase name="testProgressBarStyleSwitchOnNonWindows(Path)"
                  classname="com.devonfw.tools.ide.io.IdeProgressBarConsoleTest"
                  time="130.352">
    <system-out><![CDATA[20:33:18.673 [main] - DEBUG - c.d.tools.ide.log.IdeSubLoggerSlf4j - Configuration directory \non-existing-user-home-for-testing\.ide does not exist.
]]></system-out>
  </testcase>
```

With this story the effect should be reproduced, analysed and hopefully fixed.

hohwille:
fixed #618 (not yet)

* cleanup test and removed obvious nonsense
* still the `testProgressBarStyleSwitchOnNonWindows` that testes linux on Windows seems to cause 2 Minutes of think-time if called from a real shell. My question is what is the benefit of this if the costs are that high.
* This PR is a small but nice improvement. However, it does not solve the problem described in #618

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/io/AbstractIdeProgressBar.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/io/AbstractIdeProgressBar.java
@@ -17,9 +17,7 @@ public abstract class AbstractIdeProgressBar implements IdeProgressBar {
     this.maxLength = maxLength;
   }
 
-  /**
-   * @return the maximum length of a progress bar.
-   */
+  @Override
   public long getMaxLength() {
 
     return maxLength;
@@ -32,6 +30,15 @@ public abstract class AbstractIdeProgressBar implements IdeProgressBar {
    * @param currentProgress current progress state (used for tests only).
    */
   protected abstract void doStepBy(long stepSize, long currentProgress);
+
+  protected void stepTo(long stepPosition) {
+
+    if ((this.maxLength > 0) && (stepPosition > this.maxLength)) {
+      stepPosition = this.maxLength; // clip to max avoiding overflow
+    }
+    this.currentProgress = stepPosition;
+    doStepTo(stepPosition);
+  }
 
   /**
    * Sets the progress bar to given step position.
@@ -48,12 +55,18 @@ public abstract class AbstractIdeProgressBar implements IdeProgressBar {
       // check if maximum overflow
       if (this.currentProgress > this.maxLength) {
         this.currentProgress = this.maxLength;
-        doStepTo(this.maxLength);
+        stepTo(this.maxLength);
         return;
       }
     }
 
     doStepBy(stepSize, this.currentProgress);
+  }
+
+  @Override
+  public long getCurrentProgress() {
+
+    return this.currentProgress;
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


jan-vcapgemini:
I think the whole test can be omitted because currently it does nothing which we can track inside the test itself. Would need a more complicated solution for the capturing of the prompt with a check for the expected style.

hohwille:
> I think the whole test can be omitted because currently it does nothing which we can assert inside the test itself. Would need a more complicated solution for the capturing of the prompt with a check for the expected style.

At least we are testing the progress-bar does not crash and the steps are aggregated both in the abstract super-class as well that they arrive in the progress bar implementation. So I would still say the entire JUnit test class adds some value.

hohwille:
From CI logs:
```
[INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.253 s -- in com.devonfw.tools.ide.io.IdeProgressBarConsoleTest
```

So on linux it took `0.25` seconds what is perfectly fine.
Seems to be a Windows only problem. I will add a "conditional ignore".

cli/src/main/java/com/devonfw/tools/ide/io/AbstractIdeProgressBar.java
```diff
@@ -33,6 +31,15 @@ public long getMaxLength() {
    */
   protected abstract void doStepBy(long stepSize, long currentProgress);
 
+  protected void stepTo(long stepPosition) {
```
jan-vcapgemini, line 34: Please add Javadoc.

jan-vcapgemini (changes requested):
LGTM. Thanks for fixing my faulty additions. I've added just one CR (Javadoc).

jan-vcapgemini:
> > I think the whole test can be omitted because currently it does nothing which we can assert inside the test itself. Would need a more complicated solution for the capturing of the prompt with a check for the expected style.
> 
> At least we are testing the progress-bar does not crash and the steps are aggregated both in the abstract super-class as well that they arrive in the progress bar implementation. So I would still say the entire JUnit test class adds some value.

I've meant just the `testProgressBarStyleSwitchOnNonWindows` which you've removed already, not the whole class. So everything is fine now.

cli/src/main/java/com/devonfw/tools/ide/io/AbstractIdeProgressBar.java
jan-vcapgemini, line 34:
```suggestion
  /**
   * Sets the progress bar to given step position while making sure to avoid overflow.
   *
   * @param stepPosition position to set to.
   */
  protected void stepTo(long stepPosition) {
```

jan-vcapgemini (changes requested):

PR #622, opened 2024-09-17
#621: fixed ID fallback for XML merger

### LINKED ISSUE #621: Xml merger ID fallback not working
### Expected behavior

See #43, #403 etc. We specified the ID fallback as following:
https://github.com/devonfw/IDEasy/blob/main/documentation/configurator.adoc#element-identification

### Actual behavior

The implementation is wrong and never tested:
https://github.com/devonfw/IDEasy/blob/f296ef9df6074764ad00834fd73ae99585deef1e/cli/src/main/java/com/devonfw/tools/ide/merge/xmlmerger/matcher/ElementMatcher.java#L56-L71

While it tries to handle the cases it does not actually initialize a fallback at all and IMHO the only fallback for the name is incorrect.

1. write a XML template and omit `merge:id` attributes for elements that have name or id attribute
2. run the XML merger on that

### Related/Dependent Issues

#403

### Comments/Hints:

As stated above the fallback needs to be `name()`, `@id` or `@name` according to the attribute presence.

### Affected version:

* latest 2024.09.002-beta-SNAPSHOT

hohwille:
Fixes #621

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -8,6 +8,7 @@ Release with new features and bugfixes:
 
 * https://github.com/devonfw/IDEasy/issues/553[#554]: Missmatch of IDE_ROOT
 * https://github.com/devonfw/IDEasy/issues/557[#557]: Failed to update tomcat: Cannot find a (Map) Key deserializer for type VersionRange
+* https://github.com/devonfw/IDEasy/issues/621[#621]: Xml merger ID fallback not working
 
 The full list of changes for this release can be found in https://github.com/devonfw/IDEasy/milestone/13?closed=1[milestone 2024.09.002].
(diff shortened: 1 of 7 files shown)


jan-vcapgemini (commented):
LGTM. If possible it would be nice if you could also add another test for the repeated usage of merge:ids for similar elements.

hohwille:
> If possible it would be nice if you could also add another test for the repeated usage of merge:ids for similar elements.
I added a test for the bug I found and fixed. Your suggestion is not related to #621 
For the record: repeated usage of `merge:ids` is already covered by existing tests (e.g. `differentStrategies` or `intellijWorkspaces`).
So IMHO no further change required.

PR #624, opened 2024-09-17
#623: fix CliArgument prepend/append

### LINKED ISSUE #623: CliArgument prepand and append methods inconsistent
### Expected behavior

The class `CliArgument` has the methods `prepand` and `append`.
These methods should do what is implied by their name and JavaDoc.

### Actual behavior

Both method implementations are identical while according to name and JavaDoc they should behave differently.
Further, in our code the `prepend` method is used to actually append CLI arguments what seems inconsistent and confusing.

1. Write a test for prepend and append and test the expected behaviour.

### Comments/Hints:

Implementation can also be simplified.

### Affected version:

* current (2024.09.002-beta-SNAPSHOT)

hohwille:
Fixes #623 
Intentionally I did not change Eclipse from prepend to append since I think this is desired here.


slskiba (approved):
Thanks for the cleanup. Looks good 👍

jan-vcapgemini (approved):
LGTM. Ready for merge.

Issue #312, opened 2024-05-02
Ability to prefer `git` (ssh) protocol instead of `https` for cloning repos


hohwille:
As a IDEasy, I want to be able to configure my preference for `https` vs. `ssh` for cloning my projects so that I can benefit from authentication advantages.

Most git repositories can be cloned and synced both via `https` or `git` (`ssh`) protocol and developers have the freedom of choice between those two. See e.g. on github:
![image](https://github.com/devonfw/IDEasy/assets/1923119/8a87e8e0-35e7-4e75-b46f-49c4cc1e19fe)

Some enterprises host their git repos behind some general SSO for all `https` URLs that requires 2FA, cookies, etc.
Therefore, developers might have less trouble using `git` (`ssh`) protocol but in order to use it, you first need to do some manual preparations:
* generate an SSH keypair
* login manually into the git service and register the SSH public key in your profile
* potentially more stuff (e.g. ssh agent for unlocking private key with passphrase)

Therefore simple users can start easily with `https` that is default choice while advanced users often prefer `git` (`ssh`).

This story is about creating a property `PREFERRED_GIT_PROTOCOL` that allows to automatically configure a preferred git protocol.
The default is that this property is not defined and then the behave is like the current state where the git repository URL is used as is for cloning.
However, if this property is defined, then a git repository URL is converted to the preferred protocol:
* `PREFERRED_GIT_PROTOCOL=git` will automatically change `https` git URLs to `git`
* `PREFERRED_GIT_PROTOCOL=https` will automatically change `git` git URLs to `https`

The transformation can be seen from our github url:
* https://github.com/devonfw/IDEasy.git
* git@github.com:devonfw/IDEasy.git

So the prefix is either `https://` or `git@` and after the hostname there is a slash (`/`) for https but a colon (`:`) for git.
In case `PREFERRED_GIT_PROTOCOL` is defined and the git URL starts with its value, then nothing is to be done, otherwise the transformation.
One excuse I would see as hardcoded build-in is that for the hostname `github.com` we should not change the protocol, because we only use this for public OSS repositories that do not need any authentication at all and most of our users will not have a github account with SSH setup. So `https://github.com/...` will just work out of the box but via git typical users will end up with this:
```
Cloning into 'D:\projects\devonfw\workspaces\test\IDEasy'...
Warning: Permanently added 'github.com' (ED25519) to the list of known hosts.
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

[...]

PR #625, opened 2024-09-17
#312: Added ability to prefer `git` (ssh) protocol instead of `https`for cloning repos 

diiinesh:
Adds #312 

- Introduced `GIT_PREFERRED_PROTOCOL` property to allow configuration of preferred Git protocol (SSH or HTTPS).
- Default behavior remains unchanged if the property is not set, using the provided Git URL as is.
- Implemented GitUrlSyntax enum to manage and convert between SSH and HTTPS URL formats.
- Updated GitUrl class for URL parsing and protocol conversion based on the property.
- Added tests to validate URL conversion and ensure no changes for specific domains like GitHub.


cli/src/main/java/com/devonfw/tools/ide/context/GitUrl.java
```diff
@@ -11,6 +13,76 @@
  */
 public record GitUrl(String url, String branch) {
 
+  // Regex pattern to match SSH URLs
+  private static final Pattern SSH_URL_PATTERN = Pattern.compile("git@([a-zA-Z0-9._-]+):([a-zA-Z0-9._/-]+)\\.git", Pattern.CASE_INSENSITIVE);
```
hohwille, line 17: It is fragile to make assumptions about these segments for host, etc.
At least for the first segment/group there is a simple solution:
```suggestion
  private static final Pattern SSH_URL_PATTERN = Pattern.compile("git@([^:]+):([a-zA-Z0-9._/-]+)\\.git", Pattern.CASE_INSENSITIVE);
```

cli/src/main/java/com/devonfw/tools/ide/context/GitUrlSyntax.java
```diff
@@ -0,0 +1,99 @@
...
+   * <p>
+   * If the preferred protocol is not valid or the domain should not be converted, the original URL is returned.
+   * </p>
+   *
+   * @param gitUrl the original {@link GitUrl} to be converted.
+   * @param preferredProtocol the preferred protocol as a {@link String}, either "ssh" or "https".
+   * @return the converted {@link GitUrl} or the original {@link GitUrl} if no conversion is required.
+   * @throws IllegalArgumentException if the protocol is not supported.
+   */
+  public static GitUrl convertToPreferredProtocol(GitUrl gitUrl, String preferredProtocol) {
```
hohwille, line 75: Why using `String` for protocol?
```suggestion
  public static GitUrl convertToPreferredProtocol(GitUrl gitUrl, GitUrlSyntax syntax) {
```

Why not creating this as a non-static method in `GitUrl` instead?
```java
public class GitUrl ... {
  ...
  public GitUrl convert(GitUrlSyntax syntax) {
    ...
  }
}
```

cli/src/test/java/com/devonfw/tools/ide/context/GitContextTest.java
```diff
@@ -236,4 +236,65 @@ public void testIsRepositoryUpdateAvailableNoUpdates(@TempDir Path tempDir) {
...
   }
+
+  /**
+   * Tests the conversion of a Git URL from HTTPS to SSH protocol.
+   * <p>
+   * Given a Git URL in HTTPS format, this test ensures that it is correctly converted to the SSH format using the
+   * {@link GitUrlSyntax#convertToPreferredProtocol(GitUrl, String)} method.
+   */
+  @Test
+  public void testConvertGitUrlFromHTTPSToSSH() {
```
hohwille, line 247: https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc#naming
```suggestion
  public void testConvertGitUrlFromHttpsToSsh() {
```

cli/src/test/java/com/devonfw/tools/ide/context/GitContextTest.java
```diff
@@ -236,4 +236,65 @@ public void testIsRepositoryUpdateAvailableNoUpdates(@TempDir Path tempDir) {
...
+  }
+
+  /**
+   * Tests the conversion of a Git URL from SSH to HTTPS protocol.
+   * <p>
+   * Given a Git URL in SSH format, this test ensures that it is correctly converted to the HTTPS format using the
+   * {@link GitUrlSyntax#convertToPreferredProtocol(GitUrl, String)} method.
+   */
+  @Test
+  public void testConvertGitUrlFromSSHToHTTPS() {
```
hohwille, line 263: https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc#naming

cli/src/test/java/com/devonfw/tools/ide/context/GitContextTest.java
```diff
@@ -236,4 +236,65 @@ public void testIsRepositoryUpdateAvailableNoUpdates(@TempDir Path tempDir) {
...
+  }
+
+  /**
+   * Tests that the conversion does not alter the Git URL when no valid protocol is provided.
+   * <p>
+   * Given an SSH Git URL and a null preferred protocol, this test ensures that the original URL remains unchanged.
+   */
+  @Test
+  public void testConvertGitUrlNoValidProtocol() {
+    String url = "git@testgitdomain.com:devonfw/IDEasy.git";
```
hohwille, line 279: The protocol is valid here. This JavaDoc and method name seems inconsistent to me.

cli/src/main/java/com/devonfw/tools/ide/context/GitContextImpl.java
```diff
@@ -332,6 +330,26 @@ IdeContext getContext() {
...
+  /**
+   * Converts the given Git URL to the preferred protocol if a preferred protocol is set. If no protocol is set (null or empty) or the protocol is invalid, the
+   * original URL is returned.
+   *
+   * @param gitUrl the original {@link GitUrl} object.
+   * @return the converted {@link GitUrl} with the preferred protocol, or the original if no conversion is needed.
+   */
+  private GitUrl convertGitUrlToPreferredProtocol(GitUrl gitUrl) {
+    EnvironmentVariables envVars = this.context.getVariables().getByType(EnvironmentVariablesType.CONF);
+    String preferredProtocol = envVars.get("PREFERRED_GIT_PROTOCOL");
```
hohwille, line 343: Do not hardcode where a variable is configured.
```suggestion
    EnvironmentVariables envVars = this.context.getVariables();
    String preferredProtocol = envVars.get("PREFERRED_GIT_PROTOCOL");
```
Also we at least define a constant for such variable, but as a best practice create a variable in `IdeVariables` class and use it here. Then you can even make it typed to return `GitUrlSyntax` instead of `String`.

cli/src/main/java/com/devonfw/tools/ide/context/GitUrl.java
```diff
@@ -30,4 +102,11 @@ public URL parseUrl() {
     }
     return validUrl;
   }
+
+  private String[] getSshComponents() {
+    if (isSshUrl()) {
+      return parseSshLink();
+    }
+    return null;
```
hohwille, line 110: If we need to decompose the URL and want to have this here, we should already do that at construction time.
However, I would suggest to remove all this code here and move everything that is really needed into `GitUrlSyntax`.
Also then you can match the REGEX only once instead of 3 times (for `isSshUrl()`, `getHost()`, and `getPath()`).
Please note that matching a REGEX can be quite expensive. Also compiling so great that you already did this nicely via a `Pattern` constant.

hohwille (commented):
@diiinesh I already had a look even though still in draft mode.
You did already a great job with `GitUrlSyntax` and the integration into `GitContextImpl` plus adding a wonderful JUnit test. 👍 
I already left some suggestions for improvement and refactoring early on...
Please have a look.

cli/src/main/java/com/devonfw/tools/ide/context/GitContextImpl.java
hohwille: See here: 
https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/variable/IdeVariables.java

alfeilex (approved):
LGTM

hohwille (changes requested):
@diiinesh thanks for your rework. 👍 
I left some small review comments for improvement.
However, this PR does not integrate the new feature so story #312 is not yet implemented.
You need to change `GitContextImpl` to actually make use of the new infrastructure that you have created in order to transform the `GitUrl` before calling the actual git clone command.

Also I would suggest to add `GitUrlSyntax.DEFAULT` and provide this via default lambda in `IdeVariables` as default so that you can avoid NullPointerException and do not always need to check if configured `GitUrlSyntax` is `null`.

cli/src/main/java/com/devonfw/tools/ide/variable/IdeVariables.java
```diff
@@ -73,6 +75,11 @@ public interface IdeVariables {
   /** {@link VariableDefinition} for {@link com.devonfw.tools.ide.context.IdeContext#getProjectName() PROJECT_NAME}. */
   VariableDefinitionString PROJECT_NAME = new VariableDefinitionString("PROJECT_NAME", null, c -> c.getProjectName());
 
+  /** Preferred Git protocol (HTTPS/SSH) as defined by {@link GitUrlSyntax}. */
+  VariableDefinitionEnum<GitUrlSyntax> PREFERRED_GIT_PROTOCOL = new VariableDefinitionEnum<>("PREFERRED_GIT_PROTOCOL", null, GitUrlSyntax.class,
+      c -> GitUrlSyntax.HTTPS
```
hohwille, line 80: The default should be not to change the URL. Only if the user sets the property explicitly, the URL should be changed.
See #625:
> Default behavior remains unchanged if the property is not set, using the provided Git URL as is.

alfeilex: Resolved here: https://github.com/devonfw/IDEasy/pull/724#discussion_r1821054880

alfeilex: Done in https://github.com/devonfw/IDEasy/pull/724

hohwille:
This PR is replaced by PR #724 - hence closing.

Issue #628, opened 2024-09-17
ide update fails on first error


hohwille:
### Expected behavior

As a IDEasy, I want that ide update applies all updates so that I can continue working and even if one of the step fails this should not block other steps.

### Actual behavior

IDEasy stops on the first error (IMHO also the same for `ide create`).
This was already a problem in devonfw-ide and the reason why we invented steps (#282) what was too complex to implement in Bash.
Now with Java we have all this, however we do not properly apply the nice infrastructure we have created.

As you can read in #102:
> That is if we iterate over tools in 2.,3., and 4. and some tool update fails with an error, we should not abort the entire process and prevent other tools to get updated. Instead in Java we can now properly catch and collect the errors and include a report at the end of the process.

However, we do not properly test that all aspects of a story have really been implemented before closing and also never did a sprint review of that story.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. ide update (with a tool that for some reason fails on update)

```
$ ide update
Updates are available for the settings repository. If you want to pull the latest changes, call ide update.
Start: Pull settings repository
Successfully updated settings repository.
Start: Copy configuration templates
Copy configuration templates
Start: Install or update software
Start: Install node
Version v18.20.4 of tool node is already installed
Install node
Start: Install mvn
Version 3.9.9 of tool mvn is already installed
Install mvn
Start: Install aws
Trying to download aws-2.17.52-windows-x64.msi from https://awscli.amazonaws.com/AWSCLIV2-2.17.52.msi
Failed to download from https://awscli.amazonaws.com/AWSCLIV2-2.17.52.msi
java.lang.IllegalStateException: Failed to download file from URL https://awscli.amazonaws.com/AWSCLIV2-2.17.52.msi to D:\projects\_ide\tmp\Downloads\aws-2.17.52-windows-x64.msi
        at com.devonfw.tools.ide.io.FileAccessImpl.download(FileAccessImpl.java:121)
        at com.devonfw.tools.ide.repo.AbstractToolRepository.download(AbstractToolRepository.java:160)
        at com.devonfw.tools.ide.repo.AbstractToolRepository.download(AbstractToolRepository.java:94)
        at com.devonfw.tools.ide.repo.AbstractToolRepository.download(AbstractToolRepository.java:64)
        at com.devonfw.tools.ide.tool.LocalToolCommandlet.installTool(LocalToolCommandlet.java:222)
[...]

PR #638, opened 2024-09-19
#628: Fixed update fails on first error

jan-vcapgemini:
Fixes: #628 

Implements:
* added catch of exceptions for ToolCommandlet installations
* converted steps to try-with-resources
* added new test for failed update substeps
* added mvn and java repositories to update test-project

### DIFF
--- a/cli/src/test/java/com/devonfw/tools/ide/commandlet/UpdateCommandletTest.java
+++ b/cli/src/test/java/com/devonfw/tools/ide/commandlet/UpdateCommandletTest.java
@@ -56,4 +56,30 @@ class UpdateCommandletTest extends AbstractIdeContextTest {
     Files.delete(templates.getParent());
     Files.delete(templates.getParent().getParent());
   }
+
+  /**
+   * Tests if a sub step (installation of software) of update failed, the overall update process will not fail too.
+   * <p>
+   * See: <a href="https://github.com/devonfw/IDEasy/issues/628">#628</a> for reference.
+   */
+  @Test
+  public void testRunUpdateSoftwareDoesNotFailOnFailedSoftwareInstallations() {
+
+    // arrange
+    IdeTestContext context = newContext(PROJECT_UPDATE);
+    Path javaRepository = context.getToolRepositoryPath().resolve("default").resolve("java");
+    context.getFileAccess().delete(javaRepository);
+    Path javaDownload = context.getIdeRoot().resolve("repository").resolve("java");
+    context.getFileAccess().delete(javaDownload);
+    UpdateCommandlet uc = context.getCommandletManager().getCommandlet(UpdateCommandlet.class);
+
+    // act
+    uc.run();
+
+    // assert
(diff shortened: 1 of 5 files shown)


cli/src/test/java/com/devonfw/tools/ide/commandlet/UpdateCommandletTest.java
```diff
@@ -56,4 +56,30 @@ private void deleteTemplatesFolder(IdeContext context) throws IOException {
...
+    context.getFileAccess().delete(javaRepository);
+    Path javaDownload = context.getIdeRoot().resolve("repository").resolve("java");
+    context.getFileAccess().delete(javaDownload);
+    UpdateCommandlet uc = context.getCommandletManager().getCommandlet(UpdateCommandlet.class);
+
+    // act
+    uc.run();
+
+    // assert
+    assertThat(context).logAtError().hasMessage("Installation of java failed!");
```
hohwille, line 80: We could even assert on the step structure here via `this.context.getCurrentStep()`.
In `StepImpl` we could expose `getChildren()` for testing.
Just a thought since here we do not assert the status of aggregated steps nor the overall success or failure.
Also by thinking this further and just by reading the code I am now confident that there is an error and this will not work:
https://github.com/devonfw/IDEasy/blob/2e601cd6b0efcb7d6b77e88155543fa2e7cbb569/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L766-L779

First of all the overall status should not be success but error and we should get exit code != 0 in such case.
My assumption is even that we will fail with an assertion error if we succeed the root step if it has child steps that are not successful but I have not verified this. At least my expectation is that a step should only be successful if all its child steps are successful.
This is all not your fault and your PR and changes are great. It is just that this review reveals new related problems that I now get aware of. IMHO we should also fix that together. But if you want you can also create a new bug issue and merge this first...

hohwille (commented):
@jan-vcapgemini thanks for your PR. Great work with nice and consistent improvements. I very much appreciate if you care for clean code like improving all places to try-with-resource syntax and extending with a test of the according situation. 👍 

During the review I thought further and IMHO identified a deeper problem that goes beyond the initial demand of story #628. Could you have a look and we can then discuss if it makes sense to extend this PR or if we want to create a new story for that...?

hohwille:
> You have already added the issue implemented by your PR in [CHANGELOG.adoc](https://github.com/devonfw/ide/blob/master/CHANGELOG.adoc) to the next open release

cli/src/test/java/com/devonfw/tools/ide/commandlet/UpdateCommandletTest.java
jan-vcapgemini: I've create a new issue now: #639

hohwille: Thanks

Issue #640, opened 2024-09-20
ide-settings broken for IntelliJ or bug in XML merger


hohwille:
I updated my `IDEasy` project and wanted to start intellij:
```
$ ide intellij
Configuring workspace main for IDE intellij
An unexpected error occurred!
We are sorry for the inconvenience.
Please check the error below, resolve it and try again.
If the error is not on your end (network connectivity, lack of permissions, etc.) please file a bug:
https://github.com/devonfw/IDEasy/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=IllegalStateException%3A+No+merge%3Aid+value+defined+for+element+%2Fprofile%2Finspection_tool+in+document+D%3A%5Cprojects%5CIDEasy%5Csettings%5Cintellij%5Cworkspace%5Cupdate%5C.intellij%5Cconfig%5Cinspection%5CDefault.xml
java.lang.IllegalStateException: No merge:id value defined for element /profile/inspection_tool in document D:\projects\IDEasy\settings\intellij\workspace\update\.intellij\config\inspection\Default.xml
        at com.devonfw.tools.ide.merge.xmlmerger.model.MergeElement.getId(MergeElement.java:95)
        at com.devonfw.tools.ide.merge.xmlmerger.matcher.ElementMatcher.matchElement(ElementMatcher.java:51)
        at com.devonfw.tools.ide.merge.xmlmerger.MergeStrategy.combineChildNodes(MergeStrategy.java:155)
        at com.devonfw.tools.ide.merge.xmlmerger.MergeStrategy$1.merge(MergeStrategy.java:30)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMerger.merge(XmlMerger.java:103)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMerger.merge(XmlMerger.java:78)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:65)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:69)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:69)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:69)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:69)
[...]

PR #644, opened 2024-09-22
#640: added option --skip-workspace-update

hohwille:
implemented suggestion from #640 to add `--skip-workspace-update` option and fixed CLI arg parsing error.

Discussion:
* Should the option actually be named `--skip-configure-workspace`?
* Would it be better to define a step for workspace update instead and make only that step fail?
* Or should we just catch the potential exception if workspace update fails and ask the user to continue on error?


cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
```diff
@@ -18,6 +19,8 @@
  */
 public abstract class IdeToolCommandlet extends PluginBasedCommandlet {
 
+  public final FlagProperty skipWorkspaceUpdate;
```
jan-vcapgemini, line 22: Please add Javadoc.

jan-vcapgemini (approved):
LGTM.

jan-vcapgemini:
Please add new feature to changelog.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
hohwille: Since all that code was reverted and is gone, I will set this to resolved so we can proceed.

hohwille:
> Please add new feature to changelog.

Since that feature is not contained anymore and we only do internal changes here, no need to add to changelog in this PR.

Issue #627, opened 2024-09-17
Still log messages break processable command output


hohwille:
```
$ ide -v
Updates are available for the settings repository. If you want to pull the latest changes, call ide update.
2024.09.002-beta-SNAPSHOT
bash: Updates: command not found
```

I have fixed #531 already two times.
It seems that also issue #161 with PR #467 also broke this.
I still do not understand how this happens since I thought my second fix was bullet proof.
Is that output printed with `System.out` or what could I be missing?

alfeilex:
As I understand, the problem relies in https://github.com/devonfw/IDEasy/blob/7ac8cabede1fb1ebb4313b9c5e68d03a6105ba22/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L853 because after executing `env --bash`, bash thinks that that log message is an executable command. 

Maybe one of these solutions can help us to prevent that behavior: 
- Append `#` to the log message and treat it as a comment
- Redirect the info log level to `stderr` if `env --bash` is run
- Using some `sed`, `grep` workarounds but I don't know if that makes things more complicated

alfeilex:
I tried to use `interaction()` instead of `info()` but the result is the same. First of all `interaction` log level is supressed by 
https://github.com/devonfw/IDEasy/blob/7ac8cabede1fb1ebb4313b9c5e68d03a6105ba22/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L829-L830 and I had to extend the condition with `.. && level != IdeLogLevel.INTERACTION `

But the end result is the same after executing `ide -v` with output `bash: $'\E[96mUpdates': command not found`

Am I missing something?

hohwille:
> and I had to extend the condition with .. && level != IdeLogLevel.INTERACTION 

You should not do that. This code should deactivate all loglevels EXCEPT the Log-Level INFO.
Hence, when you are logging in INTERACTION instead of INFO, the output should be hidden.

BTW: Even smarter would be to move this check to a method that is executed only if the command to run returns `false` on `isProcessableOutput()`. In that case the log-level does not matter and we also do not waste performance to determine information from GIT that we then omit in the logging.

hohwille:
fixed by changing log-level.

PR #649, opened 2024-09-24
#627: Fix info-log message that breaks processable context output

alfeilex:
Fixes: #627 

This PR changes `info.log` to `interaction.log` message output to prevent an bug that causes the message to be interpreted as a command by bash


jan-vcapgemini (approved):
LGTM.

hohwille (approved):
Ready for merge.

Issue #608, opened 2024-09-11
Enhance error messsages of ProcessBuilder


jan-vcapgemini:
As an IDEasy user, I want to see what exactly failed when an application was not run successfully.

Currently IDEasy only shows the exit code of a failed process and omits the error messages in the log.

See: https://github.com/devonfw/IDEasy/blob/e73c7ae38f397c01fd8131bd62d0323b72b5a3d5/cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java#L284-L287

We could easily add `result.getErr()` here too.

UPDATE (JH): This should depend on `ProcessMode` so e.g. for `DEFAULT` all output was already redirected to the shell. However, for `DEFAULT_CAPTURE` we should output what has been captured (IMHO both stdout and stderr).
Somehow this is a duplicate or at least strongly overlaps with #593.

hohwille:
What is the status here?
According to #593 most of this issue is already implemented.

hohwille:
IMHO #593 problem went away since output is not captured anymore.
But in case of capture this story still needs to be implemented.

PR #653, opened 2024-09-25
#608: enhance error messages

jan-vcapgemini:
Fixes: #608 

Implements:
* reformatted process out and error messages
* added test with expected Exception and error messages
* performLogOnError will only be used on DEFAULT_CAPTURE
* added Npm run test (disabled)
* adjusted Dotnet run test (Windows only, does not work on CI)


hohwille:
Sorry for interfering with PR #652. I am not sure if #608 is already fixed. I assume I only fixed it partially.
What if we have `DEFAULT_CAPTURE` combined with `THROW_CLI`?
IMHO then still the output would not be printed so we still need a small change for #652.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -313,22 +316,34 @@ private String addExecutable(String exec, List<String> args) {
   private void performLogOnError(ProcessResult result, int exitCode, String interpreter) {
 
     if (!result.isSuccessful() && (this.errorHandling != ProcessErrorHandling.NONE)) {
-      String message = createCommandMessage(interpreter, " failed with exit code " + exitCode + "!");
+      IdeSubLogger subLogger = this.context.error();
+      IdeLogLevel ideLogLevel;
+      String message = createCommandMessage(interpreter, "\nfailed with exit code " + exitCode + "!");
+
       if (this.errorHandling == ProcessErrorHandling.THROW_CLI) {
+        subLogger.log(message);
```
hohwille, line 324: if we still hardcode the log-level to `error` why then hiding this information making the code in-transparent?
```suggestion
        this.context.error(message);
```

cli/src/main/java/com/devonfw/tools/ide/process/ProcessResultImpl.java
```diff
@@ -55,4 +58,25 @@ public List<String> getErr() {
...
 
+  @Override
+  public void log(IdeLogLevel level, IdeContext context) {
+
+    if (this.out != null && !this.out.isEmpty() && level == IdeLogLevel.INFO) {
+      doLog(level, this.out, context);
+    }
+    if (this.err != null && !this.err.isEmpty() && level == IdeLogLevel.ERROR) {
+      doLog(level, this.err, context);
+    }
```
hohwille, line 69: You are not implementing what you specified in the JavaDoc.
Why should you ignore `out` and `err` if the passed log-level is not as expected/hardcoded here?
This code is really confusing me...

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -313,22 +316,34 @@ private String addExecutable(String exec, List<String> args) {
...
+        result.log(IdeLogLevel.ERROR, context);
         throw new CliProcessException(message, result);
       } else if (this.errorHandling == ProcessErrorHandling.THROW_ERR) {
+        subLogger.log(message);
+        result.log(IdeLogLevel.ERROR, context);
         throw new IllegalStateException(message);
       }
```
hohwille, line 331: Why not just moving this code after the logging code below?
Than we have two clear aspects:
1. logging of errors with log-level according to error handling
2. throwing exception according to error handling

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -313,22 +316,34 @@ private String addExecutable(String exec, List<String> args) {
...
+        subLogger = this.context.warning();
       } else {
-        level = this.context.error();
+        ideLogLevel = IdeLogLevel.ERROR;
+        subLogger = this.context.error();
         this.context.error("Internal error: Undefined error handling {}", this.errorHandling);
       }
-      level.log(message);
+
+      subLogger.log(message);
```
hohwille, line 345: why first logging on the `subLogger` and then again via `result.log`?
I like that we introduce this `result.log` method to centralize the logging of `ProcessOutput`.
However, this can result in duplicated log output and if we have many lines or even pages of output, we do not want to get it twice.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -313,22 +316,34 @@ private String addExecutable(String exec, List<String> args) {
...
+      result.log(ideLogLevel, context);
```
hohwille, line 346: do we probably need two log-levels? One for `out` and one for `err` or should `err` always go on log-level `ERROR`?

cli/src/test/java/com/devonfw/tools/ide/tool/npm/NpmTest.java
```diff
@@ -37,13 +38,41 @@ public void testNpmInstall(String os) {
...
+
+    // act
+    commandlet.run();
+
+    // assert
+    if (context.getSystemInfo().isWindows()) {
+      assertThat(context).logAtInfo().hasMessage("npmcmdbin ");
+    } else {
+      assertThat(context).logAtInfo().hasMessage("npmcmd ");
+    }
```
hohwille, line 66: I am not really getting this but this looks odd. I would not like to assert and expect that such strange output is logged.

hohwille (changes requested):
@jan-vcapgemini thanks for your PR. You covered the relevant spots and did some good improvements. 👍 
However, some things here IMHO need some rework.
Further, `ProcessContextTestImpl` needs to be adopted since you now have added reasonable logging and you should remove the hack there to log the output again.

cli/src/test/java/com/devonfw/tools/ide/tool/npm/NpmTest.java
jan-vcapgemini: After executing the npm.bat, this message will be shown on the log (basically echo of the bat script).

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
jan-vcapgemini: Adjusted.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessResultImpl.java
jan-vcapgemini: Optimized and removed check for log level now.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
jan-vcapgemini: Move throwing of exceptions after logging.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
jan-vcapgemini: Removed the subLogger entirely.

jan-vcapgemini: I've used the subLogger only for the message of the exitcode the result.log logs err and out only. Need to re-add it now.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -313,22 +315,28 @@ private String addExecutable(String exec, List<String> args) {
...
-        throw new IllegalStateException(message);
-      }
-      IdeSubLogger level;
+      IdeLogLevel ideLogLevel;
+      String message = createCommandMessage(interpreter, "\nfailed with exit code " + exitCode + "!");
+
       if (this.errorHandling == ProcessErrorHandling.LOG_ERROR) {
-        level = this.context.error();
+        ideLogLevel = IdeLogLevel.ERROR;
+        context.error(message);
```
hohwille, line 323: IMHO the log-level is the only code dependent on the `ProcessErrorHandling` in order to control the log-level at which the message is logged after the `if` block.
Why do you still log the message manually in each condition block?
```
this.context.level(ideLogLevel).log(message);
```
BTW: The enum `ProcessErrorHandling` could define an abstract method `getLogLevel()` so that I do not need the `if ... else if ...` logic here and when creating a new enum constant, I get a compile error unless I implement `getLogLevel()`, so I do not accidentally end up in the `else` block with probably undesired behaviour.

cli/src/main/java/com/devonfw/tools/ide/tool/eclipse/Eclipse.java
```diff
@@ -84,21 +83,11 @@ public void installPlugin(ToolPluginDescriptor plugin, Step step) {
         }
       }
     }
-    log(IdeLogLevel.DEBUG, result.getOut());
-    log(IdeLogLevel.ERROR, result.getErr());
+    result.log(IdeLogLevel.DEBUG, context);
+    result.log(IdeLogLevel.ERROR, context);
```
hohwille, line 87: before, `out` and `error` have been logged on different log-levels.
Now, you IMHO log both on both levels causing duplications.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
jan-vcapgemini: Removed subLogger and just used result.log now instead.

cli/src/test/java/com/devonfw/tools/ide/tool/npm/NpmTest.java
hohwille: I disagree here. This is not what we are doing there:
https://github.com/devonfw/IDEasy/blob/main/cli/src/test/resources/ide-projects/npm/repository/npm/npm/default/windows/bin/npm

Therefore this should not be the expected result.
What we are asserting here is actually buggy output appearing for yet unclear reasons.
We should not assert buggy behaviour but what is really expected behaviour.

hohwille:
There are tests revealing that something is still broken:

> [ERROR]   MvnTest.testMvnRun:69 Could not find log message equal to 'mvn foo bar' on level INFO
[ERROR]   TomcatTest.testTomcat:50 Found expected log entries:
SUCCESS:Successfully installed java in version 8u402b06
The first entry that was not matching from a block of 10 expected log-entries was:
INFO:OpenJDK version 8u402b06

The message `OpenJDK version 8u402b06` that is the output of the faked `java --version` is not logged anymore so the process output is somehow not logged here anymore. Probably we have the same problem with `mvn` where the test now also failed.

cli/src/test/java/com/devonfw/tools/ide/tool/npm/NpmTest.java
hohwille: Thanks for creating issue #700 and linking everything together.
I will therefore resolve this conversation and we will address it outside of this PR. 👍

cli/src/test/java/com/devonfw/tools/ide/context/ProcessContextTestImpl.java
```diff
@@ -21,7 +22,11 @@ public ProcessContextTestImpl(IdeContext context) {
 
   @Override
   public ProcessResult run(ProcessMode processMode) {
-
-    return super.run(ProcessMode.DEFAULT_CAPTURE);
+    ProcessResult result = super.run(ProcessMode.DEFAULT_CAPTURE);
+    // this hack is still required to capture test script output
+    if (processMode == ProcessMode.DEFAULT || processMode == ProcessMode.BACKGROUND) {
```
hohwille, line 27: this should only happen on success since in error case it was already logged.
```suggestion
if (result.isSuccessful() && (processMode == ProcessMode.DEFAULT || processMode == ProcessMode.BACKGROUND)) {
```

cli/src/test/java/com/devonfw/tools/ide/tool/tomcat/TomcatTest.java
```diff
@@ -37,9 +38,9 @@ public void testTomcat() {
     Path javaTomcatPath = context.getToolRepositoryPath().resolve("default").resolve("java").resolve("java").resolve(javaVersionTomcat);
 
     // act
-    javaCommandlet.arguments.addValue("--version");
-    javaCommandlet.run();
-    tomcatCommandlet.run();
+    // hack to capture output of test script. Original processMode was BACKGROUND.
```
hohwille, line 41: For the record: IMHO you wrote this before you re-introduced the fallback to DEFAULT_CAPTURE for all tests. This leftover message is now slightly incorrect and could be confusing.

hohwille (approved):
@jan-vcapgemini thanks for completing this rather complicated and underestimated story 👍 
Have a look at the remaining comments. Then we can merge. Great work 🥇

cli/src/test/java/com/devonfw/tools/ide/tool/tomcat/TomcatTest.java
jan-vcapgemini: I've reverted these hacks now.

PR #658, opened 2024-09-27
#657: fix installation of java 8

### LINKED ISSUE #657: Cannot install Java 8
### Expected behavior

As a IDEasy user, I want to be able to install an older version of Java like Java 8 so that I can still use IDEasy for projects with legacy that require an older Java.

### Actual behavior

```
$ ide install java 8u412b08
8u412b08    8u412b08*   8u412b08*!
 
hohwille@CE49454 MSYS /d/projects/demo9
$ ide install java 8u412b08
Error in property version:
Given version 8u412b08 is not a valid version
 
Error: IDEasy failed with exit code 1
```

1. `ide install java 8u412b08`

### Related/Dependent Issues

#569 

### Comments/Hints:

According to our implementation of `VersionIdentifier` the version `8u412b08` was never valid since the segment `u412` is invalid since `u` is an undefined development phase.
In PR #569 a property validator was added to `VersionProperty` that requires the version to be valid.
Seemed to make sense but currently blocks the usage.

### Affected version:

* 2024.09.002-beta-SNAPSHOT (from today)

hohwille:
fixes #657 
* improved VersionIdentifierTest and also make the problem transparent
* removed version validator to fix the actual bug #657


jan-vcapgemini (approved):
Looks good to me, ready to merge. Will we reactivate it later? if yes, we need to add a new issue with the removed code part.

PR #666, opened 2024-09-30
#652: properly report step success/error

hohwille:
addon to #652 that introduced step to plugin installation but code needs to report success/error to step now


jan-vcapgemini (approved):
LGTM.

Issue #669, opened 2024-09-30
Make JUnit Tests reliable again


hohwille:
As a IDEasy developer, I want to have deterministic JUnit tests so that I can rely on their results and CI status.

It all started with #648 where we changed our generated start-script that delegates to the actual tool to run in background.

Since that is now processing asynchronous in the background the process may terminate for Java before the output is printed and received.
As a workaround we started writing the output to a file instead of echoing it and assuming that it will be captured and logged.
Therefore also the assertions all had to be refactored to assert the message not from the logs anymore but from the file where the script is writing the output to.
Now in #667 we had the effect that even that failed because the file was created in background while the JUnit in the foreground ran the assertion for the file before it was actually created in the background.

Thoughts:
* For KISS we could question if the `&` is really that important for the end-user or if we simply want to prevent all this background and async processing. Still we start the real process in BACKGROUND mode - but however in JUnits we mock the process mode in `ProcessContextTestImpl` so we never actually run processes directly with `ProcessMode.BACKGROUND`.
* For maintainability we should also consider more design for writing JUnits and stop the copy&paste mentality also in tests: We need to understand that we have a common pattern in all these tests and instead of copy&pasting this to many many test methods, we need to write reusable code for such checks. Then we can adjust in a central place to stabilise the tests if they are flickering instead of finding all the copy&pasted spots spread around multiple JUnit classes and methods.

hohwille:
I assume the error will come back. We can simply reopen and remove milestone if that happens.

jan-vcapgemini:
Issue still occurs as seen in: https://github.com/devonfw/IDEasy/actions/runs/11340850609/job/31537989280?pr=674
So we need to re-open this issue.

julia-cap:
Bug does no occur anymore, therefore closing this issue again.

PR #671, opened 2024-10-01
#669: Attempt to fix background script

jan-vcapgemini:
Fixes: #669 

Implements:
* moved checkInstallation before run test validation


hohwille (approved):
@jan-vcapgemini thanks for the PR. Surely a quickfix of #669 👍 
IMHO not a long term solution but will surely reduce the probability / likelihood that the test fails due to async effects.

Happy to merge asap.

Issue #315, opened 2024-05-06
Process stays hanging after installing Docker


salimbouch:
### Expected behavior

As a user of IDEasy, I want to be able to install global tools (like docker) and get a response that it was installed successfully when it's done.

### Actual behavior

The global tool (like docker) is installed, but the process stays hanging and the user can't tell whether the tool was installed successfully or not.

### Steps to reproduce (bug) 

1. ide install docker
 (tested with rancher desktop)

### Comments/Hints:

when debugging the process stays hanging in `run()` in [GlobalToolCommandlet](https://github.com/devonfw/IDEasy/blob/d16b3828562cae2550783b61150363fea375761e/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java#L134C17-L134C29)

### Affected version:
tested on Windows

alfeilex:
Right now the installation of `Rancher Desktop` needs user interactions for installation. After the installation, the process of `msiexec` not terminated properly and the ide process hangs, because it's waiting.

PR #684, opened 2024-10-09
#315: Fix frozen process after global tool installation with msiexec

alfeilex:
Fixes: #315 

~~This PR adds an argument for `msiexec` to install `msi` files. With `/qb` the installation requires no user interaction, but there is still a UI to see the progress of the installation. This way the process no longer freeze~~

This PR change the installation process of global tools so that they are run in the background.

Affected tools: `docker, pgadmin`


cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -304,6 +304,7 @@ private String addExecutable(String exec, List<String> args) {
     }
     if ("msi".equalsIgnoreCase(fileExtension)) {
       args.add(0, "/i");
+      args.add(0, "/qb");
```
hohwille, line 307: https://docs.revenera.com/installshield25helplib/helplibrary/IHelpCmdLineMSI.htm
> qb creates a basic user interface

might fix the problem but in case we need to run an installer for a global tool like docker why don't we just run that in `BACKGROUND`? I do not know if this `/qb` option fixes this bug but adds another bug to some other tool using an MSI...

hohwille (approved):
@alfeilex thanks for testing, filing bug and creating PR (including CHANGELOG update) 👍 
I left a comment with an alternative suggestion to fix this.
If you agree, please update the PR.
In case you disagree, we could at least explain and document more about why we magically add this `/qb` option.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
alfeilex: revert

alfeilex:
If the installation process of global tools should happen in the background, then this implementation doesn't make sense, does it?  We are waiting for an exit code to proceed but if the setup is done by the user in the background, the IDE should output something like `Setup of <tool> started.` without any success or error output. But what are we doing with `postInstall()`. Is it even necessary for a global tool when the user is handling the entire installation?

https://github.com/devonfw/IDEasy/blob/54a8b476ffbcc12f64041fadf2fe3a571e03662b/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java#L134-L147

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -131,7 +132,7 @@ public boolean install(boolean silent, EnvironmentContext environmentContext) {
       executable = fileAccess.findFirst(downloadBinaryPath, Files::isExecutable, false);
     }
     ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.LOG_WARNING).executable(executable);
-    int exitCode = pc.run();
+    int exitCode = pc.run(ProcessMode.BACKGROUND).getExitCode();
```
alfeilex, line 135: If the process runs in background mode, then message `Successfully installed docker in version 1.16.0` will be output immediately after the setup installer window has opened.

alfeilex: Could switching to background mode cause problems for installations on other operating systems? Should it only be adapted for windows MSI installations?

alfeilex:
> But what are we doing with `postInstall()`

https://github.com/devonfw/IDEasy/blob/071e732e6c74aed7b12ccc1d522faf1c4a015dad/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java#L142-L144

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
hohwille: We can change the log message but this is the only reasonable behaviour. We trigger the installation and the end-user needs to click through the installation process. This can take a minute or several hours if the user has fallen asleep. Not much that we can do about it...

PR #688, opened 2024-10-11
#687: prevent checksum recheck

### LINKED ISSUE #687: UrlUpdater rechecks all URLs if one OS is missing
# Problem
From https://github.com/hohwille/IDEasy/pull/1:
Here @alfeilex discovered an important flaw of `AbstractUrlUpdater`:
https://github.com/devonfw/IDEasy/blob/3e6fc6529157613639caa8d00a4e09d4a132e67f/cli/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java#L621-L624

If we think that an URL is missing for an OS, we then revalidate all URLs of that version causing a lot of overhead and waste.
In the worst case, that URL is broken and this revalidation will happen every night.
This can accumulate to a huge amount of time, bandwidth and compute being wasted.

At least we have to call the individual logic of the updater from `addVersion` since that contains the tool specific logic to generate the URLs for the tool and that is calling `doAddVersion` what finally calls `checkDownloadUrl`.
https://github.com/devonfw/IDEasy/blob/3e6fc6529157613639caa8d00a4e09d4a132e67f/cli/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java#L268-L287

# Suggested solution
We should rename `checkDownloadUrl` since that method is also adding the URL (e.g. to `addAndVerifyUrlIfNew`).
Please note that the adding of the URL is currently hidden here:
https://github.com/devonfw/IDEasy/blob/3e6fc6529157613639caa8d00a4e09d4a132e67f/cli/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java#L327

We should then first check if the URL is already contained in the `UrlDownloadFile` and if yes immediately return without doing anything else.

hohwille:
fixes #687

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/UrlStatusState.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/UrlStatusState.java
@@ -21,7 +21,17 @@ public final class UrlStatusState {
    */
   public UrlStatusState() {
 
-    this.timestamp = Instant.now();
+    this(Instant.now());
+  }
+
+  /**
+   * The constructor.
+   *
+   * @param timestamp the {@link #getTimestamp() timestamp}.
+   */
+  public UrlStatusState(Instant timestamp) {
+
+    this.timestamp = timestamp;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


hohwille:
@jan-vcapgemini since a test that was written by you and that failed after my fix and that I finally decided to replace with 3 test methods instead of 1, you should do a review on it.
Since I need to merge this asap due to the impact of #687, I would still ask you to do the review next week after this has already been merged. If you find any concerns, feel free to create another issue or PR. Thanks in advance.

Issue #423, opened 2024-06-25
Create installer for MacOS


hohwille:
As a Mac user, I want to have a simple installer for IDEasy so that I get guided through the installation and cannot get it wrong.

Analog to #420 we want to have the same thing for MacOS (Installer with the same checks and screens)
* license agreement
* installation location defaulting to `~/projects` (due to MacOS permission restrictions home directory seems the best default)
* Installation with
  * Extending PATH with `:$IDE_ROOT/_ide/bin`
  * Setting IDE_ROOT to $IDE_ROOT
  * extending `~/.bashrc` like described in #420
  * install workflows for Finder integration

For the integration workflows see old devonfw-ide solutions:
* https://github.com/devonfw/ide/tree/master/scripts/src/main/resources/system/mac/terminal/Open_devonfw-Terminal_here.workflow
* https://github.com/devonfw/ide/tree/master/scripts/src/main/resources/system/mac/iterm/Open_devonfw-iTerm_here.workflow (only if iterm2 is installed)

hohwille:
Retested with official Release 2026.08.001 on my MacBookPro:
I uninstalled IDEasy, downloaded the PKG and ran it.
Still is not signed and I had to use gatekeeper workaround (see #451) but after using the common workaround the installation worked perfect.
I even checked the details of the license and could find the latest changes we did and also the entire table with the licenses was readable.
After the installation IDEasy worked fine.
Great job 👍

PR #693, opened 2024-10-15
#693: Setup file not working on Mac

blizzarac:
Tested on mac

1. ide alias is not working like this -> Changing alias from source to ide to directly to ide
2. IDE_URL was missing -> Added for current folder


hohwille:
@blizzarac thanks for testing IDEasy (in beta phase already) and trying to help by creating a PR 👍 
Could you please first describe the error you were experiencing on MacOS?
It is definitely wrong to remove the `source` statement from the script.
I can explain that in detail why we need to source the script (see also #63) but first I would like to understand what was going wrong.

> IDE_URL was missing -> Added for current folder

That is definitely a known issue for Linux and Mac you we were already thinking about fixing this in the way you did.
However, currently we have a OS specific solution for that on Windows setting this in the Windows System Environment Variables after calling the setup Bash script (see https://github.com/devonfw/IDEasy/blob/main/cli/src/main/package/setup.bat). Therefore we should add an OS check and avoid doing that on Windows.
```
if [ "${OSTYPE}" = "cygwin" ] || [ "${OSTYPE}" = "msys" ]; then
...
else
...
fi
```
Do you want to do that and update your PR? Otherwise, we are happy to take over your work and create a new PR fixing that and hopefully include that into the planned 2024.10.001-beta release.

hohwille:
See also #423

hohwille:
@blizzarac as you seem to be quite busy with other stuff, I would suggest the following:
* we create another PR for adding `IDE_ROOT` during setup for Linux and MacOS.
* you only need to provide a small explanation of the error that you were facing that lead you to removing the `source` instruction. If there is some other "bug" than the missing IDE_ROOT, we can fix that. A comment would even be sufficient if you have no time. Otherwise creating a new bug would be even better.

blizzarac:
@hohwille sorry about, was a busy week. I think the problems that I had, had to do with my chaotic setup before I talked to @jan-vcapgemini and found out that I had to put everything in a _ide folder. Maybe this is the bug here (this was missing in the documentation)

blizzarac:
I have pushed a new version removing the source parts and added the IDE_ROOT with OS guards, hope that works for you

hohwille (approved):
@blizzarac thank you so much for the update of your PR. This is now perfect and ready for merge 👍

Issue #689, opened 2024-10-11
fix ide-projects to be consisent for tests


hohwille:
In PR #665 I introduced an improvement of our test infrastructure by removing some hack.
There I added this log message to spot and highlight inconsistent ide-projects and tests:
https://github.com/devonfw/IDEasy/blob/51a16e1e9e58ef79ed78f4b731b74066fac63ae2/cli/src/test/java/com/devonfw/tools/ide/repo/ToolRepositoryMock.java#L39

If you now run all our tests you will find lots of such exceptions spamming our logs.
We need to get rid of this since it is indicating inconsistent test projects.

In order to fix this all we need to do is create the according tool versions in `urls` of the according project.
The tool versions missing are reported in the exception message.
Examples how to fix this are included in PR #665 and can therefore be found in the diff (Files changed tab).

The job of this story is to complete this cleanup until all such errors are gone.
If that is the case, then please also remove the try catch block from the above linked code point so that in the future no further such inconsistencies can be created.

jan-vcapgemini:
There are still some tests with missing urls.
Please check the latest build log and search for "Invalid test project using version".
We also need to remove the try-catch block when all tests were adjusted.

PR #695, opened 2024-10-15
#689: Added urls to tests

leonrohne27:
fixes:#689

Implements:  
* Added Urls to tests to be consistent


jan-vcapgemini (approved):
@leonrohne27 thanks for the addition of the missing URLs. This will help us a lot. Finally no more cluttered logs.

jan-vcapgemini:
@leonrohne27 I was a bit too fast in merging this PR it seems. It appears that there are quite some tests still missing.
You can check out the latest build log and search for "Invalid test project using version".
We also need to remove the try catch block as stated in the issue.

PR #698, opened 2024-10-16
#689: Added url and removed try-catch-block

leonrohne27:
Fixes: #689
Added url to ide-projects/build and removed try-catch-block


hohwille (commented):
@leonrohne27 thanks for your PR and fixing this. Great work: you removed the "hack" and all tests are still green. 👍 
Before we merge, I would like to discuss how we handle such test `urls` files.
Its seems you have copied the `urls` file of Java to all the other tools and then the URL in there does not make any sense. So:
* should we either have a generic dummy URL in all these files?
* or should we have something reasonable there? And if so should that include both the tool and its version in the URL? Or is that all pointless since that URL is most likely never ever relevant in the test and will always be ignored.

I was quite lax in the beginning but after we ended up in a partial mess of quite inconsistent test-data that sooner or later broke our tests, I want to think a step further.
However, IMHO it is also fine to go for a simply dummy URL for all these files but lets discuss and agree in the team first.
Apart from that all is fine and ready for merge.

hohwille:
> Before we merge, I would like to discuss how we handle such test `urls` files...

I have created issue #706 for the idea of serving the software as compressed archive via wiremock.
As a preparation the suggested format for the `urls` files is as written in that issue:
```
${testbaseurl}/download/«tool»/«edition»/«version»/«tool»-«version».tgz
```
The values for `«tool»`, `«edition»` and `«version»` need to be filled in while `${testbaseurl}` remains as variable in the files and will later be resolved dynamically by implementation of #706.

cli/src/test/resources/ide-projects/dotnet/_ide/urls/dotnet/dotnet/6.0.419/urls.txt
jan-vcapgemini, line 1: Shouldn't git detect a rename instead of a delete here instead?

jan-vcapgemini (approved):
LGTM.

hohwille (approved):
@leonrohne27 thanks for changing all the URLs. 👍 
For Java I changed `.jar` to `.tgz` extension.
Now ready for merge.

Issue #81, opened 2023-10-10
Implement ToolCommandlet for Kubernetes


hohwille:
Create ToolCommandlet for Kubernetes (`kubectl`)
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/kubectl
ATTENTION: Blocked by issue #16 that should be implemented first or together with this story.
NOTE: there is no installation for kubernetes/kubectl itself so we need to override `install(boolean)` and just delegate installation to `Docker` without calling `super.install(silent)`.

slskiba:
Some work done at https://github.com/cwieschnat/IDEasy/tree/feature/81-implement-tool-commandlet-for-kubernetes

PR #702, opened 2024-10-18
#81: Implement ToolCommandlet for Kubernetes

KianRolf:
implemented #81 ToolCommandlet for Kubernetes.

Discussion:

- Has no test, since it is a GlobalToolCommandlet.


cli/src/main/resources/nls/Help.properties
```diff
@@ -62,6 +62,8 @@ cmd.kotlinc=Tool commandlet for Kotlin (compiler for JRE language).
 cmd.kotlinc.detail=Kotlin Compiler (kotlinc) is the command-line tool for compiling Kotlin code. Detailed documentation can be found at https://kotlinlang.org/docs/home.html
 cmd.kotlincnative=Tool commandlet for Kotlin-Native (compiler for JRE language).
 cmd.kotlincnative.detail=Kotlin/Native Compiler (kotlincnative) compiles Kotlin code to native executables. Detailed documentation can be found at https://kotlinlang.org/docs/reference/native-overview.html
+cmd.kubectl=Tool commandlet for kubernetes. Detailed documentation can be found at https://kubernetes.io/docs/home/
+cmd.kubectl.detail=The kubectl commandlet allows to install and use kubernetes. On Windows WSL 2 (Windows Subsystem for Linux) has to be installed properly as a prerequisite. The setup on windows will then install kubernetes with K3D. K3D will create a cluster with a single node with a default name as "ideasy-cluster"
```
hohwille, line 66:
```suggestion
cmd.kubectl.detail=The kubectl commandlet allows to install and use kubernetes. On Windows WSL 2 (Windows Subsystem for Linux) has to be installed properly as a prerequisite.
```

I suggest to remove the last 2 sentences since it is wrong. Rancher-Desktop or Docker-Desktop installs kubernetes. What "implementation" they use will depended on the edition (rancher vs. docker desktop) and already changed over time so it also depends on the version. Also I am not aware that we create a custom cluster name anywhere.

cli/src/main/java/com/devonfw/tools/ide/tool/kubectl/KubeCtl.java
```diff
@@ -0,0 +1,28 @@
...
+  public KubeCtl(IdeContext context) {
+
+    super(context, "kubectl", Set.of(Tag.DOCKER));
+  }
+
+  @Override
+  public boolean install(boolean silent, EnvironmentContext environmentContext) {
+    // TODO create kubectl/kubectl/dependencies.json file in ide-urls and delete this method
+    return getCommandlet(Docker.class).install(silent, environmentContext);
+  }
```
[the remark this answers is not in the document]
hohwille: We will have the same pattern multiple times.
Either we consider something like this as a proxy commandlet (create abstract class?) and just define the actual delegate commandlet via some protected method that we override or we would override a protected method like `protected boolean skipOwnInstallation()`. IMHO we should discuss this with the team in a next meeting.

hohwille (changes requested):
@KianRolf thanks for implementing #81. 
You properly implemented the steps to create a new commandlet. 👍 
I left some comments to improve and we still need some discussion here to complete this.

cli/src/main/java/com/devonfw/tools/ide/tool/kubectl/KubeCtl.java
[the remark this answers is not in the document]
hohwille: We can agree that here is a case where we do not need `dependencies.json` and a hardcoded dependency makes sense. There is no kubectl in ide-urls and there is no need to ever add it. This is and will always be a commandlet delegating to `Docker`.
Did you already try to implement such abstract class `DelegatingCommandlet` (if you think `ProxyCommandlet` might be confusing since it has nothing to do with an HTTP proxy)?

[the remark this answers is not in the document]
KianRolf: Yes, i already implemented a ProxyCommandlet but i am not sure if i have done so correctly. I would like to discuss it in the next meeting

cli/src/main/java/com/devonfw/tools/ide/tool/DelegatingToolCommandlet.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.Set;
+
+import com.devonfw.tools.ide.common.Tag;
+import com.devonfw.tools.ide.context.IdeContext;
+import com.devonfw.tools.ide.process.EnvironmentContext;
+
+public abstract class DelegatingToolCommandlet extends ToolCommandlet {
+
+  public DelegatingToolCommandlet(IdeContext context, String tool, Set<Tag> tags) {
```
hohwille, line 11: Either in constructor or via a `protected abstract` method you need to request the commandlet to delegate to.
I would even consider using a generic type `<D extends ToolCommandlet>` and then you can add:
```java
protected abstract Class<D> getDelegateClass();
```

This allows you to centralize the workaround:
```java
@Override
  public final boolean install(boolean silent, EnvironmentContext environmentContext) {
    return getCommandlet(getDelegateClass()).install(silent, environmentContext);
  }
```

jan-vcapgemini, line 11: Please add Javadoc.

jan-vcapgemini (changes requested):
Please add an install and run test for kubectl too. I've added some CRs (mostly missing javadocs)

cli/src/main/java/com/devonfw/tools/ide/tool/DelegatingToolCommandlet.java
```diff
@@ -0,0 +1,35 @@
...
+   */
+  public DelegatingToolCommandlet(IdeContext context, String tool, Set<Tag> tags, Class<D> delegateClass) {
+
+    super(context, tool, tags);
+    this.delegateClass = delegateClass;
+  }
+
+  @Override
+  public final boolean install(boolean silent, EnvironmentContext environmentContext) {
+    return getCommandlet(delegateClass).install(silent, environmentContext);
```
jan-vcapgemini, line 32:
```suggestion
    return getCommandlet(this.delegateClass).install(silent, environmentContext);
```

CHANGELOG.adoc
```diff
@@ -7,7 +7,9 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 Release with new features and bugfixes:
 
 * https://github.com/devonfw/IDEasy/issues/708[#708]: Open vscode in workspace path
-* https://github.com/devonfw/IDEasy/issues/608[#608]: Enhanced error messages. Now logs missing command output and error messages
+* https://github.com/devonfw/IDEasy/issues/608[#608]: Enhanced error messages.
+Now logs missing command output and error messages
```
jan-vcapgemini, line 11:
```suggestion
* https://github.com/devonfw/IDEasy/issues/608[#608]: Enhanced error messages. Now logs missing command output and error messages
```

jan-vcapgemini (approved):
LGTM. Please check CRs for cleanups.

jan-vcapgemini (approved):
Thanks, ready for review.

hohwille (approved):
@KianRolf thanks for the PR and updates. 👍 
I left 2 comments for rework.
After addressing them this is ready for merge.

cli/src/main/java/com/devonfw/tools/ide/tool/DelegatingToolCommandlet.java
```diff
@@ -0,0 +1,35 @@
...
+  public DelegatingToolCommandlet(IdeContext context, String tool, Set<Tag> tags, Class<D> delegateClass) {
+
+    super(context, tool, tags);
+    this.delegateClass = delegateClass;
+  }
+
+  @Override
+  public final boolean install(boolean silent, EnvironmentContext environmentContext) {
+    return getCommandlet(this.delegateClass).install(silent, environmentContext);
+  }
```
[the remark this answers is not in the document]
KianRolf: done

Issue #704, opened 2024-10-21
settings-security.xml not found


jan-vcapgemini:
### Expected behavior

Password encryption and decryption is not working properly.
As a < type of user >, I want < some goal > so that < some reason >

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. 

### Affected version:

* OS: Windows/Linux/Mac?
* Browser: Chrome/Firefox/Safari?

PR #705, opened 2024-10-21
#704: fix settings-security.xml not found

KianRolf:
connected to #704: With this the program successfully locates the settings-security file


cli/src/main/java/com/devonfw/tools/ide/tool/mvn/Mvn.java
```diff
@@ -163,6 +163,7 @@ private String getEncryptedPassword(String variable) {
 
     ProcessContext pc = this.context.newProcess().executable("mvn");
     pc.addArgs("--encrypt-password", input);
+    pc.addArg("-Dsettings.security=" + this.context.getMavenRepository().getParent().resolve(SETTINGS_SECURITY_FILE).toString().replace("\\", "\\\\"));
```
hohwille, line 166: @KianRolf this will fix the password encryption but not the decryption that later needs to happen in order to make use of the password.

hohwille (changes requested):
@KianRolf thanks for your PR. You found the `settings.security` that is not even documented in the [maven-encrpytion-doc](https://maven.apache.org/guides/mini/guide-encryption.html). Very nice job for finding this hidden feature what is the major job to solve this story. You also added the config to fix the password encryption. 👍 

As I commented, we also need to ensure that we also can decrypt the password when it is to be used. Therefore you need to add this magic system property here:
https://github.com/devonfw/IDEasy/blob/2024c35457ed14f0f57cbe2b5bb3848fbf47f26b/cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java#L441

For a better understanding see also here:
https://github.com/devonfw/IDEasy/blob/2024c35457ed14f0f57cbe2b5bb3848fbf47f26b/cli/src/main/java/com/devonfw/tools/ide/variable/IdeVariables.java#L45

So this way it will be exported as property and work also work if `mvn` is called directly by the end-user from his shell after initializing with IDEasy so in scenarios such as this:
```bash
ide
mvn clean deploy
```
compared to 
```bash
ide mvn clean deploy
```

Since this will only apply once the `settings.xml` is created, your existing change in `Mvn` is still needed and therefore fully correct.

To avoid redundancies, you might want to create a method in `Mvn` creating the `-Dsettings.security=...` String so it can be re-used from both places. If you want to make this method `private` what seems to make sense, you could delegate from `IdeContext` to a new method `getMavenArgs()` in `Mvn` that then returns the entire value of `MAVEN_ARGS` variable and `getMavenArgs()` in `IdeContext` is maily just delegating to that method in `Mvn`. That would be more clean since maven specific logic should be in `Mvn` rather than `IdeContext` (see "Separation of Concerns" pattern).

hohwille (approved):
@KianRolf thanks for your rework. Now looking perfect to me. 👍 
Ready for merge.

hohwille:
issue was not added to CHANGELOG but I will fix that after merge so we can progress with the release.
Please double check [DoD](https://github.com/devonfw/IDEasy/edit/main/documentation/DoD.adoc) on next PR.

PR #707, opened 2024-10-22
#533: Add autocompletion of exit in ide shell

### LINKED ISSUE #533: Add autocompletion of 'exit' to shell
As an IDEasy `shell` user I want to get the offer to exit the autocompletion mode when I type 'e'.

ATTENTION: For bash completion it does not make sense to suggest `exit` for completion of `e` since `exit` is not a commandlet and `ide exit` does not make any sense. Hence for the shell specific suggestion this needs to be implemented in a special way.

## Hints:
A good start would be to use the `AggregateCompleter` like this in the `ShellCommandlet.java`:
```
Completer completer = new AggregateCompleter(
            new StringsCompleter("exit"), new IdeCompleter((AbstractIdeContext) this.context));
```
We could also try to use the `SystemRegistry` as this already has an `exit` command.
https://github.com/jline/jline3/wiki/Command-Registries#systemregistry

WorkingAmeise:
Closes #533 

Implements: 

- When in ide shell 'exit' is now suggested when e is typed and tab is pressed
- Pressing enter successfully aborts the shell


cli/src/main/java/com/devonfw/tools/ide/commandlet/ShellCommandlet.java
```diff
@@ -86,6 +90,10 @@ public void run() {
         while (true) {
           try {
             line = reader.readLine(prompt, rightPrompt, (MaskingCallback) null, null);
+            line = line.trim();
```
jan-vcapgemini, line 93: Is this required for this feature?

cli/src/main/java/com/devonfw/tools/ide/commandlet/ShellCommandlet.java
```diff
@@ -66,7 +69,8 @@ public void run() {
       try (Terminal terminal = TerminalBuilder.builder().build()) {
 
         // initialize our own completer here
-        IdeCompleter completer = new IdeCompleter((AbstractIdeContext) this.context);
+        Completer completer = new AggregateCompleter(
```
jan-vcapgemini, line 72: Maybe add a comment, that `exit` was added here.

jan-vcapgemini (approved):
Looks good to me. I've added some questions, but overall this can get into review now.

jan-vcapgemini (approved):

cli/src/main/java/com/devonfw/tools/ide/commandlet/ShellCommandlet.java
WorkingAmeise: The tab of the auto completion is registered as a character in the line-string so if we want to match the string with "exit" we need to either trim it down or use another method like `line.contains("exit")` however these would probably also match some nonsense like : "fireexit" . If there is another method we could easily use here i am happy to now about.

hohwille: You could do `if (line.trim().equals("exit"))` if the `trim` is undesired for the rest of the code.
However, I do not see this `trim()` as bad but rather as a general improvement since also `runCommand` should not cause problems in case of a training tab or space.

hohwille (approved):
@WorkingAmeise thanks for your PR. Looks like you found the correct solution. All looking fine and nothing to rework. Also you added the issue to the changelog so you really nailed it - great job👍

Issue #708, opened 2024-10-23
Open vscode in workspace path


leonrohne27:
As an IDEasy user, I want my project to be automatically set up in vscode, so that I don't have to manually pass the directory of my workspace.

By calling the "ide vscode" - command, vscode opens without a project set up. In order to do that, the user has to manually pass the directory of the workspace ("ide vscode workspaces/main").  We want that this happens automatically, just like in Intellij.

To do that, we need to add a method that reads the workspace directory and passes it.

Issue #654, opened 2024-09-26
Improve plugin support


hohwille:
As a IDEasy user, I want to be able to push out a new required plugin to my teams with a simple push to my settings git repo so that everybody in my team can automatically get this new plugin installed.

Currently we already have support for automated plugin installation in all our major IDEs.
However there are some problems:
- When a new plugin has later been added to my settings as active neither `ide update` nor pulling the settings and restarting the IDE will install the plugin. As a workaround, I have to uninstall the IDE and install it again (maybe after manually deleting the according plugins folder). This should happen automatically.
- When an automated plugin installation failed, it will not be retried again. E.g. if there is a temporary network issue or an unavailability of the plugin provider download site, it will not be repeated. We should either implement a logic that can detect if a specific plugin is installed per IDE individually or we could write a marker file after successful installation of the plugin based on the plugin name (not ID since for eclipse that is a comma separate list that can get very long and maybe cause errors in Windows filesystem) and then implement a generic mechanism in PluginBasedCommandlet. Then when we launch the IDE, we can (re)install missing plugins automatically. Ideally we also provide an option to disable this feature.
- We should improve the error handling for plugin installations. At least for Eclipse, I gets pages of crazy HEX output logged to the console. If that is retried for every launch this can be considered as annoying spam. E.g. we could just log the last max N lines of std-error and then refer to a logfile for further details. I already changed the loglevel of std-out to debug in PR #652
- When the user manually removes or uninstalls a plugin, we need to still define what should happen. At least a warning should be logged. We should clarify if we instead want to reinstall it anyway - the problem is that then the user cannot easily get rid of a plugin and may be annoyed if he uninstalled a plugin that he does not like if it comes back on IDE restart. A possibility for the user is to override the plugin by copying the plugin properties file into `~/.ide/plugins/«ide»/«plugin-name».properties` and set `plugin_active=false`.

Solution design and acceptance criteria:
- [ ] for every plugin we create a marker file after successful installation in `$IDE_HOME/.ide/plugin.«ide».«plugin-name»`. Here `«ide»` is the name of the IDE like `intellij` or `eclipse` and `«plugin-name»` is the name of the properties file for the plugin excluding the extension.
- [ ] when we setup an IDE (e.g. `ide install intellij` or `ide update`) even if it is already installed in the correct version end edition, then we still check all active plugins. If the above marker file is already present, nothing has to be done for the plugin, but otherwise if will be installed.
[...]

PR #709, opened 2024-10-23
#708: Added method for vscode

leonrohne27:
Fixes:#708

### implements:

- Added method that reads and passes the workspace path when calling "ide vscode", so that the user doesn't need to hand over the directory manually.


cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -131,6 +131,7 @@ protected void configureToolArgs(ProcessContext pc, ProcessMode processMode, Pro
     pc.addArg("--user-data-dir=" + vsCodeConf);
     Path vsCodeExtensionFolder = this.context.getIdeHome().resolve("plugins/vscode");
     pc.addArg("--extensions-dir=" + vsCodeExtensionFolder);
+    pc.addArg(this.context.getWorkspacePath());
```
hohwille, line 134: Did you test this in combination with plugin installations?
This argument will be added to any invocation of `runTool` and that is not only used to start the IDE but also to install plugins. In the latter case this argument is IMHO undesired and might even break VScode.
In case I am right, you could make this conditional on `processMode` or you could move this extra arg to `run()` method (overriding it).

hohwille (changes requested):
@leonrohne27 thanks for your PR. You found the right way to solve this problem 👍 
https://code.visualstudio.com/docs/editor/command-line#_opening-files-and-folders

However, the place in the code where you integrated the change may be incorrect.
Could you have a look and double-check it? Thanks.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
leonrohne27: Yes I just tested it and it worked out regardless. Anyways, I've found a bug not related to my change:
When vscode is already installed, plugins won't be reinstalled when a user manually deleted the plugins from the plugins/vscode- folder. Instead it just create an empty extensions.json- file.

hohwille: That is covered by #654.
Thanks for testing.

hohwille (approved):
@leonrohne27 this is fine now after your comment that resolved my doubt.
Could you please follow DoD and add this to our [changelog](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc#202411001).
The first thing then would be to pull your feature branch so you have the latest CHANGELOG (I already updated your branch to main).

leonrohne27:
> @leonrohne27 this is fine now after your comment that resolved my doubt. Could you please follow DoD and add this to our [changelog](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc#202411001). The first thing then would be to pull your feature branch so you have the latest CHANGELOG (I already updated your branch to main).

thanks, I added the entry.

Issue #302, opened 2024-04-19
Add plugin support for GraalVM


hohwille:
As a IDEasy and GraalVM user, I want to have plugins for GraalVM preinstalled so that I can directly build native-images without manually calling `gu install native-image`.

The idea is to do the following:
* Make `GraalVm` commandlet extend from `PluginBasedCommandlet`
* Make plugin installation and removal to delegate to `gu install` / `gu uninstall`.
* Also create a separate PR for ide-settings adding `graalvm/plugins/native-image.properties` to pre-configure that `native-image` gets installed by default.
Please note that for GraalVM we do not need an extra plugin-id as already the plugin-name from `«plugin-name».properties` is sufficient, so all we need is active flag set to true.
Further, plugin installation is an additive operation and never destructive.
Therefore, I suggest we can simplify if we do not keep the GraalVM installation in the software repository pristine by extracting plugins to a separate plugin folder (`$IDE_HOME/plugins/graalvm/`).
To document this you should extend the help page of GraalVM to state that adding a plugin will change the physical installation and therefore also effect other IDEasy projects using the same GraalVM version.

PR #711, opened 2024-10-25
#302: Added plugin support for GraalVM

KianRolf:
Implements: #302

Enables the install-plugin and uninstall-plugin command for graalvm.


cli/src/main/java/com/devonfw/tools/ide/tool/graalvm/GraalVm.java
```diff
@@ -51,4 +58,24 @@ public void postInstall() {
     envVars.save();
     super.postInstall();
   }
+
+  @Override
+  public void installPlugin(ToolPluginDescriptor plugin, Step step) {
+    String[] args = { "install", plugin.name() };
+    ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.THROW_CLI);
+    pc.executable(getToolPath() + "\\bin\\gu");
+    configureToolArgs(pc, ProcessMode.DEFAULT, ProcessErrorHandling.THROW_CLI, args);
```
hohwille, line 67: The method `configureToolArgs` exists if generic executions of the tool itself (here that would actually be `graalvm` that even does not exist as binary) needs some extra options so tools like `Eclipse` can override it to easily add some internally needed arguments.

Here, this will do nothing else but `pc.addArgs(args)`.
Therefore, you can simplify your code and kick out the `args` array variable.
Instead simply do this:
```
pc.addArgs("install", plugin.name());
```

BTW: ProcessContext has a fluent API. You could even write all in single line without the `pc` variable at all.

cli/src/main/java/com/devonfw/tools/ide/tool/graalvm/GraalVm.java
```diff
@@ -51,4 +58,24 @@ public void postInstall() {
...
+    pc.run(ProcessMode.DEFAULT);
+  }
+
+  @Override
+  public void uninstallPlugin(ToolPluginDescriptor plugin) {
+    String[] args = { "remove", plugin.name() };
+    ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.THROW_CLI);
+    pc.executable(getToolPath() + "\\bin\\gu");
+    configureToolArgs(pc, ProcessMode.DEFAULT, ProcessErrorHandling.THROW_CLI, args);
```
hohwille, line 76: same here.

hohwille (approved):
@KianRolf thanks for your PR. Nice job and looks already fully functional 👍 

I double-checked and meanwhile figured out the following:
* `plugin_id` is not needed: https://github.com/KianRolf/IDEasy/blob/2024c35457ed14f0f57cbe2b5bb3848fbf47f26b/cli/src/main/java/com/devonfw/tools/ide/tool/plugin/ToolPlugins.java#L66-L69
* `plugin_url` is also not needed by default: https://github.com/KianRolf/IDEasy/blob/2024c35457ed14f0f57cbe2b5bb3848fbf47f26b/cli/src/main/java/com/devonfw/tools/ide/tool/plugin/PluginBasedCommandlet.java#L69-L72

The latter fact actually confused me since I expected it the other way around.
However, then we have a bug at least in `Eclipse` that is not overriding this method setting this to `true` since for Eclipse the `plugin_url` is needed. However, not your bug and not part of this story.

I left some very small comments for rework. It would be nice, if you can improve and then we merge this PR.

hohwille:
@KianRolf 
I should not do coding in the browser during review but I wanted to do a constructive review to get this resolved and merged asap.

I found this in the final review:

> pc.executable(getToolPath() + "\\bin\\gu");

Using backslashes in path only works on windows so that would fail on Linux or Mac.
Using slashes works on any OS so if you really want to code slashes always prefer slashes over backslash.
This only got a green build on github action CI build since this feature is not covered by a test.
I extracted a private method since both install and remove methods more or less shared the exact same code except for the String "install" or "remove" passed to `gu`.
There I also avoiding coding a slash or backslash at all using the `resolve` method what seems even more elegant.

PR #719, opened 2024-10-28
#710: make workspace configuration robust

### LINKED ISSUE #710: Make IDE workspace configuration robust
Issue #640 revealed that an error in workspace configuration can entirely prevent the IDE (IntelliJ, Eclipse, VSCode, etc.) from being started.
While there will still be workarounds to continue working but average users of IDEasy will then be lost and blocked and do not know how to start their IDE.
Hence, we should invest into robustness and improve UX:
- [x] when a single workspace template file could not be read/parsed (e.g. invalid JSON, XML, ... or invalid option for `merge:strategy`, etc.), we should log an error with details about the corrupted file but continue with the workspace merging. However, the workspace merging procedure should then still fail.
- [x] if the workspace merging procedure fails, IDEasy should log an error that the workspace could not be properly merged and either settings repo is wrong or a bug in IDEasy was hit and ask the user if he wants to continue. This should be done via Step action so that the overall status gets transparent. The user may still run in `--batch` mode in case such error cannot be fixed for some days to avoid being annoyed by the question. However, the question should raise attention that something is wrong.

See also #644

hohwille:
fixes #710

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/merge/xmlmerger/XmlMerger.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/merge/xmlmerger/XmlMerger.java
@@ -54,7 +54,7 @@ public class XmlMerger extends FileMerger {
   }
 
   @Override
-  public void merge(Path setup, Path update, EnvironmentVariables resolver, Path workspace) {
+  protected void doMerge(Path setup, Path update, EnvironmentVariables resolver, Path workspace) {
 
     Document document = null;
     Path template = setup;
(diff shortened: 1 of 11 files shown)


cli/src/main/java/com/devonfw/tools/ide/merge/xmlmerger/XmlMerger.java
```diff
@@ -82,6 +82,7 @@ public void merge(Path setup, Path update, EnvironmentVariables resolver, Path w
       resolve(document, resolver, false, template);
       save(document, workspace);
     }
+    return;
```
jan-vcapgemini, line 85: void return?

jan-vcapgemini (commented):
LGTM. I've added a small question.

PR #724, opened 2024-10-29
#312: Added ability to prefer git (ssh) protocol instead of https for cloning repos

alfeilex:
This PR finishes PR #625

Fixes https://github.com/devonfw/IDEasy/issues/312

- Introduced `PREFERRED_GIT_PROTOCOL` property to allow configuration of preferred Git protocol (SSH or HTTPS).
- Default behavior remains unchanged if the property is not set, using the provided Git URL as is.
- Implemented GitUrlSyntax enum to manage and convert between SSH and HTTPS URL formats.
- Updated GitUrl class for URL parsing and protocol conversion based on the property.
- Added tests to validate URL conversion and ensure no changes for specific domains like GitHub.
- When `PREFERRED_GIT_PROTOCOL` is not defined then URL doesn't change


cli/src/main/java/com/devonfw/tools/ide/variable/IdeVariables.java
```diff
@@ -73,6 +75,9 @@ public interface IdeVariables {
   /** {@link VariableDefinition} for {@link com.devonfw.tools.ide.context.IdeContext#getProjectName() PROJECT_NAME}. */
   VariableDefinitionString PROJECT_NAME = new VariableDefinitionString("PROJECT_NAME", null, c -> c.getProjectName());
 
+  /** Preferred Git protocol (HTTPS/SSH) as defined by {@link GitUrlSyntax}. */
+  VariableDefinitionEnum<GitUrlSyntax> PREFERRED_GIT_PROTOCOL = new VariableDefinitionEnum<>("PREFERRED_GIT_PROTOCOL", null, GitUrlSyntax.class);
```
alfeilex, line 79: I removed the default value for `PREFERRED_GIT_PROTOCOL` 

Related to: https://github.com/devonfw/IDEasy/pull/625#discussion_r1817459404

cli/src/main/java/com/devonfw/tools/ide/context/GitContextImpl.java
```diff
@@ -138,7 +139,8 @@ private void handleErrors(Path targetRepository, ProcessResult result) {
   @Override
   public void clone(GitUrl gitRepoUrl, Path targetRepository) {
 
-    URL parsedUrl = gitRepoUrl.parseUrl();
+    GitUrlSyntax gitUrlSyntax = IdeVariables.PREFERRED_GIT_PROTOCOL.get(getContext());
+    URL parsedUrl = gitRepoUrl.convert(gitUrlSyntax).parseUrl();
```
alfeilex, line 143: Added syntax convert in the clone function

cli/src/main/java/com/devonfw/tools/ide/variable/IdeVariables.java
```diff
@@ -73,6 +75,10 @@ public interface IdeVariables {
   /** {@link VariableDefinition} for {@link com.devonfw.tools.ide.context.IdeContext#getProjectName() PROJECT_NAME}. */
   VariableDefinitionString PROJECT_NAME = new VariableDefinitionString("PROJECT_NAME", null, c -> c.getProjectName());
 
+  /** Preferred Git protocol (HTTPS/SSH) as defined by {@link GitUrlSyntax}. */
+  VariableDefinitionEnum<GitUrlSyntax> PREFERRED_GIT_PROTOCOL = new VariableDefinitionEnum<>("PREFERRED_GIT_PROTOCOL", null, GitUrlSyntax.class,
+      c -> GitUrlSyntax.DEFAULT);
+
```
alfeilex, line 81: Changed default function

cli/src/main/java/com/devonfw/tools/ide/context/GitUrlSyntax.java
```diff
@@ -0,0 +1,95 @@
...
+    String url = gitUrl.url();
+
+    // Prevent conversion for domains in the no-conversion list
+    if (isDomainWithNoConversion(url.toLowerCase())) {
+      return gitUrl;
+    }
+
+    switch (this) {
+      case DEFAULT:
+        break;
```
alfeilex, line 53: Added default case

hohwille: FYI: Java meanwhile supports a nice case syntax so we should stop using this ugly C-legacy-style `switch` statements with `break`.

Example:
https://github.com/devonfw/IDEasy/blob/d146098fa03789dcdf6e3b0f6a78b8f7cfd3320d/cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java#L38-L42

cli/src/main/java/com/devonfw/tools/ide/context/GitUrlSyntax.java
```diff
@@ -0,0 +1,95 @@
...
+  public GitUrl format(GitUrl gitUrl) {
+    String url = gitUrl.url();
+
+    // Prevent conversion for domains in the no-conversion list
+    if (isDomainWithNoConversion(url.toLowerCase())) {
+      return gitUrl;
+    }
+
+    switch (this) {
+      case DEFAULT:
```
hohwille, line 52: Why creating a new `GitUrl` instance if nothing changes?
For `DEFAULT` I would recommend to override this method this way:
```java
DEFAULT("") {
  @Override
  public GitUrl format(GitUrl gitUrl) {
    return gitUrl;
  }
}
```

You can also handle this here at the top:
```
public GitUrl format(GitUrl gitUrl) {
    if (this == DEFAULT) {
      return gitUrl;
    }
    String url = gitUrl.url();
    ...
}
```

cli/src/main/java/com/devonfw/tools/ide/context/GitContextImpl.java
hohwille, line 143: No bug but if we convert the `GitUrl` we should replace the old one with the new one - otherwise subsequent code may still use attributes from the old instance and lead to bugs:
```suggestion
    GitUrlSyntax gitUrlSyntax = IdeVariables.PREFERRED_GIT_PROTOCOL.get(getContext());
    gitRepoUrl = gitUrlSyntax.format(gitUrlSyntax);
    URL parsedUrl = gitRepoUrl.parseUrl();
```
Assuming the conversion would change the branch attribute, the current code would already be buggy otherwise.

Further, I would kick out this `parsedUrl` here. We should directly use `gitRepoUrl.url()` instead. The only place we use this is for logging. So we convert a `String` to `URL` and then the logger converts it back to `String` what seems pointless waste and I am always confused if I read such code because I am searching for the reason why this has been done until I figure out there is no such reason.

hohwille (approved):
@alfeilex thanks for taking over this PR and completing it. All seems functionally correct and complete now. 👍 
I left some comments for final improvement. Please have a look. We can merge if resolved.

cli/src/main/java/com/devonfw/tools/ide/context/GitContextImpl.java
alfeilex: I also could not find a reason to parse the url into `URL`, the only requirement for `URL` is in `CliOfflineException`. 

Finally, I removed the `parsedUrl` and replaced it with `gitRepoUrl.url().` In the `CliOfflineException` I use `gitRepoUrl().parseUrl()`.

Solved with [3db08c4](https://github.com/devonfw/IDEasy/pull/724/commits/3db08c449a458b8fc06fece127f892dc78378756)

cli/src/main/java/com/devonfw/tools/ide/context/GitUrlSyntax.java
alfeilex: Solved with https://github.com/devonfw/IDEasy/pull/724/commits/eabd33873b5f0dac2facd826ceeabf356567bf35

cli/src/main/java/com/devonfw/tools/ide/context/GitUrlSyntax.java
```diff
@@ -0,0 +1,95 @@
...
+      }
+      case HTTPS -> {
+        if (url.startsWith(SSH.prefix)) {
+          int index = url.indexOf(":");
+          if (index > 0) {
+            url = HTTPS.prefix + url.substring(SSH.prefix.length(), index) + "/" + url.substring(index + 1);
+          }
+        }
+      }
+      default -> throw new IllegalArgumentException("Unsupported protocol: " + this);
```
alfeilex, line 71: I am not sure if the default case can actually occur, because if the enum is set incorrectly, then a `IllegalArgumentException` is thrown by the `Enum` class before.

hohwille: This is a common problem pattern. The best way to solve this in an enum is to avoid the `switch` at all:
simply make the method `abstract` and create a method implementation for each enum constant.
This way, if new enum constants are added, the compiler will force the developer to implement the `abstract` method and he cannot forget it due to the compiler error.
I will create a small cleanup story for this improvement but we can already merge this PR as planned.

hohwille (approved):
@alfeilex thanks for the rework. All perfect now 👍 
Ready for merge.

cli/src/main/java/com/devonfw/tools/ide/context/GitUrlSyntax.java
hohwille: Created #750

PR #725, opened 2024-10-29
#673: fixed tomcat and improved logging

### LINKED ISSUE #673: Tomcat still not working (JAVA_HOME not set)
### Expected behavior

As a IDEasy user, I want to be able to install and run tomcat so that I can use it for my legacy WAR app.

### Actual behavior

```
$ ide tomcat run
Tomcat is running at localhost on HTTP port 8080:
http://localhost:8080
Successfully completed ide (tomcat,run)
Neither the JAVA_HOME nor the JRE_HOME environment variable is defined
At least one of these environment variable is needed to run this program
```

Even though IDEasy said the launch was successful, it actually failed (as it was started in the background).
The actual bug is that `JAVA_HOME` does not seem to be set for the process context.

1. `ide tomcat run` 

### Related/Dependent Issues

#36
#659 

### Comments/Hints:

Do we need to export it in the `tomcat` wrapper script?
Or is the problem that I introduced this `tomcat` wrapper script and we need to call `catalina.bat` or `catalina.sh` directly?

### Affected version:

* OS: Windows
* IDEasy: 2024.09.002-beta

hohwille:
fixes #673 
* One fix was also in ide-urls [commit](https://github.com/devonfw/ide-urls/commit/a33c2eee78a489df424ff94be071ec6fe25ac09c)
* The main fix here was that `environmentContext` was not passed to recursive `install` method (in such situations a NPE would have been preferred that is prevented by the EMPTY instance of EnvironmentContext)
* And for the record: IntelliJ has a severe flaw. When you run this use-case from IntelliJ, there is no way to stop the process in Intellij. The only way to end it and being able to start another run is to find and kill the process manually in task manager (what requires ProcessExplorer). This works flawless in Eclipse.

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
@@ -251,7 +251,11 @@ public abstract class LocalToolCommandlet extends ToolCommandlet {
 
   private void installToolDependencies(VersionIdentifier version, String edition, EnvironmentContext environmentContext, ToolRepository toolRepository) {
     Collection<ToolDependency> dependencies = toolRepository.findDependencies(this.tool, edition, version);
+    String toolWithEdition = getToolWithEdition(this.tool, edition);
+    int size = dependencies.size();
+    this.context.debug("Tool {} has {} other tool(s) as dependency", toolWithEdition, size);
     for (ToolDependency dependency : dependencies) {
+      this.context.trace("Ensuring dependency {} for tool {}", dependency.tool(), toolWithEdition);
       LocalToolCommandlet dependencyTool = this.context.getCommandletManager().getRequiredLocalToolCommandlet(dependency.tool());
       dependencyTool.installAsDependency(dependency.versionRange(), environmentContext);
     }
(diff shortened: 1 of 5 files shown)


cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/ToolDependencies.java
```diff
@@ -55,10 +63,10 @@ public static ToolDependencies of(Path file) {
 
     if (Files.exists(file)) {
       try (BufferedReader reader = Files.newBufferedReader(file)) {
-        TypeReference<HashMap<VersionRange, List<ToolDependency>>> typeRef = new TypeReference<>() {
+        TypeReference<TreeMap<VersionRange, List<ToolDependency>>> typeRef = new TypeReference<>() {
```
jan-vcapgemini, line 66: Why was this converted to a TreeMap?

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -251,7 +251,11 @@ public boolean installAsDependency(VersionRange version, EnvironmentContext envi
 
   private void installToolDependencies(VersionIdentifier version, String edition, EnvironmentContext environmentContext, ToolRepository toolRepository) {
     Collection<ToolDependency> dependencies = toolRepository.findDependencies(this.tool, edition, version);
+    String toolWithEdition = getToolWithEdition(this.tool, edition);
+    int size = dependencies.size();
```
jan-vcapgemini, line 255: Does not look like we need this variable. Only used once.

jan-vcapgemini (commented):
I've added some questions, please check.

PR #729, opened 2024-10-30
#646: Add variable for intellij to disable console log output

### LINKED ISSUE #646: Add variable for intellij to disable/modify log output
As an intellij user I want to be able to disable/modify the log output of intellij.

When running intellij using `ide intellij`, intellij logs errors directly into the shell and clutters the console with it.

We need to add a new exported environment variable with which we can set intellij to log into a file or even completely disable its log output.

Might be required for #669 if we remove the background process.

KianRolf:
fixes #646

- gives the user the possibility to enable or disable console log output via IDE_ENABLE_LOG
- is set to false by default
- so far is only affecting intellij but intended as a switch for all ide

tied to: https://github.com/devonfw/ide-settings/pull/66


CHANGELOG.adoc
```diff
@@ -6,6 +6,8 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/646[#646]: Add variable for intellij to disable log output
+
```
jan-vcapgemini, line 10: Please remove.
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/variable/IdeVariables.java
```diff
@@ -31,6 +31,9 @@ public interface IdeVariables {
   /** {@link VariableDefinition} for list of tools to install by default. */
   VariableDefinitionStringList IDE_TOOLS = new VariableDefinitionStringList("IDE_TOOLS", "DEVON_IDE_TOOLS", c -> List.of("mvn", "npm"));
 
+  /** {@link VariableDefinition} for enabling logs of ides. */
+  VariableDefinitionBoolean IDE_ENABLE_LOG = new VariableDefinitionBoolean("IDE_DISABLE_LOG", null, c -> false);
```
jan-vcapgemini, line 35: Be aware, the name has a discrepancy.
```suggestion
  VariableDefinitionBoolean IDE_ENABLE_LOG = new VariableDefinitionBoolean("IDE_ENABLE_LOG", null, c -> true);
```

jan-vcapgemini (changes requested):
@KianRolf  thanks for your addition of the IDE_ENABLE_LOG variable. I've added some CRs, please resolve.

jan-vcapgemini (changes requested):
Thanks for addressing the change requests. I think we should also add an entry to our documentation, so the user becomes aware of this feature.

jan-vcapgemini:
I'm not sure about the naming yet. IDE_ENABLE_LOG could be misinterpreted that the logging is now being done into a file. Maybe we need 2 variables instead? IDE_ENABLE_FILE_LOG (logs intellij errors into log file) and IDE_ENABLE_LOG (disables complete logging)

jan-vcapgemini (approved):
Nice solution for the naming. Ready for review.

jan-vcapgemini (approved):

hohwille (changes requested):
@KianRolf this PR should fix #646 and you nicely introduced and documented a new variable for this feature.
However, that variable is not used anywhere in the code so it has no effect yet.
Before merging this PR, we also need to add this effect to make the story complete.
Otherwise after merge, we have a documentation promising a feature that is actually not there and we had stuff like that recently what confused our pilot users.
Please add a change that will redirect the output to a log-file according to the value of this new property.
IMHO this should not only happen for IntelliJ but for any tool we run as `BACKGROUND` as you already implied by the documentation.
Maybe we might even think of a more precise name for the variable since `IDE_ENABLE_CONSOLE_LOG` may imply that all console output is written to a logfile what is actually not the case.

KianRolf:
PR will be closed for now, since it is still unclear how we will handle console logging from other IDEs, and intellij console log will be disabled by default.

Issue #415, opened 2024-06-25
Log command before asking user for sudo password


hohwille:
From Review of PR #409:
What I think we need to consider for runWithPackageManager is that we need to add logging of the executed command in ProcessContext before actually executing the process.

# Rationale

A linux user cannot be blindly prompted for its root password from sudo without knowing what is actually happening and what command is to be executed.

# Requirement

Then runWithPackageManager should log an according message on log-level interaction. So that you get something like this:
```
We need to run the following privileged command(s):
sudo «somecommand»
...maybe more commands...
This will require root permissions!
```

Please note that we also need this for other things than just global tool installations with `runWithPackageManager` but also e.g. for #453 
Therefore the logging itself should be implemented in a method that can be reused for other use-cases than `executePackageManagerCommand`.

# Hint

Needs to be added here:
https://github.com/devonfw/IDEasy/blob/f1b1801c2bfcc795307e78abce9dd61ca9d7b528/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java#L82

PR #730, opened 2024-10-30
Fix/415 log command before sudo pw

leonrohne27:
**Fixes:** #415

**implements:**

- Added a method that will inform the user for what process he will need to enter his sudo- password

note: [cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java] hasn't been changed, idk why it shows here, just ignore it :)


leonrohne27:
Needs to be closed because it was not branched from main

PR #731, opened 2024-10-30
#632: added .editorconfig  to settings-workspaces

### LINKED ISSUE #632: Add .editorconfig to settings workspaces
As an IDEasy user I expect `intellij` to automatically install an `.editorconfig` (for our code reformattings) into my workspace.

At the moment we've already added the `.editorconfig` to our project root in git.
For projects that do not follow such best-practice, we still need the settings to automatically install it to the current workspace.

See #735 for a clean solution of the approach.

See: #465 for reference

WorkingAmeise:
closes #632 

- When the workspace is configured now the editor config located in the settings-workspaces folder is copied to the current workspace


cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
```diff
@@ -77,6 +85,14 @@ protected void configureWorkspace() {
     }
     try (Step step = this.context.newStep("Configuring workspace " + ideWorkspacePath.getFileName() + " for IDE " + this.tool)) {
       int errors = this.context.getWorkspaceMerger().merge(setupFolder, updateFolder, this.context.getVariables(), ideWorkspacePath);
+      try {
+        Files.copy(editorConfigPath, ideWorkspacePath.resolve(".editorconfig"),
+            StandardCopyOption.REPLACE_EXISTING);
+      } catch (IOException e) {
+        step.error("Failed to copy the .editorconfig file into the current workspace. \n"
```
jan-vcapgemini, line 92: Maybe add the workspace path to the message here too.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
```diff
@@ -77,6 +85,14 @@ protected void configureWorkspace() {
     }
     try (Step step = this.context.newStep("Configuring workspace " + ideWorkspacePath.getFileName() + " for IDE " + this.tool)) {
       int errors = this.context.getWorkspaceMerger().merge(setupFolder, updateFolder, this.context.getVariables(), ideWorkspacePath);
+      try {
+        Files.copy(editorConfigPath, ideWorkspacePath.resolve(".editorconfig"),
+            StandardCopyOption.REPLACE_EXISTING);
```
jan-vcapgemini, line 90: Should the editorconfig only be overwritten or should it be merged instead?
If it has to be merged, then we need to use our merger for it.

jan-vcapgemini (changes requested):
@WorkingAmeise thanks for your integration of the automatic editorconfig setup. It would be nice if you could add a simple test for this feature too.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
hohwille: Please always use `FileAccess` for file operations instead of low-level Java stuff with ugly `IOException`.
That is exactly why we created `FileAccess`. This will already cover error handling, logging, etc.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
hohwille: See #465 and #735 - so my expectation is also different on the long run. But to get this feature earlier, we start with this hardcoded quickfix.

hohwille (changes requested):
@WorkingAmeise thanks for your PR. Already looking nice. 👍 
However, why did you delete our IDEasy `.editorconfig` file? This file should stay.
I guess this comes from a missunderstanding what what @jan-vcapgemini wrote in the story:
> At the moment we've added the .editorconfig to our project root as a workaround but instead we need the settings to automatically install it to the currently loaded workspace.

This is not entirely correct. Having `.editorconfig` in the git code repo is the recommended way and gives direct control of the code-style in the easiest manner - you do not even need IDEasy at all and it also works with any vanilla IntelliJ (or whatever IDE properly supporting this file).
But if you clone repositories into your workspace that do not have any `.editorconfig` the one you now copy to the workspace will control the code-style.
All these combinations can be flexibly combined:
* you can only maintain `.editorconfig` in settings git and have many different code repos in your workspace all managed by that same config
* some or all code repos can have their own `.editorconfig` that can override specific features or entirely take over control.
So please revert this deletion in your feature branch.
Also make use of `FileAccess` and avoid fiddling with `IOException` and low-level file operations. You can do me a big favour and continuously spread this information to the team in internal KT sessions since I often have to do this remark in PRs. Thanks in advance.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
WorkingAmeise: Used the merger for this instead

cli/src/main/java/com/devonfw/tools/ide/merge/DirectoryMerger.java
```diff
@@ -50,7 +50,9 @@ public DirectoryMerger(IdeContext context) {
     this.extension2mergerMap.put("json", jsonMerger);
     TextMerger textMerger = new TextMerger(context);
     this.extension2mergerMap.put("name", textMerger); // intellij specific
+    this.extension2mergerMap.put("editorconfig", textMerger);
     this.fallbackMerger = new FallbackMerger(context);
+
```
jan-vcapgemini, line 55:
```suggestion
```

WorkingAmeise: removed the new line

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
```diff
@@ -62,6 +62,8 @@ protected void configureWorkspace() {
 
     Path settingsWorkspaceFolder = this.context.getSettingsPath().resolve(this.tool)
         .resolve(IdeContext.FOLDER_WORKSPACE);
+    Path workspaceUpdateFolder = this.context.getSettingsPath().resolve(IdeContext.FOLDER_WORKSPACE).resolve(IdeContext.FOLDER_UPDATE);
+    Path workspaceSetupFolder = this.context.getSettingsPath().resolve(IdeContext.FOLDER_WORKSPACE).resolve(IdeContext.FOLDER_SETUP);
```
hohwille, line 66: https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc#consider-extracting-local-variable-for-multiple-method-calls
```suggestion
    Path genericWorkspaceFolder = this.context.getSettingsPath().resolve(IdeContext.FOLDER_WORKSPACE); 
    Path workspaceUpdateFolder = genericWorkspaceFolder.resolve(IdeContext.FOLDER_UPDATE);
    Path workspaceSetupFolder = genericWorkspaceFolder.resolve(IdeContext.FOLDER_SETUP);
```
BTW: It would be even better to add a method to `DirectoryMerger` that resolves the two sub-folders. I will add this as leftover to #735

hohwille (approved):
@WorkingAmeise thanks for your PR. You found a very elegant way to also implement #735 with a KISS approach that I did not think of. All you need to do is reverse the order of the `merge` statements and we are done with both stories. Awesome solution and great work 👍 
Ready for merge when suggestion is applied.

jan-vcapgemini (approved):

PR #732, opened 2024-10-30
#415: log command before sudo pw

leonrohne27:
**Fixes**: #415 

**implements**:

- Added a method that will inform the user for what process he will need to enter his sudo- password

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
@@ -55,6 +55,8 @@ public abstract class GlobalToolCommandlet extends ToolCommandlet {
    */
   protected boolean runWithPackageManager(boolean silent, List<PackageManagerCommand> pmCommands) {
 
+    logPackageManagerCommands(pmCommands);
+
     for (PackageManagerCommand pmCommand : pmCommands) {
       PackageManager packageManager = pmCommand.packageManager();
       Path packageManagerPath = this.context.getPath().findBinary(Path.of(packageManager.getBinaryName()));
@@ -68,6 +70,19 @@ public abstract class GlobalToolCommandlet extends ToolCommandlet {
       }
     }
     return false; // None of the package manager commands were successful
+  }
+
+  private void logPackageManagerCommands(List<PackageManagerCommand> pmCommands) {
+    StringBuilder commandLog = new StringBuilder("We need to run the following privileged command(s):\n");
+
+    for (PackageManagerCommand pmCommand : pmCommands) {
+      for (String command : pmCommand.commands()) {
+        commandLog.append(command).append("\n");
+      }
+    }
+    commandLog.append("This will require root permissions!");
+
+    this.context.info(commandLog.toString());
   }
 
   /**
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
@@ -54,9 +54,7 @@ public abstract class GlobalToolCommandlet extends ToolCommandlet {
    * @return {@code true} if installation or uninstallation succeeds with any of the package manager commands, {@code false} otherwise.
    */
   protected boolean runWithPackageManager(boolean silent, List<PackageManagerCommand> pmCommands) {
-
-    logPackageManagerCommands(pmCommands);
-
+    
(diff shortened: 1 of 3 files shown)


cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -55,6 +55,8 @@ protected boolean runWithPackageManager(boolean silent, String... commandStrings
    */
   protected boolean runWithPackageManager(boolean silent, List<PackageManagerCommand> pmCommands) {
 
+    logPackageManagerCommands(pmCommands);
```
hohwille, line 58: You may want to look at the usages of `runWithPackageManager` before to see examples how this is used and how to implement this story:
This `List` (`pmCommands`) contains a list of potential ways to install the tool using a specific package manager for each supported linux distribution.
So debian based linux (Debian, Ubunit, Mint, etc.) will use `apt install ...` while SuSE will use `zypper install ...` and others may use `yam ...` or whatever.
In the `for` loop we iterate of these "options" and check if the package manager (apt/zypper/yum/...) is present on the actual system and then we will execute the command.
Therefore, only this executed option will be relevant for the end-user while all other options are candidate for other Linux distrubtions that he does not care about.

Hence, you IMHO should implement your change only in `executePackageManagerCommand` and only print the output for the single `PackageManagerCommand` that will be executed rather than for all of them.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -70,6 +72,19 @@ protected boolean runWithPackageManager(boolean silent, List<PackageManagerComma
     return false; // None of the package manager commands were successful
   }
 
+  private void logPackageManagerCommands(List<PackageManagerCommand> pmCommands) {
+    StringBuilder commandLog = new StringBuilder("We need to run the following privileged command(s):\n");
```
hohwille, line 76: you do not need a `StringBuilder` here. Simply call `info` method 3 times and remove the `\n`s will do.

hohwille (approved):
@leonrohne27 thanks for your PR. Your solution should already work slightly 👍 
However, I left some comments for improvement since you are logging too much information what will confuse or mislead the user.
Since I am sure, you will rework this easily, I already approve so if the review comments are resolved, we can merge this.

hohwille (approved):
@leonrohne27 thanks for your rework. Now perfect 👍 
Ready for merge.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -54,7 +54,7 @@ protected boolean runWithPackageManager(boolean silent, String... commandStrings
    * @return {@code true} if installation or uninstallation succeeds with any of the package manager commands, {@code false} otherwise.
    */
   protected boolean runWithPackageManager(boolean silent, List<PackageManagerCommand> pmCommands) {
-
+
```
hohwille, line 57: check your IntelliJ settings?
Our formatter should remove such trailing whitespaces.
And the "Actions on Save" feature should be automatically enabled invoking the formatter and organize imports.
If that is not the case, we still struggle with the heritage of #52 and you should consult @jan-vcapgemini

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -70,6 +70,15 @@ protected boolean runWithPackageManager(boolean silent, List<PackageManagerComma
     return false; // None of the package manager commands were successful
   }
 
+  private void logPackageManagerCommands(PackageManagerCommand pmCommand) {
+
+    this.context.info("We need to run the following privileged command(s):");
+    for (String command : pmCommand.commands()) {
+      this.context.info(command);
+    }
+    this.context.info("This will require root permissions!");
```
hohwille, line 79: Sorry for interfering once more but I just read my own story:
> Then runWithPackageManager should log an according message on log-level interaction.

So please change the loglevel from INFO to INTERACTION.

hohwille (commented):
Sorry - I just wanted to merge but checked the story again and found that you missed an explicit requirement given in #415. Please update so we can merge.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
hohwille: Thanks for rework 👍

PR #733, opened 2024-10-31
#715: Cygwin not supported warning.

### LINKED ISSUE #715: Show "cygwin is not supported" message for cygwin users
As a `cygwin` user I expect IDEasy to show me that this program does not work properly on `cygwin` and I have to use `bash` instead.
Adjust documentation accordingly.

nmCapG:
Fixes #715 .

### DIFF
--- a/cli/src/main/package/bin/ide
+++ b/cli/src/main/package/bin/ide
@@ -18,6 +18,15 @@ if [ $? = 0 ]; then
     echo "IDE environment variables have been set for ${IDE_HOME} in workspace ${WORKSPACE}"
   fi
 fi
+
+if [ "${OSTYPE}" = "cygwin" ]; then
+    echo -e "\n--- WARNING: CYGWIN IS NOT SUPPORTED ---"
+    echo "Cygwin console is not supported by IDEasy. Consider using the git-bash terminal"
+    echo "instead. If you want to use Cygwin with IDEasy, you will have to configure it"
+    echo "yourself. A few suggestions and caveats can be found here:"
+    echo -e "https://github.com/devonfw/IDEasy/blob/main/documentation/Not_supported_Cygwin.adoc\n"
+fi
+
 unset _IDEASY
 unset ide_env
 unset return_code
--- a/cli/src/main/package/bin/ide
+++ b/cli/src/main/package/bin/ide
@@ -20,11 +20,7 @@ if [ $? = 0 ]; then
 fi
 
 if [ "${OSTYPE}" = "cygwin" ]; then
-    echo -e "\n--- WARNING: CYGWIN IS NOT SUPPORTED ---"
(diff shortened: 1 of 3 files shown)


cli/src/main/package/bin/ide
```diff
@@ -18,6 +18,15 @@ if [ $? = 0 ]; then
     echo "IDE environment variables have been set for ${IDE_HOME} in workspace ${WORKSPACE}"
   fi
 fi
+
+if [ "${OSTYPE}" = "cygwin" ]; then
+    echo -e "\n--- WARNING: CYGWIN IS NOT SUPPORTED ---"
+    echo "Cygwin console is not supported by IDEasy. Consider using the git-bash terminal"
+    echo "instead. If you want to use Cygwin with IDEasy, you will have to configure it"
+    echo "yourself. A few suggestions and caveats can be found here:"
+    echo -e "https://github.com/devonfw/IDEasy/blob/main/documentation/Not_supported_Cygwin.adoc\n"
```
hohwille, line 27: Even though nice-to-have and syntactic sugar but we use ANSI color codes and log warnings in a special orange color.
Generic solution for bash can be found here:
https://github.com/devonfw/ide/blob/15e75c01ce3740b8f52a8a76c257dd190726161e/scripts/src/main/resources/scripts/functions#L246-L248

So please do not introduce another function here (since this file is sourced into the bash shell of the user) but instead join all into a single `echo -e` statements using `\n` for newlines.
Also it is a nice style if you can place newlines at the end of a sentence instead of before the last word of a sentence what makes it not as good to read.

FYI: Java equivalent is here:
https://github.com/devonfw/IDEasy/blob/881c978e4440c72cf7a690caa788285653d3ebf7/cli/src/main/java/com/devonfw/tools/ide/log/IdeLogLevel.java#L33

documentation/Not_supported_Cygwin.adoc
```diff
@@ -0,0 +1,21 @@
+:toc:
+toc::[]
+
+= Not supported: Cygwin
```
hohwille, line 4: For our automated AsciiDoc processing we must not use special characters like colon in section titles:
https://github.com/devonfw/docgen/blob/master/documentation/guidelines.asciidoc

could you also rename this file to just `cygwin.adoc`.
Over time our support for cygwin may increase one day and then we do not need to rename the file, etc.

After merging, I can add some hints here how cygwin user can make it still work.

hohwille (approved):
@nmCapG thanks for your PR. That was quite fast 👍 
I left some comments for further improvement.
Please have a look. Once resolved, we can merge.

jan-vcapgemini (approved):
LGTM.

hohwille:
Thanks for the rework. I just merged it. Now I realize that CHANGELOG was not updated. Please check complete DoD next time ;) Will update the CHANGELOG for you directly outside PR.

PR #746, opened 2024-11-05
#745: fix loading and evaluation of variables

### LINKED ISSUE #745: Maven install fails with NPE
### Expected behavior

As a IDEasy user, I want to get helpful error messages in case of config errors and a robust behaviour so that I do not waste my time with deep debugging of the product internals.

### Actual behavior

```
$ ide install mvn
Start: Install mvn
Step 'ide (install,mvn)' failed: java.lang.NullPointerException
 Step 'Install mvn (null)' failed: java.lang.NullPointerException
2 step(s) failed out of 3 steps.
An unexpected error occurred!
https://github.com/devonfw/IDEasy/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=java.lang.NullPointerException
java.lang.NullPointerException
        at com.devonfw.tools.ide.tool.LocalToolCommandlet.install(LocalToolCommandlet.java:82)
        at com.devonfw.tools.ide.tool.ToolCommandlet.install(ToolCommandlet.java:223)
        at com.devonfw.tools.ide.commandlet.InstallCommandlet.run(InstallCommandlet.java:49)
        at com.devonfw.tools.ide.context.AbstractIdeContext.applyAndRun(AbstractIdeContext.java:862)
        at com.devonfw.tools.ide.context.AbstractIdeContext.run(AbstractIdeContext.java:771)


```

1. edit your config (`devon.properties` or `ide.properties`):
```
MVN_VERSION=3.9.6
MAVEN_VERSION=${MVN_VERSION}
```
3. ide install mvn

### Comments/Hints:

I hit multiple problems here. First of all `MAVEN_VERSION=${MVN_VERSION}` is a legacy workaround that could be used in `devonfw-ide` to make the configuration work also with older versions of `devonfw-ide` that expected `MAVEN_VERSION` what was changed in newer releases.
Further with #104 we implemented that the value of a variable can reference other variables including itself so `${MVN_VERSION}` can be evaluated by `EnvironmentVariablesResolved`.
However, we also implemented variable legacy support via `IdeVariableDefinition` with legacy name for fallbacks:
https://github.com/devonfw/IDEasy/blob/d967d69d36dccd211b7a02af8929fe695e734959/cli/src/main/java/com/devonfw/tools/ide/variable/VariableDefinition.java#L24

The goal is that we can read legacy variables from devonfw-ide easily even though they may not be physically defined in any properties file by "redirecting" it from the legacy name to the new name.
To simplify variable resolution, we also consider that when loading properties and already store them under the new instead of the legacy name since otherwise we might have two variables defined for the same thing with different values (MVN_VERSION=3.9.9 and MAVEN_VERSION=3.0.0). However, if we then only use the new one for installation and processing but have template configuration files with `${MAVEN_VERSION}` or `$[MAVEN_VERSION]`, we might end up with the wrong value.

Further, a lot of edge-cases can happen here.
Assuming we have:
```
LEGACY_X=legacyX
NEW_X=newX
NEW_X=newX2
NEW_Y=newY
LEGACY_Y=legacyY
```
The expected result should be the same as for
```
NEW_X=newX2
NEW_Y=newY
```

Further while loading these strange assumed config, we should log warnings so the user gets informed that his configuration is odd and he should probably do a cleanup.
In the future we might want to add a commandlet to automatically cleanup all configs and migrate to IDEasy:
* convert devon.properties to ide.properties
* remove legacy variables and only have new/official variables
* convert curly variable syntax for supported variables in all templates to bracket variable syntax (in combination `${DEVON_IDE_HOME}` -> `$[IDE_HOME]`).
* etc.

### Affected version:

* 2024.11.001-beta-SNAPSHOT

hohwille:
fixes #745

* fix loading of variables with handling of legacy variables
* log warnings for legacy variables or duplicate variable definition in same file
* allow overriding of variable with empty value instead of ignoring
* fix `EnvironmentVariables.getToolVersion` and prevent NPE
* improved tests to cover all these edge-cases


cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
```diff
@@ -6,40 +6,53 @@
...
+        propertiesFilePath, context);
     // assert
     assertThat(variables.getType()).isSameAs(TYPE);
     assertThat(variables.get("MVN_VERSION")).isEqualTo("3.9.0");
     assertThat(variables.get("IDE_TOOLS")).isEqualTo("mvn, npm");
     assertThat(variables.get("CREATE_START_SCRIPTS")).isEqualTo("eclipse");
     assertThat(variables.get("KEY")).isEqualTo("value");
-    assertThat(variables.getVariables()).hasSize(4);
+    assertThat(variables.get("IDE_MIN_VERSION")).isEqualTo("2024.11.001");
+    //assertThat(variables.get("JAVA_VERSION")).isEqualTo("illegal");
```
jan-vcapgemini, line 47:
```suggestion
```

cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
```diff
@@ -6,40 +6,53 @@
...
 /**
  * Test of {@link EnvironmentVariablesPropertiesFile}.
  */
 @SuppressWarnings("javadoc")
-class EnvironmentVariablesPropertiesFileTest extends Assertions {
+class EnvironmentVariablesPropertiesFileTest extends AbstractIdeContextTest {
 
   /**
    * Test of {@link EnvironmentVariablesPropertiesFile} including legacy support.
    */
```
jan-vcapgemini, line 27: What happened here? Shouldn't this be moved?

cli/src/test/resources/com/devonfw/tools/ide/environment/var/devon.properties
```diff
@@ -3,5 +3,10 @@
 #************************************************************************************************
 DEVON_IDE_TOOLS=(mvn npm)
 DEVON_CREATE_START_SCRIPTS=(eclipse)
-MAVEN_VERSION=3.9.0
+MVN_VERSION=undefined
+MVN_VERSION=3.9.0
+MAVEN_VERSION=3.6.1
+DEVON_IDE_MIN_VERSION=2023.07.003
+IDE_MIN_VERSION=2024.11.001
+JAVA_VERSION=
```
jan-vcapgemini, line 11: What about the case of: `MVN_VERSION=${MAVEN_VERSION}`?

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
```diff
@@ -75,12 +76,35 @@ private void load() {
...
           if (name != null) {
-            variableLine = migrateLine(variableLine, false);
-            name = variableLine.getName();
-            String value = variableLine.getValue();
-            this.variables.put(name, value);
+            VariableLine migratedVariableLine = migrateLine(variableLine, false);
+            if (migratedVariableLine == null) {
+              this.context.warning("Illegal variable definition: {}", variableLine);
+              continue;
+            }
```
jan-vcapgemini, line 83: This should be improved later I think. Maybe add a TODO with a link to a new issue so we don't forget it.

jan-vcapgemini (approved):
Thanks for this bugfix. I've added some CRs, please check. Approving so it won't be blocked.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
hohwille: What should be improved here? I do not see the reason for a TODO.

cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
hohwille: Indeed an undesired left-over. Thanks. Removed.

cli/src/test/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFileTest.java
hohwille: This was cause by some git merge since that is not merging semantically. I moved the JavaDoc back to the method where it was before.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
hohwille: To make this clear because IMHO you did not properly understand this code:
* most probably this if can never be entered.
* if it should ever be entered then the `migrateLine` method returned `null` and therefore eliminated the original `variableLine` since it was somehow illegal syntax. Instead of causing an NPE it makes sense to log this and continue.
* you could remove this `if` block entirely but then stupid IntelliJ complains that in the next line there is a potential NPE.

As I need to merge all stories assigned to our milestone today EOB, I can only resolve this.
If I am missing something and you still see a problem here, please create a new issue.

cli/src/test/resources/com/devonfw/tools/ide/environment/var/devon.properties
hohwille: This was the other way around...
To make you happy, I changed this `devon.properties` accordingly.
However, it will not make any difference. You need to add a `EnvironmentVariablesResolvedTest` to really test this case.
Currently that does not exist and is not in scope of this story.

Issue #737, opened 2024-11-04
Add cd command to shell commandlet


jan-vcapgemini:
This is a follow up issue of: #168 
We need to implement a `cd` command into our `shell` commandlet. This could be done in a similar way as the `exit` command #533.

PR #748, opened 2024-11-06
#737: Added cd command to shell commandlet

leonrohne27:
### Implements:

- Added the command "cd" to the shell commandlet that allows the user to switch the directory inside the ide shell


cli/src/main/java/com/devonfw/tools/ide/commandlet/ShellCommandlet.java
```diff
@@ -129,9 +132,40 @@ private int runCommand(String args) {
...
+      this.context.error("Error: 'cd' requires a directory argument.");
+      return -1;
+    }
+
+    String targetDir = String.valueOf(cliArgs.next());
+    Path path = Paths.get(targetDir);
+
+    // If the given path is relative, resolve it relative to the current directory
+    if (!path.isAbsolute()) {
+      path = Paths.get(System.getProperty("user.dir")).resolve(path).normalize();
```
jan-vcapgemini, line 154:
```suggestion
      path = context.getCwd();
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/ShellCommandlet.java
```diff
@@ -129,9 +132,40 @@ private int runCommand(String args) {
...
+    }
+
+    // Check if the path exists and is a directory
+    if (Files.exists(path) && Files.isDirectory(path)) {
+      // Set the current working directory to the new path
+      System.setProperty("user.dir", path.toAbsolutePath().toString());
```
jan-vcapgemini, line 160:
```suggestion
      context.setCwd(path, context.getWorkspaceName(), context.getIdeHome());
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/ShellCommandlet.java
```diff
@@ -129,9 +132,40 @@ private int runCommand(String args) {
...
+      return changeDirectory(cliArgs);
+    }
+
     return ((AbstractIdeContext) this.context).run(cliArgs);
   }
 
+  private int changeDirectory(CliArguments cliArgs) {
+    if (!cliArgs.hasNext()) {
+      this.context.error("Error: 'cd' requires a directory argument.");
+      return -1;
```
jan-vcapgemini, line 146: I'm not sure about the error code.

jan-vcapgemini (changes requested):
Thanks for implementing the cd command to our Shell. I've added some CRs, please resolve.

jan-vcapgemini (approved):
Thanks, LGTM. Ready for review.

cli/src/main/java/com/devonfw/tools/ide/commandlet/ShellCommandlet.java
leonrohne27: should be fine now.

cli/src/main/java/com/devonfw/tools/ide/commandlet/ShellCommandlet.java
```diff
@@ -79,13 +81,13 @@ public void run() {
...
         // TODO: implement TailTipWidgets, see: https://github.com/devonfw/IDEasy/issues/169
 
-        String prompt = "ide> ";
         String rightPrompt = null;
         String line;
 
         AnsiConsole.systemInstall();
         while (true) {
           try {
+            String prompt = "ide " + context.getCwd() + "> ";
```
hohwille, line 90: I see some general problem here with the design of our shell feature:
* Initially we only supported IDEasy commandlets.
* Therefore we changed the promt to `ide>`
* In other words if submit `ide> «args»` in our shell, you will get more or less the same effect as if you call `ide «args»` outside the shell.
* Now we start integration other non `ide` commands to the shell like `cd` (how about `icd`, `find`, `mv`, `cp`, `grep`, `cat`, etc. ?) this does not make sense anymore
* Instead, I would suggest that we pre-define the user input to `"ide "`  as if the user would have typed this manually.
* Then the user can use our shell like before
* But he also has the possibility to use Backspace to remove this pre-defined input and change it to `cd some-folder`.

If you want to do a quick fix, you can do a KISS solution and switch:
```suggestion
            String prompt = context.getCwd() + "$ ide ";
```
But IMHO it is inconsistent that the user does `ide cd some-directory` while the is no `cd` commandlet.

Further, it is a great idea to show the current working directory like in other shells.
However, the CWD/PWD can get very long:
image
This is a common problem discussed about Linux forums
Simple solution is to only print the file-name of the CWD (foldername instead of full path).
Or to limit the length - see e.g.
https://superuser.com/questions/1346795/multi-line-bash-prompt-if-path-is-longer-than-x-characters
We should create a separate ticket for this last thought, but it would be good to also address this.

My suggestion would be that you create two issues (one for the pre-define the user "ide " and one for the handling of long path to avoid line wrapping) and then just apply some pragmatic solution like my proposed change that already prepares for the first story. Then we merge this PR and other developers can take over implementing the new stories.

cli/src/main/java/com/devonfw/tools/ide/commandlet/ShellCommandlet.java
```diff
@@ -129,9 +131,39 @@ private int runCommand(String args) {
...
+    if ("cd".equals(arguments[0])) {
+      return changeDirectory(cliArgs);
+    }
+
     return ((AbstractIdeContext) this.context).run(cliArgs);
   }
 
+  private int changeDirectory(CliArguments cliArgs) {
+    if (!cliArgs.hasNext()) {
+      this.context.error("Error: 'cd' requires a directory argument.");
```
hohwille, line 144: Not entirely true.
In bash/Linux/Posix the `cd` command without arguments goes to the users home directory (`cd $HOME`).
I would suggest to implement that since it is very easy.

hohwille (approved):
@leonrohne27 thanks for your PR. This is nice work and you digged to find the right code locations to implement the story. 👍 
I left 2 simple review comments. If you have the time left to address them this would be lovely and we can then merge. Thanks!

cli/src/main/java/com/devonfw/tools/ide/commandlet/ShellCommandlet.java
leonrohne27: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/ShellCommandlet.java
```diff
@@ -129,9 +131,37 @@ private int runCommand(String args) {
...
+    if ("cd".equals(arguments[0])) {
+      return changeDirectory(cliArgs);
+    }
+
     return ((AbstractIdeContext) this.context).run(cliArgs);
   }
 
+  private int changeDirectory(CliArguments cliArgs) {
+    if (!cliArgs.hasNext()) {
+      Path homeDir = Paths.get(System.getProperty("user.home"));
```
hohwille, line 144: Not testable this way.
```suggestion
      Path homeDir = this.context.getUserHome();
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/ShellCommandlet.java
hohwille: Thanks for creating the issues:

* #821
* #822

👍

Issue #563, opened 2024-08-26
Final attempt to improve deployment workflows


hohwille:
While I was stumbling over this topic again with my PR #562 I want to finally get this topic done once and for all.
I thought I had made my goals clear with #427 but this issue was closed while important aspects were never implemented.
- [ ] In the matix OS build all we should do is compile the code and build the native image - no documentation, not test, no packaging/assembly or whatever. That native image should be uploaded via github action upload as executable (not as tar.gz archive) to `${matrix-os}/ideasy[.exe]`.
- [ ] In the main build we should download the native images via github action download and then trigger the full deployment build. Here the assembly should be build and then be attached as artifact. These assembly can be changed so that they refer the native image directly from the downloaded directory `${matrix-os}/ideasy[.exe]` so we do not need additional mapping in the github actions scripts. No need for build-helper-plugin and pointless file-mappings.
- [ ] rename the assembly files to reasonable filenames again as they are not connected to `${matrix.os}` anymore. So `release-win-x64.xml`, `release-mac-arm.xml`, etc.
- [ ] ensure to update the `pom.xml` so all assembly descriptors are run during the build. You do not even need to list them all: https://maven.apache.org/plugins/maven-assembly-plugin/single-mojo.html#descriptorSourceDirectory
- [ ] We aggreed that the entire build process gets documented as already described in #427 but never implemented yet.

This ticket shall not be closed before all tasks are implemented (or moved to other tickets that do not get closed before implemented).
Sorry for saying the obvious but it seemed necessary.

PR #749, opened 2024-11-06
#563 Optimize workflows

jan-vcapgemini:
Fixes: #563

Implements:
* improved nightly build deployment workflow
* adjusted upload and download artifact actions
* changed native image build phase from package to compile
* added new build profile for assemblies
* renamed assembly xml files accordingly
* added workflow documentation


cli/pom.xml
```diff
@@ -174,6 +174,33 @@
...
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-assembly-plugin</artifactId>
+            <version>${assembly.maven.plugin.version}</version>
+            <executions>
+              <execution>
+                <id>create-distribution</id>
+                <phase>install</phase>
+                <goals>
+                  <goal>single</goal>
```
hohwille, line 191: dont we have to change this now? With the new design we need to build all releases in one maven run now so we need to tell assembly plugin to build all assemblies in a "loop".

hohwille (commented):
@jan-vcapgemini thanks for working on #563 and bringing this cleanup ahead. 👍

jan-vcapgemini:
I'm not sure about the deploy step of the release job yet. Currently I start two profiles in the deploy phase for the release. I fear that the assembly profile is not being started in the correct order.
The deploy step of the nightly build uses another approach though. First I start the assembly using the install phase and then I skip the maven install in the deploy step. I want to prevent the repeated build of the documentation.

documentation/github-workflows-reference.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+=== native
+
+For building the native images using graalvm.
+
+=== assembly
+
+For the creation of release archives (SNAPSHOT releases for SonarType and official releases for Maven Central).
+The Maven Assembly Plugin uses specific assembly configuration files located in `cli/src/main/assembly`.
+These assembly files are specify the name of the respective archive to create by its `id`.
```
alfeilex, line 42: Maybe specify what is meant by `id`, as there is no explicit mention of about how the different OSs are identified.

documentation/github-workflows-reference.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+The nightly build iterates over each OS/architecture runner e.g. `windows-latest` (Windows x64), `macos-latest` (Mac ARM) etc. and builds a binary image.
+Then the job uploads the binaries to GitHub as an artifact using the GitHub `actions-upload` tool.
+Each artifact archive is then named by taking a prefix `natives-` and the runner name taken from the `matrix:os` e.g. `natives-windows-latest`.
+
+=== deploy
+
+The `deploy` job downloads the image artifacts from GitHub into the `cli/target` directory.
+Then it builds the documentation PDF file and creates release archives by packing the downloaded binaries and the `IDEasy.pdf` file into a separate archive file for each OS/architecture e.g. `bin/ideasy.exe` for Windows.
+This behaviour is controlled by the Maven Assembly Plugin which is being described later.
```
alfeilex, line 24: maybe add a direct link to the description

documentation/github-workflows-reference.adoc
jan-vcapgemini: I've added some examples now.

cli/pom.xml
jan-vcapgemini: Can be closed as it is still a single process, using `descriptorSourceDirectory` now though.

hohwille: This was a misunderstanding on my end.
I was expecting the `single` goal to process a single Assembly XML file. However, it already builds all Assembly XML files since we configure the directory containing them.

documentation/github-workflows-reference.adoc
```diff
@@ -0,0 +1,47 @@
...
+We have two main workflow files:
+
+== nightly-build.yml
+
+For the creation of SNAPSHOT releases on our OSS Nexus using two steps:
+
+=== build-natives
+
+The nightly build iterates over each OS/architecture runner e.g. `windows-latest` (Windows x64), `macos-latest` (Mac ARM) etc. and builds a binary image.
+Then the job uploads the binaries to GitHub as an artifact using the https://github.com/actions/upload-artifact[GitHub `actions-upload`] tool.
```
alfeilex, line 17: A small adjustment on my side would be to link `actions-upload` without Github, as that would look better in the docs

hohwille (approved):
@jan-vcapgemini thank you so much for implementing and testing this refactoring as well as writing documentation. I am very happy to see out deployment finally in this great state. Excellent work 🥇 

There is some room for documentation improvements: I would rather focus on release/deployment in the docu and rename it to `github-release.adoc`.
I am happy to contribute to improve the documentation but IMHO we can already merge and I can then improve doc later.

Issue #739, opened 2024-11-04
Replace red "You are not inside an IDE installation" message


jan-vcapgemini:
Currently IDEasy displays a big red `You are not inside an IDE installation: D:\Projects` message when the `ide` command is being used.
We should remove this message. When IDEasy is installed, and I open a shell I always get this message that is not really adding value but is annoying some users. If I just type `ide` and I am in a project, I get the according success message if environment was setup. In case I call `ide` outside of a project and do not get this message but no output I can derive the same information that nothing happened.

PR #753, opened 2024-11-12
#739: removed "not inside an IDE installation message"

leonrohne27:
### Implements:

- removed the message "you are not inside an IDE installation"


jan-vcapgemini (approved):
Thanks, LGTM. Ready for review.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -277,10 +272,10 @@ private static String getMessageIdeRootNotFound() {
...
   public String getMessageIdeHome() {
 
-    if (this.ideHome == null) {
-      return getMessageIdeHomeNotFound();
+    if (this.ideHome != null) {
+      return getMessageIdeHomeFound();
     }
-    return getMessageIdeHomeFound();
+    return null;
   }
```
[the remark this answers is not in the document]
hohwille: This is incorrect IMHO.
If I call a commandlet that requires IDE_HOME the user needs to still get this error message to understand that his command failed.
```
$ ide install mvn
You are not inside an IDE installation: D:\projects

Error: IDEasy failed with exit code 2
```
This is still correct. However, this should be changed
```
$ ide
You are not inside an IDE installation: D:\projects
```

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -819,9 +803,7 @@ private ValidationResult applyAndRun(CliArguments arguments, Commandlet cmd) {
     }
     if (result.isValid()) {
       debug("Running commandlet {}", cmd);
-      if (cmd.isIdeHomeRequired() && (this.ideHome == null)) {
-        throw new CliException(getMessageIdeHomeNotFound(), ProcessResult.NO_IDE_HOME);
-      } else if (cmd.isIdeRootRequired() && (this.ideRoot == null)) {
```
hohwille, line 824: You should not remove this.
It will cause strange errors otherwise and therefore this handling must stay.

leonrohne27: Thanks, I readded it

cli/src/main/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandlet.java
```diff
@@ -53,7 +54,9 @@ public boolean isProcessableOutput() {
 
   @Override
   public void run() {
-
+    if (context.getIdeRoot() == null) {
```
hohwille, line 57: IDE_ROOT will always be defined after proper setup of IDEasy.
It is IDE_HOME that may not be defined and that is why you changed `isIdeHomeRequired()` above (there is also isIdeRootRequired()).
```suggestion
    if (context.getIdeHome() == null) {
```

cli/src/main/java/com/devonfw/tools/ide/cli/NullIdeRootException.java
```diff
@@ -0,0 +1,15 @@
+package com.devonfw.tools.ide.cli;
+
+/**
+ * {@link CliException} that is thrown if IDE_ROOT == null
+ */
+public class NullIdeRootException extends CliException {
```
hohwille, line 6: If you really want to express what you stated in the JavaDoc, it would be better to rename it to `IdeRootIsNullException` so one can better understand and guess what all this is about.
However, an exception that we use to abort without any message should not be named after the condition it was caused by but rather than the problem type it represents and potential handling it may imply.
Therefore I would suggest renaming to `CliCancelException` or `CliExitException`.
Also it would be good if you define a custom exit code like already done in `CliAbortException`.

leonrohne27: done

hohwille (approved):
@leonrohne27 thanks for the rework. Nice job 👍 
Ready for merge.

hohwille:
CHANGELOG entry is still missing. Could you quickly add that?
Otherwise I will merge in ~1h and add it after the merge...

PR #755, opened 2024-11-12
#754: quickfix via lazy init

### LINKED ISSUE #754: Again messages break processable command output
```
$ ide -v
Duplicate variable definition MVN_VERSION with old value '0.1' and new value '3.9.6' in D:\projects\salog\settings\devon.properties
Both legacy variable MAVEN_VERSION and official variable MVN_VERSION are configured in D:\projects\salog\settings\devon.properties - ignoring legacy variable declaration!
2024.11.001-beta-SNAPSHOT
bash: $'\E[93mDuplicate': command not found
bash: $'\E[93mBoth': command not found
```
See #627 and #531.
This error was introduce by me with PR #746

Adding a log warning to some reasonable place should not lead to this error.

It is not acceptable that before adding a log warning you always have to understand if that may be triggered during context initialization since this way this error will come back again and again.

We need to find a way to finally fix this problem once and forever!

hohwille:
fixes #754 
but only as quickfix.
For a "once and forever" fix, we might need to logger that is initially collecting log statements and only activated if the commandlet to run is determined and the apply the collected log events.
We can still discuss how to solve this properly.
However, for our release the current quickfix would be enough and we can merge this and leave #754 open.

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
@@ -247,10 +247,8 @@ public abstract class AbstractIdeContext implements IdeContext {
     this.userHomeIde = this.userHome.resolve(".ide");
     this.downloadPath = this.userHome.resolve("Downloads/ide");
 
-    this.variables = createVariables();
     this.path = computeSystemPath();
     this.customToolRepository = CustomToolRepositoryImpl.of(this);
-    this.workspaceMerger = new DirectoryMerger(this);
   }
 
   private String getMessageIdeHomeFound() {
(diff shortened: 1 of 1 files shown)


jan-vcapgemini:
I've tested this change with a newly compiled binary in a testproject with some "broken" variables. Now if I use `ide -v` I get only the version without your warnings. Only if I use `ide -d -v` I get your warnings now. Is this intended? The parsing error is gone though.

jan-vcapgemini (commented):
Looks good to me. I've added a question though, will approve if everything is ok.

hohwille:
> I've tested this change with a newly compiled binary in a testproject with some "broken" variables. Now if I use `ide -v` I get only the version without your warnings. Only if I use `ide -d -v` I get your warnings now. Is this intended? The parsing error is gone though.

Yes, that is the planned behaviour:
https://github.com/devonfw/IDEasy/blob/64a73e49e30865d3fc04a4a647d17b49bab71d8e/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L823-L830

Issue #404, opened 2024-06-20
enhance logging concept


hohwille:
In IDEasy we have created our own logger below `IdeContext`.
This has some special feature and somehow makes sense since what we want for the regular end-user is something different than what a regular logger provides...

However, from a design PoV (and during JUnits) regular logging via SLF4J has several advantages:
* logger can be defined as static final constant simplifying dependencies (no `IdeContext` must be passed to constructor of every class that wants to log something)
* during the tests we can see which class was triggering the logging

On the other hand we have some edge-cases that cannot be easily addressed by SLF4J:
* custom log-levels like `step` or `interaction`
* advanced test-support where we can assert that specific lines or fragments have been logged

The idea of this story is to analyze and after agreement also implement some improvement to our logging.
Currently I am thinking about a hybrid approach:
* What it any class in IDEasy could also log with SLF4J - During JUnits we would then use logback see the regular log output including the Class containing the log statement. 
* By implementing our own SLF4J Bridge Implementation, we could make SLF4J log to our custom logger in regular IDEasy CLI usage.
* We could investigate if SLF4J API & core would work in GraalVM (what should already be sure) and still not increase our footprint (native-image size, performance, memory consumption, etc. - what is also 99% likely). If results are fine, we could decide for this apprach and include SLF4J into IDEasy.
* We could even evaluate if SLF4J API can somehow support custom log-levels (e.g. via Markers). Otherwise I would suggest an hybrid approach where the current ldeContext logger remains as is and we add a SLF4J Bridge that delegates to it (in the worst case we would again need to go back to a static instance variable holding the latest `IdeContext` instance like we had before).

Wouldn't that be a great improvement?

PR #760, opened 2024-11-13
Enhance/404 logging concept

KianRolf:
draft pr for: https://github.com/devonfw/IDEasy/issues/404

Attempt to create custom slf4j provider by implementing 3 slf4j interfaces. Will result in the following error
Screenshot 2024-11-13 161610


hohwille:
This PR is still in new so I never looked at it before.
Most probably Jan-V. wanted to do a team review and probably complete things together with Kian.
Anyway, here is my Review feedback:
* Great that you found this trick with `META-INF/services` to control which implementation SLF4J will use. This is the concept used by Java ServiceLoader. A lot of Java developers with years of development experience have never heard of such things before so it is great that you figured this out :)
* You properly found a way to provide our own logger implementation for SLF4J 👍 
* Please add JavaDoc: Your methods are already annotated with `@Override` - that is great so nothing is to be done here but your classes should have a JavaDoc:
```java
/**
 * Implementation of {@Logger} for this project. Allows logging to the console with coloured output.
 */
public class TestLogger implements Logger {
```
BTW: you should avoid fully-qualified types in your code and use `import` statements instead (only use fully-qualified types when necessary because you are using 2 different types with the same simple name).
* Further, you should rename this class: It is located in `src/main/java` and that is correct, because it is to be used for the productive usage by the end-user. The name `TestLogger` implies that is will only be used for testing. A better name could be e.g. `IdeLoggerAdapter`. or `IdeSlf4jLoggerImpl`.
* Finally, your logger implementation is more or less empty. I guess your are fully aware of that but since the PR is not in draft state, I just wanted to explicitly mention this because here comes the tricky part: You need to delegate to `IdeLogger` and do additional things. Therefore I assume this PR cannot be finished by you and someone would need to take over.

jan-vcapgemini:
Closing and continuing in #1017

Issue #449, opened 2024-07-05
Consider integration of continue GenAI coding assistant


hohwille:
As an IDEasy user, I want to make use of AI tools so that I can get smart coding assistance.
Similar to #412 we want to offer various state-of-the-art solutions and alternatives to projects.
This story is about [continue](https://github.com/continuedev/continue).
It comes as IntelliJ plugin and as VSCode plugin.
Both a linked from their website and should be pre-configured for IDEasy (with `active=false` and tagged with `ai`).

alfeilex:
Related PR in `ide-settings`: https://github.com/devonfw/ide-settings/pull/67

alfeilex:
There is a `config.json` file that is created after the plugin has been installed. But it is located in `%USERPROFILE%\.continue\config.json`.

For specific configuration there is possibility to create a `.continuerc.json` file in the root folder of the project. 
Reference: https://docs.continue.dev/customize/overview#per-workspace-configuration

How should we add the pre-configuration of this plugin or is it even reasonable?

alfeilex:
For the pre-configuration, the model that is hosted with `vllm` could be pre-configured by the user in `.continuerc.json`. 

I try to figure out when it's possible to use a `vllm` hosted model for autocompletion. 

But how I can test the functionality. My hardware is probably to weak to handle `vllm` hosted models

hohwille:
Only documentation was merged so far. Hence, reopening until plugin config is also merged.

alfeilex:
We have think about how to design a template mechanism for the settings

See here: https://github.com/devonfw/ide-settings/pull/67#pullrequestreview-2470331591

PR #761, opened 2024-11-13
#449: Add Continue coding assistant plugin and docs for Intellij and VScode

alfeilex:
Fixes: #449

This PR adds a small documentation about coding-assistants and specially a guideline to pre-configure `continue` plugin that is available for `intellij` and `vscode`

Further information: 
- The `.continuerc.json` file has to be in the root folder, so it can't be in the code repository
- Tested on VSCode
- Latest version `0.75` ignores the pre-configuration file in Intellij but version `0.80` worked out
- There is a remote url configuration that could be interesting to share configuration in teams

Related settings PR: https://github.com/devonfw/ide-settings/pull/67


alfeilex:
During my test in `intellij`, I mentioned that the pre-confguration file `.continuerc.json` is being ignored. I need to determine if this is a bug by `continue` itself or something is wrong with our intellij config.

alfeilex:
> I am wondering if JetBrains removed the version from the marketplace.

I figured out that version `0.80` had some bugs and was therefore removed from the marketplace.

Issue #750, opened 2024-11-11
Small cleanup from PR #724


hohwille:
From PR #724 see the review comment discussion:
https://github.com/devonfw/IDEasy/blob/55db46d0e936fd7e81aaa353b786bc8b825a04e8/cli/src/main/java/com/devonfw/tools/ide/context/GitUrlSyntax.java#L71
> I am not sure if the default case can actually occur, because if the enum is set incorrectly, then a IllegalArgumentException is thrown by the Enum class before.
> > This is a common problem pattern. The best way to solve this in an enum is to avoid the switch at all:
simply make the method abstract and create a method implementation for each enum constant.
This way, if new enum constants are added, the compiler will force the developer to implement the abstract method and he cannot forget it due to the compiler error.
I will create a small cleanup story for this improvement but we can already merge this PR as planned.

This should be cleaned up with this story.

PR #762, opened 2024-11-14
#750: Created method implementation for each enum constant

leonrohne27:
### Implements:

- Removed switch and created a method implementation for each enum constant instead
- Changed the method to abstract


hohwille (approved):
@leonrohne27 thanks for this PR and implementing this code-improvement. Very nice 👍

Issue #751, opened 2024-11-11
Small cleanup of GlobalToolCommandlet from PR #684


hohwille:
From PR #684 we have some small left-overs from the review discussion:
- [ ] Move `postInstall()` to `LocalToolCommandlet` and remove it from `[Global]ToolCommandlet`:
> But what are we doing with `postInstall()`
https://github.com/devonfw/IDEasy/blob/071e732e6c74aed7b12ccc1d522faf1c4a015dad/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java#L142-L144
- [x] Improve log statement of GlobalToolCommandlet:
https://github.com/devonfw/IDEasy/blob/55db46d0e936fd7e81aaa353b786bc8b825a04e8/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java#L149
> If the process runs in background mode, then message Successfully installed docker in version 1.16.0 will be output immediately after the setup installer window has opened.
> > We can change the log message but this is the only reasonable behaviour. We trigger the installation and the end-user needs to click through the installation process. This can take a minute or several hours if the user has fallen asleep. Not much that we can do about it...

PR #763, opened 2024-11-14
#751: Moved postInstall() and changed toll installation message

leonrohne27:
### Implements:

- Moved the method postInstall() from GlobalToolCommandlet to LocalToolCommandlet
- Changed the installation message so that the user is not getting confused

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
@@ -106,6 +106,7 @@ public abstract class LocalToolCommandlet extends ToolCommandlet {
       } else {
         step.success("Successfully installed {} in version {} replacing previous version {}", this.tool, resolvedVersion, installedVersion);
       }
+      postInstall();
       return true;
     } catch (RuntimeException e) {
       step.error(e, true);
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
@@ -147,12 +147,11 @@ public abstract class GlobalToolCommandlet extends ToolCommandlet {
       fileAccess.delete(tmpDir);
     }
     if (exitCode == 0) {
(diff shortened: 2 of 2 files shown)


jan-vcapgemini (approved):
LGTM. Ready for review.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -106,6 +106,7 @@ public final boolean install(boolean silent, EnvironmentContext environmentConte
       } else {
         step.success("Successfully installed {} in version {} replacing previous version {}", this.tool, resolvedVersion, installedVersion);
       }
+      postInstall();
```
hohwille, line 109: `postInstall` is already called above in line 103. Please remove this.
```suggestion
```
See also 
https://github.com/devonfw/IDEasy/blob/73c1efcb20e441201e80356fb261b2ded122548f/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java#L242-L247

hohwille (changes requested):
@leonrohne27 thanks for your PR. You solved the second task of #751 nicely. 👍 
For the first task there is a slight misunderstanding - sorry, if I was not precise enough:
> Move postInstall() to LocalToolCommandlet and remove it from [Global]ToolCommandlet:

I expect that the `postInstall` method(s) (with its body/implementation) gets moved to `LocalToolCommandlet`. If `GlobalToolCommandlet` does not support `postInstall`, then it does not make sense anymore if it has a method `postInstall`.
It is fully correct that you removed the invocation of `postInstall()` from `GlobalToolCommandlet` but you also need to move `postInstall()` and `postInstall(boolean)` from `ToolCommandlet` to `LocalToolCommandlet`.
Then sub-classes of `GlobalToolCommandlet` cannot override `postInstall` anymore and since we always use `@Override` annotations in such case, we would even get compile errors if we missed something.
Please have a look and update your PR. Thanks.

jan-vcapgemini (approved):

hohwille (approved):
@leonrohne27 thanks for rework, ready for merge now 👍

Issue #637, opened 2024-09-19
Option to disable updates


hohwille:
As a IDEasy user, I want the ability to start my IDE fast without checking for and installing of updates so that I can start develop asap.

Currently when I set a version to `*` then I always get the latest version.
In general this is fine: latest is greatest. I can still fix the version, if I do not want updates.

However, as a result every time I start a tool (e.g. `intellij` or `eclipse` or `vscode`, etc.), IDEasy will check for updates and install it if found what means I have to wait for download, extraction, etc.
Sometimes there are frequent updates of `intellij` and this behaviour can get annoying.
E.g. if I am in a meeting and doing a screensharing, I do not want to wait the audience for 5 minutes or more to watch the update being installed.

As a solution, I suggest to add a new global option e.g. `--no-updates` (does someone have a shorter/better name?).
If that option is active, then we check if the installed version is matching the configured version.
If yes, we skip any check for newer versions and updates.
However, if a tool is not yet installed or the version does not match the configured version, then I still get a download & installation.
Therefore, the existing `--offline` option is not a suitable workaround.
If I change `JAVA_VERSION=17*` to `JAVA_VERSION=21*` and then do `ide --no-updates intellij`, I still expect Java to be updated from 17 to 21. This example also indicates that a better and more intuitive name for the option would be nice.

Issue #664, opened 2024-09-30
Tool dependencies and environment variables still not working properly


hohwille:
### Expected behavior

As a IDEasy user, I want to install and run tools seamlessly even if they have complex dependencies so that I do not have to worry about it.

### Actual behavior

If I run the tool (e.g. tomcat) for the first time it is installed together with its dependencies.
This is also setting the environment variables and PATH extensions of dependent tools:
https://github.com/devonfw/IDEasy/blob/42201edae4aecf0687ae4c8437d0ba692401dda8/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java#L372

This is called from `installTool` method but this is skipped if the tool is already installed:
https://github.com/devonfw/IDEasy/blob/42201edae4aecf0687ae4c8437d0ba692401dda8/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java#L80-L84

Therefore on the first run of tomcat the environment variables will be set for tomcat including java but on the second run this does not happen.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. ide tomcat
2. ide tomcat

### Related/Dependent Issues

#131 
#439

### Comments/Hints:

IMHO we have to run installTool also if the tool is already installed.
That will imply that additional things like MacOSWorkaround, etc. have to be done even if the tool is already installed wasting some performance. But fist we need to have things functionally correct and then we can think of how to optimize for performance.
After all such optimizations may save less than 1 second most probably and are therefore less important - much more important is the boostrapping performance we also waste for completion that massively slows down tab-completion that needs to be much faster whereas saving 1s on top of tool installation (check) is not very relevant.

### Affected version:

* 2024.09.002-beta-SNAPSHOT

PR #765, opened 2024-11-14
#637: Option to disable updates

KianRolf:
fixes https://github.com/devonfw/IDEasy/issues/637

Allows the user to call tools with --skipUpdates or -s, which skips updates for the tool if the configured version and the installed version match.


jan-vcapgemini (changes requested):
Thanks for adding an option to disable the updates. I've added some CRs, please resolve.

jan-vcapgemini (approved):
Thanks. Ready for review.

cli/src/main/java/com/devonfw/tools/ide/commandlet/ContextCommandlet.java
```diff
@@ -41,6 +43,7 @@ public ContextCommandlet() {
     this.debug = add(new FlagProperty("--debug", false, "-d"));
     this.quiet = add(new FlagProperty("--quiet", false, "-q"));
     this.offline = add(new FlagProperty("--offline", false, "-o"));
+    this.skipUpdates = add(new FlagProperty("--skip-updates", false, "-s"));
```
hohwille, line 46: I would not add the short option `-s` as alias because it is unintuitive. Who would guess that `-s` stands for `--skip-updates`? Also we need to be aware that we can have at max 26 short options (`-a` to `-z`) and consider this as a valuable reserved space for important and common options. Maybe we once introduce `--safe` or `--secure` or `--sudo` or whatever where we might want to use `-s` as an alias.
```suggestion
    this.skipUpdates = add(new FlagProperty("--skip-updates", false));
```

cli/src/test/java/com/devonfw/tools/ide/commandlet/HelpCommandletTest.java
```diff
@@ -145,13 +145,14 @@ public void testEnsureAllNlsPropertiesPresent(String locale) throws IOException
...
-        "-o | --offline  enable offline mode (skip updates or git pull, fail downloads or git clone).",
-        "-q | --quiet    disable info logging (only log success, warning or error).",
-        "-t | --trace    enable trace logging.");
+        "--locale             the locale (e.g. '--locale=de' for German language).",
+        "-b | --batch         enable batch mode (non-interactive).",
+        "-d | --debug         enable debug logging.",
+        "-f | --force         enable force mode.",
+        "-o | --offline       enable offline mode (skip updates or git pull, fail downloads or git clone).",
+        "-q | --quiet         disable info logging (only log success, warning or error).",
+        "-s | --skip-updates  disables tool updates if the configured versions match the installed versions.",
```
hohwille, line 154:
```suggestion
        "--skip-updates       disables tool updates if the configured versions match the installed versions.",
```

cli/src/test/java/com/devonfw/tools/ide/completion/CompleteTest.java
```diff
@@ -75,7 +75,7 @@ public void testCompleteShortOptsCombined() {
     // act
     List<CompletionCandidate> candidates = context.complete(args, true);
     // assert
-    assertThat(candidates.stream().map(CompletionCandidate::text)).containsExactly("-f", "-fb", "-fd", "-fo", "-fq",
+    assertThat(candidates.stream().map(CompletionCandidate::text)).containsExactly("-f", "-fb", "-fd", "-fo", "-fq", "-fs",
```
hohwille, line 78:
```suggestion
    assertThat(candidates.stream().map(CompletionCandidate::text)).containsExactly("-f", "-fb", "-fd", "-fo", "-fq",
```

cli/src/test/java/com/devonfw/tools/ide/completion/CompleteTest.java
```diff
@@ -85,11 +85,11 @@ public void testCompleteShortOptsCombinedAllButVersion() {
 
     // arrange
     AbstractIdeContext context = newContext(PROJECT_BASIC, null, false);
-    CliArguments args = CliArguments.ofCompletion("-fbdoqt");
+    CliArguments args = CliArguments.ofCompletion("-fbdoqts");
```
hohwille, line 88:
```suggestion
    CliArguments args = CliArguments.ofCompletion("-fbdoqt");
```

cli/src/test/java/com/devonfw/tools/ide/completion/CompleteTest.java
```diff
@@ -85,11 +85,11 @@ public void testCompleteShortOptsCombinedAllButVersion() {
...
     // act
     List<CompletionCandidate> candidates = context.complete(args, true);
     // assert
-    assertThat(candidates.stream().map(CompletionCandidate::text)).containsExactly("-fbdoqt", "-fbdoqtv");
+    assertThat(candidates.stream().map(CompletionCandidate::text)).containsExactly("-fbdoqts", "-fbdoqtsv");
```
hohwille, line 92:
```suggestion
    assertThat(candidates.stream().map(CompletionCandidate::text)).containsExactly("-fbdoqt", "-fbdoqtv");
```

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -78,7 +78,8 @@ public final boolean install(boolean silent, EnvironmentContext environmentConte
       // TODO https://github.com/devonfw/IDEasy/issues/664
       boolean enableOptimization = false;
       // performance: avoid calling installTool if already up-to-date
-      if (enableOptimization & configuredVersion.equals(installedVersion)) { // here we can add https://github.com/devonfw/IDEasy/issues/637
+      if (enableOptimization & configuredVersion.equals(installedVersion)
+          || configuredVersion.matches(installedVersion) & context.isSkipUpdatesMode()) {
```
hohwille, line 82: Please have a look at #664.
There is a bug triggered when we use this optimization to exit already here with `return toolAlreadyInstalled...`.
Sorry for the mess but IMHO the comment `// here we can add 637` was added before `// TODO 664` was added.
Can you test this with tool dependencies (e.g. with `tomcat` if tomcat update is available but installed version matches configured but also java was uninstalled or a matching version is not available)?

IMHO we should do the following:
* remove this entire block from `// TODO` until the end of the `if` block that before could never be entered.
* move your feature to the next `if` (line 89/90)

hohwille (approved):
@KianRolf thanks for your PR. You implemented this feature nicely but stepped into a pitfall I caused by accident with a comment. 👍 
I also suggested and explained not to add `-s` short option and by doing so, you can also revert changes you needed to apply to completion tests.
When the review comments are resolved, we can merge.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
KianRolf: I moved my feature accordingly and also did some tests with tool dependencies, coming to the conclusion that --skip-updates does not hinder (for example) the installation of a different java version because the tool dependencies requires a different version.

hohwille: Thanks for your rework and testing.
IMHO before your change it would not have worked since the skip would have returned before the tool dependencies had been processed.
After your change it now looks perfect and your test confirms that it works as expected. Great 👍

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -75,18 +75,13 @@ public final boolean install(boolean silent, EnvironmentContext environmentConte
...
-        return toolAlreadyInstalled(silent, installedVersion, step);
-      }
       // install configured version of our tool in the software repository if not already installed
       ToolInstallation installation = installTool(configuredVersion, environmentContext);
 
       // check if we already have this version installed (linked) locally in IDE_HOME/software
       VersionIdentifier resolvedVersion = installation.resolvedVersion();
-      if (resolvedVersion.equals(installedVersion) && !installation.newInstallation()) {
+      if (resolvedVersion.equals(installedVersion) && !installation.newInstallation()
+          || configuredVersion.matches(installedVersion) & context.isSkipUpdatesMode()) {
```
hohwille, line 84: I do not know why Java ever allowed this syntax but by convention `&&` or `||` should always be used for logical AND or OR while `&` and `|` should be used for bit-wise `AND` or `OR`. It turns out that for `boolean` values the compiler accepts both variants and produces the same result but IMHO it is confusing and raises the question if the developer did some magic trick by first using `&&` and later using `&` that the reader may not understand. Also even though not required due to operator priority, it IMHO improves readability to explicitly use parenthesis to express how this combination of expressions will be evaluated (reader could even think left-to-right what would be very wrong).
```suggestion
      if ((resolvedVersion.equals(installedVersion) && !installation.newInstallation())
          || (configuredVersion.matches(installedVersion) && context.isSkipUpdatesMode())) {
```

hohwille (approved):
@KianRolf thanks for your rework. All complete now. Great work!👍 
We can use the daily to discuss if we accept my final improvement suggestion and then merge.

Issue #734, opened 2024-11-01
Improve ProcessResult: get out and err in order


hohwille:
With #653 we now log the standard output and error messages of a process so the use can see it (e.g. in case of errors) even if it was captured.
However, this now opens a new problem:
We have modelled this as two independent `List`s:
https://github.com/devonfw/IDEasy/blob/d146098fa03789dcdf6e3b0f6a78b8f7cfd3320d/cli/src/main/java/com/devonfw/tools/ide/process/ProcessResultImpl.java#L17-L19

With this design it is impossible to print these messages in their original order how it was printed by the actual process.
Let us look at the following example:
```bash
#!/usr/bin/env bash

echo "out1"
echo "err1" >&2
sleep 5
echo "out2"
echo "err2" >&2
```

This program will let you see this on your console:
```
out1
err1
out2
err2
```

However, with IDEasy and DEFAULT_CAPTURE we would get:
```
out1
out2
err1
err2
```

The idea and expectation of this story is to change the model (e.g. having a single list of LogEvents) and the way we read the out and err from the thread to preserve the order (as far as possible).

NOTE: With Java Process API it is impossible to create a perfect solution as you can [read here](https://stackoverflow.com/questions/4959548/why-does-output-appear-in-wrong-order).

PR #773, opened 2024-11-19
#734: Improve Process Result Model

alfeilex:
Fixes: #734 

This PR improves the process result capturing of `stdout` and `stderr` output by preserving the order of the output.


cli/src/main/java/com/devonfw/tools/ide/process/LogEvent.java
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * Represent a log event.
+ *
+ * @param error A boolean flag that indicates whether the log event represents and error or standard output
+ * @param message A string containing the log message
+ */
+public record LogEvent(boolean error, String message) {
+
+}
```
alfeilex, line 11: Should it be located in the process or log package or elsewhere?

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -214,11 +214,19 @@ public ProcessResult run(ProcessMode processMode) {
...
     return CompletableFuture.supplyAsync(() -> {
 
       try (InputStreamReader isr = new InputStreamReader(is); BufferedReader br = new BufferedReader(isr)) {
+
+        String line;
+        while ((line = br.readLine()) != null) {
+          synchronized (logs) {
+            LogEvent logEvent = new LogEvent(errorStream, line);
+            logs.add(logEvent);
+          }
```
alfeilex, line 228: I am not sure, if I am using `LogEvent` correctly. In this implementation, a single `LogEvent` instance represents one line from either `stout` or `sterr`. I don't see another solution to capture the entire log while keeping the order of the log streams.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -167,14 +167,15 @@ public ProcessResult run(ProcessMode processMode) {
...
 
       Process process = this.processBuilder.start();
 
       if (processMode == ProcessMode.DEFAULT_CAPTURE) {
-        CompletableFuture<List<String>> outFut = readInputStream(process.getInputStream());
-        CompletableFuture<List<String>> errFut = readInputStream(process.getErrorStream());
+        CompletableFuture<List<String>> outFut = readInputStream(process.getInputStream(), false, logs);
+        CompletableFuture<List<String>> errFut = readInputStream(process.getErrorStream(), true, logs);
         out = outFut.get();
         err = errFut.get();
```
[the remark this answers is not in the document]
hohwille: Yes, that is the plan but we should keep the existing API methods `getOut()` and `getErr()` to avoid that all using code also needs to get refactored. But you should take a `List<OutputMessage> output` instead and make that available via a new getter. The old getters can be implemented with streams and lambdas.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
alfeilex: Now its `OutputMessage` instead of `LogEvent`. We can still discuss the name

cli/src/main/java/com/devonfw/tools/ide/process/ProcessResultImpl.java
```diff
@@ -23,23 +24,25 @@ public class ProcessResultImpl implements ProcessResult {
...
     super();
     this.executable = executable;
     this.command = command;
     this.exitCode = exitCode;
-    this.out = Objects.requireNonNullElse(out, Collections.emptyList());
-    this.err = Objects.requireNonNullElse(err, Collections.emptyList());
+    this.outputMessages = Objects.requireNonNullElse(outputMessages, Collections.emptyList());
+    this.out = this.outputMessages.stream().filter(outputMessage -> !outputMessage.error()).map(OutputMessage::message).collect(Collectors.toList());
+    this.err = this.outputMessages.stream().filter(OutputMessage::error).map(OutputMessage::message).collect(Collectors.toList());
   }
```
alfeilex, line 46: Should we fill the `out` and `err` lists in the constructor or elsewhere?

cli/src/main/java/com/devonfw/tools/ide/process/ProcessResultImpl.java
```diff
@@ -80,11 +90,8 @@ public void log(IdeLogLevel level, IdeContext context) {
...
   public void log(IdeLogLevel outLevel, IdeContext context, IdeLogLevel errorLevel) {
 
-    if (!this.out.isEmpty()) {
-      doLog(outLevel, this.out, context);
-    }
-    if (!this.err.isEmpty()) {
-      doLog(errorLevel, this.err, context);
+    if (!this.outputMessages.isEmpty()) {
+      doLog(outLevel, getOutputMessages().stream().map(OutputMessage::message).collect(Collectors.toList()), context);
     }
```
alfeilex, line 95: Are two separate log levels still necessary with a single lists of captures?

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -220,11 +219,19 @@ public ProcessResult run(ProcessMode processMode) {
...
-  private static CompletableFuture<List<String>> readInputStream(InputStream is) {
+  private static CompletableFuture<List<String>> readInputStream(InputStream is, boolean errorStream, List<OutputMessage> outputMessages) {
 
     return CompletableFuture.supplyAsync(() -> {
 
       try (InputStreamReader isr = new InputStreamReader(is); BufferedReader br = new BufferedReader(isr)) {
+
+        String line;
+        while ((line = br.readLine()) != null) {
+          synchronized (outputMessages) {
```
hohwille, line 230: This may create a DeadLock.
Consider using a concurrent collection.
I would suggest to use `ConcurrentLinkedQueue` for collecting the `OutputMessage`s and when completed, convert to a regular `List`.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessResultImpl.java
```diff
@@ -23,23 +24,25 @@ public class ProcessResultImpl implements ProcessResult {
...
 
     super();
     this.executable = executable;
     this.command = command;
     this.exitCode = exitCode;
-    this.out = Objects.requireNonNullElse(out, Collections.emptyList());
-    this.err = Objects.requireNonNullElse(err, Collections.emptyList());
+    this.outputMessages = Objects.requireNonNullElse(outputMessages, Collections.emptyList());
+    this.out = this.outputMessages.stream().filter(outputMessage -> !outputMessage.error()).map(OutputMessage::message).collect(Collectors.toList());
+    this.err = this.outputMessages.stream().filter(OutputMessage::error).map(OutputMessage::message).collect(Collectors.toList());
```
hohwille, line 45: Small simplification:
```suggestion
    this.out = this.outputMessages.stream().filter(outputMessage -> !outputMessage.error()).map(OutputMessage::message).toList();
    this.err = this.outputMessages.stream().filter(OutputMessage::error).map(OutputMessage::message).toList();
```

cli/src/main/java/com/devonfw/tools/ide/process/ProcessResultImpl.java
hohwille: Good point:
We could compute out and err lazily in the getters so if these getters are never called and we have a larger output, this may be pointless waste otherwise...

cli/src/main/java/com/devonfw/tools/ide/process/ProcessResultImpl.java
hohwille: There are special cases at least Eclipse for where we want to make err visible to the user but have out on debug that is hidden by default but can be enabled.
I would therefore recommend to:
1. first check if `outLevel == errorLevel` and then do as you suggest/implemented
2. otherwise simply use the old approach but then also first check if the log-level is actually enabled. So if it is disabled, we do nothing (not even calling `getOut()` or `getErr()` what would compute the extra `List`).

cli/src/test/java/com/devonfw/tools/ide/context/ProcessContextGitMock.java
```diff
@@ -61,19 +60,30 @@ public void setExitCode(int exitCode) {
   }
 
   /**
-   * @return the {@link List} of mocked error messages.
+   * @return the {@link List} of mocked stderr messages.
    */
   public List<String> getErrors() {
 
-    return errors;
+    return this.outputMessages.stream().filter(OutputMessage::error).map(OutputMessage::message).collect(Collectors.toList());
```
hohwille, line 67: This will not work. Current code is calling `getErrors()` or `getOuts()` in order to add mocked messages to such `List`.
Streams will create immutable collections that cannot be modified after creation.
In case we leave it like this, also use `toList()` simplification in all places.

Could you discuss with @jan-vcapgemini who IMHO created this class about the design?
Wouldn't it be smarter to create a regular `ProcessResult` here and make it accessible so instead of tests calling
`processContextGitMock.getErrors().add("fake error message")` we could call `processContextGitMock.getProcessResult().getOutputMessages().add(new OutputMessage("fake error message", true))`
to make it work.
There is a lot of pointless redundancy in this class related to `ProcessResultImpl` that can be avoided this way.

BTW: I just noticed that you already found this problem and had to update an existing test-case.

hohwille: If you want a convenience method for adding faked out or err messages, you could still provide them here but delegating to the ProcessResult. Also this way we could even allow a setter for the mocked `ProcessResult` that also gives us more flexibility (resetting the result for a next faked git call, etc.).

hohwille (commented):
@alfeilex thanks for your PR. Already on its way and looking good. Thanks 👍 
I left some review comments for improvement. Please have a look.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessResultImpl.java
alfeilex: done

cli/src/test/java/com/devonfw/tools/ide/context/ProcessContextGitMock.java
alfeilex: done

cli/src/main/java/com/devonfw/tools/ide/process/ProcessResultImpl.java
alfeilex: done

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
alfeilex: done

cli/src/test/java/com/devonfw/tools/ide/context/ProcessContextGitMock.java
```diff
@@ -181,7 +182,8 @@ public ProcessResult run(ProcessMode processMode) {
       }
     }
     this.arguments.clear();
-    return new ProcessResultImpl("git", command.toString(), this.exitCode, this.outs, this.errors);
+    setProcessResult(getExitCode(), this.processResult.getOutputMessages());
+    return this.processResult;
```
alfeilex, line 186: `setExitCode()` and `getExitCode()` are currently not used in the implementation. The exitcode is set differently in conditions with e.g. `this.exitcode = 0`

What is here the best solution? Is it the right way to use getters and setters inside a class, also for mock classes or can we just assign the attributes with values directly?

cli/src/test/java/com/devonfw/tools/ide/process/ProcessContextImplTest.java
```diff
@@ -221,6 +221,18 @@ public void enablingCaptureShouldRedirectAndCaptureStreamsWithErrorsCorrectly()
...
+    // arrange
+    IdeTestContext context = newContext(PROJECT_BASIC, null, false);
+
+    // act
+    context.newProcess().executable(TEST_RESOURCES.resolve("process-context").resolve("log-order.sh")).run();
+
+    // assert
+    assertThat(context).log(IdeLogLevel.INFO).hasEntries("out1", "err1", "out2", "err2");
+  }
+
```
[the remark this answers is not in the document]
alfeilex: If `synchronized` is to inefficent and we stick to `ConcurrentLinkedQueue` then we have to accept the compromise that the order is not 100% guaranteed. Also, the test-case doesn't work and testing the order like in the implemented test-case isn't possible and has to be removed before merging

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -218,13 +219,21 @@ public ProcessResult run(ProcessMode processMode) {
...
 
       try (InputStreamReader isr = new InputStreamReader(is); BufferedReader br = new BufferedReader(isr)) {
+
+        String line;
+        while ((line = br.readLine()) != null) {
+          OutputMessage outputMessage = new OutputMessage(errorStream, line);
+          outputMessages.add(outputMessage);
+        }
+
         return br.lines().toList();
```
hohwille, line 237: why do we try to read the lines again from the buffered reader after it has already been consumed to the end?
IMHO you should change the signature from `CompletableFuture<List<String>>` to `CompletableFuture<Void>` and simply return `null` here.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
hohwille: OMG. Sorry, if I caused a mess here.
Lets have a look at this in January after vacation time.
Maybe we can resolve the problem and make it work deterministic and reliable...
In the worst case, we have to revert the Concurrent-Collection approach and need to go back to your original solution using `synchronized` (or better using `ReentrantLock`).

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
alfeilex: done

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
alfeilex: It was the right decision to try out `ConcurrentLinkedQueue` because, by doingso,  I have tried out the `synchronized` and `ReentrantLock` approaches in multiple test runs again. Unforunately, I mentioned that there are occasionally outliers where the order is incorrect

So in the end, the problem still remains...

alfeilex:
Finally, I chose `ConcurrentLinkedQueue` since the runtime  is similar for all approaches and this approach is non-blocking which could be more efficient when handling with more lines of logs. Further, with the added sleep between each `echo` in the test file, also no more failing tests occur.

cli/src/test/java/com/devonfw/tools/ide/context/ProcessContextGitMock.java
hohwille: The file `exitcode` should be deleted together with its getters and setters. Actually it is now already in `processResult.getExitCode()`.

hohwille (approved):
@alfeilex thanks for bringing this PR to perfection. Great work! 👍 
If you address your own review comment that I also commented with a hint, we can finally merge this.

cli/src/test/java/com/devonfw/tools/ide/context/ProcessContextGitMock.java
hohwille: Maybe we do not even need `processResult` varaible (unless we want to be able to set a mocked `processResult` from the outside via setter that would later be returned by the next git operation.

alfeilex: > Maybe we do not even need `processResult` varaible (unless we want to be able to set a mocked `processResult` from the outside via setter that would later be returned by the next git operation.

Is this an example for a described case?
https://github.com/devonfw/IDEasy/blob/6f167c101b1181be14cc8ecb199a5791ba48c46a/cli/src/test/java/com/devonfw/tools/ide/git/GitContextTest.java#L91-L94

Finally, I removed the mock for the `ProcessResult` and I am using an attribute `outputMessages` to collect the messages from the tests and create a new `ProcessResult` with a local `exitCode` and the list of `outputMessages`.  I hope that these changes don't introduce any bugs. However, the tests are green.

cli/src/test/java/com/devonfw/tools/ide/context/ProcessContextGitMock.java
```diff
@@ -51,7 +48,7 @@ public ProcessResult getProcessResult() {
   }
 
   /**
-   * @param exitCode the {@link #getExitCode() exit code}.
+   * @param exitCode th exit code.
```
hohwille: accident?

cli/src/test/java/com/devonfw/tools/ide/context/ProcessContextGitMock.java
```diff
@@ -175,13 +141,13 @@ public ProcessResult run(ProcessMode processMode) {
...
         Files.writeString(newFile, this.now.toString());
-        this.exitCode = 0;
+        exitCode = ProcessResult.SUCCESS;
       } catch (IOException e) {
         throw new RuntimeException(e);
       }
     }
     this.arguments.clear();
-    return new ProcessResultImpl("git", command.toString(), this.exitCode, this.outs, this.errors);
+    return new ProcessResultImpl("git", "", exitCode, this.outputMessages);
```
hohwille, line 150: Why did you remove `command.toString()` and replace that with the empty String?

cli/src/test/java/com/devonfw/tools/ide/context/ProcessContextGitMock.java
hohwille: Was already fixed. Method is removed. Resolved.

cli/src/test/java/com/devonfw/tools/ide/context/ProcessContextGitMock.java
alfeilex: Thanks for pointing that out. It's a mistake. I have adjusted it

cli/src/test/java/com/devonfw/tools/ide/context/ProcessContextGitMock.java
hohwille: What is still fragile is that we have a state of messages (just as before) and there we collect messages by adding them.
Adding additional invocations will result in getting the previous messages back that can only grow over time unless the test-case logic explicitly takes control and does a `clear()` on the list.
I created #944 to finally being able to merge this PR. Since the tests are currently green the further cleanup is a nice-to-have that shall not block this PR.

hohwille (approved):
@alfeilex thanks - now great and ready for merge 👍

Issue #589, opened 2024-09-05
IDEasy not working on linux or Mac - Can't find bundle for base name nls.Help


hohwille:
### Expected behavior

As a linux user, I want to be able to use IDEasy on my machine so that I can make use of the product and its features.

### Actual behavior

IDEasy does not work on linux.
At least the help fails:
```
~/projects$ ide help
Variable IDE_ROOT is undefined. Please check your installation or run setup script again.
__       ___ ___  ___
\ \     |_ _|   \| __|__ _ ____ _
 > >     | || |) | _|/ _` (_-< || |
/_/ ___ |___|___/|___\__,_/__/\_, |
   |___|                       |__/

An unexpected error occurred!
We are sorry for the inconvenience.
Please check the error below, resolve it and try again.
If the error is not on your end (network connectivity, lack of permissions, etc.) please file a bug:
https://github.com/devonfw/ide/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=MissingResourceException%3A+Can%27t+find+bundle+for+base+name+nls.Help%2C+locale+en
java.util.MissingResourceException: Can't find bundle for base name nls.Help, locale en
        at java.base@21.0.2/java.util.ResourceBundle.throwMissingResourceException(ResourceBundle.java:2059)
        at java.base@21.0.2/java.util.ResourceBundle.getBundleImpl(ResourceBundle.java:1697)
        at java.base@21.0.2/java.util.ResourceBundle.getBundleImpl(ResourceBundle.java:1600)
        at java.base@21.0.2/java.util.ResourceBundle.getBundleImpl(ResourceBundle.java:1555)
        at java.base@21.0.2/java.util.ResourceBundle.getBundle(ResourceBundle.java:935)
        at com.devonfw.tools.ide.nls.NlsBundle.<init>(NlsBundle.java:55)
        at com.devonfw.tools.ide.nls.NlsBundle.<init>(NlsBundle.java:40)
        at com.devonfw.tools.ide.nls.NlsBundle.of(NlsBundle.java:131)
        at com.devonfw.tools.ide.commandlet.HelpCommandlet.run(HelpCommandlet.java:66)
        at com.devonfw.tools.ide.context.AbstractIdeContext.applyAndRun(AbstractIdeContext.java:867)
        at com.devonfw.tools.ide.context.AbstractIdeContext.run(AbstractIdeContext.java:779)
[...]

PR #781, opened 2024-11-20
#589: Fix NLS bundles for Linux and MacOS

alfeilex:
Fixes: #589 

I removed `<arg>-H:IncludeResources="nls/.*"</arg>` because it is an experimental feature and doesn't work on Linux and MacOS, see [851cdb3](https://github.com/devonfw/IDEasy/pull/781/commits/851cdb350bed6e349ef4b0500240f89842006084)

A working solution for Windows and Linux is to include a `resource-config.json` like [b5da1d4](https://github.com/devonfw/IDEasy/pull/781/commits/b5da1d42601fe6ae4a7e5934948b61145a2124ca)

Related docs:
https://www.graalvm.org/jdk21/reference-manual/native-image/metadata/#resource-metadata-in-json

This configuration also works for the latest GraalVM JDK Version (23) if we decide to upgrade our version in the workflow.

- [x] Tested on Ubuntu 20.04.01 LTS

- [ ] It still need to be tested on MacOS


hohwille (approved):
@alfeilex thank you for finding this solution and verifying that it fixes the error on Linux. Also thanks for providing the links and explanations. I am very confident that this will also fix the problem on MacOS. 🥇 
Therefore I propose to merge this now and test the next SNAPSHOT release on MacOS. I can do that on my MacBook.

Issue #764, opened 2024-11-14
IDEasy not working properly in CMD


hohwille:
### Expected behavior

As a IDEasy user on Windows, I want to be able to use IDEasy also in CMD so that I can use it from some `*.bat` scripts or keep my odd legacy habbits ;)

### Actual behavior

![grafik](https://github.com/user-attachments/assets/523e8adc-6adc-4492-917f-437d159ac130)

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Install IDEasy
2. Open CMD
3. `cd` to an IDEasy project
4. run e.g. `ide intellij`

### Affected version:

* OS: Windows
* IDEasy 2024.11.001-beta-SNAPSHOT

hohwille:
PR #784 does not fix this problem.
There is something generally broken, when we execute bash scripts in case we have been initially launched from CMD:
```
D:\projects\IDEasy\software\java\bin>ide install intellij
C:/Program Files/Git/mingw64/libexec/git-core\git-submodule: line 7: basename: command not found
C:/Program Files/Git/mingw64/libexec/git-core\git-submodule: line 7: sed: command not found
C:/Program Files/Git/mingw64/libexec/git-core\git-submodule: line 22: .: git-sh-setup: file not found
Running command 'C:\Program Files\Git\cmd\git.exe' with arguments '--no-pager' 'pull' '--quiet'
failed with exit code 1!
Git pull on branch master failed for repository D:\projects\_ide\urls.
Failed to update git repository at D:\projects\_ide\urls
See above error for details. If you have local changes, please stash or revert and retry.
Typically you should abort and fix the problem. Do you want to continue anyways?
Option 1: yes
Option 2: no
```

Workaround: Do not use CMD at all as suggested in the documentation.
However, we still need to fix this.

hohwille:
Seems to be closely related to https://github.com/devonfw/ide/issues/1395

alfeilex:
There are not problems using CMD to run `ide intellij` on Win 10. 

Tested with version: [ide-cli-2024.12.001-beta-20241210.023937-17-windows-x64.tar.gz](https://s01.oss.sonatype.org/content/repositories/snapshots/com/devonfw/tools/IDEasy/ide-cli/2024.12.001-beta-SNAPSHOT/ide-cli-2024.12.001-beta-20241210.023937-17-windows-x64.tar.gz)

alfeilex:
But if I run `ide create <project>`, I get this error: 

[...]

PR #784, opened 2024-11-21
#764: IDEasy not working properly in CMD

KianRolf:
Fixes: https://github.com/devonfw/IDEasy/issues/764

Changes the start script from bash to batch to enable ide intellij/tomcat/android-studio to be performed in the cmd


jan-vcapgemini:
@KianRolf, thanks for your adjustment to use batch files instead of bash files. Please add a check for OS e.g. linux, mac, as these still need to use the bash file. We need to enforce that only on a windows machine, these batch files get created/used.
Please check this: https://github.com/devonfw/IDEasy/blob/d706ab4b077ecca19d98872f6d97c16ce315d3c5/cli/src/test/java/com/devonfw/tools/ide/tool/dotnet/DotNetTest.java#L59 it might come in handy.

KianRolf:
The batch file now gets created additionally if the OS is windows. IDEasy automatically executes the correct file corresponding to the OS.

hohwille (approved):
@KianRolf thanks for your PR and your nice solution/workaround. This is great and I hope tests will confirm it fixes our bug 👍 
For the record I am not so sure if `start` vs. `call` should be used to run the binary from the `*.bat` script and also on Windows there are no executable permissions so `makeExecutable` will be a NO-OP on windows and could be omitted.
However, nothing wrong so I merge as is and we see how it will react.

hohwille:
needs to be reverted - see #814, etc.

Issue #810, opened 2024-11-25
setup not adding IDEasy to current shell


hohwille:
### Expected behavior

As a IDEasy user, I want to use IDEasy immediately after calling the setup script so that I can start using the product immediately and do not get confused by errors.

### Actual behavior

In case I open a shell and run `setup` script from there the first time and then call some `ide`  command, it will say
```
ide: command not found
```

Since I just ran the `setup`, I would expect this to work immediately.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. ensure IDEasy is not yet installed 
2. download and unpack IDEasy as described
3. open a shell (on Mac or Linux a bash) and `cd` to the `_ide` folder.
4. run `setup`
5. run `ide help`

### Related/Dependent Issues

See also #779 

### Comments/Hints:

The `setup` script is registering ourselves to `.bashrc` but does not apply the changes to the current existing shell.
I would expect that the `ide` command (alias or function) also gets registered to the current shell.

Workaround: type `bash` to run a new bash shell inside the existing one and then type `ide help` and it will work.

### Affected version:

* 2024.12.001-beta-SNAPSHOT

hohwille:
I tried to fix this by adding the alias also to the current shell.
However, there is a drawback that we cannot solve:
Most users will run the script like this:
```
./setup
```
This will create a sub-process from the current terminal that has its own environment variables, aliases, functions, etc. while the users shell where he invoked the setup is the parent process that cannot be influenced by the sub-process by design of Unix/Posix standards.
The only way to make it work is instead calling this:
```
. setup
```
or in other words:
```
source setup
```
As telling all our users they must run the setup this way is kind of pointless and fighting against windmills, I now have to admit that we cannot properly fix this by design.
Therefore I will instead simply add a message to the end of the setup script telling the user, that he needs to open a new shell in order to make it work.
I may document that the setup script should ideally be sourced but I will not rely on it.
Whatever a user can do wrong will happen sooner or later.
The message is the only proper way to avoid confusion and make it work.
On the long run we will have a guided installer with UI and the problem will be gone anyhow since then nobody will manually create `project` and `_ide` folder, extract the release and run a `setup` script.
But for this our users need to be patient: we have a small team developing IDEasy and it is complex to implement installers for all major OS, bring IDEasy to marketplaces/app stores, etc.

hohwille:
Was autoclosed by PR #811 - as explained above, I still need to add the according message.

Issue #782, opened 2024-11-20
Fix IDE_ROOT variable on Linux


alfeilex:
### Expected behavior

If I am setting up `IDEasy` on Ubuntu, then the variable `IDE_ROOT` should be set correctly, like on Windows `(e.g C:\projects)`

### Actual behavior

After setting up `IDEasy`, the `IDE_ROOT` variable is not set correctly. It always takes the value of the path where the terminal was started.

### Steps to reproduce (bug)

1.  Setup `ideasy` on Linux
2.  Open the terminal in any location 
3.  Run `echo $IDE_ROOT` and observe the value

### Affected version:

* OS: Ubuntu 24.04 LTS
* SNAPSHOT: 2024.11.002-beta

alfeilex:
This is the entry in my `~/.bashrc` file

`export IDE_ROOT="${PWD}"`

Issue #778, opened 2024-11-20
Add icd command


hohwille:
As a IDEasy power user, I want to have advanced support for switching directories, projects and environments so that I gain efficiency during development.

The idea is to introduce a new command `icd` additionally to `ide`.
This will be an IDEasy variant of the standard `cd` command but with the following features:
* whenever I changed a directory with `icd`, I do not have to type `ide` again to update my environment variables.
* when I call `icd` without arguments, I will go to `$IDE_HOME` (instead of `$HOME` when calling `cd`). In case I am not in an IDEasy project (and have not set IDE_HOME), I will go to `$IDE_ROOT`.
* when I call `icd -p «project»` I directly jump to that project (`cd $IDE_ROOT/«project» && ide`)
* when I call `icd -w [«workspace»]` I directly go to the specified `«workspace»`. The default is `main` if omitted. If I am not in an IDE project, this command will fail.
* when I call `icd 

Advanced features (might come in a later PR):
* when I call `icd -m [«module»]` I directly go to the specified `«module»` in my current project. That means if I am in a maven, gradle, npm, or yarn project, I directly go to the specified module. If I omit the module name I go to the current module I am in so if I am in `«module»/src/main/java/com/devonfw/tools/ide` I go to `«module»` (`cd ../../../../../../.. && ide`). When I call `icd -m /` I go to the top-level module (project root). This feature will be rather tricky - to support multiple build systems we need some flexibility. For `«module»` we should always support `artifactId` in case of a maven/gradle project but also the module directory name if unique. In order to get the entire list of modules (Map from `artifactId` as String to `Path`) quite some parsing might be required so maybe we need caching to better support this feature efficiently. The basic approach is to traverse directories upwards until the first `pom.xml`, `build.gradle[.kts]`, or `package.json` is found. This way `icd -m` and `icd -m /` can be implemented without parsing any of these files.
* I want to have autocompletion for `icd` so `icd -p TAB` will suggest me all available projects (and not `_ide`).

ATTENTION: IMHO it makes sense to first implement #779 before starting with this story since the refactoring will make it easier to add `icd` and otherwise both stories might cause conflicts if developed in parallel.

Implementation hint: I would suggest to implement `icd` in BASH.
[...]

Issue #774, opened 2024-11-19
HTTP proxy support not working properly


sven-binnig:
As a developer, I want create a new project.

I'm working with a windows-10-pc behind a proxy.
The command `ide --debug create test-project` ends with an error message
`could not clone [url] to [path] because you are offline`

But I'm not offline - I'm only behind a proxy.
Setting up the environment variables `http_proxy`and `https_proxy`doesn't help.
I short look into the code makes it obvious, that no proxy configuration is used to determine, if I'm online or offline.

As a result of this, I cannot setup a project-folder.

sven-binnig:
Additionally: please make sure to handle this:
`javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target`

sven-binnig:
2nd addition belongs to the `.gitconfig`
Setting up a project include clone the url for the installable tools.
This only works for me while temporaryly setting `http.sslVerify` to  `false`

sven-binnig:
After I have downloaded the zip archive of the code I tried to implement a dirty fix, that tries to take care about the existance of an proxy. I'm sure the this fix doesn't fit the needs of code-quality of this projekt. Beside this, I just want to share what I have done:

```
public boolean isOnline() {

    if (this.online == null) {
      // we currently assume we have only a CLI process that runs shortly
      // therefore we run this check only once to save resources when this method is called many times
      try {
        int timeout = 1000;
        //open a connection to github.com and try to retrieve data
        //getContent fails if there is no connection
        URL url = new URL("https://www.github.com");
        ProxyContext proxyContext = new ProxyContext(this);
        Proxy httpsProxy = proxyContext.getProxy(url.toExternalForm());
        URLConnection connection;
        if(proxyContext.equals(Proxy.NO_PROXY)){
          connection = url.openConnection();
        } else {
          // this is very evil - but I need to fix the 'PKIX'-Issue quickly 
          System.setProperty("javax.net.ssl.trustStore", "C:/Windows/win.ini");
          System.setProperty("javax.net.ssl.trustStoreType", "Windows-ROOT");
          connection = url.openConnection(httpsProxy);
        }
        connection.setConnectTimeout(timeout);
        connection.getContent();
        this.online = Boolean.TRUE;
      } catch (Exception ignored) {
        this.online = Boolean.FALSE;
      }
    }
```

hohwille:
[...]

PR #811, opened 2024-11-26
#778: add icd #779: use functions instead of alias #810: setup in current shell #782: fix IDE_ROOT on linux/mac

### LINKED ISSUE #779: Consider functions instead of alias
In IDEasy we currently define `ide` as an alias to sourcing a script.
For auto-completion we also hook in via a sourced bash script (`source $IDE_ROOT/_ide/completion`).
With #778 a new such command is to be added that also needs autocompletion.

An alternative could be to use bash functions instead:
* Instead of having an `ide` script in `$IDE_ROOT/_ide/bin` we can have `$IDE_ROOT/functions`.
* Instead of adding aliases, sourcing multiple scripts, etc. we would just add a `source $IDE_ROOT/functions` to `.bashrc` / `.zshrc`.
* In `functions` we can also check if we are in `zsh` and install autocompletion to simplify setup and uninstallation of IDEasy if all we need might be a single line of code in `.bashrc` / `.zshrc`.
* This `functions` "script" that is sourced can define bash functions such as `ide` and `icd` instead that we use `alias` for that.

IMHO this will simplify and increase flexibility for future extension (e.g. if we have more ideas like #778), since we only need to tweak `.bashrc` once and forever and newer releases of IDEasy can add new commands and completions all via `functions`.
The first step with this story should be an analysis and PoC testing if this plan works well or there are some drawbacks of this idea.
If all works as planned, we should implement it this way since it seems the superior design.

hohwille:
fixes #779 #810 #782

* #778: add icd command but not yet with auto-completion.
* #779: use functions instead of alias 
* #810: setup in current shell
* #782: fix IDE_ROOT on linux/mac
* #774: fix HTTP proxy support - see https://github.com/devonfw/IDEasy/blob/66b6d6bbdb5609995bda70133179e27cdd999099/documentation/proxy-support.adoc


hohwille:
I have meanwhile tested this on MacOS and it works.
@leonrohne27 testet this on Linux.
So as it works on all platforms with this new approach and both in `bash` and `zsh`, I will merge this.
For the record: This will also mean that CMD support will remain poor. We would have to reimplement everything redundantly in CMD as BAT files what is pure pain. There will be no support for `icd` in CMD (unless some external users contribute this as PR or our team run out of stories and gets bored what will not happen before 2027).

cli/src/main/package/setup
```diff
@@ -6,34 +6,27 @@ function doSetupInConfigFile() {
...
-    if ! grep -q "IDE_ROOT" "${cfg}"
+    if ! grep -q "export IDE_ROOT=" "${cfg}"
     then
-     echo -e 'export IDE_ROOT="${PWD}"' >> "${cfg}"
+     echo "export IDE_ROOT=\"${IDE_ROOT}\"" >> "${cfg}"
     fi
   fi
+  if ! grep -q 'source "$IDE_ROOT/_ide/functions"' "${cfg}"; then
+    echo 'source "$IDE_ROOT/_ide/functions"' >> "${cfg}"
+    echo "ide init" >> "${cfg}"
```
hohwille, line 18: has to be removed, already done in functions

Issue #759, opened 2024-11-13
Add UpgradeSettingsCommandlet for the upgrade of legacy devonfw-ide settings to IDEasy


hohwille:
As a IDEasy user comming from a devonfw-ide project, I want to have a command to upgrade my settings from devonfw-ide to IDEasy so that I can get rid of all legacy after all my team members have already switched to IDEasy.

An `UpgradeSettingsCommandlet` shall be added so I can run `ide upgrade-settings` to upgrade my settings:
- [x] all `devon.properties` are renamed to `ide.properties`. In case a corresponding `ide.properties` already exists, the config from the `devon.properties` is merged into the existing `ide.properties` where `ide.properties` is the new one and therefore has priority if different values for the same property are defined in both files, the one from `devon.properties` is ignored/lost. This not only applies to `settings` but also to `conf` and all `workspaces`.
- [x] We ensure all legacy variables in `ide.properties` are replaced with their new name (if both legacy and new variable name is defined, the legacy one is removed). 
- [x] Duplicated variables in the same `ide.properties` are removed (only the last occurrence is kept).
- [x] If variable `DEVON_IDE_CUSTOM_TOOLS` exists, it is converted into a `ide-custom-tools.json` file in `settings` and the variable is removed (see [here](https://github.com/devonfw/IDEasy/blob/main/documentation/software.adoc#custom-tools) and [legacy](https://github.com/devonfw/ide/blob/master/documentation/software.asciidoc#custom)).
- [x] if `settings/devon` folder exists, it is renamed to `settings/templates`
- [x] #807
- [x] if `settings/projects` folder exists, it is renamed to `settings/repositories`
- [x] we recursively traverse all files in `settings/workspace` and `settings/*/workspace` and replace all known `IdeVariables` from CURLY into ANGLED syntax and legacy variable names are replaced with their new names. So e.g. `${DEVON_IDE_HOME}` is replaced with `$[IDE_HOME]` or `${MAVEN_VERSION}` is replaced with `$[MVN_VERSION]`. We also need to replace `${SETTINGS_PATH}` with `$[IDE_HOME]/settings`.
- [x] if we hit an `xml` file that is not containing the XML namespace of our merger, we print a warning with the filepath saying that this file is most-likely outdated and print a link to a documentation page about this settings migration feature and linking to our XML merger docs in configurator and telling the project to sync their workspace config manually with the official `ide-settings` git repo on github provided by us as template for the settings.
[...]

PR #820, opened 2024-11-28
#759: upgrade settings commandlet

leonrohne27:
### Implements:

- all devon.properties are renamed to ide.properties. In case both files exists in the same directory, devon.properties gets merged into ide.properties. ide.properties has priority here.
- duplicated variables in the same ide.properties file are being removed
- If variable DEVON_IDE_CUSTOM_TOOLS exists, it is converted into a custom-tools.json file in settings
- if settings/devon folder exists, it is renamed to settings/templates, if settings/projects folder exists, it is renamed to settings/repositories
- recursively traverses all files in settings/workspace and settings/*/workspace and replacse all known IdeVariables from CURLY into ANGLED syntax and legacy variable names are replaced with their new names.
- for all .xml files that do not contain the XML namespace of our merger,  we print a warning with the filepath saying that this file is most-likely outdated and print a link to a documentation page about this settings migration feature
- documentation page with small instructions for those who migrate from devonfw-ide to IDEasy
- if a [replacement-patterns.properties](https://github.com/devonfw/ide-settings/blob/devonfw-ide/eclipse/workspace/replacement-patterns.properties) is found in a workspace folder a warning will be logged if the file is not empty and the file will be deleted.
- if settings/ide.properties does not contain IDE_VARIABLE_SYNTAX_LEGACY_SUPPORT_ENABLED=false it will be added.


cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,366 @@
...
+  private void checkForXMLNameSpace() {
+    this.context.info("Scanning XML files...");
+    Path settingsDirectory = context.getIdeHome().resolve("settings");
+    AtomicBoolean missingNamespaceFound = new AtomicBoolean(false);
+    try {
+      Files.walk(settingsDirectory)
+          .filter(path -> Files.isDirectory(path) && path.getFileName().toString().equals("workspace"))
+          .forEach(workspaceDir -> {
+            try {
+              Files.walk(workspaceDir)
```
hohwille, line 177: why are you nesting two file walkers? File worker already work recursively.
https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#walk-java.nio.file.Path-java.nio.file.FileVisitOption...-
https://medium.com/@AlexanderObregon/javas-files-walk-method-explained-570d8a67247d

IMHO I would iterate the toplevel `workspace` folder and then just go to every `*/workspace` (`workspace` folder inside any toplevel folder) rather than `**/workspace` (any `workspace` folder in the entire directory tree).

Maybe the code would get more readable if you do the following:
1. find all these workspace folders as discussed above
2. for each of them call a new private method with a speaking name that will find any XML file in there
3. for each XML file call a new private method that does the namespace check.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,366 @@
...
+    } catch (IOException e) {
+      this.context.error("Error walking through the 'settings' directory", e);
+    }
+  }
+
+  private void createCustomToolsJson(String variable) {
+    try {
+      JSONArray tabelObject = new JSONArray();
+      variable = variable.substring(variable.indexOf("(") + 1, variable.indexOf(")") - 1);
+      String url = variable.substring(variable.indexOf("https://"), variable.indexOf(" "));
```
hohwille, line 210: such code is not robust. The `indexOf` method may return `-1` and that would then cause `IndexOutOfBoundsException`.
A valid configuration may be `()` and I am quite sure this code will fail in such case.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,366 @@
...
+      } else {
+        this.context.success("Your XML files are up to date");
+      }
+
+    } catch (IOException e) {
+      this.context.error("Error walking through the 'settings' directory", e);
+    }
+  }
+
+  private void createCustomToolsJson(String variable) {
```
hohwille, line 206: `variable` is not a very good name here. IMHO this is the value of the `DEVON_IDE_CUSTOM_TOOLS` variable.
How about renaming to `devonIdeCustomToolsValue`?

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
hohwille: FYI: You can extract a `static` method that does the mapping from the String value to a Java object representing the JSON. Then this can be easily tested with many different variants.
https://github.com/devonfw/IDEasy/blob/83dd243202290c224b6a10295803cca2e2e7eb1b/cli/src/main/java/com/devonfw/tools/ide/repo/CustomToolRepositoryImpl.java#L169-L195

We can extract an issue for this and address this later but this is the way to good design and code quality.
We already have [CustomTool](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/repo/CustomTool.java). We need a class representing the entire JSON containing multiple `CustomTool`s and would offer methods to load from JSON and write to JSON.
Then we can use this here to simplify this code and make everything better testable (Separation of Concerns -> one test for reading from JSON and writing to JSON, one test to convert the legacy String to the Java object representing the JSON, etc.).

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,366 @@
...
+          this.context.success("Successfully merged and updated ide.properties: " + file_path);
+
+          Files.delete(file_path);
+
+        } catch (IOException e) {
+          throw new RuntimeException(e);
+        }
+      } else {
+        try {
+          String comment = "#********************************************************************************\n"
```
hohwille, line 331: If you need something like this, you should use [multi-line-strings](https://www.baeldung.com/java-multiline-string).

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,366 @@
...
+              + "#********************************************************************************\n"
+              + "\n"
+              + "# Uncomment the following line to use your shared maven repository.\n"
+              + "# This will save diskspace but comes with the risk that other projects could\n"
+              + "# accidentally (or due to an security attack) access and use artifacts from this project.\n"
+              + "# If you are working in a project that is highly sensitive never use this feature.\n"
+              + "#export M2_REPO=~/.m2/repository\n"
+              + "\n"
+              + "# In case you are sitting behind a proxy these JVM options may help:\n"
+              + "#export JAVA_OPTS=-Dhttp.proxyHost=myproxy.com -Dhttp.proxyPort=8080\n";
```
hohwille, line 347: we should not have something like this in our code. This is just template content.
Most probably users will already have this:
https://github.com/devonfw/ide-settings/blob/bc23bf79838bfdef796607f03a60b660ff6974ce/devon/conf/devon.properties#L14
https://github.com/devonfw/ide-settings/blob/ab94f201befe7b8a66a21996e521f1f80649fe21/templates/conf/ide.properties#L14

documentation/migration-from-devonfw.adoc
```diff
@@ -0,0 +1,16 @@
...
+Step 2: Get rid of all legacy by calling the following command:
+
+[source]
+----
+upgrade-settings
+----
+
+Step 3: After running the command, you might need to update your xml files for our merger.
+Please consider reading our documentation for that topic:
+https://github.com/devonfw/IDEasy/blob/main/documentation/configurator.adoc--
```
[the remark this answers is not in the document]
leonrohne27: I removed it, thanks

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,366 @@
...
+    return "upgrade-settings";
+  }
+
+  @Override
+  public void run() {
+    updateDevonProperties();
+    replaceLegacyVariablesAndBracketsInWorkspace();
+    checkIfLegacyFolderExists();
+    handleReplacementPatternsFiles();
+    checkForXMLNameSpace();
```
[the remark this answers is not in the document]
leonrohne27: Thank you !

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,366 @@
...
+  }
+
+  private void checkIfLegacyFolderExists() {
+    this.context.info("Scanning for legacy folders...");
+
+    Path devonFolder = context.getIdeHome().resolve("settings/devon");
```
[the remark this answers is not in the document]
leonrohne27: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,366 @@
...
+    List<Path> pathList = new ArrayList<>();
+    EnvironmentVariables devonPropertiesPath = context.getVariables();
+    while (devonPropertiesPath != null) {
+      if (devonPropertiesPath.getPropertiesFilePath() != null) {
+        pathList.add(devonPropertiesPath.getPropertiesFilePath().getParent().resolve("devon.properties"));
+      }
+      devonPropertiesPath = devonPropertiesPath.getParent();
+    }
+
+    for (Path file_path : pathList) {
```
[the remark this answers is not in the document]
leonrohne27: Thanks, I changed it

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,366 @@
...
+            } catch (IOException e) {
+              this.context.error("Error processing 'replacement-patterns.properties' file: " + file, e);
+            }
+          });
+    } catch (IOException e) {
+      this.context.error("Error scanning for 'replacement-patterns.properties' file", e);
+    }
+  }
+
+  private void checkForXMLNameSpace() {
```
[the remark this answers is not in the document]
leonrohne27: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,366 @@
...
+      }
+      JSONObject finalObject = new JSONObject();
+      finalObject.put("tools", tabelObject);
+      finalObject.put("url", url);
+
+      FileWriter writer = new FileWriter(context.getIdeHome().resolve("settings/custom-tools.json").toString());
+      writer.write(finalObject.toJSONString());
+
+      writer.flush();
+      writer.close();
```
hohwille, line 234: don't call  `close()` manually but use try-with-resource instead that is the only safe pattern to ensure you do not cause resource leaks.
https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc#closing-resources

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,366 @@
...
+            }
+            if (split.length == 2) {
+              devonProperties.put(split[0], split[1]);
+            }
+          }
+        }
+
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
```
hohwille, line 284: You are duplicating the code to read a `devon.properties` and a `ide.properties` file.
https://github.com/devonfw/IDEasy/blob/76124f73b5008293ca845d94c65850c648e2a0d1/cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java#L71-L86

We should use our existing infrastructure to do that.

Also your implementation contains some problems. E.g. `#` can also be used inside a value (e.g. to separate branch of a git-url) and I guess you are trying to ignore comments here. Using String "fiddling" with methods such as `contains` is also dangerous since it will also match in unexpected cases. Surely not a very relevant scenario but rather an odd edge-case but in case a user has `DISABLED_DEVON_IDE_CUSTOM_TOOLS=(...)` in his config, this will cause undesired effects (and I have seen users who were not aware that you can outcomment a line by adding `#` at the beginning but were simply adding `DISABLED_` as prefix). As you can see in the linked code we already have better infrastructure to parse a line and separate variable name and value. This should be reused.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,366 @@
...
+              + "\n"
+              + "# In case you are sitting behind a proxy these JVM options may help:\n"
+              + "#export JAVA_OPTS=-Dhttp.proxyHost=myproxy.com -Dhttp.proxyPort=8080\n";
+          Files.write(filePath, comment.getBytes());
+          for (Entry<Object, Object> set : devonProperties.entrySet()) {
+            if (set.getValue() instanceof String) {
+              Files.write(filePath, ("\n" + set.getKey().toString() + "=" + set.getValue().toString()).getBytes(), StandardOpenOption.APPEND);
+            }
+            if (set.getValue() instanceof String[]) {
+              String[] values = (String[]) set.getValue();
```
hohwille, line 354: https://www.baeldung.com/java-pattern-matching-instanceof

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,366 @@
...
+              Files.walk(workspaceDir)
+                  .filter(file -> Files.isRegularFile(file) && file.toString().endsWith(".xml"))
+                  .forEach(file -> {
+                    try {
+                      String content = Files.readString(file);
+
+                      if (!content.contains("xmlns:merge=\"https://github.com/devonfw/IDEasy/merge\"")) {
+                        this.context.warning("The XML file " + file + " does not contain the required 'xmlns:merge' attribute.");
+                        missingNamespaceFound.set(true);
+                      }
```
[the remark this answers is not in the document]
leonrohne27: I have changed it
https://github.com/devonfw/IDEasy/pull/820/commits/afaa4f7ce27c4f305a0286c00c3ef28cfa247d08

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
hohwille: After re-reading all the code of this method I understand this was an important blind spot.
You tried to reimplement a lot of things that have already been solved in our code but could not find this.
Therefore, you ran into all these challenges how to deal with comments, how to split into property name and value, etc.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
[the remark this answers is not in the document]
hohwille: Thanks but also read the link, please.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
hohwille: actually it seems to be the entire line from the properties file.
IMHO we could close this. If you consider my comment about reusing our infrastructure with `VariableLine` and all the things already there, this will already be obsolete.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
hohwille: Will also be entirely removed IMHO...

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -863,8 +863,12 @@ private IdeProgressBar getProgressbarForUnpacking(long sizeFile) {
   }
 
   @Override
-  public List<Path> listChildren(Path dir, Predicate<Path> filter) {
+  public List<Path> listChildren(Path dir, Predicate<Path> filter, boolean recursive) {
```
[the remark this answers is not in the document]
KianRolf: removed again

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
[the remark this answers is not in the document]
KianRolf: only calls the method once now

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
KianRolf: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
leonrohne27: I've tried to follow your instructions and implemented it:
https://github.com/devonfw/IDEasy/pull/820/commits/694a7b8f0fccfa8d2508ddbcf24ad8bd6c1341e5
Now it should be better readable.
I am open for further improvement suggestions if needed.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -863,8 +863,12 @@ private IdeProgressBar getProgressbarForUnpacking(long sizeFile) {
...
+    return listChildrenRecursive(dir, filter, recursive);
+  }
 
+  private List<Path> listChildrenRecursive(Path dir, Predicate<Path> filter, boolean recursive) {
```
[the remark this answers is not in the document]
hohwille: All changes in FileAccess[Impl] have been reverted. Thanks. Resolving this conversation.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,409 @@
...
+                    try {
+                      String content = Files.readString(file);
+                      String originalContent = content;
+
+                      for (Map.Entry<String, String> entry : legacyToNewMapping.entrySet()) {
+                        content = content.replace(entry.getKey(), entry.getValue());
+                      }
+
+                      content = content.replace("{", "[");
+                      content = content.replace("}", "]");
```
hohwille, line 124: this is wrong. It will destroy all JSON files.
We need to use `VariableSyntax`, like here:
https://github.com/devonfw/IDEasy/blob/886cea38611b7f5b2c097a7165c74f631c1d1509/cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java#L220-L226

cli/src/main/java/com/devonfw/tools/ide/repo/CustomToolRepositoryImpl.java
```diff
@@ -162,39 +162,46 @@ public static CustomToolRepository of(IdeContext context) {
...
+        CustomTool customTool = new CustomTool(name, VersionIdentifier.of(version), osAgnostic, archAgnostic, url,
+            checksum, context.getSystemInfo());
+        customTools.add(customTool);
       }
+      return new CustomToolsJson("tools", defaultUrl, customTools);
+    } catch (Exception e) {
+      throw new IllegalStateException("Failed to read JSON from " + customToolsJsonPath, e);
     }
-    return new CustomToolRepositoryImpl(context, tools);
   }
```
hohwille, line 205: This method and logic belong to `CustomToolsJson` class. Furhter, we can use Jackson for the mapping instead of writing the mapping code manually. We can add an internal method for post-processing that will set the `url` in all `CustomTool` instances to the parent one from `CustomToolsJson` in case it is `null`. This method can then be called both from the JSON parsing as well as from the String parsing method.

cli/src/test/resources/customtools/custom-tools.json
```diff
@@ -0,0 +1,18 @@
+{
```
hohwille, line 1: Please rename this file to `ide-custom-tools.json` to keep our own contract.

cli/src/test/java/com/devonfw/tools/ide/repo/CustomToolTest.java
```diff
@@ -64,4 +71,57 @@ public void testSpecific() {
...
 
+  @Test
+  public void testReadCustomToolsFromJson() {
+    // arrange
+    AbstractIdeTestContext context = new IdeSlf4jContext(Path.of(""));
+    SystemInfo systemInfo = SystemInfoMock.of("linux");
+    context.setSystemInfo(systemInfo);
+    Path testPath = Path.of("src/test/resources/customtools");
+    // act
+    CustomToolsJson customToolsJson = CustomToolRepositoryImpl.readCustomToolsFromJson(context, testPath.resolve("custom-tools.json"));
```
hohwille, line 82: Great that you added new test methods to increase coverage. 👍 
As I already commented on the main code please follow SoC and put the code into the classes responsible for what is done by the according method.
Then do the same for JUnits:
If you test a class from `src/main/java/my/package/MyClass.java` then the unit tests shall be in `src/test/java/my/package/MyClassTest.java`.
Hence you should create `CustomToolsJsonTest` instead of extending this test.

cli/src/test/java/com/devonfw/tools/ide/repo/CustomToolTest.java
```diff
@@ -64,4 +71,57 @@ public void testSpecific() {
...
+        "https://some-file-server.company.com/projects/my-project/jboss-eap/7.1.4.GA/jboss-eap-7.1.4.GA.tgz");
+    assertThat(customToolsJson.tools().get(1).getUrl()).isEqualTo(
+        "https://some-file-server.company.com/projects/my-project2/firefox/70.0.1/firefox-70.0.1-linux-x64.tgz");
+  }
+
+  @Test
+  public void testReadCustomToolsFromLegacyConfig() {
+    // arrange
+    IdeContext context = new IdeTestContext();
+    String legacyProperties = "DEVON_IDE_CUSTOM_TOOLS=(jboss-eap:7.1.4.GA:all:https://host.tld/projects/my-project firefox:70.0.1:all:https://host.tld/projects/my-project2)";
```
hohwille, line 95: I already left review comments to Kian and Leon that the context of that method is kind of odd and also we need to reuse our existing infrastructure to parse `ide.properties` and `devon.properties` files.
Then we can just get the value of a variable and only need to pass that value to this method instead of fiddling the value from the variable inside a class that is not responsible for that and code gets more redundant.
BTW: We also have existing code to parse variable array syntax into a regular Java `List`. This is currently in `VariableDefinitionStringList` e.g. used by `IdeVariables.IDE_TOOLS` to parse things like `IDE_TOOLS=(java mvn node npm)` (see https://github.com/devonfw/IDEasy/blob/main/documentation/variables.adoc).
We could even extract that parsing code to `VariableLine` for easier reuse here.

From the test perspective I would then only pass the value and not the entire variable declaration line.

cli/src/test/resources/ide-projects/upgrade-settings/_ide/urls/java/java/17.0.6/linux_x64.sha256
```diff
@@ -0,0 +1 @@
+c2de7dfbd9f8faaa21b4cdd8518f826dd558c9ab24a0616b3ed28437a674a97b
```
hohwille, line 1: since when do we add checksums for test projects? Are they actually used? I can not see how this could be used since the `java` installation in our tests is not yet downloaded as archive but only copied as directory tree.
If these `*.sha256` files are all not needed or used, please remove them.

hohwille (commented):
@jan-vcapgemini thanks for your nice improvements 👍

cli/src/test/java/com/devonfw/tools/ide/repo/CustomToolTest.java
jan-vcapgemini: Adjusted.

cli/src/test/java/com/devonfw/tools/ide/repo/CustomToolTest.java
jan-vcapgemini: Adjusted.

cli/src/test/resources/customtools/custom-tools.json
jan-vcapgemini: Adjusted.

cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
```diff
@@ -136,6 +136,12 @@ public interface IdeContext extends IdeStartContext {
   /** Legacy folder name used as compatibility fallback if {@link #FOLDER_TEMPLATES} does not exist. */
   String FOLDER_LEGACY_TEMPLATES = "devon";
 
+  /** The filename of the configuration file in the settings for this {@link CustomToolRepository}. */
+  String FILE_CUSTOM_TOOLS = "ide-custom-tools.json";
+
+  String DEVON_IDE_CUSTOM_TOOLS = "DEVON_IDE_CUSTOM_TOOLS";
```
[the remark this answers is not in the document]
jan-vcapgemini: Adjusted.

cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
```diff
@@ -136,6 +136,12 @@ public interface IdeContext extends IdeStartContext {
...
+  String IDE_CUSTOM_TOOLS = "IDE_CUSTOM_TOOLS";
```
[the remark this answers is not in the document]
jan-vcapgemini: Adjusted.

hohwille (approved):
from my PoV ready for merge now.

cli/src/test/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandletTest.java
```diff
@@ -0,0 +1,108 @@
...
+   * Ensure that all devon.properties are renamed to ide.properties and that the variables inside have been adjusted.
+   */
+  @Test
+  public void testDevonPropertiesUpgrade() {
+    // arrange
+    UpgradeSettingsCommandlet upgradeSettingsCommandlet = new UpgradeSettingsCommandlet(context);
+    // act
+    upgradeSettingsCommandlet.run();
+    // assert that files where renamed
+    // FIXME assertThat(UPGRADE_SETTINGS_PATH.resolve("home/ide.properties")).exists();
```
hohwille, line 31: This is a remaining open problem.
* devonfw-ide: `~/devon.properties` (see [here](https://github.com/devonfw/ide/blob/master/documentation/configuration.asciidoc))
* IDEasy: `~/.ide/ide.properties` (see [here](https://github.com/devonfw/IDEasy/blob/main/documentation/configuration.adoc))

Should we change the new location to `~/ide.properties`.
Currently we could still do such a breaking change in BETA.
It would simplify this migration and avoid pointless differences and edge-cases.
Otherwise we need to add some very special extra logic for this case.

hohwille:
> [ERROR] Failures: 
[ERROR]   UpgradeSettingsCommandletTest.testIfFolderAreRenamed:72 
Expecting path:
  target/test-projects/upgrade-settings/project/settings/templates/conf/mvn/settings.xml
to exist (symbolic links were followed).
[INFO] 
[ERROR] Tests run: 387, Failures: 1, Errors: 0, Skipped: 1

Not reproducible on my Windows machine and does not make any sense to me.
Maybe it is just too late and I need to sleep...

cli/src/test/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandletTest.java
jan-vcapgemini: I'd rename the file to `ideasy.properties` then. `ide.properties` sounds a bit like a file a user might create in his home folder manually. So yes, we should move the location to the root of the home directory.

hohwille: I would either rename all properties to `ideasy.properties` or keep all as `ide.properties`

hohwille:
>  fixed test

Thanks for helping out but you just added a new empty file in the location where it is asserted to make the test green.
The test is already green on my windows machine. The reason why it is failing on github needs to be clarified and understanding this will most probably reveal an important problem.
Just adding the expected file is more like removing the assertion.

The update legacy folders functionality is doing the following:
* settings/projects -> settings/repositories
* settings/devon -> settings/templates
* settings/templates/conf/.m2 -> settings/templates/conf/mvn

The third one is what I added the assertion for.
The implementation also seems correct.
So we need to focus on why this test is failing in github actions.

hohwille:
> After checking the `ide.properties` file after the update, I've found an outdated property which probably won't work. `INTELLIJ_EDITION_TYPE=U` this needs to be converted to `INTELLIJ_EDITION=ultimate`. We need to add a new issue for that.

I fixed this by implementing additional migration logic and extended the test accordingly. Please have a look.

hohwille:
For the failing test:
On Windows:
```
10:05:50.901 [main] - INFO  - c.d.tools.ide.log.IdeSubLoggerSlf4j - Updating legacy folders if present...
10:05:50.903 [main] - INFO  - c.d.tools.ide.log.IdeSubLoggerSlf4j - SUCCESS:Successfully renamed folder 'projects' to 'repositories' in D:\projects\IDEasy\workspaces\main\IDEasy\cli\target\test-projects\upgrade-settings\project\settings.
10:05:50.905 [main] - INFO  - c.d.tools.ide.log.IdeSubLoggerSlf4j - SUCCESS:Successfully renamed folder 'devon' to 'templates' in D:\projects\IDEasy\workspaces\main\IDEasy\cli\target\test-projects\upgrade-settings\project\settings.
10:05:50.905 [main] - INFO  - c.d.tools.ide.log.IdeSubLoggerSlf4j - SUCCESS:Successfully renamed folder '.m2' to 'mvn' in D:\projects\IDEasy\workspaces\main\IDEasy\cli\target\test-projects\upgrade-settings\project\settings\templates\conf.
0:05:51.201 [main] - DEBUG - c.d.tools.ide.log.IdeSubLoggerSlf4j - Removed variable name of 'INTELLIJ_EDITION_TYPE' in D:\projects\IDEasy\workspaces\main\IDEasy\cli\target\test-projects\upgrade-settings\project\conf\ide.properties
```
So we are getting a little closer.
Anyhow why is the `.m2` folder not found and renamed on linux. Will debug this in ubuntu once I have some time.
On Github Action:
```
08:48:38.595 [main] - INFO  - c.d.tools.ide.log.IdeSubLoggerSlf4j - Updating legacy folders if present...
08:48:38.595 [main] - INFO  - c.d.tools.ide.log.IdeSubLoggerSlf4j - SUCCESS:Successfully renamed folder 'projects' to 'repositories' in /home/runner/work/IDEasy/IDEasy/cli/target/test-projects/upgrade-settings/project/settings.
08:48:38.596 [main] - INFO  - c.d.tools.ide.log.IdeSubLoggerSlf4j - SUCCESS:Successfully renamed folder 'devon' to 'templates' in /home/runner/work/IDEasy/IDEasy/cli/target/test-projects/upgrade-settings/project/settings.
08:48:38.604 [main] - DEBUG - c.d.tools.ide.log.IdeSubLoggerSlf4j - Removed variable name of 'DEVON_IDE_CUSTOM_TOOLS' in /home/runner/work/IDEasy/IDEasy/cli/target/test-projects/upgrade-settings/project/workspaces/main/devon.properties
```

cli/src/test/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandletTest.java
jan-vcapgemini: > This is a remaining open problem.
> 
>     * devonfw-ide: `~/devon.properties` (see [here](https://github.com/devonfw/ide/blob/master/documentation/configuration.asciidoc))
> 
>     * IDEasy: `~/.ide/ide.properties` (see [here](https://github.com/devonfw/IDEasy/blob/main/documentation/configuration.adoc))
> 
> 
> Should we change the new location to `~/ide.properties`. Currently we could still do such a breaking change in BETA. It would simplify this migration and avoid pointless differences and edge-cases. Otherwise we need to add some very special extra logic for this case.

So we've decided to add the special extra logic, keeping the `~/.ide/ide.properties`.

hohwille: I fixed this. Therefore I had to move the logic that finds the properties (ide.properties and devon.properties) from `AbstractIdeContext` to `EnvironmentVariablesPropertiesFile`. Now we just pass the configuration directory and from there `EnvironmentVariablesPropertiesFile` locates the properties files. If the `type` is `USER` (home directory) then we can do the workaround to use the parent folder (of `.ide`) for the legacy properties.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesResolved.java
```diff
@@ -26,6 +26,11 @@ public EnvironmentVariablesType getType() {
     return EnvironmentVariablesType.RESOLVED;
   }
 
+  @Override
+  public void remove(String string) {
+
```
jan-vcapgemini, line 31: We might have to add an exception here as the whole implementation is missing.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesSystem.java
```diff
@@ -22,6 +22,11 @@ public EnvironmentVariablesType getType() {
     return EnvironmentVariablesType.SYSTEM;
   }
 
+  @Override
+  public void remove(String string) {
+
```
jan-vcapgemini, line 27: We might have to add an exception here as the whole implementation is missing.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  private void updateLegacyFolder(Path folder, String legacyName, String newName) {
+
+    Path legacyFolder = folder.resolve(legacyName);
+    Path newFolder = folder.resolve(newName);
+    if (Files.isDirectory(legacyFolder)) {
+      try {
+        if (!Files.exists(newFolder)) {
+          Files.move(legacyFolder, newFolder, StandardCopyOption.REPLACE_EXISTING);
```
jan-vcapgemini, line 64: Maybe check FileAccess and use this instead.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
```diff
@@ -27,14 +26,18 @@ public final class EnvironmentVariablesPropertiesFile extends EnvironmentVariabl
...
+
+  private final Path legacyPropertiesFilePath;
 
   private final Map<String, String> variables;
 
   private final Set<String> exportedVariables;
 
   private final Set<String> modifiedVariables;
 
+  private Boolean legacyConfiguration;
```
jan-vcapgemini, line 39: Why not primitive type?

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
```diff
@@ -46,29 +49,80 @@ public final class EnvironmentVariablesPropertiesFile extends EnvironmentVariabl
...
+        this.propertiesFilePath = propertiesFilePath;
+        assert (propertiesFilePath.getParent().equals(propertiesFolderPath));
+      }
+      Path legacyPropertiesFolderPath = propertiesFolderPath;
+      if (type == EnvironmentVariablesType.USER) {
+        // ~/devon.properties vs. ~/.ide/ide.properties
+        legacyPropertiesFolderPath = propertiesFolderPath.getParent();
+      }
+      this.legacyPropertiesFilePath = legacyPropertiesFolderPath.resolve(LEGACY_PROPERTIES);
+    }
```
jan-vcapgemini, line 87: Maybe move into another method to distinct its use case

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
```diff
@@ -184,9 +219,31 @@ public void save() {
...
       this.modifiedVariables.clear();
     } catch (IOException e) {
-      throw new IllegalStateException("Failed to save properties to " + newPropertiesFilePath, e);
+      throw new IllegalStateException("Failed to save properties to " + this.propertiesFilePath, e);
     }
-    this.propertiesFilePath = newPropertiesFilePath;
+    this.legacyConfiguration = Boolean.FALSE;
+  }
+
+  private List<VariableLine> loadVariableLines(Path file) {
```
jan-vcapgemini, line 227: Why do we need this and can't use the Properties library?

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -264,22 +264,33 @@ private boolean isJunction(Path path) {
   @Override
   public void backup(Path fileOrFolder) {
 
-    if (Files.isSymbolicLink(fileOrFolder) || isJunction(fileOrFolder)) {
+    if ((fileOrFolder != null) && (Files.isSymbolicLink(fileOrFolder) || isJunction(fileOrFolder))) {
       delete(fileOrFolder);
-    } else {
+    } else if ((fileOrFolder != null) && Files.exists(fileOrFolder)) {
       // fileOrFolder is a directory
```
jan-vcapgemini, line 270:
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
```diff
@@ -237,7 +241,19 @@ default void extract(Path archiveFile, Path targetDir, Consumer<Path> postExtrac
...
+  default List<Path> listChildren(Path dir, Predicate<Path> filter) {
+
+    return listChildrenMapped(dir, child -> (filter.test(child)) ? child : null);
+  }
+
+  /**
+   * @param dir the {@link Path} to the directory where to list the children.
+   * @param filter the filter {@link Function} used to {@link Function#apply(Object) filter and transform} children to include. If the {@link Function}
+   *     returns  {@code null}, the child will be filtered, otherwise the returned {@link Path} will be included in the resulting {@link List}.
+   * @return all children of the given {@link Path} returned by the given {@link Function}. Will be the empty list of the given {@link Path} is not an existing
```
jan-vcapgemini, line 253:
```suggestion
   * @return all children of the given {@link Path} returned by the given {@link Function}. Will be the empty list if the given {@link Path} is not an existing
```

cli/src/main/java/com/devonfw/tools/ide/merge/FileMerger.java
```diff
@@ -60,4 +64,76 @@ public final int merge(Path setup, Path update, EnvironmentVariables variables,
...
+    String content = Files.readString(workspaceFile);
+    String migratedContent = upgradeWorkspaceContent(content);
+    boolean modified = !migratedContent.equals(content);
+    if (modified) {
+      Files.writeString(workspaceFile, migratedContent);
+    }
+    return modified;
+  }
+
+  protected String upgradeWorkspaceContent(String content) {
```
jan-vcapgemini, line 109: Please add Javadoc.

cli/src/main/java/com/devonfw/tools/ide/merge/JsonMerger.java
```diff
@@ -264,4 +360,34 @@ private Status(boolean inverse, boolean addNewProperties) {
 
   }
 
+  public static class JsonPrettyPrinter extends DefaultPrettyPrinter {
```
jan-vcapgemini, line 363: Please add Javadoc.

cli/src/main/java/com/devonfw/tools/ide/merge/WorkspaceMerger.java
```diff
@@ -21,9 +21,12 @@ public interface WorkspaceMerger {
...
    * @param workspace the workspace {@link Path} where to get the changes from.
    * @param variables the {@link EnvironmentVariables} to {@link EnvironmentVariables#inverseResolve(String, Object) inverse resolve variables}.
-   * @param addNewProperties - {@code true} to also add new properties to the {@code updateFile}, {@code false} otherwise (to only update existing properties).
+   * @param addNewProperties - {@code true} to also add new properties to the {@code updateFile}, {@code false} otherwise (to only update existing
+   *     properties).
    * @param update the update {@link Path}
    */
   void inverseMerge(Path workspace, EnvironmentVariables variables, boolean addNewProperties, Path update);
 
+  void upgrade(Path workspace);
```
jan-vcapgemini, line 30: Please add Javadoc.

cli/src/main/java/com/devonfw/tools/ide/repo/CustomToolMetadata.java
```diff
@@ -0,0 +1,119 @@
...
+import com.devonfw.tools.ide.os.OperatingSystem;
+import com.devonfw.tools.ide.os.SystemArchitecture;
+import com.devonfw.tools.ide.url.model.file.UrlDownloadFileMetadata;
+import com.devonfw.tools.ide.version.VersionIdentifier;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+/**
+ * Representation of a {@link CustomToolMetadata} from a {@link CustomToolRepository}.
+ */
+public final class CustomToolMetadata implements UrlDownloadFileMetadata {
```
jan-vcapgemini, line 14: Couldn't this be changed to a record too?

cli/src/main/java/com/devonfw/tools/ide/repo/CustomToolsJsonMapper.java
```diff
@@ -0,0 +1,186 @@
...
+  private static final ObjectMapper MAPPER = JsonMapping.create();
+
+  /**
+   * @param customTools the {@link CustomToolsJson} to save.
+   * @param path the {@link Path} of the file where to save as JSON.
+   */
+  public static void saveJson(CustomToolsJson customTools, Path path) {
+
+    try (BufferedWriter writer = Files.newBufferedWriter(path)) {
+      MAPPER.writerWithDefaultPrettyPrinter().writeValue(writer, customTools);
```
jan-vcapgemini, line 32: Could we use the JsonPrettyPrinter here too?

jan-vcapgemini (changes requested):
LGTUS. Co-Reviewed by: @salimbouch; @
We've added some questions and CRs to resolve before the merge.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesResolved.java
jan-vcapgemini: Added default implementation to EnvironmentVariables.

jan-vcapgemini (commented):
Removed unnecessary files.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  @Override
+  public void run() {
+    updateLegacyFolders();
+    updateProperties();
+    updateWorkspaceTemplates();
+  }
+
+  void updateLegacyFolders() {
```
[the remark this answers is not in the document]
jan-vcapgemini: Adjusted.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
jan-vcapgemini: Adjusted.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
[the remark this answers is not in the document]
hohwille: This was not `private` so it could be referenced from the JUnit in the same package.
If you change visibility here, please also update the JUnit to avoid compiler warnings.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
hohwille: Always a good idea to encapsulate low-level file operations via `FileAccess`.
Will you do that or can we just create a new story for that?
We have lot of critical PRs on our table and need to move forward with IDEasy.
Nice to have improvements like this should not block our progress.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
hohwille: If you could do that please go ahead.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
hohwille: Are you talking about `java.util.Properties`?
This does not support `export` nor does it support comments so if you do `Properties.load(...)` and then `Properties.save(...)` all the comments are lost and the order of your file is random.
I started with `java.util.Properties` since I also do not like reinventing the wheel but it is unfortunately not helpful for our problem.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
hohwille: Thanks for finding this inconsistent code and cleaning up.
Howere, the missleading comment is still present.

cli/src/main/java/com/devonfw/tools/ide/repo/CustomToolMetadata.java
hohwille: Surely it could. But what is the value. We have to implement the interface `UrlDownloadFileMetadata` that defines getter methods.
So try it if you like but there is no reasonable benefit in doing so.

cli/src/main/java/com/devonfw/tools/ide/repo/CustomToolsJsonMapper.java
hohwille: Very good question.
It was introduced for `upgrade-settings`.
If we consider integrating it here (what totally makes sense), we should rather consider adding it directly to `JsonMapping` and also move the entire inner class `JsonPrettyPrinter` there as well.
But we have some JUnits that assert unformatted JSON on purpose that most likely need to be adjusted.
You are more than welcome to further improve directly on this feature-branch.
I spent last Friday till after 2 am in the night to get this PR working good enough.
So I just want this story to get completed and merged as we have some pressure and important blockers that needs to be fixed asap and releases to be shipped.
I am also a big fan of perfect and clean code.
But if we reach the finishing line for releasing and already start shifting release dates that end-users and projects rely on, we should go to either create extra issues with the nice-to-have improvements and resolve this later or we need constructive reviews that immediately address nice-to-haves.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesResolved.java
hohwille: BTW: Does it make sense to add a default implementation in the interface throwing an exception that says `Not yet implemented!`? It will IMHO never be implemented in `EnvironmentVariablesSystem` since it cannot work by design.

So if we want to cleanup here, remove it entirely from the interface and only add this method in `EnvironmentVariablesPropertiesFile`. That would be a reasonable design improvement to prevent users from calling this method in situations where it will not work.

jan-vcapgemini (approved):
Thanks for resolving the CRs. LGTM.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesResolved.java
jan-vcapgemini: Thanks for elaborating, I've adjusted it now.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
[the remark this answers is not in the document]
jan-vcapgemini: These methods were only referenced in the Javadocs of the test methods. Settings these private should be ok.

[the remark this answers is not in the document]
jan-vcapgemini: Adjusted the javadocs now.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
jan-vcapgemini: I've created a new issue for that: #929

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesPropertiesFile.java
jan-vcapgemini: Nothing to do here, resolved.

cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesResolved.java
hohwille: Thanks. Maybe for `set` we should do the same thing, but for `remove` it is now solved hence closing this thread as resolved.

Issue #830, opened 2024-12-02
ide -v is not displaying the correct version


jan-vcapgemini:
### Expected behavior

As an IDEasy user I expect the tool to return its proper version if I run `ide -v`

### Actual behavior
Instead of printing the complete version it just returns:
SNAPSHOT

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Download and install the following SNAPSHOT: https://s01.oss.sonatype.org/content/repositories/snapshots/com/devonfw/tools/IDEasy/ide-cli/2024.12.001-beta-SNAPSHOT/ide-cli-2024.12.001-beta-20241130.023556-10-windows-x64.tar.gz
2. run `ide -v`

### Affected version:

* OS: Windows
* IDEasy: ide-cli-2024.12.001-beta-20241130.023556-10-windows

PR #831, opened 2024-12-02
#830: fixed missing version

jan-vcapgemini:
Fixes: #830 

Implements:
* changed `native-maven-image` plugin phase from `compile` back to `package`


hohwille (approved):
@jan-vcapgemini thanks for your PR and rescuing us from all this bug. 👍

hohwille:
Since this error was not present in an official release, we do not need to document it explicitly in our CHANGELOG. Merging now.

Issue #832, opened 2024-12-02
Start script does not work on Windows for various tools


jan-vcapgemini:
### Expected behavior

As an IDEasy user I expect that if I run `ide intellij` that intellij starts properly.
Same issue applies for `android-studio` and `tomcat`.

### Actual behavior
When running `ide intellij`, the workspace setup works properly but intellij does not start.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Download and install this SNAPSHOT: [ide-cli-2024.12.001-beta-20241130.023556-10-windows](https://s01.oss.sonatype.org/content/repositories/snapshots/com/devonfw/tools/IDEasy/ide-cli/2024.12.001-beta-SNAPSHOT/ide-cli-2024.12.001-beta-20241130.023556-10-windows-x64.tar.gz)
2. run `ide intellij`

### Comments/Hints:

We need to re-introduce our OS `idea64.exe` etc. switch into `getBinaryName()`
We need to replace postExtract and the createStartScript part with getBinaryName

### Affected version:

* OS: Windows
* IDEasy: ide-cli-2024.12.001-beta-20241130.023556-10-windows

jan-vcapgemini:
It looks like this is preventing IDEasy from starting IDEs atm.
https://github.com/devonfw/IDEasy/blob/e849467f5321f044a0390b7f3b12e0ea3c07d519/cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java#L197-L199
When I start intellij while debugging and this finally block is not entered yet, intellij starts just fine.

PR #839, opened 2024-12-05
#832: fixed tools not starting on Windows

jan-vcapgemini:
Fixes: #832

Implements:
* replaced postExtract with getBinaryName for intellij, android-studio and tomcat
* added check for background processes to finally block of ProcessContext (prevents auto destroy of tool)


hohwille (approved):
@jan-vcapgemini thanks for your fix. Awesome work. 👍

Issue #840, opened 2024-12-05
git fetch does not work with multiple remotes


jan-vcapgemini:
### Expected behavior

As an IDEasy developer I expect that my git fetch of settings and urls works properly.

### Actual behavior
If I have multiple remotes for settings (need to adjust these as an admin). IDEasy fails to use git fetch.

```
$ ide intellij
Failed to determine current origin of git repository. - Expected single line of output but received 2 lines from C:\Program Files\Git\mingw64\bin\git.exe remote
An unexpected error occurred!
We are sorry for the inconvenience.
Please check the error below, resolve it and try again.
java.lang.IllegalStateException: Running command 'C:\Program Files\Git\mingw64\bin\git.exe' with arguments 'fetch' 'null' 'main' failed: NullPointerException
        at com.devonfw.tools.ide.process.ProcessContextImpl.run(ProcessContextImpl.java:210)
        at com.devonfw.tools.ide.git.GitContextImpl.runGitCommand(GitContextImpl.java:290)
        at com.devonfw.tools.ide.git.GitContextImpl.runGitCommand(GitContextImpl.java:282)
        at com.devonfw.tools.ide.git.GitContextImpl.fetch(GitContextImpl.java:168)
        at com.devonfw.tools.ide.git.GitOperation$1.execute(GitOperation.java:22)
        at com.devonfw.tools.ide.git.GitOperation.executeIfNeeded(GitOperation.java:125)
        at com.devonfw.tools.ide.git.GitContextImpl.fetchIfNeeded(GitContextImpl.java:46)
        at com.devonfw.tools.ide.git.GitContextImpl.fetchIfNeeded(GitContextImpl.java:40)
        at com.devonfw.tools.ide.context.AbstractIdeContext.applyAndRun(AbstractIdeContext.java:886)
        at com.devonfw.tools.ide.context.AbstractIdeContext.run(AbstractIdeContext.java:798)
        at com.devonfw.tools.ide.cli.Ideasy.runOrThrow(Ideasy.java:89)
        at com.devonfw.tools.ide.cli.Ideasy.run(Ideasy.java:52)
        at com.devonfw.tools.ide.cli.Ideasy.main(Ideasy.java:29)
        at java.base@22.0.2/java.lang.invoke.LambdaForm$DMH/sa346b79c.invokeStaticInit(LambdaForm$DMH)
Caused by: java.lang.NullPointerException
[...]

PR #841, opened 2024-12-05
#840: fixed fetch determine remotes

jan-vcapgemini:
Fixes: #840 

Implements:
* added fallback to origin if no remotes were found


hohwille (approved):
@jan-vcapgemini thanks for the fast fix of this error 👍 

I am still convinced that `determineRemote` method does not work and should either be removed or fixed. However, lets approve this minimal change since it will definitely fix our current problem.

PR #849, opened 2024-12-05
#791: unique SNAPSHOT versions

### LINKED ISSUE #791: unique SNAPSHOT versions
in our `nightly-build.yaml` workflow, we should determine the current SNAPSHOT version and tweak it. My suggestion would be to replace `SNAPSHOT` with the current timestamp (e.g. `date +%m_%d_%H`) and the suffix `-SNAPSHOT` so `2024.12.001-beta-SNAPSHOT` would become `2024.12.001-beta-12_21_12-SNAPSHOT (for the time of writing this). We do not need to modify any files but could pass this via `-Drevision=${tweakedVersion}`to the maven build command what should override the`revision`from`maven.config`.

hohwille:
fixes #791

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -23,8 +23,12 @@ jobs:
       - name: Build native image
         shell: bash
         run: |
+          maven_config="$(cat .mvn/maven.config)"
+          current_version="${maven_config/#*-Drevision=}"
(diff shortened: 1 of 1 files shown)


hohwille:
Works:
```
$ ide -v
2024.12.001-beta-12_05_21-SNAPSHOT/
```
but I added a slash due to a bash expression syntax issue.

PR #850, opened 2024-12-05
#757: Settings in code repository

### LINKED ISSUE #757: Support to allow settings in code repository
As a IDEasy admin, I want to have the ability to keep my settings together with the code in the same git repo so that I can change IDE setttings on feature branches (e.g. java or maven updates together with according code-changes in the same Pull-Request).

# Warning

This will not be the default. We designed the settings repo separated from the code repo for good reasons:

* An (IDEasy) project may have multiple code git repos (for many microservices) but want to have one IDE setup for the entire team
* On the other hand code git repos can grow big and it may take a long time to clone them (I have supported monolithic projects with multi million LOC where cloning alone takes more than 10 minutes). Getting an IDE change onto that git repo and testing it will be pure pain.
* We have the feature to configure code git repos in the settings-git repo so that you can get one or multiple code git repos cloned into various workspaces as needed by the project. This would all not make any sense if the settings are inside the code repo.
* So to summarize you currently have full flexibility: You can have 1 IDEasy project with 1 code repo, 1 IDEasy project with N code repos, N IDEasy projects with M code repos, etc. all in the same global development initiative.

# Design

In order to support this new feature request, we should add an option to `ide create` that will tell that we are providing a git URL that is actually the code repo (e.g. `ide create myproject --code https://git.company.com/domain/project.git[#branch]`).
This will first create a `$IDE_ROOT/myproject/workspaces/main/` and clone the code git repo into that folder.
Next IDEasy would verify that inside the cloned project there is a `settings` folder.
It will create a symlink `$IDE_HOME/settings` pointing to that `settings` folder.
All existing features to update our settings need to be modified since there will be no `.git` folder in `$IDE_ROOT/settings` and `git fetch` or `git pull` will not work there. Therefore `ide update` should skip the (alternatively the entire code git repo could be pulled, but I would leave that up to the end-user).
Further, we should track the git commit id (hash) of our settings.
Our feature to check for updates in `settings` should compare the current git commit id (determined from the root code git repo if `settings` is a symlink) with the tracked one. If the commit id does not match there are changes and the user should be informed that he may want to call `ide update`. This is also a helpful improvement if the regular and recommended settings git repo approach is used. If the end user manually calls `git pull` in `settings`, then IDEasy is not updated unless `ide update` is called but the user is not notified that updates are available (compare story #161). At the end of `ide update` process the git revision hash needs to be updated in the tracking file (e.g. a simple text file containing the revision let say in `$IDE_ROOT/.settings.id`).
On the other hand, if the git commit id is unchanged and nothing could be pulled from settings, we might know that the user is quite up-to-date (except for version patterns where new updates are available now).

# Acceptance criteria

- [ ] IDEasy still works with a regular settings git repository
- [ ] after `ide create` or `ide update` the `$IDE_ROOT/.settings.id` is updated with the current commit id of the git repository hosting our `settings` (may be from a symlinked code-repo)
- [ ] if the current git commit id behind the settings repo differes from the tracked one (see above) the user gets the existing notification that changes are available and `ide update` should be called.
- [ ] https://github.com/devonfw/IDEasy/blob/main/documentation/settings.adoc has been updated with the description of this feature
- [ ] ide help create shows additional information about the feature and new options that have been added
- [ ] If I create a a git repo with a `settings` folder containing the current content of the `ide-settings` repo and some additional files (e.g. a `README.adoc` and something like `src/test.txt` and I call `ide create test-code-settings --code https://«url-to-that-created-repo»` then I can create an IDEasy project `test-code-settings` just like it would have been created using `ide create test-code-settings -`. However, in `$IDE_ROOT/test-code-settings/workspaces/main/` I can find my repo cloned and `$IDE_ROOT/test-code-settings/settings` is a link pointing to the `settings` folder inside that cloned git repo inside `main` workspace.
- [ ] In case I later push an update to that git repo (e.g. change of `MVN_VERSION` in `settings/ide.properties`) and I pull these changes in the clode in `main` workspace and then 1. if I call a command like `ide help`, I can see that updates of settings are available and when I call `ide update` the new `MVN_VERSION` is applied and maven gets updated accordingly.
- [ ] the message `Updates are available for the settings repository. If you want to pull the latest changes, call ide update.` has been changed to `Updates are available for the settings repository. If you want to apply the latest changes, call "ide update".`

# Discussion

We can even think about auto-detecting this and omitting the `--code` option.
If we clone our settings to `$IDE_ROOT/settings` and detect that our expected settings-structure is not directly inside but in a `settings` subfolder, we could then move the `settings` folder to the `main` workspace and and rename it to the name of the project that was contained in the git URL, create the symlink, etc.
Maybe this will prevent errors since the average user cannot make mistakes.

### LINKED ISSUE #826: git settings check can break every commandlet
```
$ ide -d -v
Running commandlet VersionCommandlet[--version]
Configuration directory C:\Users\hohwille\.ide does not exist.
Configuration directory D:\projects\misc\conf does not exist.
Running command 'C:\Program Files\Git\cmd\git.exe' with arguments 'rev-parse' 'HEAD' ...
Running command 'C:\Program Files\Git\cmd\git.exe' with arguments 'rev-parse' 'HEAD'
failed with exit code 128!
fatal: not a git repository (or any of the parent directories): .git
Failed to get the local commit hash.
Missing .git folder in D:\projects\misc\settings.
Running command 'C:\Program Files\Git\cmd\git.exe' with arguments 'branch' '--show-current' ...
Running command 'C:\Program Files\Git\cmd\git.exe' with arguments 'branch' '--show-current'
failed with exit code 128!
fatal: not a git repository (or any of the parent directories): .git
Failed to determine current branch of git repository D:\projects\misc\settings
Running command 'C:\Program Files\Git\cmd\git.exe' with arguments 'remote' ...
Running command 'C:\Program Files\Git\cmd\git.exe' with arguments 'remote'
failed with exit code 128!
fatal: not a git repository (or any of the parent directories): .git
Failed to determine current origin of git repository D:\projects\misc\settings
Running command 'C:\Program Files\Git\cmd\git.exe' with arguments 'fetch' 'null' 'null' ...
Step 'ide' ended with failure.
An unexpected error occurred!
https://github.com/devonfw/IDEasy/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=IllegalStateException%3A+Running+command+%27C%3A%5CProgram+Files%5CGit%5Ccmd%5Cgit.exe%27+with+arguments+%27fetch%27+%27null%27+%27null%27+failed%3A+NullPointerException
java.lang.IllegalStateException: Running command 'C:\Program Files\Git\cmd\git.exe' with arguments 'fetch' 'null' 'null' failed: NullPointerException
...
Caused by: java.lang.NullPointerException
        at java.base@21.0.2/java.lang.ProcessBuilder.start(ProcessBuilder.java:1108)
        at com.devonfw.tools.ide.process.ProcessContextImpl.run(ProcessContextImpl.java:173)
```

To reproduce this bug, create a new project folder with an empty `settings` folder and `workspaces` folder.
Fix is easy: check if `.git` folder exists before calling such git commands.

salimbouch:
fixes #757 
fixes #826


cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -115,6 +116,13 @@ private void updateSettings() {
...
       if (Files.isDirectory(settingsPath) && !this.context.getFileAccess().isEmptyDir(settingsPath)) {
         step = this.context.newStep("Pull settings repository");
+        if (!Files.isDirectory(settingsPath.resolve(GitContext.GIT_FOLDER))) {
+          try {
+            settingsPath = settingsPath.toRealPath();
+          } catch (IOException e) {
+            throw new IllegalStateException("Could not resolve settings folder to real path.", e);
+          }
+        }
         gitContext.pull(settingsPath);
```
hohwille, line 126: If you want to pull here, you need to also traverse the real path up to where the `.git` directory is.
So `settingsPath.getParent()` would be required after `toRealPath()` (or you do a loop upwards until you have found a `.git` folder).

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -28,6 +36,8 @@ public CreateCommandlet(IdeContext context) {
     super(context);
     this.newProject = add(new StringProperty("", true, "project"));
     this.skipRepositories = add(new FlagProperty("--skip-repositories"));
+    this.codeRepositoryFlag = add(new FlagProperty("--code"));
+    this.codeRepository = add(new StringProperty("", false, "codeRepository"));
```
hohwille, line 40: We do not need an extra `codeRepository` since then the user can provide two URLs what will be rather confusing.

cli/src/main/java/com/devonfw/tools/ide/git/GitContext.java
```diff
@@ -58,13 +58,20 @@ public interface GitContext {
    * Checks if there are updates available for the Git repository in the specified target folder by comparing the local commit hash with the remote commit
    * hash.
    *
-   * @param repository the {@link Path} to the target folder where the git repository is located. This should be the folder containing the ".git"
-   *     subfolder.
+   * @param repository the {@link Path} to the target folder where the git repository is located.
```
hohwille, line 61: From a design perspective, `GitContext` is a generic helper for git.
This story is first of all something specific for what we want to do with our `settings`.
This method may also be used for other git repos than `settings` and then overriding and reusing the same `.commit.id` file will cause trouble.
Either you need to expose the lower level methods to get the local commit ID and the remote commit ID and do the saving and comparing of the `.commit.id` outside. Besides the fact that an update is applied from our settings and the fact that the remote and the local commit ID differs are two very different things.
Therefore I would recommend not to integrate this here.

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
```diff
@@ -51,12 +52,38 @@ public boolean fetchIfNeeded(Path repository, String remote, String branch) {
...
+
+  @Override
+  public void saveCurrentCommitId(Path repository) {
+
+    if (!Files.exists(repository.resolve(GitContext.GIT_FOLDER))) {
+      try {
+        repository = repository.toRealPath();
+      } catch (IOException e) {
+        throw new IllegalStateException("Couldn't resolve settings to real path.", e);
+      }
```
hohwille, line 78: you have duplicated this code fragment in multiple places. I have made some suggestions to improve this.
This already indicates that this should go to a central place and should be reused.
IMHO this feature is something special for our `settings` and I would recommend to keep the logic out of `GitContext`.
You could add a `getSettingsGitRepository()` method to `IdeContext` that currently would just delegate to `getSettingsPath()` but could then be extended by you to do this extra logic to find the real git repo with the `.git` folder.
Also you can then detect if no `.git` folder was found at all to resolve #826.
IMHO you should then log an error but return `null`.
In the use-case of #826 we just wanted to check for an update of the settings. If that is not possible, we could still continue regular commands. IMHO commands like `--version` or `--help` should not trigger online network activity after all and especially not fail because of that.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -58,13 +68,40 @@ public void run() {
...
     this.context.success("Successfully created new project '{}'.", newProjectName);
+
+  }
+
+  private void initializeCodeRepository(String repoUrl) {
+
+    // clone the given repository into IDE_HOME/workspaces/main
+    GitUrl gitUrl = GitUrl.of(repoUrl);
+    Path codeRepoPath = this.context.getWorkspacePath().resolve(gitUrl.getProjectName());
+    this.context.getGitContext().pullOrClone(gitUrl, codeRepoPath);
```
hohwille, line 95: Just as a hint. We have currently implemented `GitContext.pull` such that we provide the exact directory into which to clone the project since we must specify the name. Therefore we explicitly provide the folder to clone into as `.` and call the git command inside that directory. You can also omit that last argument and let git do the job to determine the project name from the git URL. However, then you also need to find the sub-directory that has been created by git afterwards in order to find the settings to link. Since the workspace was just created and should be empty, this may be very easy but consider this just as a comment to consider or ignore. You approach to explicitly implement `getProjectName()` should also be fine as long as we do not hit edge-cases that we are not aware yet where git would do a proper job but our code might get the project name wrong...

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -58,13 +68,40 @@ public void run() {
...
     this.context.setIdeHome(newProjectPath);
+    if (codeRepositoryFlag.isTrue()) {
+      String repoUrl = codeRepository.getValue();
+      if (repoUrl.isEmpty()) {
+        this.context.info("Please provide a repository URL after using --code");
+      } else {
+        initializeCodeRepository(repoUrl);
+      }
+    }
     super.run();
```
hohwille, line 79: this will still call `updateSettings()`. Is your intention to initialize the `settings` before so it will afterwards `update` the repo that has just been cloned? IMHO this is still confusing since milliseconds after cloning we do not expect changes and updating again does not make much sense.
It would IMHO make more sense to override `updateSettings()` method or to extract a new protected method out of `updateSettings()` that does the cloning part and potentially asks for settings URL. Then you can override only this method here.

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
salimbouch: I found something pretty interesting. When the settings repository is a symlink and I execute a git command in that repository, then the git command works. Even if the repository is just a folder inside a real repository, git commands work there too. With this I think we dont need to differentiate between the two cases (symlink or not), we can directly update the settings repository.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
salimbouch: done, overrode updateSettings()

cli/src/main/java/com/devonfw/tools/ide/git/GitContext.java
salimbouch: I now made the comparison and saving of commit Id more generic and not specific to our settings repository by just providing these as args to these methods.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
salimbouch: is there an example where `.` is used?

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -132,7 +136,10 @@ private void updateSettings() {
         }
         gitContext.pullOrClone(GitUrl.of(repository), settingsPath);
       }
+      gitContext.saveCurrentCommitId(settingsPath);
       step.success("Successfully updated settings repository.");
+    } catch (IOException e) {
+      throw new IllegalStateException("Could not update the settings repository.");
```
hohwille: if I am not mistaken, this exception can only come from `toRealPath()`.
It it better to do such exception handling closer to the actual operation.
That is why we abstract such things typically in `FileAccess` for reuse in multiple places.
Further, you did the famous evil bug pattern:
https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc#catching-and-handling-exceptions

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
hohwille: Please push your changes. They did not yet arrive on github so it is not yet done from my PoV.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
hohwille: Did you forget to push your changes?

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
hohwille: As I stated in `GitContext.pull`:
https://github.com/devonfw/IDEasy/blob/11de3ca8cc9b4956b3900ac9b3a7d1d0ab8c078b/cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java#L135

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
salimbouch: Not really, its really weird that you are not seeing the latest changes

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
hohwille: Was a refresh problem. Sorry for confusion. Thanks for the great rework. Resolved.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
salimbouch: Im trying to clone that repository, how would I achieve this exactly? The java-doc of `pullOrClone` says that in this given path git will "extract" the repo and not create a sub-folder and clone it there. See [here](https://github.com/devonfw/IDEasy/blob/11de3ca8cc9b4956b3900ac9b3a7d1d0ab8c078b/cli/src/main/java/com/devonfw/tools/ide/git/GitContext.java#L80)

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
hohwille: Was a refresh problem. Sorry for confusion. Thanks for the great rework. Resolved.

cli/src/main/java/com/devonfw/tools/ide/git/GitContext.java
hohwille: Nice solution. Thanks.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
hohwille: We would need to enhance `GitContext` so we have a new variant of the clone method taking a flag saying that we want to clone by letting git decide how to name and create a subfolder with the cloned repo.
However, this does not work well with the design of `cloneOrPull[IfNeeded]`.
Since this was just a nice to have idea, I would simply suggest that we leave this out of this PR.
It may come later as a separate improvement or maybe we will never do that at all.
You current implementation should work fine.
When users run into edge-cases that will not work as desired, they can raise a new ticket and we can react.
Lets not over-engineer in advance but wait if that is ever required.

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
salimbouch: forgot to mark this as done too

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
salimbouch: was due to the refresh problem, can be resolved

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -79,4 +105,15 @@ private void updateRepositories() {
 
     this.context.getCommandletManager().getCommandlet(RepositoryCommandlet.class).run();
   }
+
+  @Override
+  protected void updateSettings() {
+
+    if (codeRepositoryFlag.isTrue() && !settingsRepo.getValue().isBlank()) {
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -79,4 +106,26 @@ private void updateRepositories() {
...
+
+    if (codeRepositoryFlag.isTrue()) {
+      String codeRepository;
+      if (!Objects.isNull(settingsRepo.getValue())) {
+        codeRepository = settingsRepo.getValue();
+      } else {
+        String message = """
+          No code repository was given after '--code'.
+          Please give the code repository below that includes your settings folder.
+          Further details can be found here: https://github.com/devonfw/IDEasy/blob/main/documentation/settings.asciidoc
```
hohwille, line 121:
```suggestion
          Further details can be found here: https://github.com/devonfw/IDEasy/blob/main/documentation/settings.adoc
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -79,4 +106,26 @@ private void updateRepositories() {
 
     this.context.getCommandletManager().getCommandlet(RepositoryCommandlet.class).run();
   }
+
+  @Override
+  protected void updateSettings() {
+
+    if (codeRepositoryFlag.isTrue()) {
+      String codeRepository;
+      if (!Objects.isNull(settingsRepo.getValue())) {
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -79,4 +106,26 @@ private void updateRepositories() {
...
+        codeRepository = settingsRepo.getValue();
+      } else {
+        String message = """
+          No code repository was given after '--code'.
+          Please give the code repository below that includes your settings folder.
+          Further details can be found here: https://github.com/devonfw/IDEasy/blob/main/documentation/settings.asciidoc
+          """;
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
```diff
@@ -262,7 +278,7 @@ private void runGitCommand(Path directory, List<String> args) {
     runGitCommand(directory, args.toArray(String[]::new));
   }
 
-  private String runGitCommandAndGetSingleOutput(String warningOnError, Path directory, String... args) {
+  public String runGitCommandAndGetSingleOutput(String warningOnError, Path directory, String... args) {
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -1123,4 +1154,21 @@ public IdeStartContextImpl getStartContext() {
...
+    }
+    String currentCommitId = getGitContext().runGitCommandAndGetSingleOutput("Failed to get current commit id.", repository, "rev-parse", "HEAD");
+    if (currentCommitId != null) {
+      try {
+        Files.writeString(trackedCommitIdPath, currentCommitId);
+      } catch (IOException e) {
+        throw new IllegalStateException("Failed to save commit ID", e);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
salimbouch: done

cli/src/main/resources/nls/Help.properties
```diff
@@ -114,6 +114,7 @@ cmd.vscode=Tool commandlet for Visual Studio Code (IDE).
 cmd.vscode.detail=Visual Studio Code (VS Code) is a popular code editor developed by Microsoft. Detailed documentation can be found at https://code.visualstudio.com/docs/
 commandlets=Available commandlets:
 opt.--batch=enable batch mode (non-interactive).
+opt.--code=use git repository as both code and settings repository
```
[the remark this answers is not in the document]
salimbouch: done

hohwille:
As the status of the last review feedback and if/when we will address it was unclear to be, I created story #931.
Therefore this PR is now ready for merge. We can implement #931 as further improvement later.

Issue #851, opened 2024-12-06
Replace idea script with native binary


jan-vcapgemini:
As an IDEasy intellij and android-studio user I constantly get the message that I need to use the idea binary instead of the idea.sh script.
We need to change the constants used in getBinaryName.

Issue #808, opened 2024-11-25
devonfw-ide regression: cannot launch intellij


hohwille:
### Expected behavior

As a devonfw-ide user, I want to be able to use IDEasy in my existing project so that I can test if it works better.

### Actual behavior

```
$ ide intellij
Both legacy variable MAVEN_VERSION and official variable MVN_VERSION are configured in D:\projects\example-project\settings\devon.properties - ignoring legacy variable declaration!
Start: Configuring workspace main for IDE intellij
Successfully ended step 'Configuring workspace main for IDE intellij'.
Step 'ide (intellij)' failed: java.lang.IllegalStateException: Running command 'intellij' with arguments 'C:\Program Files\Git\bin\bash.exe' '-c' '/d/projects/example-project/workspaces/main & disown' failed: Cannot run program "intellij": CreateProcess error=2, Das System kann die angegebene Datei nicht finden
1 step(s) failed out of 4 steps.
An unexpected error occurred!
We are sorry for the inconvenience.
Please check the error below, resolve it and try again.
If the error is not on your end (network connectivity, lack of permissions, etc.) please file a bug:
https://github.com/devonfw/IDEasy/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=IllegalStateException%3A+Running+command+%27intellij%27+with+arguments+%27C%3A%5CProgram+Files%5CGit%5Cbin%5Cbash.exe%27+%27-c%27+%27%2Fd%2Fprojects%2Fexample-project%2Fworkspaces%2Fmain+%26+disown%27+failed%3A+Cannot+run+program+%22intellij%22%3A+CreateProcess+error%3D2%2C+Das+System+kann+die+angegebene+Datei+nicht+finden
java.lang.IllegalStateException: Running command 'intellij' with arguments 'C:\Program Files\Git\bin\bash.exe' '-c' '/d/projects/example-project/workspaces/main & disown' failed: Cannot run program "intellij": CreateProcess error=2, Das System kann die angegebene Datei nicht finden
        at com.devonfw.tools.ide.process.ProcessContextImpl.run(ProcessContextImpl.java:204)
        at com.devonfw.tools.ide.tool.ToolCommandlet.runTool(ToolCommandlet.java:169)
        at com.devonfw.tools.ide.tool.ToolCommandlet.runTool(ToolCommandlet.java:152)
        at com.devonfw.tools.ide.tool.ide.IdeToolCommandlet.runTool(IdeToolCommandlet.java:55)
[...]

PR #852, opened 2024-12-06
#851: replaced script with binary

jan-vcapgemini:
Fixes: #851

Implements:
* replaced idea.sh with idea
* replaced studio.sh with studio

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/androidstudio/AndroidStudio.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/androidstudio/AndroidStudio.java
@@ -18,8 +18,6 @@ public class AndroidStudio extends IdeaBasedIdeToolCommandlet {
 
   private static final String STUDIO64_EXE = STUDIO + "64.exe";
 
-  private static final String STUDIO_BASH_SCRIPT = STUDIO + ".sh";
-
   /**
    * The constructor.
    *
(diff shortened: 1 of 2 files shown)


salimbouch:
LGTM, you have a build error tho

cli/src/main/java/com/devonfw/tools/ide/tool/androidstudio/AndroidStudio.java
```diff
@@ -35,10 +33,8 @@ protected String getBinaryName() {
 
     if (this.context.getSystemInfo().isWindows()) {
       return STUDIO64_EXE;
-    } else if (this.context.getSystemInfo().isMac()) {
-      return STUDIO;
     } else {
-      return STUDIO_BASH_SCRIPT;
+      return STUDIO;
```
hohwille, line 37: This is a new feature of Intellij and older releases do not contain an `idea` script but only `idea.sh`.
So we should IMHO check if `idea.sh` exists but `idea` does not exist and make our choice dependent on that:
* `idea` script exists --> use `idea`
* `idea` script does not exist but `idea.sh` exists --> use `idea.sh`
* none of these scripts exists --> use `idea` or fail. I would suggest not to fail in case `getBinaryName()` is called for whatever reason before or without the tool being installed.

Same for android-studio.

hohwille (approved):
@jan-vcapgemini thanks for the PR and fixing these problems. Great work 👍 
I am quite confident that we now have found a good approach. This will also fix #808

Only if Jetbrains ships many more releases and every time changes things here and we reach a state where our code needs to consider the INTELLIJ_VERSION to determine the binary name then we should have stayed with the old approach.
In the end Jetbrains is already going in the direction we prefer only that the tool is called `intellij` but they use the internal name `idea` and that is not the binary name.
Assuming on windows there would also be a `idea.exe` instead of `idea64.exe` we would not need such quirks.

cli/src/main/java/com/devonfw/tools/ide/tool/androidstudio/AndroidStudio.java
hohwille: Thanks 👍

PR #880, opened 2024-12-09
#879: allow `-` as value

### LINKED ISSUE #879: cannot omit default settings URL in ide create
### Expected behavior

As a devonfw-ide user, I want to also be able to create a project with default settings by just passing `-` as settings URL parameter so that I can save typing and reuse my habits without learning new CLI syntax for IDEasy.

In devonfw-ide I did:
```
$ devon ide create testproject -
```
So I expect the same to work with IDEasy:
```
$ ide create testproject -
```

### Actual behavior
```
$ ide create testproject -
 
No matching property found
Invalid arguments: "create" "testproject" "-"
```

1. `ide create testproject -`

### Related/Dependent Issues

#101

### Comments/Hints:

We have implemented this feature as specified:
https://github.com/devonfw/IDEasy/blob/b8c670541760c28e75dcff69413f1069dbc6e445/cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java#L130

However, we never tested it and it is not working.
The reason is that `-` is considered as an option that cannot be passed to a value.
As a workaround the user would need to call `ide create testproject -- -` what somehow makes sense but is not very intuitive.

### Affected version:

* 2024.12.001-beta-12_06_02-SNAPSHOT

hohwille:
fixes #879
minimum fix to allow `-` as value.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -23,6 +23,7 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/799[#799]: binaries from zip download lack executable flags
 * https://github.com/devonfw/IDEasy/issues/81[#81]: Implement Toolcommandlet for Kubernetes
 * https://github.com/devonfw/IDEasy/issues/737[#737]: Adds cd command to ide shell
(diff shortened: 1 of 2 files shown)


jan-vcapgemini (approved):
LGTM.

Issue #827, opened 2024-11-29
ide status fails outside IDE_HOME


hohwille:
```
$ ide -t status

Error: IDEasy failed with exit code 2

hohwille@CE49454 MSYS /d/projects
$ ideasy -t status

hohwille@CE49454 MSYS /d/projects
$ echo $?
2
```

See #758

PR #885, opened 2024-12-10
#827: Made status commandlet available outside of IDE_HOME

WorkingAmeise:
Closes: #827 
- Made the status commandlet not require IDE_HOME or IDE_ROOT 
- Wrote a test for given szenario

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
@@ -332,17 +332,8 @@ public class ProcessContextImpl implements ProcessContext {
   private void performLogging(ProcessResult result, int exitCode, String interpreter) {
 
     if (!result.isSuccessful() && (this.errorHandling != ProcessErrorHandling.NONE)) {
-      IdeLogLevel ideLogLevel;
+      IdeLogLevel ideLogLevel = this.errorHandling.getLogLevel();
       String message = createCommandMessage(interpreter, "\nfailed with exit code " + exitCode + "!");
-
-      if (this.errorHandling == ProcessErrorHandling.LOG_ERROR) {
-        ideLogLevel = IdeLogLevel.ERROR;
-      } else if (this.errorHandling == ProcessErrorHandling.LOG_WARNING) {
-        ideLogLevel = IdeLogLevel.WARNING;
-      } else {
-        ideLogLevel = IdeLogLevel.ERROR;
-        this.context.error("Internal error: Undefined error handling {}", this.errorHandling);
-      }
 
       context.level(ideLogLevel).log(message);
       result.log(ideLogLevel, context);
--- a/cli/src/main/java/com/devonfw/tools/ide/process/ProcessErrorHandling.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/process/ProcessErrorHandling.java
@@ -1,5 +1,7 @@
 package com.devonfw.tools.ide.process;
 
+import com.devonfw.tools.ide.log.IdeLogLevel;
+
 /**
  * {@link Enum} with the available handling if a {@link Process#exitValue() status code} was not {@link ProcessResult#SUCCESS successful}.
  */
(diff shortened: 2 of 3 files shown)


cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -841,6 +842,7 @@ private ValidationResult applyAndRun(CliArguments arguments, Commandlet cmd) {
       debug("Running commandlet {}", cmd);
       if (cmd.isIdeHomeRequired() && (this.ideHome == null)) {
         throw new CliException(getMessageIdeHomeNotFound(), ProcessResult.NO_IDE_HOME);
+      } else if (cmd.isIdeRootRequired() && (this.ideRoot == null)) {
```
hohwille, line 845: Did you duplicate this entry by accident? Or maybe commited this change by accident but had it as a workaround before overriding `isIdeRootRequired()`?
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -171,6 +171,7 @@ public AbstractIdeContext(IdeStartContextImpl startContext, Path workingDirector
       if (nameCount >= 1) {
         name1 = currentDir.getName(nameCount - 1).toString();
       }
+      //TODO: make sure, that this does not check above IDE_ROOT
```
hohwille, line 174: I do not get this comment. What should we do here? Who is going to address this TODO? you?
```suggestion
```

cli/src/test/java/com/devonfw/tools/ide/commandlet/StatusCommandletTest.java
```diff
@@ -0,0 +1,30 @@
...
+    //arrange
+    IdeTestContext context = new IdeTestContext();
+    CliArguments args = new CliArguments("status");
+    args.next();
+
+    //act
+    context.run(args);
+
+    //assert
+    assertThat(context).logAtWarning().hasMessageContaining("You are not inside an IDE installation: C:");
```
hohwille, line 28: This test may work on your windows machine but it may already fail on mine if I am on `D:` and it will fail on every Linux or Mac that does not have a concept of drive letters.

cli/src/main/java/com/devonfw/tools/ide/commandlet/StatusCommandlet.java
```diff
@@ -80,4 +80,16 @@ private void logOnlineStatus() {
       this.context.warning("You are offline. Check your internet connection and potential proxy settings.");
     }
   }
+
+  @Override
+  public boolean isIdeHomeRequired() {
```
hohwille, line 85: Nothing wrong but if not even IDE_ROOT is required, then it does not make to require IDE_HOME.
Therefore you could omit overriding also this method. See here:
https://github.com/devonfw/IDEasy/blob/95e7d15357559c2d84334511cf1d5badcbf453dc/cli/src/main/java/com/devonfw/tools/ide/commandlet/Commandlet.java#L183-L186

hohwille (approved):
@WorkingAmeise thanks for fixing the problem and also adding a test. 👍 
I left some review comments. Please address them so we can merge.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
WorkingAmeise: When i ran ide status in trace mode, while being in C:\Projects i got these lines : 
```
Looking for IDE_HOME in C:\Projects
Looking for IDE_HOME in C:\
```
While IDE_ROOT was set to C:\Projects. Jan-V noticed, that this could be unwanted behavior, since the user maybe does not want IDEasy to read outside of its "territory" . But since the code snippet was to complex for me to implement the change right away i added this TODO.

hohwille: Correct but
1. it is complicated since IDE_ROOT is IMHO set after this code was processed
2. if we want to address this it is better to create an issue than adding a TODO to the code.

cli/src/main/java/com/devonfw/tools/ide/commandlet/StatusCommandlet.java
hohwille: Maybe I was not clear enough. The code I referenced is from the `Commandlet` class that you are extending here so the entire method could be removed. It seems you miss-understood this and change the implementation.

jan-vcapgemini (approved):
LGTM. Added a small CR.

PR #886, opened 2024-12-10
#764: Fix IDEasy for CMD by adding Git related entries into PATH

alfeilex:
Fixes: #764 

Tested using:
- Windows 10
- CMD, PowerShell
- Git-Version: 2.47.1.windows.1
- `ide create <project>`, `ide -f intellij`, `ide intellij`

### DIFF
--- a/cli/src/main/package/bin/ide.bat
+++ b/cli/src/main/package/bin/ide.bat
@@ -11,14 +11,12 @@ for %%H in ( HKEY_LOCAL_MACHINE HKEY_CURRENT_USER ) do for /F "usebackq tokens=2
 set "GIT_BIN=%GIT_HOME%\usr\bin"
 set "GIT_CORE=%GIT_HOME%\mingw64\libexec\git-core"
 
-if exist "%GIT_HOME%\bin\bash.exe" (
+if exist "%GIT_BIN%" (
 	echo "%PATH%" | find /i "%GIT_BIN%">nul  || set PATH=%PATH%;%GIT_BIN%
 )
 
-if exist "%GIT_HOME%\bin\bash.exe" (
+if exist "%GIT_CORE%" (
 	echo "%PATH%" | find /i "%GIT_CORE%">nul  || set PATH=%PATH%;%GIT_CORE%
-	REM KEEP the whitespace after call to clear ERRORLEVEL back to 0 otherwise it's 1 when PATH entries not found in find
-	call
 )
 
 if not "%1%" == "" (
@@ -34,6 +32,6 @@ for /f "tokens=*" %%i in ('ideasy %IDE_OPTIONS% env') do (
   call set %%i
 )
 
-if %ERRORLEVEL% == 0 (
+if not %ERRORLEVEL% == 0 (
   echo IDE environment variables have been set for %IDE_HOME% in workspace %WORKSPACE%
 )
--- a/cli/src/main/package/bin/ide.bat
+++ b/cli/src/main/package/bin/ide.bat
@@ -8,8 +8,17 @@ Set _RESET=[0m
 rem Add Git PATH entries for CMD - https://github.com/devonfw/IDEasy/issues/764
 for %%H in ( HKEY_LOCAL_MACHINE HKEY_CURRENT_USER ) do for /F "usebackq tokens=2*" %%O in (`call "%SystemRoot%"\system32\reg.exe query "%%H\Software\GitForWindows" /v "InstallPath" 2^>nul ^| "%SystemRoot%\system32\findstr.exe" REG_SZ`) do set GIT_HOME=%%P
 
+set "GIT_BIN=%GIT_HOME%\usr\bin"
+set "GIT_CORE=%GIT_HOME%\mingw64\libexec\git-core"
+
 if exist "%GIT_HOME%\bin\bash.exe" (
-  set "PATH=%PATH%;%GIT_HOME%\usr\bin\;%GIT_HOME%\mingw64\libexec\git-core"
+	echo "%PATH%" | find /i "%GIT_BIN%">nul  || set PATH=%PATH%;%GIT_BIN%
+)
+
+if exist "%GIT_HOME%\bin\bash.exe" (
+	echo "%PATH%" | find /i "%GIT_CORE%">nul  || set PATH=%PATH%;%GIT_CORE%
+	REM clear ERRORLEVEL back to 0 otherwise it's 1 when PATH entries not found
+	call
 )
 
 if not "%1%" == "" (
--- a/cli/src/main/package/bin/ide.bat
+++ b/cli/src/main/package/bin/ide.bat
@@ -5,6 +5,8 @@ Set _fBGreen=[92m
 Set _fBRed=[91m
 Set _RESET=[0m
 
+set PATH=%PATH%;C:\Program Files\Git\usr\bin;C:\Program Files\Git\mingw64\libexec\git-core
+
 if not "%1%" == "" (
   ideasy %IDE_OPTIONS% %*
   if not %ERRORLEVEL% == 0 (
--- a/cli/src/main/package/bin/ide.bat
+++ b/cli/src/main/package/bin/ide.bat
@@ -33,6 +33,6 @@ REM https://stackoverflow.com/questions/61888625/what-is-f-in-the-for-loop-comma
 for /f "tokens=*" %%i in ('ideasy %IDE_OPTIONS% env') do (
   call set %%i
 )
-if not %ERRORLEVEL% == 0 (
+if %ERRORLEVEL% == 0 (
   echo IDE environment variables have been set for %IDE_HOME% in workspace %WORKSPACE%
 )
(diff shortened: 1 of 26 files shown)


cli/src/main/package/bin/ide.bat
```diff
@@ -17,6 +24,6 @@ REM https://stackoverflow.com/questions/61888625/what-is-f-in-the-for-loop-comma
 for /f "tokens=*" %%i in ('ideasy %IDE_OPTIONS% env') do (
   call set %%i
 )
-if not %ERRORLEVEL% == 0 (
```
alfeilex, line 20: Is it a correct fix?

alfeilex:
https://github.com/devonfw/IDEasy/blob/d36317e3436615b18d0be121ee96e5201b896677/cli/src/main/package/bin/ide.bat#L27-L29

Is the condition right here?

cli/src/main/package/bin/ide.bat
```diff
@@ -5,6 +5,13 @@ Set _fBGreen=[92m
 Set _fBRed=[91m
 Set _RESET=[0m
 
+rem Add Git PATH entries for CMD - https://github.com/devonfw/IDEasy/issues/764
+for %%H in ( HKEY_LOCAL_MACHINE HKEY_CURRENT_USER ) do for /F "usebackq tokens=2*" %%O in (`call "%SystemRoot%"\system32\reg.exe query "%%H\Software\GitForWindows" /v "InstallPath" 2^>nul ^| "%SystemRoot%\system32\findstr.exe" REG_SZ`) do set GIT_HOME=%%P
+
+if exist "%GIT_HOME%\bin\bash.exe" (
```
alfeilex, line 11: Do we have to check for Git-Bash?

hohwille: We need to do such check in case the registry entry was not found and `%GIT_HOME%` is empty or containing nonsense.
We could check for a different thing than bash here (e.g. `\usr\bin\basename.exe`), but it will not really make a difference and may even be more fragile for future git releases.

Additionally, we should check if the entries we plan to add to the PATH are already on the PATH avoiding to create duplicates on every call.
Unlike bash or zsh the CMD does not have a concept of sub-environments so we are continuously extending the `PATH` with every call making it longer and longer until we finally exceed some limit and cause an error.
Please also note that when installing git for windows you get an option in the installer GUI asking if you want to install globally into the system or not:
image
The default is the first option that is actually causing this problem we are trying to workaround.
So ideally we write CMD code here that only adds the missing PATH entries if they are not already present and that does not break if something is going wrong with this workaround to avoid blocking everything else.

hohwille (changes requested):
@alfeilex thanks for your PR and fixing this bug. 👍 
As a quickfix your changes should already work, but we need some more robustness for it.
Please have a look at my comment.

cli/src/main/package/bin/ide.bat
hohwille: https://superuser.com/questions/480238/cmd-append-to-path-without-duplicating-it

hohwille:
> https://github.com/devonfw/IDEasy/blob/d36317e3436615b18d0be121ee96e5201b896677/cli/src/main/package/bin/ide.bat#L27-L29
> 
> Is the condition right here?

Very good point. We actually want the exit code of `ideasy env` command but we are directly processing the output of it and invoking `call set %%i` in the loop. Maybe the `ERRORLEVEL` will be overridden by that command inside the loop.
Coding `CMD` is actually even 10 dimensions worse to coding `bash` what can already be some kind of nightmare itself.
But solving this problem in `bash` would be very easy and doing the same in `CMD` could be tricky - at least I have no clue how to do it.
I just did a simple test:
```
C:\Users\hohwille>ide


C:\Users\hohwille>d:

D:\>cd projects

D:\projects>cd IDEasy

D:\projects\IDEasy>ide

D:\projects\IDEasy>
```
So it seems this condition is never met and the message is never printed out at all.
We need to create a separate bug issue for this.

cli/src/main/package/bin/ide.bat
alfeilex: Done, I had to create two IF blocks where each entry is added to the PATH, because the PATH doesn't update in the same block. With that approach, we dont, have to use DelayedExpansion.

cli/src/main/package/bin/ide.bat
```diff
@@ -17,6 +33,7 @@ REM https://stackoverflow.com/questions/61888625/what-is-f-in-the-for-loop-comma
 for /f "tokens=*" %%i in ('ideasy %IDE_OPTIONS% env') do (
   call set %%i
 )
-if not %ERRORLEVEL% == 0 (
+
+if %ERRORLEVEL% == 0 (
   echo IDE environment variables have been set for %IDE_HOME% in workspace %WORKSPACE%
```
alfeilex, line 38: > > https://github.com/devonfw/IDEasy/blob/d36317e3436615b18d0be121ee96e5201b896677/cli/src/main/package/bin/ide.bat#L27-L29
> > 
> > Is the condition right here?
> 
> Very good point. We actually want the exit code of `ideasy env` command but we are directly processing the output of it and invoking `call set %%i` in the loop. Maybe the `ERRORLEVEL` will be overridden by that command inside the loop. Coding `CMD` is actually even 10 dimensions worse to coding `bash` what can already be some kind of nightmare itself. But solving this problem in `bash` would be very easy and doing the same in `CMD` could be tricky - at least I have no clue how to do it. I just did a simple test:
> 
> ```
> C:\Users\hohwille>ide
> 
> 
> C:\Users\hohwille>d:
> 
> D:\>cd projects
> 
> D:\projects>cd IDEasy
> 
> D:\projects\IDEasy>ide
> 
> D:\projects\IDEasy>
> ```
> 
> So it seems this condition is never met and the message is never printed out at all. We need to create a separate bug issue for this.
> 
> UPDATE: I created bug #894 for this problem since it is unrelated to the GIT PATH issue.

I think the removing of `not` is the only fix we need. I get the `ERRORLEVEL` of 17, when I run the `ide env` outside a project folder and there is not output. The only thing remains, is that if we use `ide` in a project folder, we get the `echo IDE environment variables have been set for %IDE_HOME% in workspace %WORKSPACE%` message as output

cli/src/main/package/bin/ide.bat
```diff
@@ -5,6 +5,22 @@ Set _fBGreen=[92m
...
+
+if exist "%GIT_HOME%\bin\bash.exe" (
+	echo "%PATH%" | find /i "%GIT_BIN%">nul  || set PATH=%PATH%;%GIT_BIN%
+)
+
+if exist "%GIT_HOME%\bin\bash.exe" (
+	echo "%PATH%" | find /i "%GIT_CORE%">nul  || set PATH=%PATH%;%GIT_CORE%
+	REM KEEP the whitespace after call to clear ERRORLEVEL back to 0 otherwise it's 1 when PATH entries not found in find
+	call
+)
```
alfeilex, line 22: For clarification... the `call_ ` is there to set the errorlevel to 0 since the `find /i "%GIT_CORE%">nul` is neq 0 when the entry is not found in the PATH and we get `Error: IDEasy failed with exit code %ERRORLEVEL%`

cli/src/main/package/bin/ide.bat
```diff
@@ -5,6 +5,22 @@ Set _fBGreen=[92m
...
+
 if not "%1%" == "" (
   ideasy %IDE_OPTIONS% %*
   if not %ERRORLEVEL% == 0 (
```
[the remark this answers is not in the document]
alfeilex: Will be handled in #894

cli/src/main/package/bin/ide.bat
```diff
@@ -5,6 +5,20 @@ Set _fBGreen=[92m
...
 Set _RESET=[0m
 
+rem Add Git PATH entries for CMD - https://github.com/devonfw/IDEasy/issues/764
+for %%H in ( HKEY_LOCAL_MACHINE HKEY_CURRENT_USER ) do for /F "usebackq tokens=2*" %%O in (`call "%SystemRoot%"\system32\reg.exe query "%%H\Software\GitForWindows" /v "InstallPath" 2^>nul ^| "%SystemRoot%\system32\findstr.exe" REG_SZ`) do set GIT_HOME=%%P
+
+set "GIT_BIN=%GIT_HOME%\usr\bin"
+set "GIT_CORE=%GIT_HOME%\mingw64\libexec\git-core"
+
+if exist "%GIT_BIN%" (
+	echo "%PATH%" | find /i "%GIT_BIN%">nul  || set PATH=%PATH%;%GIT_BIN%
```
hohwille, line 15: This is not working. Please test your code before providing your solution for review. When I merge and we release without further testing end-users will get blocked.
```suggestion
	echo "%PATH%" | find /i "%GIT_BIN%">nul  || set "PATH=%PATH%;%GIT_BIN%"
```

cli/src/main/package/bin/ide.bat
```diff
@@ -5,6 +5,20 @@ Set _fBGreen=[92m
...
+)
+
+if exist "%GIT_CORE%" (
+	echo "%PATH%" | find /i "%GIT_CORE%">nul  || set PATH=%PATH%;%GIT_CORE%
```
hohwille, line 19:
```suggestion
	echo "%PATH%" | find /i "%GIT_CORE%">nul  || set "PATH=%PATH%;%GIT_CORE%"
```

Issue #894, opened 2024-12-12
ide.bat not printing if IDEasy was initialized


hohwille:
### Expected behavior

As a IDEasy user who for some odd reason is ignorant and still using CMD, I want that after setting up my environment with the `ide` command the according success message gets printed so that I get the appropriate feedback.

### Actual behavior

In `CMD` no message is printed at all no matter if I was inside an IDEasy project or not after calling `ide` command (`ide.bat`).
```
D:\projects>ide


D:\projects>cd IDEasy

D:\projects\IDEasy>ide

D:\projects\IDEasy>
```

The interesting thing is that there is a tiny difference: If the setup failed because I was not inside an IDEasy project, I get an extra empty line logged.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. open `CMD` inside your IDEasy project
2. run `ide` command

### Related/Dependent Issues

See #886

### Comments/Hints:

https://github.com/devonfw/IDEasy/blob/75f5043e5279ccb9ee5bcdefaa76692bd8f13e75/cli/src/main/package/bin/ide.bat#L17-L22

So we want to get the exit code of `ideasy ... env` command but the way we are doing this seems to be wrong.
The question is if the `call set %%i` command is also setting (overriding) the `ERRORLEVEL` variable.
Finally, the major question is why the `ERRORLEVEL` is never zero - so is there always an error happening?.
The general variable and pattern was already used before in `devonfw-ide` and used to work:
https://github.com/devonfw/ide/blob/8698d2ba7468d34aec82c86fabdb09b5ee84d9c8/scripts/src/main/resources/scripts/devon.bat#L94

### Affected version:

* OS: Windows
* 2024.12.001-beta-12_10_02-SNAPSHOT

PR #902, opened 2024-12-17
#894: Fix ide.bat printing for initialization and error output

alfeilex:
Fixes: #894 

This PR fixes ide.bat to print the correct output when an error occurs or ideasy is initialized.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -6,6 +6,7 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/894[#894]: Fix ide.bat printing for initialization and error output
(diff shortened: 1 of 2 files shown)


alfeilex:
23a6924 fixes that the correct errorlevel is used. Here an example to see the difference
```
if not "%1%" == "" (
  ideasy %IDE_OPTIONS% %*
  echo test %ERRORLEVEL% 
  goto :output_error
)

:output_error
echo inside %ERRORLEVEL%
if not %ERRORLEVEL% == 0 (
	echo %_fBRed%Error: IDEasy failed with exit code %ERRORLEVEL% %_RESET%
	exit /b %ERRORLEVEL%
  )
```

hohwille (approved):
@alfeilex thanks for this fix. I tested it on my Win11 X64 machine and it is now working correctly. Great job👍

Issue #790, opened 2024-11-21
Fix intellij plugin installation


jan-vcapgemini:
### Expected behavior

As an IDEasy intellij user I expect IDEasy to install and activate plugins for intellij properly if the active flag is set.

### Actual behavior
IDEasy downloads the plugins to the `project/plugins/intellij` folder but does not activate it automatically.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Modify the `settings/intellij/plugins/` files and activate a plugin
2. Start intellij by typing `ide intellij`
3. Check if the plugin was activated in intellij

### Comments/Hints:
We need to check the new intellij CLI installPlugins feature: https://www.jetbrains.com/help/idea/working-with-the-ide-features-from-command-line.html#arguments

### Affected version:

* OS: Windows
* IDEasy: 2024.11.001-beta

hohwille:
This doc is IMHO the better link: https://www.jetbrains.com/help/idea/install-plugins-from-the-command-line.html

PR #909, opened 2025-01-08
#790: Fix intellij plugin installation

alfeilex:
Fixes: #790 

This PR fixes the plugin installation for intelllij by using the available cli command `installPlugins` provided by `intellij` itself


cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeaBasedIdeToolCommandlet.java
```diff
@@ -38,7 +38,6 @@ public IdeaBasedIdeToolCommandlet(IdeContext context, String tool, Set<Tag> tags
   protected void configureToolArgs(ProcessContext pc, ProcessMode processMode, ProcessErrorHandling errorHandling, String... args) {
 
     super.configureToolArgs(pc, processMode, errorHandling, args);
-    pc.addArg(this.context.getWorkspacePath());
```
alfeilex, line 41: I removed that part and I am executing `run()` after the plugin installation is complete (in case that is the idea of that part). I am not sure if that change breaks anything. Otherwise, I need some more clarification on why the workspace path is added to the arguments and where I could move it

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -63,4 +71,28 @@ protected void installDependencies() {
...
+  public void installPlugin(ToolPluginDescriptor plugin, Step step) {
+
+    // In case of plugins with a custom repo url
+    boolean customRepo = plugin.url() != null;
+    List<String> args = new ArrayList<>();
+    args.add("installPlugins");
+    args.add(plugin.id());
+    if (customRepo) {
+      args.add(plugin.url());
+    }
```
alfeilex, line 84: I have tried to run `... installPlugins <plugin_id> [custom_repo_url]` as described in the docs https://www.jetbrains.com/help/idea/install-plugins-from-the-command-line.html#-ci6zah_17

Unfortunately, I could not make it work. I build my own dummy plugin for testing and hosted a local server where the plugin and the required `updatePlugins.xml` are located, but I get a error message stating that  the plugin could not be found....

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeaBasedIdeToolCommandlet.java
```diff
@@ -70,6 +72,13 @@ public void installPlugin(ToolPluginDescriptor plugin, Step step) {
     }
   }
 
+  @Override
+  public void runTool(String... args) {
+    List<String> extendedArgs = new ArrayList<>(Arrays.asList(args));
+    extendedArgs.add(this.context.getWorkspacePath().toString());
+    super.runTool(extendedArgs.toArray(new String[0]));
+  }
```
alfeilex, line 80: I moved the workspace configuration into this location. So when the tool runs, then the workspace is added as an argument.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
hohwille: OK. Got it - so with your new approach you install each plugin by a separate invocation of Intellij.
That might be slower if many plugins need to be installed but should also be KISS and robust.
Also if a plugin installation fails, the error handling should already be in place telling the user via `Step`s which plugin was successfully installed and which not.
We can crate a new story that will even add a progress bar if multiple plugins are to be installed and move it forward by the number of plugins processed.

> maybe the problem is with the local hosting

I assume so. Most probably the URL is just the base-URL of a repository and the full download URL is automatically derived from that and you need to read the documentation from Jetbrains about this.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeaBasedIdeToolCommandlet.java
hohwille: An alternative could be to check if the `args` start with `installPlugins` but it is also OK to remove it here and move the argument to `runIde`. But then you should remove the enitre method since just calling `super` is kind of pointless now.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeaBasedIdeToolCommandlet.java
hohwille: IMHO there is `runIde` for IDE commandlets that could override for this use-case what would make it a little bit more obvious.

hohwille (approved):
@alfeilex thanks for your PR and improvements. Good work 👍 
IMHO when your own review comments are addressed, we can merge.
I added some comments to give some hints/suggestions on that.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeaBasedIdeToolCommandlet.java
jan-vcapgemini: I can only find `run`, which is final. So I guess we need to keep `runTool` here.

hohwille: It was present once:
https://github.com/devonfw/IDEasy/blob/12cdd6406aea09a8f451592f3abcba57a5ee9b67/cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java#L56

However, someone removed it meanwhile so we need to stay with this approach.

hohwille (approved):
@alfeilex and @jan-vcapgemini thanks for completing this PR. Looks good now and great work 👍 
What I still dislike is this stuff in the JUnit:
```java
    String content = "plugin_id=mockedPlugin\nplugin_active=true\nplugin_url=http://customRepo";
    Files.write(this.context.getSettingsPath().resolve("intellij").resolve("plugins").resolve("MockedPlugin.properties"),
        content.getBytes(StandardCharsets.UTF_8));
```

We have solved this problem in other tests by adding plugin configuration properties file to the test project using a variable for the wiremock base URL and have some generic solution to replace the variable.
We should align `IntellijTest` with other tests using the better pattern, but let us not block this PR. Ready for merge.

hohwille:
> What I still dislike is this stuff in the JUnit:
> 
> ```java
>     String content = "plugin_id=mockedPlugin\nplugin_active=true\nplugin_url=http://customRepo";
>     Files.write(this.context.getSettingsPath().resolve("intellij").resolve("plugins").resolve("MockedPlugin.properties"),
>         content.getBytes(StandardCharsets.UTF_8));
> ```
> 
> We have solved this problem in other tests by adding plugin configuration properties file to the test project using a variable for the wiremock base URL and have some generic solution to replace the variable. We should align `IntellijTest` with other tests using the better pattern, but let us not block this PR. Ready for merge.

Created issue #970

PR #917, opened 2025-01-13
#916: download status code error handling

### LINKED ISSUE #916: download is missing status code error handling
### Expected behavior

As a IDEasy user, I want to get reasonable error message with details if a download failed so that I can understand what is going on and trace down the error.

### Actual behavior

IDEasy download code is only handling status code `200` and otherwise ignores the response doing nothing.
As a result no file is downloaded and the file where download is expected simply does not exist and subsequent errors show up like move fails with `java.nio.file.NoSuchFileException`. The real reason of the error is lost and the end-user is left in the rain.

1. create `ide-custom-tools.json` with some invalid repository URL and push it to your fork of ide-settings (see #915 for details)
2. create according project and ensure custom tools is trying to get installed (see #915 for details)
3. when the download failed you should get a `java.nio.file.NoSuchFileException` for the move of the downloaded file that was actually not downloaded because of an invalid HTTP response.

### Related/Dependent Issues

IMHO this bug existed from the beginning of our git history when the initial code was added.
Slightly related with PR #53

### Comments/Hints:

The bug is very obvious here:
https://github.com/devonfw/IDEasy/blob/8e971e1a86eaf51ce2ccd586ead1c8147a3ac48b/cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java#L101-L103
There is no `else` block for error handling of status codes like 404, 500, etc.

### Affected version:

* 2024.12.002-beta

hohwille:
fixes #916

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -6,6 +6,7 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/916[#916]: download is missing status code error handling
 * https://github.com/devonfw/IDEasy/issues/894[#894]: Fix ide.bat printing for initialization and error output
 
 The full list of changes for this release can be found in https://github.com/devonfw/IDEasy/milestone/18?closed=1[milestone 2025.01.001].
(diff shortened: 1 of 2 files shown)


jan-vcapgemini (approved):
LGTM.

PR #932, opened 2025-01-14
#894: Fix ide.bat printing for initialization and error output

alfeilex:
Fixes: #894 and undo some wrong changes in PR #902


hohwille (approved):
@alfeilex thanks for this fix. Should be fine now. 👍

Issue #758, opened 2024-11-13
Create status commandlet


hohwille:
As a IDEasy user, I want to have a simple command so that I can quickly see outstanding notifications and warnings.

Currently we log such notifications and warnings on the fly while processing arbitrary commandlets.
However, there are also some commandlets that are considered to produce "processable output" that need to suppress such output.

Question: What commandelt should the end-user invoke in order to see if there are updates to `settings` or there are warnings for his configuration?
Answer: There is currently no clean answer this this question. You may call `ide help` and get the output on top of the help but need to scroll up in order to find it. You may call `ide -v` or `ide` or `ide get-version java` and notice that all of these have processable output and suppress the information. You might get the output if you add `--debug` option but then you also get lots of debug output. So you might call `ide install java` or something but just to get such information you do not want to invoke something like an installation.

So in other words: 
- [ ] We need a `StatusCommandlet` that can be called via `ide status` to get only such status information without applying any changes.
- [ ] I also propose that just typing `ide` should invoke `ideasy status` if we are inside an IDEasy project.
- [ ] Whenever I call this `StatusCommandlet` it will print me all warnings about my `settings` (e.g. duplicated or legacy variables) as well as potential information about updates to `settings` I should apply with `ide update`.

PR #935, opened 2025-01-14
remove init sub-command

hohwille:
with #779 we introduced `functions` script that defines the `ide` "command" as a function instead of an alias.
There we decided to add a virtual sub-command `init` so that we distinguish `ide` from `ide init`.
The idea was that when `ide init` is called we do only a minimum initialization and print no extra messages because this is what happens when a new shell is started. On the other hand we had the idea that when the user just calles `ide` that he gets more verbose output like #758. But as we later decided that the user should explicitly call `ide status` to get such extra details we consider that calling jsut `ide` should still remain minimal and only initialize my environment variables.

ATTENTION: This PR is to be discussed with the team.
We had the idea to have the following extra features when `ide` is called explicitly by the user (compared to if `ide init` is called when the shell is started):
* `You are not inside an IDEasy project: $PWD` (see also issue #739)
* Cygwin warning (see also #715)

If we agree that we do not want or need such feedback messages if just `ide` is called, we can merge this PR to simplify our script and avoid confusion with `ide init` that does not show up in `ide help` since there is no `InitCommandlet`.

### DIFF
--- a/cli/src/main/package/functions
+++ b/cli/src/main/package/functions
@@ -3,7 +3,7 @@
 function ide() {
   local return_code
   local ide_env
-  if [ $# != 0 ] && [ "$1" != "init" ]; then
+  if [ $# != 0 ]; then
     ideasy ${IDE_OPTIONS} "$@"
     return_code=$?
     if [ $return_code != 0 ]; then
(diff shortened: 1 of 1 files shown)


hohwille:
No review after 1 week - merging now.

Issue #914, opened 2025-01-13
Setup needs to create .bashrc if not existent


hohwille:
### Expected behavior

As an IDEasy user, I want `setup[.bat]` to do all that is needed for the setup so that everything works afterwards.

### Actual behavior

Recent versions of Git for Windows do not create `.bashrc` by default.
If `.bashrc` file is not present then IDEasy `setup` does not add required initializations there.
Hence the installation is then still incomplete:
```
$ ide -v
bash: ide: command not found
```

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. setup a fresh windows 11 machine
2. install git for windows
3. download IDEasy
4. unpack it to `projects/_ide`
5. run `setup.bat` 

### Comments/Hints:

We simply should render a warning if `.bashrc` does not exist but then create it (`touch`).
https://github.com/devonfw/IDEasy/blob/8e971e1a86eaf51ce2ccd586ead1c8147a3ac48b/cli/src/main/package/setup#L6

### Affected version:

* OS: Windows but theoretically all OS
* 2024.12.002-beta

PR #940, opened 2025-01-15
#914: Setup creates .bashrc and .zshrc if not existent 

julia-cap:
Closes https://github.com/devonfw/IDEasy/issues/914

### Implements
* Setup script checks if files exist
* Setup script creates .bashrc and/or .zshrc if not existent
* Info about file creation is logged


cli/src/main/package/setup
```diff
@@ -2,6 +2,18 @@
...
+   if [ "$1" = "${fileBashrc}" ] && [ ! -f "${fileBashrc}" ]; then
+       echo "File .bashrc not found."
+       touch "$fileBashrc"
+       echo "Created .bashrc file in directory ~/."
+     fi
+ 	if [ "$1" = "${fileBashrc}" ] && [ ! -f "${fileZshrc}" ]; then
+       echo "File .zshrc not found."
+       touch "$fileZshrc"
+       echo "Created .zshrc file in directory ~/."
+     fi
```
jan-vcapgemini, line 16: Please reformat this file, it looks a bit distorted.
Also please try to remove redundancies using the variables like `File ${fileBashrc} not found.`  f.e.
The if condition could be moved into a separate function and the log messages could be re-used.

jan-vcapgemini (changes requested):
@julia-cap thanks for your implementation of the `.bashrc` and `.zshrc` checks.
I've added a small CR, please check, should be ready for review when this is resolved.

cli/src/main/package/setup
hohwille: I see two different approaches:
1. only ensure `.bashrc` is created if missing, then lets do it before calling `doSetupInConfigFile ~/.bashrc` outside the function.
2. create every missing file (even `.zshrc` though on Windows it is unlikely that users will use z-Shell) but then do not hardcode the filenames in `doSetupInConfigFile` but just use `$cfg` variable.

FYI: Shell scripting is always a little magic. To get started, it may help to add something like `echo $cfg` temporary to understand what is going on.

julia-cap: I have refactored the code according to your suggestions. Please have a look.

cli/src/main/package/setup
```diff
@@ -1,7 +1,22 @@
...
+function createFile() {
+  local fileName=$1
+  echo "File $fileName not found."
+  touch "$fileName"
+  echo "Created ${fileName}"
+}
+
+function checkIfFileExists() {
+  local fileName=$1
+  if [ ! -f "${fileName#~/}" ]; then
```
hohwille, line 12: why do we have to strip a potential `~/` prefix from the path here?
I do not get this one.
Bash globbing should actually resolve `~` with the value of `$HOME`.

hohwille (approved):
@julia-cap thanks for your PR. Looks good to me now. 👍 
I guess only very little windows users will use z-Shell but actually it is good that we create `.zshrc` anyway so those users that later install z-Shell will have IDEasy working without the need to rerun setup again.

I added a small comment just for curiosity and clarification.
IMHO the `#~/` is not needed but maybe you also found an edge-case I am missing.

BTW: In `functions` we should not create new bash functions since those will pollute the users namespace but in `setup` script this is fine and makes the code more readable.

cli/src/main/package/setup
```diff
@@ -1,7 +1,22 @@
 #!/bin/bash
 
+function createFile() {
+  local fileName=$1
+  echo "File $fileName not found."
+  touch "$fileName"
+  echo "Created ${fileName}"
+}
+
+function checkIfFileExists() {
```
hohwille, line 10: probably this function should be renamed to `createFileIfNotExisting` since it is not only checking if a file exists but also creating it then.

cli/src/main/package/setup
```diff
@@ -1,7 +1,22 @@
 #!/bin/bash
 
+function createFile() {
+  local fileName=$1
+  echo "File $fileName not found."
```
hohwille, line 5: BTW: If we already print out that we created the file in line 7, then this message might only cause additional confusion. WDYT?

cli/src/main/package/setup
```diff
@@ -1,7 +1,22 @@
...
+    createFile "$fileName"
+  fi
+}
+
 function doSetupInConfigFile() {
   local cfg=$1
+  checkIfFileExists "${cfg}"
   if [ ! -f "${cfg}" ]; then
     echo "${cfg} not found - skipping."
     return
```
hohwille, line 22: shouldnt we remove this if now?
It can never happen anymore after your change.

cli/src/main/package/setup
julia-cap: When I created a variable for the filePath, IntelliJ complained that the tilde cannot be resolved. It might be different in other places in the script, I saw it's being used elsewhere. I will check if it works without trimming and use $cfg as is.

cli/src/main/package/setup
julia-cap: True. I will refactor.

cli/src/main/package/setup
julia-cap: Yes, quite right. I removed it.

cli/src/main/package/setup
julia-cap: It works, just like you said. I have committed the changes accordingly.

PR #949, opened 2025-01-17
#815: decouple from docgen and fix ant adoc processing

### LINKED ISSUE #815: Links within IDEasy.pdf generally do not work
### Expected behavior

As a IDEasy user, I want to use the links in IDEasy.pdf so that i can quickly get to the reverenced sites.

### Actual behavior
Most links just throw you back to the introduction

1. Open IDEasy.pdf in ...projects/_ide and find and use the "settings" link

hohwille:
fixes #815

* I decoupled our documentation build from docgen (devonfw)
* I found various bugs in the regex rules of the ant AsciiDoc pre-precessing
* I did some small improvements to the documentation content itself
* Currently the actual bug #815 is not yet fixed nor understood


hohwille:
I figured out some interesting things:
* there were bugs in the ant logic using regex replacements and even a bean-shell script (Java code) to transform the AsciiDoc Syntax.
* This ant logic is required as a preprocessor because of the link syntax:
  1.  On github we have many single AsciiDoc pages that can link to each other and also into sections of the same or other AsciiDoc files
  1. When we link to a section in the same document, then we refer to the section via the syntax `xref:«section-name»[«linktext»]` if the section in that document has the title `«section-name»`. This works on github because the github HTML renderer of AsciiDoc automatically adds anchors for linking and bookmarking to each section this way (e.g. to support ToC).
  1. When we link to a different AsciiDoc file, we instead use the syntax `link:«filename.adoc»[«linktext»]` and we can even link to a section in that other document via `link:«filename.adoc»#«section-name»[«linktext»]`. What both also works fine on github.
  1. Now if we render multiple AsciiDoc files into a single PDF or HTML output things are slightly different. AsciiDoc(tor) considers everything as one file. Whenever we do an `include::«filename.adoc»[...]` what behaves as if we copy&pasted the AsciiDoc file `«filename.adoc»` into our root document (`IDEasy.adoc`) instead of the include and this also works recursively.
  1. Since now everything is treated as a single document, the `link:` syntax does not work anymore and all document internal links need to be done with `xref:` syntax. To put explicit anchors into AsciiDoc, we use the syntax `[[«anchor»]]` so we can link there via `xref:«anchor»[«linktext»]`.
  1. Now, there is yet another problem: In individual small AsciiDoc files we can link a section `Overview` via `xref` and also we may have a second AsciiDoc file that by coincident also has a section `Overview` and may link to it via `xref` and maybe even a third document may link to `Overview` in one of these documents e.g. via `link:«filename.adoc»#Overview[«linktext»]`.
  1. So in other words in our virtual document where everything is included together we need to make anchors unique as otherwise the anchor `Overview` would be ambiguous.
  1. We solved this in docgen by prepending the document filename to every anchor.
  1. For every `xref` or `link` that has no `«linktext»` (e.g. `xref:«section-name»[]`) we add the link itself as `«linktext»` (so the example will become `xref:«section-name»[«section-name»]`). This would be pointless so far since this is the default behaviour of AsciiDoc but is required since we change the link itself in the next steps.
  1. So for the top-level section title of each AsciiDoc file, we automatically add `[[«filename.adoc»]]` before that section title. And for each other section, we add `[[«filename.adoc»_«section-name»]]`.
  1. Now for each `«filename.adoc»` we replace each `xref:«section-title»[«linktext»]` with `xref:«filename.adoc»_«section-title»[«linktext»]`.
  1. Further, we replace every `link:«filename.adoc»#«section-name»[«linktext»]` with `xref:«filename.adoc»_«section-name»[«linktext»]`.
  1. Finally, we replace every `link:«filename.adoc»[«linktext»]` with `xref:«filename.adoc»[«linktext»]`.
* At this point conceptually everything should work file.
* The step `xii` was broken (in docgen and therefore in devonfw-ide documentation generation) and I fixed it.
* What still confuses me is that now links of that format and initial `xref` links seem to work but links to files without anchor (`xi`) do not work. And that seems to be a bug caused by `asciidoctor`.
* For further analysis, I just reran the pure AsciiDoc generation (`mvn org.asciidoctor:asciidoctor-maven-plugin:1.6.0:process-asciidoc`) and tweaked the transformed AsciiDoc files that looked correct to me although the links from form `xi` did not work (also when I generate to HTML instead of PDF, then the broken links just have `#` as target).
* The crazy thing is that when I changed both the anchor (`[[usage.adoc]]`) and the link (`xref:usage.adoc[...]`) from `usage.adoc` to `usage.adoc_test` and reran the generation then it worked. So is asciidoctor thinking that `xref:usage.adoc` is actually still pointing to a file instead of an anchor and has some obscure build-in heuristic rules for this???
* So maybe we just change this to add a `_title` suffix to the top anchors and the according xrefs and can workaround the problem then. Still the behaviour of asciidoctor does not make any sense to me.

hohwille:
Hurray, this was a tricky issue with quite some detective puzzle work, but I finally made it.
IMHO asciidoctor is kind of flawed but the workaround solves our problem and the links in the PDF work now.

jan-vcapgemini:
I've tested a local build of the documentation.
It appears that some links are still not working, f.e. `link:setup.adoc[setup]` in lombok

jan-vcapgemini (approved):
LGTM. Ready for merge.

Issue #942, opened 2025-01-15
Faulty legacy warnings about non existing devon.properties


hohwille:
```
$ ide
IDE_ROOT is set to C:\Users\joerg\projects
IDE_HOME is set to C:\Users\joerg\projects\IDEasy
You are online.
Your settings are up-to-date.
Found legacy properties C:\Users\joerg\projects\IDEasy\conf\devon.properties
Found legacy properties C:\Users\joerg\projects\IDEasy\workspaces\main\devon.properties
Found legacy properties C:\Users\joerg\projects\IDEasy\settings\devon.properties
Found legacy properties C:\Users\joerg\devon.properties
Your settings are outdated and contain legacy configurations. Please consider upgrading your settings:
https://github.com/devonfw/IDEasy/blob/main/documentation/settings.adoc#upgrade
IDE environment variables have been set for /c/Users/joerg/projects/IDEasy in workspace main
```

Please see the warnings about `Found legacy properties....`

But the fact is that none of these legacy properties even exists:
```
$ ls /c/Users/joerg/projects/IDEasy/conf/
ide.properties  mvn/  npm/
```

# Comments

I guess we had some code checking for this that was making his choice via `EnvironmentVariablesProperties.getLegacyPropertiesFilePath() != null`.
This has changed but this has always been wrong then since only if the file really exists, we need to print the warning. Earlier the method was returning `null` if it internally found out that the `Path` does not exist.

hohwille:
Here is the bug:
https://github.com/devonfw/IDEasy/blob/6f167c101b1181be14cc8ecb199a5791ba48c46a/cli/src/main/java/com/devonfw/tools/ide/commandlet/StatusCommandlet.java#L44-L48

Add `Files.exists` to the `if` to fix it.

PR #953, opened 2025-01-20
#942: Added check for existence of devon.properties file

julia-cap:
Closes https://github.com/devonfw/IDEasy/issues/942

### Implements:
* Added check to verify if devon.properties file exists 
* Warning is only logged when file exists


hohwille (approved):
@julia-cap thanks for the PR and the correct fix 👍 
Ready for merge.

PR #903, opened 2024-12-19
#898: Improved output of get-version/edition and uninstall/-plugin

WorkingAmeise:
Closes: #898 
Edition/VersionGetCommandlet changed them to log the configured option if the tool is not installed and the installed option if it is installed.
Changed the toolInstallInfo method to log into IdeSublogger instead of using context.info or context.warning.
Changed the uninstall commandlet to log a warning in case the requested commandlet is not installed.
Changed the uninstallPlugin commandlet to give feedback wether the plugin could be deleted or if there was not installation found.
Adjusted Tests for Edition/VersionGetCommandlet as well as Uninstall commandlet to expect the correct loglevel or content of logged messages.
Added logAtProccessable to IdeTestContextAssertions to check if a log-message was logged on PROCESSABLE-level.
Added .ide.software.version file to the npm software in the basic ide-test-project-ressource, since it was needed to test my implementation.

PR #955, opened 2025-01-20
#903: fix test to avoid modification in src/test/resources causing other tests to turn red as side effect

hohwille:
Test-quickfix introduced by PR #903


jan-vcapgemini (approved):
LGTM.

PR #957, opened 2025-01-20
#786: Upgrade commandlet

### LINKED ISSUE #786: Add UpgradeCommandlet to automatically update to the latest release of IDEasy
As a IDEasy user, I want to be able to upgrade to the latest version of IDEasy so that I do not have to check for new versions, manually download and re-install IDEasy.
This should happen via `ide upgrade` (what will install the latest IDEasy product version if not already installed - unlike `ide update` what will update my current project to the latest state).
Therefore an `UpgradeCommandlet` shall be implemented.

The old mechanism in devonfw-ide for this feature was `devon ide update scripts` and the code can be found here:
https://github.com/devonfw/ide/blob/9efaf926f1daa4e1e37a22fc89081255fbf55dab/scripts/src/main/resources/scripts/command/ide#L47-L66
The key to find the latest version is https://repo1.maven.org/maven2/com/devonfw/tools/IDEasy/ide-cli/maven-metadata.xml
and for SNAPSHOTs https://s01.oss.sonatype.org/content/repositories/snapshots/com/devonfw/tools/IDEasy/ide-cli/maven-metadata.xml

- [x] As a preparation a first PR could create a `MavenRepository` implementing [ToolRepository](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/repo/ToolRepository.java) with a mapping from `tool` and `edition` to `groupId` and `artifactId` (as a starting point this can be hardcoded). It should support parsing `maven-metadata.xml` to implement `resolveVersion`.
The `download` method can be implemented straight forward reusing existing download functionality we already have implemented after mapping and building the download URL. Theoretically `findDependencies` could be implemented via `pom` but we do not want or need this yet and simply return an empty `List` here.
Then we can make the `MavenRepository` accessible via `IdeContext`.
- [ ] Once all this is done (in another PR after merging the first or in the same PR), the `run` method of `UpgradeCommandlet` can use the `MavenRepository` to resolve the latest version, compare it with the current version (`IdeVersion.get()` converted to `VersionIdentifier`). If a newer version is available, then download it. Finally extract the download to `$IDE_ROOT/_ide` overwriting all files. The latter part may be tricky on Windows due to file locks.
- [ ] As an advanced feature if the current version of IDEasy is a SNAPSHOT, then the MavenRepository should be created with `https://s01.oss.sonatype.org/content/repositories/snapshots` as base URL instead of `https://repo1.maven.org/maven2`.
Further, we currently cannot distinguish different SNAPSHOT versions.
Therefore I propose a new sub-task story:
- [x] #791

salimbouch:
fixes #786


cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
```diff
@@ -0,0 +1,158 @@
...
+    String classifier = sys.getOs() + "-" + sys.getArchitecture();
+
+    String pathVersion = version.toString();
+    // If it's a resolved snapshot version (contains timestamp), convert back to -SNAPSHOT format for the path
+    if (pathVersion.matches(".*-\\d{8}\\.\\d{6}-\\d+")) {
+      pathVersion = pathVersion.replaceAll("-\\d{8}\\.\\d{6}-\\d+", "-SNAPSHOT");
+    }
+
+    // hardcoding the file extension seems wrong
+    String fileName = artifactId + "-" + version + "-" + classifier + ".tar.gz";
```
salimbouch, line 74: right now just hardcoding the file extension, that means this doesnt work for every maven artifact

cli/src/test/java/com/devonfw/tools/ide/commandlet/UpgradeCommandletTest.java
```diff
@@ -0,0 +1,22 @@
...
+    IdeTestContext context = newContext(PROJECT_BASIC);
+    UpgradeCommandlet uc = context.getCommandletManager().getCommandlet(UpgradeCommandlet.class);
+
+    assertThat(uc.isSnapshotNewer("2024.12.002-beta-12_18_02-SNAPSHOT", "2025.01.001-beta-20250118.022832-8")).isTrue();
+    assertThat(uc.isSnapshotNewer("2024.12.002-beta-01_01_02-SNAPSHOT", "2024.12.002-beta-20241218.023429-8")).isTrue();
+    assertThat(
+        uc.isSnapshotNewer("2024.12.002-beta-12_18_02-SNAPSHOT", "2024.12.002-beta-20241218.023429-8")).isFalse();
+    assertThat(uc.isSnapshotNewer("SNAPSHOT", "2024.12.002-beta-20241218.023429-8")).isFalse();
+    assertThat(uc.isSnapshotNewer("someUnknownFormat1", "someUnknownFormat2")).isFalse();
+  }
```
salimbouch, line 21: only testing the snapshot comparison method as I cant really test the run method with all the download and the script launching and extraction and so on ..

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeCommandlet.java
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  private Path createUpgradeScript() throws IOException {
+
+    Path scriptPath = this.context.getIdeRoot().resolve("upgrade.bat");
+
+    String scriptContent =
+        "@echo off\n" + "ping -n 4 127.0.0.1 > nul\n" + "C:\\Windows\\System32\\tar.exe -xzf \"%~1\" -C \"%~2\"\n"
+            + "del \"%~f0\"";
+
```
salimbouch, line 162: doesnt seem like the most elegant bat file but works. The `timeout` command doesnt work for some reason so using `ping`, and I needed to get the whole path of the `tar.exe` for it to extract.

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
hohwille: Simply add an extra parameter for the extension - it should default to `.jar`.

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
```diff
@@ -0,0 +1,158 @@
...
+    return version != null ? basePath + "/" + version + "/" + fileName : basePath + "/" + fileName;
+  }
+
+  private String getBaseUrl(String groupId, String artifactId) {
+
+    if (isIdeasySnapshot(groupId, artifactId, IdeVersion.get())) {
+      return MAVEN_SNAPSHOTS;
+    }
+    return MAVEN_CENTRAL;
+  }
```
hohwille, line 59: this and its usage is wrong.
You implemented `MavenRepository` specific for our use-case of downloading IDEasy.
`IdeVersion.get()` needs to be used outside of this class to do the version comparison.
In `getMetadata` you need to use the `version` argument that you ignoring here.

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
```diff
@@ -0,0 +1,158 @@
...
+  protected UrlDownloadFileMetadata getMetadata(String groupId, String artifactId, VersionIdentifier version) {
+
+    SystemInfo sys = this.context.getSystemInfo();
+    String classifier = sys.getOs() + "-" + sys.getArchitecture();
+
+    String pathVersion = version.toString();
+    // If it's a resolved snapshot version (contains timestamp), convert back to -SNAPSHOT format for the path
+    if (pathVersion.matches(".*-\\d{8}\\.\\d{6}-\\d+")) {
+      pathVersion = pathVersion.replaceAll("-\\d{8}\\.\\d{6}-\\d+", "-SNAPSHOT");
+    }
```
hohwille, line 71: This is also misplaced here.
`getMetadata(IDEASY_GROUP_ID, IDEASY_ARTIFACT_ID, VersionIdentifier.of("2024.12.002-beta"))` should give us
https://repo1.maven.org/maven2/com/devonfw/tools/IDEasy/ide-cli/2024.12.002-beta/ide-cli-2024.12.002-beta.jar
The variant `getMetadata(IDEASY_GROUP_ID, IDEASY_ARTIFACT_ID, VersionIdentifier.of("2024.12.002-beta"), "windows-x64", ".tar.gz")` should give
https://repo1.maven.org/maven2/com/devonfw/tools/IDEasy/ide-cli/2024.12.002-beta/ide-cli-2024.12.002-beta-windows-x64.tar.gz
Where `windows-x64` is called `classifier` in maven terminology and `.tar.gz` is the extension (aka type in maven POM but there excluding the dot if you want to make it POM compliant just pass `tar.gz` with `jar` as default and add the dot internally in method impl).

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
```diff
@@ -0,0 +1,158 @@
...
+      return MAVEN_SNAPSHOTS;
+    }
+    return MAVEN_CENTRAL;
+  }
+
+  @Override
+  protected UrlDownloadFileMetadata getMetadata(String groupId, String artifactId, VersionIdentifier version) {
+
+    SystemInfo sys = this.context.getSystemInfo();
+    String classifier = sys.getOs() + "-" + sys.getArchitecture();
```
hohwille, line 65: Also this code is misplaced here and needs to be moved outside of `MavenRepository`.

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
```diff
@@ -0,0 +1,158 @@
...
+    try {
+      String metadataUrl =
+          getBaseUrl(groupId, artifactId) + "/" + getPath(groupId, artifactId, null, MAVEN_METADATA_XML);
+      Path tmpDownloadFile = createTempDownload(MAVEN_METADATA_XML);
+
+      try {
+        this.context.getFileAccess().download(metadataUrl, tmpDownloadFile);
+        DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+        DocumentBuilder builder = factory.newDocumentBuilder();
+        Document doc = builder.parse(tmpDownloadFile.toFile());
```
hohwille, line 91: It does not make sense to write the XML to the disc just to then read it.
We should directly load the file into XML in memory without writing temporary files to the disc.

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
```diff
@@ -0,0 +1,158 @@
...
+
+        String version = Optional.ofNullable(doc.getElementsByTagName("latest").item(0)).map(Element.class::cast)
+            .map(Element::getTextContent).orElseGet(
+                () -> Optional.ofNullable(doc.getElementsByTagName("release").item(0)).map(Element.class::cast)
+                    .map(Element::getTextContent)
+                    .orElseThrow(() -> new IllegalStateException("No latest or release version found in metadata")));
```
hohwille, line 97: wouldn't XPath help here to make the code more compact and elegant?

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
```diff
@@ -0,0 +1,158 @@
...
+
+        if (isIdeasySnapshot(groupId, artifactId, version)) {
+          version = resolveSnapshotVersion(groupId, artifactId, version);
+        }
```
hohwille, line 101: this should also not be specific for IDEasy.
If someone provides a version ending with `-SNAPSHOT` then you should resolve it the way you are doing that for any artifact no matter if IDEasy or not.
Outside when you use all this specific for IDEasy and you have version `2025.01.001-beta-01_21_02-SNAPSHOT` simply convert this back to `2025.01.001-beta-SNAPSHOT` what seems easy and pass that as version to this method to make it work.

Following Separation of Concerns this class should be generic and make sense for any maven artfact so we can reuse it also for `gcviewer`, `gcloganalyzer`, `cobigen`, or whatever we like.
Everything that is specific for IDEasy and the way we build our version printed by `ide -v` needs to go to `UpgradeCommandlet` and related code outside of this `repo` package.

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
```diff
@@ -0,0 +1,158 @@
...
+
+        return VersionIdentifier.of(version);
+      } finally {
+        this.context.getFileAccess().delete(tmpDownloadFile);
```
hohwille, line 105: will also be pointless then.

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
```diff
@@ -0,0 +1,158 @@
...
+        DocumentBuilder builder = factory.newDocumentBuilder();
+        Document doc = builder.parse(tmpDownloadFile.toFile());
+
+        String timestamp = Optional.ofNullable(doc.getElementsByTagName("timestamp").item(0)).map(Element.class::cast)
+            .map(Element::getTextContent)
+            .orElseThrow(() -> new IllegalStateException("No timestamp found in snapshot metadata"));
+
+        String buildNumber = Optional.ofNullable(doc.getElementsByTagName("buildNumber").item(0))
+            .map(Element.class::cast).map(Element::getTextContent)
+            .orElseThrow(() -> new IllegalStateException("No buildNumber found in snapshot metadata"));
```
hohwille, line 130: same here. Also this looks like code duplication that should be avoided by reusing code from a private method.

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
```diff
@@ -0,0 +1,158 @@
...
+    } catch (Exception e) {
+      throw new IllegalStateException("Failed to resolve snapshot version for " + baseVersion, e);
+    }
+  }
+
+  private boolean isIdeasySnapshot(String groupId, String artifactId, String version) {
+
+    return IDEASY_GROUP_ID.equals(groupId) && IDEASY_ARTIFACT_ID.equals(artifactId) && version != null
+        && version.contains("SNAPSHOT");
+  }
```
hohwille, line 145: As stated above
```suggestion
  private boolean isSnapshotVersion(String version) {

    return version.endsWith("-SNAPSHOT");
  }
```

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
```diff
@@ -0,0 +1,158 @@
...
+  @Override
+  public String getId() {
+
+    return "maven";
+  }
+
+  @Override
+  public Collection<ToolDependency> findDependencies(String groupId, String artifactId, VersionIdentifier version) {
+
+    return Collections.emptyList();
```
hohwille, line 156:
```suggestion
    // We could read POM here and find dependencies but we do not want to reimplement maven here.
    // For our use-case we only download bundled packages from maven central so we do KISS for now.
    return Collections.emptyList();
```

hohwille (changes requested):
@salimbouch thanks for your PR. Very nice job solving the complex things of this story 👍 
I left some review comments for improvement.
Please have a look.

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
salimbouch: Now `getMetadata` behaves as you stated. However, I still can't really get rid of this regex. The passed version to the method `getMetadata` if the version is a snapshot looks like this: `2025.01.001-beta-20250121.023134-9`.
We need to make a path out of it, that means we take that version and make `2025.01.001-beta-SNAPSHOT` out of it.

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
salimbouch: done, now passing the classifier as arg to the `download` method.

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
salimbouch: done, not really sure if I used the right libraries for it.

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
salimbouch: Since this method isn't an API (this method is only being used by the method `download`, which retrieves the download link and downloads the tool), I added a param for extension and a param for a classifier to the `download` method, so that when we want to download a tool and want to provide an extension and a classifier, we can do that directly through the download method. This caused a couple more changes to the upper hierarchy.

salimbouch:
Also changed the way I retrieve the latest version from `metadata.xml`. Have a look at [this](https://s01.oss.sonatype.org/content/repositories/snapshots/com/devonfw/tools/IDEasy/ide-cli/maven-metadata.xml).
The `latest` element doesn't give the real latest snapshot version. Now I iterate over all `version` elements and get the latest.

salimbouch:
Added a test to `MavenRepository`

cli/src/main/java/com/devonfw/tools/ide/repo/AbstractToolRepository.java
```diff
@@ -42,12 +42,19 @@ public AbstractToolRepository(IdeContext context) {
    * @param version the {@link VersionIdentifier} to download.
    * @return the resolved {@link UrlDownloadFileMetadata}.
    */
-  protected abstract UrlDownloadFileMetadata getMetadata(String tool, String edition, VersionIdentifier version);
+  abstract protected UrlDownloadFileMetadata getMetadata(String tool, String edition, VersionIdentifier version);
```
hohwille, line 45: any reason for inverting the modifiers?
Best practice is to always have `abstract` modifier after visibility.
https://www.baeldung.com/java-access-modifiers#bd-canonical-order-of-modifiers

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeCommandlet.java
```diff
@@ -0,0 +1,179 @@
...
+      throw new IllegalStateException("Failed to compare current and latest Snapshot.", e);
+    }
+  }
+
+  @Override
+  public void run() {
+
+    ToolRepository mavenRepo = this.context.getMavenSoftwareRepository();
+    VersionIdentifier currentVersion = VersionIdentifier.of(IdeVersion.get());
+    VersionIdentifier latestVersion = mavenRepo.resolveVersion(IDEASY_GROUP_ID, IDEASY_ARTIFACT_ID, currentVersion);
```
hohwille, line 119: why does `resolveVersion` give us the latest version?
According to my understanding it will resolve version patterns or SNAPSHOTs to the latest matching version number.
If we put `2024.01.001` in as `currentVersion`, then according to the JavaDoc you should also get `2024.01.001` what is not the `latestVersion`.
>    @return the resolved {@link VersionIdentifier}. If the given {@link VersionIdentifier} is NOT a {@link VersionIdentifier#isPattern() pattern} this method
     will always just return the given {@link VersionIdentifier}.

hohwille:
IMHO we need to get together and talk (and I mean live in a video call rather than chatting here):

* We are on a conflict collision course and should align and prepare that
* I did not thing about it in the first place but `ToolRepository` is taking arguments like `tool` and `edition` while a maven repository takes arguments like `groupId` and `artifactId`. Just Considerung the `tool` as `groupId` and the `edition` as `artifactId` is causing a lot of confusion.

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
```diff
@@ -0,0 +1,174 @@
...
+
+  private static final String MAVEN_METADATA_XML = "maven-metadata.xml";
+
+  private final DocumentBuilder documentBuilder;
+
+  private static final Map<String, String> TOOL_TO_GROUP_ID = Map.of(
+
+      "ideasy", "com.devonfw.tools.IDEasy",
+      "cobigen", "com.devonfw.cobigen"
+  );
```
salimbouch, line 47: This is a mapping from `tool` to `groupId`, you also mentioned a mapping from `edition` to `artifactId` but I don't get the point of it, like if we were to download a maven tool we could do `download("ideasy", "ide-cli", ...)`, it makes sense to map `ideasy` to `com.devonfw.tools.IDEasy` but what would we map "ide-cli" to?

cli/src/main/java/com/devonfw/tools/ide/repo/MavenRepository.java
```diff
@@ -0,0 +1,174 @@
...
+    List<VersionIdentifier> versions = fetchVersions(metadataUrl);
+    // beta versions of maven central are also considered unstable and "*" doesn't match them
+    VersionIdentifier resolvedVersion = this.context.getUrls().resolveVersionPattern(version, versions);
+    if (resolvedVersion.toString().endsWith("-SNAPSHOT")) {
+      String snapshotMetadataUrl = baseUrl + "/" + groupId.replace('.', '/') + "/" +
+          artifactId + "/" + resolvedVersion + "/" + MAVEN_METADATA_XML;
+      return resolveSnapshotVersion(snapshotMetadataUrl, resolvedVersion.toString());
+    }
+    return resolvedVersion;
+  }
```
salimbouch: Im using `this.context().getUrls()` just to get access to UrlMetadata, probably wrong. Plus we are supposed to pass the configured version to be resolved, the thing is maven central versions of `ideasy` are beta versions, that means `"*"` doesn't match any of those

cli/src/main/java/com/devonfw/tools/ide/repo/AbstractToolRepository.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeCommandlet.java
```diff
@@ -0,0 +1,175 @@
...
+   * "2024.12.002-beta-12_18_02-SNAPSHOT" - Latest version format: "2025.01.001-beta-20250118.022832-8"
+   * <p>
+   * First compares base versions (e.g. 2024.12.002 with 2025.01.001), then timestamps if base versions are equal. Returns false if version formats are
+   * unexpected to avoid unintended upgrades.
+   *
+   * @param currentVersion The current snapshot version
+   * @param latestVersion The latest snapshot version to compare against
+   * @return true if latestVersion is newer than currentVersion, false otherwise or if formats are invalid
+   */
+  protected boolean isSnapshotNewer(String currentVersion, String latestVersion) {
```
hohwille, line 55: I am not getting why we need this implementation.
Why can't we just use `VersionIdentifier` that already implements version comparison (greater/less)?

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeCommandlet.java
```diff
@@ -0,0 +1,175 @@
...
+    ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.LOG_WARNING).executable(scriptPath).addArgs(downloadTarget, extractionTarget);
+    pc.run(ProcessMode.BACKGROUND_SILENT);
+    this.context.success("Upgrade process has been initiated.");
+  }
+
+  private Path createUpgradeScriptForWindows() throws IOException {
+
+    Path scriptPath = this.context.getIdeRoot().resolve("upgrade.bat");
+    String scriptContent = "@echo off\n" +
+        "ping -n 4 127.0.0.1 > nul\n" +
```
hohwille, line 169: what is the `ping` command good for?
Some delay like a `sleep` command?

salimbouch: Yes exactly

hohwille:
@salimbouch thanks for your rework. This is great progress 👍 
I will take over from here since we need to have this merged today EOD if we want to get it included into our release tomorrow.

hohwille:
Replaced by PR #1002

PR #965, opened 2025-01-21
#935: followup - remove implicit status invocation

hohwille:
follow up fix of #935

do not make `ide` command without args call `ideasy status`.
User shall explicitly call `ide status` to get the status to avoid undesired log "spam".


jan-vcapgemini (approved):
LGTM. Ready for merge.

Issue #898, opened 2024-12-13
Improved output of get-version/edition and uninstall/-plugin


WorkingAmeise:
As a IDEasy User, I want understandable and descriptive feedback messages when the software performs tasks

### get-edition / get-version

- [x] calling with commandlets, that are not installed atm should result in an info message (currently prints an empty line)

### uninstall

- [x] calling with commandlets, that are not installed atm currently only prints: "Couldn't uninstall docker". Should be enhanced by : "Because docker was not installed previously"

### uninstall-plugin

- [x] Should log any success/failure message (currently does not print anything)

hohwille:
For `get-edition` and `get-version` I still recommend that without any other arguments, this should print the configured value if the tool is not already installed.
Also it IMHO makes sense that if I do `ide set-version mvn 3.9.9` and then call `ide get-version mvn` I should actually get `3.9.9` what is a natural expectations (I get back what I just have set).
However, this will currently not work and is IMHO only causing confusion.
- [ ] So to fix this this issue should be implemented such that if the tool is not installed, the configured value is printed (because when I would install the tool now that version/edition is going to be installed).
- [ ] Additionally to make the `set-*` and `get-*` commandlets really symmetric, the `set-*` commandlets should additionally be changed such that if I override the edition of version not in `settings/ide.properties` but e.g. in `conf/ide.properties` then the `set-version`/`set-edition` commandlet will actually change the configured value there (unless explicitly configured via e.g. `--cfg=settings`).

Finally, we must be aware that it is also possible to install a tool and afterwards configuring the version/edition otherwise.
In such case it will be correct that after `ide set-version mvn 3.9.9` and then call `ide get-version mvn` we get the previously installed version of maven.
The user has the following options:
```
ide -d get-version mvn`
```
This will print both the installed and the configured version including additional debug output.
```
ide get-version --configured --installed mvn`
```
This will print both the installed and the configured version without additional debug output.
Please also note that if configured and installed version/edition are identical, then still only one version/edition is printed without further details.
Also I can explicitly request only the configured (`--configured`)  or the installed (`--installed`) version by providing only one of these two long options.
And for the record:
The current implementation of these commandlets is quite buggy:
```
$ ide -d get-version mvn
Running commandlet VersionGetCommandlet[get-version]
Configuration directory C:\Users\hohwille\.ide does not exist.
The installed version for tool mvn is 3.9.9
The configured version for tool mvn is 4.0.0-rc-1
To install that version call the following command:
[...]

PR #967, opened 2025-01-21
#898: adjusted tool name in error message

jan-vcapgemini:
Follow up of #898 

### Implements:
* adjusted tool name in error message


hohwille (approved):
@jan-vcapgemini thanks for the fast fix. 👍

Issue #420, opened 2024-06-25
Create installer for Windows


hohwille:
As a IDEasy user, I want to install the product via a graphical installer so that I can easily be guided through the installtion with only few clicks.

## Acceptance criteria
* Installer first screen shows the license of IDEasy with scrollbar and the user can read it and either click `Cancel` or `Accept & Continue`. `Cancel` aborts the installation and `Accept & Continue` proceeds to the second screen.
* Installer shows second screen to choose the installation location (`$IDE_ROOT`) with and `Install` button at the bottom and above a description with the following options with radio-button selection .
  * C:\projects (pre-selected default)
  * D:\projects (if D: drive is present, a permanent disc like HDD or SSD and has at least 1GB free disc space)
  * custom with ability to browse to a custom folder. If this option is selected the `Install` button gets disabled (grayed out), the custom folder has to be chosen and the path must exist and not contain spaces or special characters (only `[a-zA-Z0-9]+` for folders). Also the path must not be longer than 32 characters. Only if all these criteria are met the `Install` button gets enabled.
* When the user has clicked on `Install` the installation is started:
  * Contents are installed to `$IDE_ROOT` (so we have `$IDE_ROOT/_ide/bin/ideasy.exe`, etc.)
  * Environment variable (user) `PATH` is extended with value of `;$IDE_ROOT\_ide\bin`
  * Environment variable (user) `IDE_ROOT` is set in Windows to value of `$IDE_ROOT`.
  * The file `%USERPROFILE%\.bashrc` is created if not existing and we append the line `alias ide="source $IDE_ROOT/_ide/bin/ide"` where `$IDE_ROOT` is the resolved path in MSYS syntax (e.g. `/c/projects` instead of `C:\projects`) followed by the line `source $IDE_ROOT/_ide/completion` and finally the line `ide`.
  * Registry entries are added for Explorer integration
* The installer features uninstallation via standard Windows Settings that reverts all changes made during the installation.

For the registry entries see old devonfw-ide sources:
* https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/system/windows/windows-terminal/devon-wt-gitbash.reg
* https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/system/windows/windows-terminal/devon-wt-cmd.reg
* https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/system/windows/windows-terminal/devon-wt-ps.reg
[...]

PR #978, opened 2025-01-22
#420: Create Windows Installer with WixToolset

alfeilex:
Fixes: #420 

This PR implements the creation of a MSI file for Windows Installer with WixToolset

It includes:
- The Windows Installer Build Package with UI Sequence:
  - Accept License
  - Continue with default installation (default `C:\projects`) or customize
    - Volume Drive
    - Root Folder or User Home Folder
  - Copy files into installation folder and run `ideasy.exe -f install`
 - The integration of the WixToolSet build logic into the `Nightly Build` and `Release` workflow


alfeilex:
I would like to start a discussion about the Windows Installer. In the current state, our UI is based on an extension thats provide the most important UI dialogs, such as the License agreement, installation path selection and when running the MSI again, the options to either repair the tool or remove it completely. It also includes a simple path validation, though not as advanced as described in the issue. 

So when we would like to integrate some custom features such as an advanced validation or space capacity of 1GB, we need to create our own UI dialogs. My suggestion would be to initially stick with the extension UI and focus more on the installation and CI itself.

build-msi/IDEASY.wxs
```diff
@@ -0,0 +1,54 @@
+<?define name = "IDEasy"?>
+<?define version = "1.0.0"?>
+<?define publisher = "devonfw-ide"?>
+<?define guid = "ba3ad5c0-94e9-42c3-9374-efaaddd3b967"?>
+
```
alfeilex, line 5: We should also define the correct meta information for the MSI packages

build-msi/IDEASY.wxs
```diff
@@ -0,0 +1,54 @@
...
+    Version="$(var.version)"
+    UpgradeCode="$(var.guid)"
+    Scope="perMachine">
+
+    <!-- Specifying a license file -->
+    <WixVariable Id="WixUILicenseRtf" Value=".\assets\LICENSE.rtf" />
+    <!-- Replacing the default bitmaps -->
+    <WixVariable Id="WixUIBannerBmp" Value=".\assets\banner.bmp" />
+    <WixVariable Id="WixUIDialogBmp" Value=".\assets\dialog.bmp" />
+
```
alfeilex, line 18: I have created some dummy bitmap files to beautify the UI dialogs

build-msi/IDEASY.wxs
```diff
@@ -0,0 +1,54 @@
...
+    <!-- Specifying default directory-->
+    <Directory Id="ROOTDIRECTORY">
+      <Directory Id="INSTALLFOLDER" Name="projects" />  
+    </Directory>
+
+    <!-- Specifying custom actions-->
+    <CustomAction Id="RunSetupAction" Directory="INSTALLFOLDER" ExeCommand="[INSTALLFOLDER]_ide\setup.bat -b" Execute="deferred" Return="check" />
+    <CustomAction Id="RunUninstallAction" Directory="INSTALLFOLDER" ExeCommand="[INSTALLFOLDER]_ide\ideeasy uninstall" Execute="deferred" Return="ignore" />
+
```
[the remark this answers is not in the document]
alfeilex: After much trial and error, I finally managed to integrate the `WixQuietExec` custom action to run scripts silently. So now, we could stick to the idea of creating Windows Installer related scripts for` installation`, `uninstallation` and `repair` mode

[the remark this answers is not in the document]
alfeilex: The `setup.bat` script in the MSI installation adds `C:\projects\_ide\bin`  in the `PATH` in each installation run even its already there

alfeilex:
How we can convert `LICENSE.adoc` in `LICENSE.rtf` ?

alfeilex:
Current State of Dialogs: 

Welcome Dialog:
image

LicenseAgreement Dialog:
image

Installation Directory Dialog:
image

Agree Installation Dialog:
image

End of Installation Dialog:
image

Change, Repair Dialog:
image

build-msi/IDEASY.wxs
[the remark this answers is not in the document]
alfeilex: Fixed in the latest Snapshot

alfeilex:
The CI/CD build is turning out to be trickier than I initially thought. For the MSI installation, we need to specify which files should be included in the installation. Since the entire IDEasy tar.gz is bundled in the deployment, we need a step before that so that the Wix Toolset can build the MSI with the content from the tar.gz. Maybe I'm missing something here, so I am looking for feedback.

Here are the commands that have to be executed to built the MSI with WixToolset:

```
- name: Build MSI for Windows
        run: |
          dotnet tool install --global wix --version 5.0.2
          cd build-msi
          wix extension add WixToolset.UI.wixext/5.0.2
          wix extension add WixToolset.Util.wixext/5.0.2
          wix build Package.wxs WixUI_IDEasySetup.wxs -loc Package.en-us.wxl -ext WixToolset.UI.wixext -ext WixToolset.Util.wixext -o ideasy.msi
        if: runner.os == 'Windows'
```

hohwille:
> How we should adapt the CI/CD workflow to integrate the MSI build?

One thing is that on the windows build (in matrix os), we need to build the MSI.
Further in maven (see [here](https://github.com/devonfw/IDEasy/blob/main/cli/pom.xml#L190)), we need to attach the artifact to the release being deployed. This does not make sense via assembly-plugin since we already have the artifact that is the MSI itself. Instead use [attach-artifact](https://www.mojohaus.org/build-helper-maven-plugin/attach-artifact-mojo.html).

hohwille:
>  Should we hide the uninstallation dialog until the uninstall command is implemented? Right now, the entire projects folder is removed if the user clicks on Uninstall. The Repair button doesn't trigger anything at all. So in my opinion, we should just hide it for now.

You could already call `ideasy uninstall`. It is easy to implement this feature and I would like to have it in the next release.
If users install IDEasy soon via MSI and they want to uninstall it in the future (maybe in 5 years) the functionality would work fine then assuming that until then the user installed an update of IDEasy containing the `uninstall` feature.
If for the moment this would lead to an error since `uninstall` is not implement, I would just consider this as a temporary known bug.

hohwille:
> The Docbook to RTF integration isn’t straightforward. I’ve created a new issue https://github.com/devonfw/IDEasy/issues/1056 for that. For now, let's stick to a statically generated LICENSE.rtf

Perfectly fine. Thanks.

hohwille:
> Although a publisher is set in the Package.wxs, it says "Unknown Publisher" in UAC. We have to sign it probably https://stackoverflow.com/questions/17589754/wix-installer-msi-publisher-unknown

Simply create a new issue for that. For the first release "Unknown Publisher" is acceptable. We can fix it later.

hohwille:
> Should we move build-msi along with the WixToolset data into cli folder or perhaps another folder?

I guess this is not a maven project but just the artifacts (config, etc.) needed for the MSI installer.
Then this does not matter so much but for me it would also be fine to have this as a top-level folder.
Maybe you can rename it to just `msi` or `windows-installer`?
Also the `cli` folder contains logic to "build" the CLI and we do not name it `build-cli`.

alfeilex:
> One thing is that on the windows build (in matrix os), we need to build the MSI.
> Further in maven (see [here](https://github.com/devonfw/IDEasy/blob/main/cli/pom.xml#L190)), we need to attach the artifact to the release being deployed. This does not make sense via assembly-plugin since we already have the artifact that is the MSI itself. Instead use [attach-artifact](https://www.mojohaus.org/build-helper-maven-plugin/attach-artifact-mojo.html).

But as I understand it, in the current workflow, we build the native image and include it in the deployment for the release. In the MSI build itself, we also need the whole package, including the binary `ideasy.exe` and also documentation, which is not built before `mvn deploy`.

alfeilex:
> @alfeilex now that my PR #1055 is merged, your MSI installer only needs to call `ideasy.exe -f install`:
> 
> https://github.com/devonfw/IDEasy/blob/4caff461795e245daeaf9a5137a488eafe05eb9e/cli/src/main/package/setup#L3

Is added but has to be tested when https://github.com/devonfw/IDEasy/issues/1084 is merged.

.github/workflows/nightly-build.yml
```diff
@@ -52,10 +73,9 @@ jobs:
       - name: Download natives and build project
         uses: actions/download-artifact@v4
         with:
-          pattern: natives-*
```
alfeilex, line 55: I removed the pattern key here, because we use all uploaded artifacts anyway. Since we also include the msi artifact, we would have to check how to extend the pattern here in the case, we want to stick with it.

.github/workflows/nightly-build.yml
```diff
@@ -34,6 +34,27 @@ jobs:
         with:
           name: natives-${{ matrix.os }}
           path: cli/target/ideasy*
+      - name: Build MSI with WixToolSet
```
alfeilex, line 37: I tried to integrate the whole MSI installation into maven, but the plugins are not compatbile with the latest version 5. https://wix-maven.github.io/wix-maven-plugin/

So I chose this place to insert the Wixtoolset installation. I would also be possible to move this to the last `deploy` job, but there we are using `ubuntu` as a runner and `dotnet` would not work properly. 

I think we could also move the whole thing into a separate Github Action. But maybe, we could create a new issue for that.

If you have any suggestions, than please comment on this

.github/workflows/nightly-build.yml
```diff
@@ -52,10 +73,9 @@ jobs:
...
         uses: actions/download-artifact@v4
         with:
-          pattern: natives-*
           path: ./cli/target/
       - name: Deploy to OSSRH nexus
         env:
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
           SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
-        run: mvn --settings .mvn/settings.xml -DskipTests=true -Darchetype.test.skip=true -Dgpg.skip=true -Dstyle.color=always -B -ntp -Passembly,deploy deploy
+        run: mvn --settings .mvn/settings.xml -DskipTests=true -Darchetype.test.skip=true -Dgpg.skip=true -Dstyle.color=always -B -ntp -Passembly,msi,deploy deploy
```
alfeilex, line 81: It looks like the MSI is attached into the deployment. I haven't change the `release` workflow yet. Maybe we can first run the Nightly Build and then adapt the changes to the release workflow

https://github.com/alfeilex/IDEasy/actions/runs/13546792586/job/37860853837#step:5:1240

cli/pom.xml
```diff
@@ -250,13 +251,60 @@
...
+            <version>${build.helper.maven.plugin.version}</version>
+            <executions>
+              <execution>
+                <goals>
+                  <goal>attach-artifact</goal>
+                </goals>
+                <configuration>
+                  <artifacts>
+                    <artifact>
+                      <file>${project.basedir}/target/${project.artifactId}-${revision}.msi</file>
```
alfeilex, line 297: Right now, its named `ide-cli-2025.02.002-beta.msi`

If you change the name or the `artifact-id,` then you have also to change line 279 because I rename the downloaded `ideasy.msi` there

hohwille (approved):
@alfeilex thanks for this PR and finding a solution to this tricky story 👍 
I checked the following aspects:
* installation is happening via new approach (`ideasy -f install`)
* the installation is triggered from the correct directory (workaround for bug #1084)
* uninstallation is happening correctly (`ideasy -f uninstall`)
* license agreement is implemented nicely with the proper license (via RTF and follow up issue #1056 was created)
* what I am unsure is if the user can pick the installation location and that will be honored by the installation (I was expecting that IDE_ROOT would be set according to the users choice before `ideasy -f install` is called)
* all is integrated into nightly build and looks sane to me. However, I do not immediately understand if the native image will be included in the MSI.
* Release download link to MSI was added
* CHANGELOG was adopted

So IMHO this is already in a great state, even though it is not entirely completed.
I would therefore merge it as is and test it with a nightly build, download MSI and test it.
Then we can improve and complete in a follow up PR.

cli/pom.xml
hohwille: Fine for now. I will resolve. Thanks for clarification.

PR #983, opened 2025-01-23
#931: enhance settings in code repository

### LINKED ISSUE #931: Enhance settings in code repository
This is a follow up story of #757:
In PR #850 there were some remaining discussions left:

* `Missing .git folder in D:\projects\code-demo\settings.` - can we prevent this warning since it is normal in this case?
* Should we skip `git pull` on the code repo on `ide update`? Or otherwise at least add an option like `ide update --skip-pull`? I would actually suggest to skip the `pull` but render a warning if the git repo is not up-to-date but let the user in control of pulling and potential conflict resolution. We could also honor `--force` flag to trigger the update so `ide -f update` would still do the git pull but fail in case of merge conflicts.

salimbouch:
fixes #931 

- No more `Missing .git folder in D:\projects\code-demo\settings.`
- When the settings repository is a symlink, `ide update` skips the pull operation. When force mode is active, the pull operation is still performed.


cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -429,15 +431,21 @@ public Path getSettingsGitRepository() {
...
       error("Settings repository exists but is not a git repository.");
       return null;
     }
 
     return settingsPath;
   }
 
+  public boolean isSettingsRepositorySymlink() {
+
+    Path settingsPath = getSettingsPath();
```
hohwille, line 445: If we are not inside an IDEasy project, IDE_HOME and all related paths are null.
```suggestion
    Path settingsPath = getSettingsPath();
    if (settingsPath == null) {
      return false;
    }
```

cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
```diff
@@ -405,6 +405,11 @@ default Path getRepositoriesPath() {
    */
   Path getSettingsGitRepository();
 
+  /**
+   * @return {@code true} if the settings repository is a symlink or a junction.
+   */
+  boolean isSettingsRepositorySymlink();
```
jan-vcapgemini, line 411: Maybe change the name to `isSettingsRepositorySymlinkOrJunction` too.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -280,11 +281,13 @@ private String getMessageIdeHomeNotFound() {
...
 
   private String getMessageIdeRootNotFound() {
+
     String root = getSystem().getEnv("IDE_ROOT");
     if (root == null) {
       return "The environment variable IDE_ROOT is undefined. Please reinstall IDEasy or manually repair IDE_ROOT variable.";
     } else {
-      return "The environment variable IDE_ROOT is pointing to an invalid path " + root + ". Please reinstall IDEasy or manually repair IDE_ROOT variable.";
+      return "The environment variable IDE_ROOT is pointing to an invalid path " + root
+          + ". Please reinstall IDEasy or manually repair IDE_ROOT variable.";
```
hohwille, line 290: We have changed our formatter to 160 characters per line for line-wrapping:
https://github.com/devonfw/IDEasy/blob/098604ee24fdf25cc1926890a3fb286065cabb96/.editorconfig#L9
It seems this config did not arrive in your IntelliJ.
Do you maybe still have the legacy eclipse formatter plugin active in IntelliJ?
If so, please remove it asap.

Additionally check if you get a log message like this from IDEasy (e.g. call help or even better type `ide status`):
```
```
Then just follow this suggestion and call `ide update`.

hohwille (approved):
@salimbouch thanks for your PR. You did a great job at got everything right 👍 
I added one small suggestion for improvement and another review comment to avoid further diff-wars. Please have a look and update. Then ready for merge.

jan-vcapgemini (approved):
LGTM.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
salimbouch: done

cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
jan-vcapgemini: @salimbouch, thanks for adding my suggestion, would have been nice if you could have answered in this CR though that the change was done.

PR #990, opened 2025-01-27
#954: improve repository support

### LINKED ISSUE #954: Improve repository support
As a IDEasy user, I want  to be able to push out a new (required) repositories to my teams with a simple push to my settings git repo so that everybody in my team can automatically get this new repositories cloned.
Further, I still want developers to have freedom and flexibility, so if for whatever reason they think that they do not need a specific repository, they are not fooled by IDEasy that recreates and clones the same repo after the user deleted it.

Therefore, we plan to introduce to track the repository status in local file(s) (e.g. `$IDE_HOME/.repository/«repository»`) so we know that we have done our job of cloning the according repository on behalf of `$IDE_HOME/settings/repositories/«repository».properties`.
This way IDEasy can check after `ide update` if a `$IDE_HOME/settings/repositories/«repository».properties` exists with `active=true` that does not have a corresponding `$IDE_HOME/.repository/«repository»`. In such case, the repository will be cloned on behalf of `ide update`.
Now after that the user could decide to remove `$IDE_HOME/workspaces/«workspace»/«repository»` manually and `ide update` will know that is was already cloned before so it will not re-create it.
It needs to be clarified if `ide create «project-name» --skip-repositories ...` will automatically mark all active repositories as "cloned" even though they are not. I would currently vote against it but this should still be discussed. Also we should consider supporting `ide update --skip-repositories`.

Instead of individual status files we also might maintain just a single status file `$IDE_HOME/.repositories.json` instead.
In both cases, we should track the timestamp when the repository was cloned and also in case of an error from `git clone` also the error message.
But in case of an error, we need to check if git actually created the repository at all.
If the error was e.g. that the user did not have permission to clone or was unauthenticated and failed 3 times to properly enter his password, we should most probably not consider the repository to be cloned at all in our status so it will be retried in the next `ide update` command.

hohwille:
fixes #954 

* track status which repos have already been setup
* add steps to project import and proper error handling
* improve step support
* implement project import that was missing
* add MvnArtifact
* add Mvn.downloadArtifact, etc.
* package refactoring to avoid confusion between git-repository and tool-repository
* etc.


cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
```diff
@@ -0,0 +1,110 @@
...
+   * The constructor.
+   *
+   * @param file the {@link Path} to the properties file.
+   * @param context the {@link IdeContext}.
+   */
+  public RepositoryProperties(Path file, IdeContext context) {
+    this(file, context, context.getFileAccess().readProperties(file));
+  }
+
+  RepositoryProperties(Path file, IdeContext context, Properties properties) {
```
jan-vcapgemini, line 33: Please add Javadoc, not clear why it is being done like this.

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
```diff
@@ -0,0 +1,110 @@
...
+  public String getProperty(String name) {
+    return getProperty(name, false);
+  }
+
+  /**
+   * @param name the name of the requested property.
+   * @param required - {@code true} if the requested property is required, {@code false} otherwise.
+   * @return the value of the requested property or {@code null} if undefined.
+   */
+  public String getProperty(String name, boolean required) {
```
jan-vcapgemini, line 53: Please rename to `retrieveProperty`.

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
```diff
@@ -0,0 +1,110 @@
...
+    if (value != null) {
+      this.context.warning("The properties file {} uses the legacy property {} instead of {}", this.file, legacyName, name);
+    }
+    return value;
+  }
+
+  /**
+   * @return the IDEs where to import the repository.
+   */
+  public Set<String> getImports() {
```
jan-vcapgemini, line 92: Maybe add some debug log messages to this method.

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java
```diff
@@ -0,0 +1,189 @@
...
+      if (!this.context.isForceMode()) {
+        this.context.info("Ignoring repository {} because it was already setup before - use --force for recreation.", repository);
+        return;
+      }
+    }
+    boolean success = cloneOrPullRepository(repositoryPath, gitUrl, repositoryCreatedStatusFile);
+    if (success) {
+      buildRepository(repositoryConfig, repositoryPath);
+      importRepository(repositoryConfig, repositoryPath, repositoryId);
+    }
```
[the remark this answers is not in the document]
hohwille: Sorry, but this suggestion does not make sense.
Line 116-120 have nothing to do with `checkStatusFile` and a `return` statement in the middle of some logic cannot be extracted to a private method without redesigning the logic.

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
hohwille: Arent we going to far with this JavaDoc everywhere?
This constructor is neither `public` nor `protected` - but if it makes you happy, I will formally add this JavaDoc even though we have the same constructor above with its parameters documented already.

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
hohwille: Why is it wrong to name this method `getProperty`?
It is delegating to `java.util.Properties.getProperty(String)`.
Do you also think that this method is named wrong?
I do not see anything wrong with this name.
There is also the same method `getProperty` without the `boolean` flag - why did you not request that this method shall also be renamed?
I would disagree and since it is after midnight and I want to complete the release, I will now set this to resolved.

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
hohwille: Done.

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
hohwille: Done.

cli/src/main/java/com/devonfw/tools/ide/tool/eclipse/Eclipse.java
```diff
@@ -18,13 +18,21 @@
...
 import com.devonfw.tools.ide.tool.plugin.ToolPluginDescriptor;
 
 /**
  * {@link IdeToolCommandlet} for <a href="https://www.eclipse.org/">Eclipse</a>.
  */
 public class Eclipse extends IdeToolCommandlet {
 
+  // version must correspond to eclipse-import.xml
+  private static final String GROOVY_VERSION = "3.0.23";
+  public static final String VMARGS = "-vmargs";
```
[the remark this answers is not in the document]
hohwille: Done.

PR #994, opened 2025-01-28
#993: create IDE start scripts

### LINKED ISSUE #993: Creation of start scripts for IDEs
As a IDEasy user, I want to have start scripts generated for my IDEs so that I can just double click them to launch the IDE instead of opening a terminal and entering `ide intellij` or `ide eclipse`,...

This was a standard feature of devonfw-ide but seems to be missed out in IDEasy.
We do have the variable with defaults in our code:
https://github.com/devonfw/IDEasy/blob/05af7d1fc8ce55f663afd74de5c7cc16055fc4fe/cli/src/main/java/com/devonfw/tools/ide/variable/IdeVariables.java#L37

However, it is not used anywhere.
It should be honoured by both `ide create ...` as well as by `ide update`.

# Notice

This feature is not to be confused with topics like #366 #540 #832 that also use the term `start script for IDE/tool` but mean a script called `intellij` inside the `software/intellij/bin` folder to avoid that on windows we need to run `idea64.exe` and on linux `idea.sh` and on Mac `idea` - these are flaws from IDE tool vendors but in the end we gave up on this idea since we also need some backward compatibility with devonfw-ide.
This story here is talking about scripts inside `$IDE_HOME` like `eclipse-main[.bat]` or `intellij-main[.bat]` that I can double click without the need to open a bash terminal and entering CLI commands to launch my IDE in daily business.

hohwille:
fixes #993

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -2,6 +2,14 @@
 
 This file documents all notable changes to https://github.com/devonfw/IDEasy[IDEasy].
 
+== 2025.01.003
+
+Release with new features and bugfixes:
+
+* https://github.com/devonfw/IDEasy/issues/993[#993] Creation of start scripts for IDEs
+
+The full list of changes for this release can be found in https://github.com/devonfw/IDEasy/milestone/20?closed=1[milestone 2025.01.003].
+
 == 2025.01.002
 
 Release with important bugfixes:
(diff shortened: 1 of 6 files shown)


jan-vcapgemini (approved):
LGTM. Tested it locally in intellij with `CREATE_START_SCRIPTS=intellij` and worked as expected.

Issue #998, opened 2025-01-30
Introduce dependencies.json for npm and remove hardcoded dependency


jan-vcapgemini:
Introduce #998 `dependencies.json` for `npm` and remove hardcoded dependency: https://github.com/devonfw/IDEasy/blob/4f2e37bf602a341666d12b62b09b633cbda05487/cli/src/main/java/com/devonfw/tools/ide/tool/npm/Npm.java#L28-L33

Sources of information about the dependencies and version compatibility:
* https://nodejs.org/en/about/previous-releases#looking-for-the-latest-release-of-a-version-branch
* https://registry.npmjs.org/npm/
Most valuable source from @yurii-yu was:
* https://www.npmjs.com/package/npm/v/10.0.0
* https://www.npmjs.com/package/npm/v/11.0.0
* ...

yurii-yu:
Hello. What should I do if there are more than 1 options for the dependent software?
for example:

for [NPM 9.9.4](https://www.npmjs.com/package/npm/v/9.9.4)

> ### Requirements
> One of the following versions of [Node.js](https://nodejs.org/en/download/) must be installed to run npm:
> 14.x.x >= 14.17.0
> 16.x.x >= 16.13.0
> 18.0.0 or higher

as far as I know, currently our json file support only one option. is that right?

PR #1015, opened 2025-02-04
998: replace npm hardcoded dependency

yurii-yu:
#998 
introduce `dependencies.json` to replace the hardcoded dependency installation


yurii-yu:
I close this pull request because some changes for maven is involved
I will create another "pure" pull request later

PR #1017, opened 2025-02-04
#404: enhance logging with 

jan-vcapgemini:
Fixes: #404

Continues PR: #760 by @KianRolf 

### Implements:
* updated WireMock dependency from 2.35.2 to 3.11.0
* added jackson-core dependency to ide-cli
* added url namespace to tools
* added LogbackServiceProvider to service
* added IdeLoggerAdapter (Factory class for slf4j)


yurii-yu:
CI failed because of concurrent modification in `AndroidStudioTest.checkInstallation`
maybe we should first make the CI work :)

cli/src/main/java/com/devonfw/tools/ide/log/AbstractIdeSubLogger.java
yurii-yu, line 1: maybe we need to re-format the code?

jan-vcapgemini: Had to enforce the removed using CTRL+SHIFT+S.
Adjusted now.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -180,7 +183,7 @@ public AbstractIdeContext(IdeStartContextImpl startContext, Path workingDirector
       if ((ideRootPath != null) && (ideRootPath.equals(currentDir))) {
         // prevent that during tests we traverse to the real IDE project of IDEasy developer
         currentDir = null;
-      }
+    }
```
hohwille, line 186: what is happening here? Is your IDE/formatter broken?

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -138,6 +138,9 @@ public abstract class AbstractIdeContext implements IdeContext {
 
   private WindowsHelper windowsHelper;
 
+  /** Context used for logging */
+  public static IdeContext loggingContext;
```
hohwille, line 142: please avoid public static (non-final + immutable) members.
Every class can now set this context.
Make this private and allow read-access via public static getter but avoid public write access.

cli/src/main/java/com/devonfw/tools/ide/cli/Ideasy.java
```diff
@@ -19,6 +22,8 @@ public final class Ideasy {
 
   private AbstractIdeContext context;
 
+  private static final Logger logger = LoggerFactory.getLogger(Ideasy.class);
```
hohwille, line 25: I do not see that this is used anywhere. What is the value then?
Leftover from local dev test?

hohwille:
Comming to the really interesting questions for the review:
* can't we remove `logback-classic` now from CLI (eliminating lots of code from CLI native image)?
* shouldn't we distinguish between our own and 3rd party loggers at least in `IdeTestLogger`?
* couldn't we somehow use `Marker` API to express custom log-levels like success or interaction (can be handled outside this PR)?
* should we present and test this together with the team in the next KT/Planning session?

hohwille:
> shouldn't we distinguish between our own and 3rd party loggers at least in IdeTestLogger?

Ah, question was already answered by the CI build that fails:

> [ERROR] Errors: 
[ERROR]   IdeProgressBarTest.testProgressBarDownloadWithMissingContentLength:71 » ConcurrentModification

Yes, this is required to make it work.

cli/src/main/java/com/devonfw/tools/ide/cli/Ideasy.java
jan-vcapgemini: Jup, removed it now.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
jan-vcapgemini: Set it to private and added a getter.

hohwille:
Thanks for this PR 👍 
I took over this PR and replaced it by PR #1710.

PR #1035, opened 2025-02-13
#1034: fix VersionRange.contains for pattern

### LINKED ISSUE #1034: version of tool is * but does not match version to install
### Expected behavior

Only if a dependency version range is not matched by the configured version an additional version of the same tool should be installed to meet the requirements of the dependency.

### Actual behavior

```
Start: Install plugin anyedit
Configured version of tool java is * but does not match version to install [22,) - need to use different version from software repository.
Successfully ended step 'Install plugin anyedit'.
Start: Install plugin startexplorer
Configured version of tool java is * but does not match version to install [22,) - need to use different version from software repository.
Successfully ended step 'Install plugin startexplorer'.
Start: Install plugin checkstyle
Configured version of tool java is * but does not match version to install [22,) - need to use different version from software repository.
```

Additionally as we can see for each plugin installation the problem is repeated what might be considered as a different bug.
Expected would be that the tool (IntelliJ) and its dependencies (Java) is installed once and then used for additional plugin installations without repeating the same warning.

1. merge https://github.com/devonfw/ide-urls/pull/23 into `urls`
2. run `ide install intellij`

### Related/Dependent Issues

#1026 
#131 

### Comments/Hints:

It seems that version * does not match any version range.
First step is to add a test to reproduce the problem and then fix the bug to make the test green.

### Affected version:

* 2025.02.001-SNAPSHOT

hohwille:
fixes #1034

### DIFF
--- a/cli/src/test/java/com/devonfw/tools/ide/version/VersionRangeTest.java
+++ b/cli/src/test/java/com/devonfw/tools/ide/version/VersionRangeTest.java
@@ -108,10 +108,10 @@ public class VersionRangeTest extends Assertions {
 
     checkContainsNot("(11,22)", "10*");
     checkContains("(11,22)", "11*");
-    checkContains("(11,22)", "11.0*");
-    checkContains("(11,22)", "22*");
-    checkContains("(11,22)", "22.*");
+    checkContains("(11,22)", "11.0");
+    checkContainsNot("(11,22)", "22.*");
     checkContainsNot("(11,22)", "23*");
+    checkContains("[22,)", "*");
   }
 
   private void checkContains(String range, String version) {
--- a/cli/src/main/java/com/devonfw/tools/ide/version/VersionRange.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/version/VersionRange.java
@@ -71,19 +71,31 @@ public final class VersionRange implements Comparable<VersionRange>, GenericVers
   public boolean contains(VersionIdentifier version) {
 
     if (this.min != null) {
-      VersionComparisonResult compareMin = version.compareVersion(this.min);
-      if (compareMin.isLess()) {
-        return false;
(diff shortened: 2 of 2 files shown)


cli/src/test/java/com/devonfw/tools/ide/version/VersionRangeTest.java
```diff
@@ -108,10 +108,10 @@ public void testContains() {
 
     checkContainsNot("(11,22)", "10*");
     checkContains("(11,22)", "11*");
-    checkContains("(11,22)", "11.0*");
```
hohwille, line 111: this one is now red and that may be still a problem...

cli/src/test/java/com/devonfw/tools/ide/version/VersionRangeTest.java
```diff
@@ -108,10 +108,10 @@ public void testContains() {
...
-    checkContains("(11,22)", "22*");
-    checkContains("(11,22)", "22.*");
```
hohwille, line 113: This we wrong assumptions and assertions.
When implementing a test, we should not assert the behaviour of the implementation but think what the correct result should be and assert that.

hohwille: I am already setting this to resolved since it is fixed with this PR.
BTW: I found the author of this anti-pattern [here](https://github.com/devonfw/IDEasy/commit/bacd9ea001b21bcc788b8267a2c5bb6ebe68a25f#diff-d8d87bce3b2e2c262d5a17b746c91778551cfcb706c4de0ae95b755b68fffcca).
I have to blame myself. Shame on me ;)

hohwille:
Replaced by PR #1036 that properly fixes the problem.

PR #1036, opened 2025-02-14
#1034: proper fix of VersionRange.contains for version patterns

### LINKED ISSUE #1034: version of tool is * but does not match version to install

hohwille:
fixes #1034
replacement for PR #1035

### DIFF
--- a/cli/src/test/java/com/devonfw/tools/ide/version/VersionRangeTest.java
+++ b/cli/src/test/java/com/devonfw/tools/ide/version/VersionRangeTest.java
@@ -109,9 +109,12 @@ public class VersionRangeTest extends Assertions {
     checkContainsNot("(11,22)", "10*");
     checkContains("(11,22)", "11*");
     checkContains("(11,22)", "11.0*");
-    checkContains("(11,22)", "22*");
-    checkContains("(11,22)", "22.*");
+    checkContains("(11,22)", "21.99*");
+    checkContainsNot("(11,22)", "22*");
+    checkContainsNot("(11,22)", "22.*");
     checkContainsNot("(11,22)", "23*");
+    checkContains("(11,22)", "21.99*");
+    checkContains("[22,)", "*");
   }
 
   private void checkContains(String range, String version) {
--- a/cli/src/main/java/com/devonfw/tools/ide/version/VersionRange.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/version/VersionRange.java
@@ -70,6 +70,10 @@ public final class VersionRange implements Comparable<VersionRange>, GenericVers
   @Override
   public boolean contains(VersionIdentifier version) {
 
+    VersionSegment start = version.getStart();
+    if ((start.getNumber() == -1) && start.isPattern()) {
+      return true; // * and *! are always contained
+    }
     if (this.min != null) {
       VersionComparisonResult compareMin = version.compareVersion(this.min);
       if (compareMin.isLess()) {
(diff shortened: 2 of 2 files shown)


hohwille:
> checkContains("(11,22)", "11*");

From a PoV of perfection this is also questionable since `11*` could resolve as `11` and then the result is actually wrong since `11` is not in `(11,22)`.
However, from a pragmatic PoV we IMHO do not need to change this since I cannot see a perfect solution for this problem.
What we should however try to change is the usage of `VersionRange` and `VersionIdentifier` in the dependency management:
We should first resolve the configured version if it is a pattern. Then `11*` would either become `11` or `11.0.1` or whatever, and from there we can correctly evaluate if the range contains the version with 100% accuracy.
This PR however want to focus only on the obvious bugs in `VersionRange` and its test to keep the change minimal.

jan-vcapgemini (approved):
LGTM. Tested with `ide set-version gradle *`, before it was returning an error, now it works fine.

cli/src/test/java/com/devonfw/tools/ide/version/VersionRangeTest.java
```diff
@@ -109,9 +109,12 @@ public void testContains() {
     checkContainsNot("(11,22)", "10*");
     checkContains("(11,22)", "11*");
     checkContains("(11,22)", "11.0*");
-    checkContains("(11,22)", "22*");
-    checkContains("(11,22)", "22.*");
+    checkContains("(11,22)", "21.99*");
+    checkContainsNot("(11,22)", "22*");
+    checkContainsNot("(11,22)", "22.*");
     checkContainsNot("(11,22)", "23*");
+    checkContains("(11,22)", "21.99*");
```
hj-lorenz, line 116:
```suggestion
    checkContains("(11,22)", "21.99.*");
```
Due this line is identical to line 122 I think this is a copy/paste error and you forgot to add a "<dot>" between "99" and "*".

cli/src/test/java/com/devonfw/tools/ide/version/VersionRangeTest.java
```diff
@@ -109,9 +109,12 @@ public void testContains() {
...
+    checkContains("[22,)", "*");
```
hj-lorenz, line 117: I did not really get the sense of this check. If you substitute "[22,)" with 
"[*,)"   , 
"(*,)"   or
"(,)" 
the check is still true. 

Could you please add a small "//" comment to explain what is special with this test?

hj-lorenz (commented):
LGTM.
Please check 2 comments I added. 

Regarding this version issue I have an additional question:
What happens if the check containing string/sign does not only contain nimbers, dots and/or regex-signes like "*", "?", "^", etc but "normal" letters?

Please have a look at checks like:
checkContains("([22,)", "X");
checkContains("(1,)", "A");
checkContains("(,)", "A");

Some of them fail. Others not. 
Could you explain this?

PR #1032, opened 2025-02-12
#1023 introduce dependencies json for eclipse

hj-lorenz:
Fixes #1023

### Implements
- Introduced dependencies.json for eclipse

Merging blocked until https://github.com/devonfw/ide-urls/pull/24 is merged

PR #1045, opened 2025-02-17
Feature/1023 introduce dependencies json for eclipse

hj-lorenz:


hj-lorenz:
I created this PR because I couldn't delete this branch because sourcetree meant that it is not fully merged yet. 
Even after creating this PR github gave me a hint that the branch is out of date. So I updated it and saw that there are no changes. 
This means that vereything was done with PR #1032

PR #1071, opened 2025-02-24
#789: implement uninstall of IDEasy

### LINKED ISSUE #789: Add support to uninstall
As a IDEasy user, I want to be able to uninstall the IDEasy product so that I can get rid of it in case I want to without further manual steps and knowing internal details how to do that.

We already support `ide uninstall «tool»` to uninstall a specific tool from my current project.
The idea is that if `ide uninstall` is called without any additional parameter, then IDEasy gets uninstalled.
To avoid accidents, an interactive question should be asked so the user must confirm his choice to uninstall the product (or to cancel in case he just forgot to specify the «tool» to uninstall).

This should more or less revert what has been done during installation/setup:
* remove ourselves from `PATH` on Windows
* remove variable `IDE_ROOT` on Windows
* remove entries we made to `.bashrc`
* remove entries we made to `.zshrc` (if exists and IDEasy entries are present)
* remove `$IDE_ROOT/_ide/` ~~except for `$IDE_ROOT/_ide/software`~~
* ~~unset `ide` or other "commands" (aliases/functions) from current bash (`unload`)~~ - since we are implementing Java and not bash code anymore, this does not make so much sense. We could do it in the wrapper script but it is extra effort for no reason.
* Log a message on `interaction` that IDEasy has been uninstalled but to avoid loss of data, the end-user should manually remove his `$IDE_ROOT` folder - but with care since all his projects are located there.

For `.bashrc` and `.zshrc` changes also see #779

hohwille:
fixed #789


jan-vcapgemini (changes requested):
We should also update the help of uninstall I think. Currently it only tells the user that it can be used to uninstall tools, but not that it can also uninstall IDEasy itself.

cli/src/main/resources/nls/Help_de.properties
```diff
@@ -102,10 +102,10 @@ cmd.terraform.detail=Terraform ist ein Tool für Infrastructure as Code zur Verw
 cmd.tomcat=Werkzeug Kommando für Tomcat
 cmd.tomcat.detail=Tomcat ist ein leistungsstarker Open-Source-Webserver und Servlet-Container für Java-Anwendungen.
 cmd.tomcat.val.command=Auszuführende Aktion ( START | STOP )
-cmd.uninstall=Deinstalliert das ausgewählte Werkzeug.
+cmd.uninstall=Deinstalliert ausgewählte Werkzeug(e).
 cmd.uninstall-plugin=Deinstalliert die selektierte Erweiterung für das selektierte Werkzeug.
 cmd.uninstall-plugin.detail=Erweiterung können nur für Werkzeuge installiert und deinstalliert werden die diese unterstützen. Mit dem Befehl "ide install-plugin" kann die Erweiterung wieder hergestellt werden.
-cmd.uninstall.detail=Wird dazu verwendet um jedwedes Werkzeug zu deinstallieren. Um z.B. Java zu deinstallieren geben Sie einfach 'uninstall java' in die Konsole ein.
+cmd.uninstall.detail=Wird dazu verwendet um ausgewählte Werkzeuge zu deinstallieren. Um z.B. Java zu deinstallieren, dient der Befehl 'ide uninstall java'. Um IDEasy selbst zu installieren, dient der Befehlt 'ide uninstall' ohne weitere Parameter.
```
jan-vcapgemini, line 108:
```suggestion
cmd.uninstall.detail=Wird dazu verwendet um ausgewählte Werkzeuge zu deinstallieren. Um z.B. Java zu deinstallieren, dient der Befehl 'ide uninstall java'. Um IDEasy selbst zu installieren, dient der Befehl 'ide uninstall' ohne weitere Parameter.
```

PR #1073, opened 2025-02-24
#38: Added LocalToolCommandlet for Python

### LINKED ISSUE #38: Implement ToolCommandlet for Python
Create `ToolCommandlet` for Python
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/python

ATTENTION: This story is blocked by #352 that needs to be implemented first. UPDATE: #352 is done so unblocked.

See also #352 that needs to be considered when implementing this story.

See also #353 (pip)

julia-cap:
Fixes #38 

Implements:
* New LocalToolCommandlet Python which sets _IgnoreSoftwareRepo_ to true
* Added property for Python to help.properties and help_de.properties
* Added entry to CHANGELOG


jan-vcapgemini (approved):
LGTM, tested locally in intellij with `python --help`. Ready for review.

cli/src/main/java/com/devonfw/tools/ide/tool/python/Python.java
```diff
@@ -0,0 +1,37 @@
...
+  public Python(IdeContext context) {
+
+    super(context, "python", Set.of(Tag.PYTHON));
+  }
+
+  @Override
+  public void printHelp(NlsBundle bundle) {
+
+    this.context.info("For a list of supported options and arguments, use \"python --help\"");
+  }
```
hohwille, line 30: I just noticed that we have some tools that override `printHelp` this way.
However, I do not think that this makes much sense.
The idea is that `printHelp` can print additional help and for `ToolCommandlet` it will call the help of the actual tool if it is installed. If we do not want to have this, we should discuss this in the team and potentially change the general behaviour implemented in `ToolCommandlet` instead of silently overriding this randomly for some of the tools and not for others.

Why overriding this behaviour with a fixed log message? We already have `cmd.python.detail` that is even internationalized. So if you request the help in German, you will still get this hardcoded message in English.

```suggestion
```

hohwille (commented):
@julia-cap thanks for implementing Python 👍 
Did you read the story and also check the old implementation?
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/python#L26-L54

cli/src/main/java/com/devonfw/tools/ide/tool/python/Python.java
julia-cap: Hi @hohwille, I took this over from Node ToolCommandlet. To be honest, I did not really question why this method was overridden. Your explanation makes it clear that when adding properties to the help.properties file, it should not be overridden. I will commit your suggestion and remove those lines of code.

Question: Should those lines for Node.java not be removed, too? Shall I add this to this PR even though it's actually out of scope?

julia-cap:
> @julia-cap thanks for implementing Python 👍 Did you read the story and also check the old implementation? https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/python#L26-L54

@hohwille, yes, I checked the old story and had a look at the old implementation. Though I am not 100% sure why you are referring to the old implementation since with the ToolCommandlet we are using the URLs repository and the provided binaries.

julia-cap:
Hi @hohwille, I have only just noticed line 63 in [old pip script](https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/pip) and I am wondering if we should add this check to the Python ToolCommandlet and install if necessary?

CHANGELOG.adoc
```diff
@@ -6,6 +6,8 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/38[#38]: Implement ToolCommandlet for Python
```
hohwille, line 9: Nice that you did care about adding this feature to CHANGELOG 👍

hohwille (approved):
@julia-cap thanks for your update. Now everything is great and ready to merge. Awesome job👍 

For the record:
> > @julia-cap thanks for implementing Python 👍 Did you read the story and also check the old implementation? https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/python#L26-L54
> 
> @hohwille, yes, I checked the old story and had a look at the old implementation. Though I am not 100% sure why you are referring to the old implementation since with the ToolCommandlet we are using the URLs repository and the provided binaries.

The latest version of Python seems to be 3.13.2 and that indeed has URLs for all OS:
https://github.com/devonfw/ide-urls/tree/master/python/python/3.13.2

This is really great. In order versions this was not the case. Check e.g.:
* https://github.com/devonfw/ide-urls/tree/master/python/python/3.9.21
* https://github.com/devonfw/ide-urls/tree/master/python/python/3.3.6

So let us look forward: New versions of python seem to support all OS now and we finally get what we wanted to archive. It may not work with older versions on some OS but python 2.x can already been considered kind of dead and 3.x versions can simply be upgraded to more recent versions supporting all OS.

hohwille:
We tested this with latest SNAPSHOT and it works in general.
But for some yet unclear issue it by default chooses the wrong version of Python if no version is configured (`3.11.11` even though there are newer versions available) and that version does not have a Windows download URL.
@julia-cap will create a bug issue for this problem.

PR #1080, opened 2025-02-25
#944: improve handling of outputMessages in ProcessContextGitMock

### LINKED ISSUE #944: Improve handling of ProcessContextGitMock.outputMessages
ProcessContextGitMock has an internal state from which it builds the `ProcessResult`.
Tests can add `outputMessages` to it that are accumulated.
The problem is that sub-sequent invocations with this approach lead to strange effects because the second invocation will additionally get output messages from the first invocation.

Therefore we should do the following:
* When we create the `ProcessResult` we create a new `List` from `outputMessages` as a copy
* At the end of the `run` method before we return the `ProcessResult` we do `this.outputMessages.clear()` to rest the state and avoid such ugly side-effects.

paz-capgemini:
Fixes #944 

- cleared the outputMessages before returning the ProcessResult


hohwille (approved):
@paz-capgemini thanks for your PR. Implemented properly and ready for merge.

Issue #1087, opened 2025-02-27
Add MSI Build Workflow Step to include ideasy.exe


alfeilex:
In the current workflows (`nightly-build` and `release`) a step is missing to include the ideasy.exe into the final MSI-files that are used for the installation.

PR #1088, opened 2025-02-27
#1087: Fix MSI Build Job in GitHub Workflows

alfeilex:
Fixes: #1087 
 
This PR fixes the missing include of `ideasy.exe` in the MSI build process for the `nightly build `and for the `release ` workflow.


hohwille (approved):
@alfeilex thank you so much for the lightning fast fix 👍

PR #1091, opened 2025-02-27
#420: exclude mac and linux content from MSI

hohwille:
improvement for #420 
since we copy the `package` content in a script code from the github action, the filters configured for assembly do not apply and linux and mac content gets included in windows MSI.
https://github.com/devonfw/IDEasy/blob/7fa23c7e687fe5015fd7149266cf4b7b598c463f/cli/src/main/assembly/release-win-x64.xml#L27-L28


alfeilex (approved):
LGTM

PR #1092, opened 2025-02-27
#420: ensure os and architecture in filename

hohwille:
improvement for #420 
have os and architecture (`windows-x64`) in filename of MSI file.


alfeilex (approved):
LGTM

Issue #692, opened 2024-10-15
Can't install Docker because of latest version


alfeilex:
### Expected behavior

As a user, I want to install the latest version of Docker (not Rancher Desktop). 

### Actual behavior and steps to reproduce the bug

Change the edition of Docker from Rancher to Docker with  `ide set-edition docker docker`

Run `ide install docker`

You should see this output:
`Version * for tool docker does not exist in edition docker.`

### Comments/Hints:
A guess: Docker has `latest` as the only version and in the code it resolves to `*`. 
Then this function tries to find `*` and the error occurs:
https://github.com/devonfw/IDEasy/blob/2b939e78f521c2b9086e88a453efad5a08ca56d0/cli/src/main/java/com/devonfw/tools/ide/url/model/folder/AbstractUrlFolder.java#L57-L62

PR #1117, opened 2025-03-07
#692 : UrlUpdater fetches every defined version

mbilda:
Fix: #692 

- deleted the new line patterns out of the regex in DockerDesktopUrlUpdater because Pattern.DOTALL already handles this
- TBH I do not know why this pattern now works, because it also should have worked with the (\n\r...) regex. Maybe both, regex and DOTALL don't work together
- changed the group to group1 because there is only one group left

With this implementation i was able to create versionzed folders in my custom repository for the defined docker versions:
{5977D301-C902-4C5E-B262-B6CA530FBE01}

This still needs to be tested with workflow because test <-> workflow isn't the same environment and so on.

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/docker/DockerDesktopUrlUpdater.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/docker/DockerDesktopUrlUpdater.java
@@ -31,13 +31,13 @@ public class DockerDesktopUrlUpdater extends WebsiteUrlUpdater {
     // get Code for version
     String body = doGetResponseBodyAsString("https://docs.docker.com/desktop/release-notes/");
     String regex = "href=#" + version
-        // .......1.........................................................2.................
-        + ".{8,12}(\r\n|\r|\n).{0,350}href=https://desktop\\.docker\\.com.*?(\\d{5,6}).*\\.exe";
+        // .....................................................(Group.1.)..........
+        + ".{8,12}.{0,350}href=https://desktop\\.docker\\.com.*?(\\d{5,6}).*\\.exe";
     Pattern pattern = Pattern.compile(regex, Pattern.DOTALL);
     Matcher matcher = pattern.matcher(body);
     String code;
     if (matcher.find()) {
-      code = matcher.group(2);
+      code = matcher.group(1);
       boolean success = doAddVersion(urlVersion,
           "https://desktop.docker.com/win/main/amd64/" + code + "/Docker%20Desktop%20Installer.exe", WINDOWS);
       if (!success) {


hohwille (changes requested):
@mbilda thank you very much for your PR. You did a great job and very well analysed the problem. Even you tested the URL Updater via GHA in your fork. Excellent 🚀 

However, I checked the logs of your test run and found e.g. this line:
```
12:28:42.793 [com.devonfw.tools.ide.url.UpdateInitiator.main()] - WARN  - c.d.t.i.u.updater.AbstractUrlUpdater - For tool docker and version 4.33.2 the download verification failed with status code 403 for URL https://desktop.docker.com/win/main/amd64/179689/Docker%20Desktop%20Installer.exe.
```

It seems that there are a lot of such errors and I can also reproduce this in my browser:
image

My assumption is that after merging this we get dedicated versions of docker desktop but no URL for Windows and therefore we will entirely break it for Windows users.
Could you confirm my observations and maybe find a solution to improve your PR?

hohwille:
See also this final summary:
```
docker/docker versions added: 0 failed, 98 succeeded, 98 total, 0.00% error - versions verified: 17 failed, 56 succeeded, 73 total, 23.29% error
```

mbilda:
@hohwille good point.
So the error occurs because there is no version to download for windows at all. We still find the version because it is listed in the release notes, but there is no download link provided for windows - only for mac.
See here (example for 179689 - 4.33.2):
{5C2B37A7-0B6A-4159-93CF-8267C2480F8A}
That means there is no download link at all and thats why we receive this error, even pasting the download url into the browser.

Why do we even try to download this version?
The regex:
```
String regex = "href=#" + version
        // .....................................................(Group.1.)..........
        + ".{8,12}.{0,350}href=https://desktop\\.docker\\.com.*?(\\d{5,6}).*\\.exe";
```
The crucial part of this is the following: 
`.*\\.exe`
It matches as many characters as needed until it finds .exe
The next match which will be found is actually the next version where a windows version exists.
See here (regex101 example matching two versions)
https://regex101.com/r/hEmRX3/2
At this point we do not know the "actual docker version" (179689) - thats why we use the regex above to extract it.
Even though it is working, we receive lots of errors with 403, because we try to download a non-existing file.

Furthermore, if the download of the non-existing windows version fails, we return out of the addVersions()-function and skipping the "maybe existing" other versions for mac.

TODO: 
* We need a more specific regex to match exactly the stuff we need - not more, not less.
** How could we do that? 
*** We could delete the .exe of the regex to match the first finding of the docker-version-number. As we see in the above example, it doesn't matter if we have the **.exe** in the regex. It just searches for anything until it finds **.exe** - which will definitely be the case at some time. But at this point we already have what we want - the docker-version-number
*** We then could use the found group-match to execute a new regex but with specific windows urls or docker urls and so on (we actually do this in a different kind of way - we just try to download it - which is kinda finding it - but it fails)
*** After another regex, we can then download it, if it exists or not

Questions:
Should we provide versions only for one OS at all? 

Like in the example above, if we wouldn't return out of the function we would just download the 4.33.2 versions for mac and failing with windows. But then only mac-versions would exist in **ide** - good or bad?

hohwille (commented):
so far we do not have dedicated versions of docker-desktop:
https://github.com/devonfw/ide-urls/tree/master/docker/docker/

That means this PR is adding this as a new feature.
I found that the web-site you are crawling (https://docs.docker.com/desktop/release-notes/) also contains checksum links pointing to SHA256 checksums.
You should also use these checksums since otherwise our fallback applies to compute them ourselves what requires that we download every entire release and compute the checksum ourselves what causes gigabytes of wasted bandwidth and makes the process also much slower.
See here for an existing example:
https://github.com/devonfw/IDEasy/blob/a44dc22f612849f5268f6c0a48cc6728d790d8a6/url-updater/src/main/java/com/devonfw/tools/ide/url/tool/intellij/IntellijUrlUpdater.java#L83-L84

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/docker/DockerDesktopUrlUpdater.java
```diff
@@ -17,6 +19,21 @@ public class DockerDesktopUrlUpdater extends WebsiteUrlUpdater {
...
+          + ".{0,300}"                            // We have to look in close range for the next part (docker lists a summary at top. If this range is to big
+          // we'll find the latest listed version with download links that doesn't match the version we are looking for
+          + "href=https://desktop\\.docker\\.com" // Start of download link
+          + ".*?"                                 // We don't care if its windows or mac - match as least as possible characters
+          + "(\\d{5,6})";                         // Associated docker-version to readable version we are looking for
+  private final static String REGEX_FOR_DOWNLOAD_URLS = "https://desktop.docker.com/%s/main/%s/%s/";
+  private final static String WIN_VERSION = "win";
+  private final static String MAC_VERSION = "mac";
+  private final static String AMD_ARCH_TYPE = "amd64";
+  private final static String ARM_ARCH_TYPE = "arm64";
```
jan-vcapgemini, line 35: Cool. :)

mbilda:
This PR extends on the fly because there are several open points to discuss to handle the new url updater for docker desktop.
Open points:
- the more regexes and compiles of the regexes, the more ide loses performance
- We need the checksum check as well
- Do we really need docker desktop? Isn't it enough to install docker (cli)
- Should we really pull all versions? (There are versions which only fixes bugs for mac -> only download links for mac)

This PR handles two individual things. The first one is the fix for the actual bug. IDE could not handle latest version. Because it is resolved to " * ". The second one is the improvement of the url updater which throws a lot of open questions.
I will split this PR into two parts.
First part: Fix the actual bug
Second part: Url updater improvement

To get things done, i will exclude the first part of this PR and create a new PR for this fix only.
I also will detach the issue from this PR and create a new issue for all open points and improvement of the url updater and link it to **this** PR.

mbilda:
> You should also use these checksums since otherwise our fallback applies to compute them ourselves what requires that we download every entire release and compute the checksum ourselves what causes gigabytes of wasted bandwidth and makes the process also much slower.


I added the checksum to the addVersion method.
For each checksum we need to curl the checksum url and get the first part out of it.
We only do this if the version exists, a check that we are already doing.

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/docker/DockerDesktopUrlUpdater.java
```diff
@@ -29,31 +52,95 @@ protected void addVersion(UrlVersion urlVersion) {
...
+   *
+   * @param urlVersion the readable version e.g. 4332 (4.33.2)
+   * @param dockerVersion the associated docker version to readable version e.g. 179689
+   * @param body the html body to search in
+   */
+  private void addVersionsForWindows(UrlVersion urlVersion, String dockerVersion, String body) {
+    boolean versionExists = checkIfVersionExists(dockerVersion, body, WIN_VERSION, AMD_ARCH_TYPE);
+    if (versionExists) {
+      String cs = getChecksum(WIN_AMD_URL + dockerVersion + "/checksum.txt");
+      doAddVersion(urlVersion, WIN_AMD_URL + dockerVersion + "/Docker%20Desktop%20Installer.exe", WINDOWS, X64, cs);
```
jan-vcapgemini, line 80: Maybe remove the "/" like this and introduce another contant instead: e.g. `... dockerVersion + "/" + CHECKSUM_FILE`.

jan-vcapgemini (approved):
Thanks for adding the checksum check too. I've added a small suggestion, please check.

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/docker/DockerDesktopUrlUpdater.java
mbilda: Fixed.

Issue #1039, opened 2025-02-14
Update Jasypt commandlet implementation to run Java version from dependencies.json


jan-vcapgemini:
The implementation of Jasypt is very old and seems to be using outdated workarounds. We should check how we can update the commandlet to use the latest features of IDEasy. 

Update:
The problem is that it is calling `Java` as commandlet what will use the project version of Java and not the one from `dependencies.json`:
https://github.com/devonfw/IDEasy/blob/e1743f8683c920ccff3131761233fe4efdbc88cc/cli/src/main/java/com/devonfw/tools/ide/tool/jasypt/Jasypt.java#L97-L98

Fix should be to just call `java` command running Jasypt via `ProcessContext`.

PR #1126, opened 2025-03-11
#1039: Updated Jasypt command implementation

leonrohne27:
### Implements:

- Update Jasypt commandlet implementation to run Java version from dependencies.json instead of calling Java as commandlet


hohwille:
Replaced by PR #1156

Issue #999, opened 2025-01-30
Make IDELogEntry comparing more flexible


yurii-yu:
As a developer, I want the `hasEntries()` method so that 

1. I do not need to list the expected entries in the right order
2. I do not need to offer the expected log entries as raw texts, instead, I can specify them as patterns.

For example

before
```
assertThat(context).log().hasEntries(IdeLogEntry.ofSuccess("Successfully installed java in version 17.0.10_7"),

IdeLogEntry.ofSuccess("Successfully installed mvn in version 3.9.6"));
```

after
```
//change the order of those expected log entries
assertThat(context).log().hasEntries(IdeLogEntry.ofSuccess("Successfully installed mvn in version 3.9.6"),

//assume that regex is supported here
IdeLogEntry.ofSuccess("Successfully installed java in version 17.[0-9._]+"));
```

julia-cap:
This issue in invalid, see linked PR #1131

PR #1131, opened 2025-03-13
#999: More flexible IDELogEntry comparison

leonrohne27:
### Implements:

- Order of IDELogEntries is now being ignored


hohwille (commented):
@leonrohne27 I am very sorry to see this and having to tell you that something went very wrong here...
Actually @yurii-yu created the story #999 with his personal expectation when he was very new in the project and did not understand our test-infrastructure.
Even worse we talked in the team about this and explained him that he can already ignore the order via fluent API: `hasEntry("last").hasEntry("first").hasEntry("middle")`
However, he forgot to update the story he created and now you created a PR removing a feature that we actually want to have and do not consider as a bug or flaw.

I once had a process that every ticket is initially created in status `Refinement` and then we do a refinement process with the team were we discuss such issues, gain a common understanding and add/adjust the story details and move it to `New` state.
However, since we are drowning in issues this process did not work out at all and many important issues I created were all still in `Refinement` and therefore nobody started working on them...
I have to think this over if we can find a better process....
However, I have to admit that IMHO we need to reject and close this PR without merging. Sorry - not your fault! 🙏

Issue #1130, opened 2025-03-13
improve behaviour on ambiguous XPath match


hohwille:
As a IDEasy user, I want to run my IDE seamlessly so that I do not have to worry about stacktraces or get blocked by errors.

## Actual behaviour

```
$ ide intellij
Start: Configuring workspace main for IDE intellij
Failed to merge workspace file D:\projects\IDEasy\workspaces\main\.idea\workspace.xml with update template D:\projects\IDEasy\settings\intellij\workspace\update\.idea\workspace.xml and setup file D:\projects\IDEasy\settings\intellij\workspace\setup\.idea\workspace.xml!
com.devonfw.tools.ide.merge.xmlmerger.XmlMergeException: Merge strategy combine failed on /project[@version='4']/component[@name='RunManager']
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMergeStrategy.merge(XmlMergeStrategy.java:69)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMergeStrategy.combineChildNodes(XmlMergeStrategy.java:113)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMergeStrategy$1.doMerge(XmlMergeStrategy.java:29)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMergeStrategy.merge(XmlMergeStrategy.java:65)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMerger.merge(XmlMerger.java:123)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMerger.doMerge(XmlMerger.java:82)
        at com.devonfw.tools.ide.merge.FileMerger.merge(FileMerger.java:50)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:69)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:73)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:73)
        at com.devonfw.tools.ide.tool.ide.IdeToolCommandlet.mergeWorkspaceSingle(IdeToolCommandlet.java:103)
        at com.devonfw.tools.ide.tool.ide.IdeToolCommandlet.mergeWorkspace(IdeToolCommandlet.java:90)
        at com.devonfw.tools.ide.tool.ide.IdeToolCommandlet.configureWorkspace(IdeToolCommandlet.java:72)
[...]

Issue #1038, opened 2025-02-14
XML merger fails in native-image on custom XPath with MissingResourceException


hohwille:
### Expected behavior

As a IDEasy user, I want to be able to run my IDE (IntelliJ) seamless without errors so that I can focus on development in my project and do not have to worry about IDEasy itself.

### Actual behavior

```
$ ide intellij
...
Failed to merge workspace file D:\projects\project\workspaces\main\.idea\workspace.xml with update template D:\projects\project\settings\intellij\workspace\update\.idea\workspace.xml and setup file D:\projects\project\settings\intellij\workspace\setup\.idea\workspace.xml!
com.devonfw.tools.ide.merge.xmlmerger.XmlMergeException: Merge strategy combine failed on /project[@version='4']/component[@name='RunManager']
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMergeStrategy.merge(XmlMergeStrategy.java:69)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMergeStrategy.combineChildNodes(XmlMergeStrategy.java:113)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMergeStrategy$1.doMerge(XmlMergeStrategy.java:29)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMergeStrategy.merge(XmlMergeStrategy.java:65)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMerger.merge(XmlMerger.java:123)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMerger.doMerge(XmlMerger.java:82)
        at com.devonfw.tools.ide.merge.FileMerger.merge(FileMerger.java:50)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:69)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:73)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:73)
        at com.devonfw.tools.ide.tool.ide.IdeToolCommandlet.configureWorkspace(IdeToolCommandlet.java:83)
        at com.devonfw.tools.ide.tool.ide.IdeToolCommandlet.run(IdeToolCommandlet.java:48)
        at com.devonfw.tools.ide.context.AbstractIdeContext.applyAndRun(AbstractIdeContext.java:926)
[...]

PR #1138, opened 2025-03-17
Fix/1130 improve behaviour on ambigous xpath match

leonrohne27:
### Implements:

- Now logs a warning instead of an error + stacktrace
- user is not being asked anymore if he wants to continue despite the error
- The code scans the child elements of the template document and uses seenElements to check whether an element has already been merged. If an element already exists, it is skipped.


cli/src/main/java/com/devonfw/tools/ide/merge/xmlmerger/XmlMergeStrategy.java
```diff
@@ -60,13 +61,13 @@ protected void doMerge(Element templateElement, Element resultElement, ElementMa
...
-
+    Logger logger = Logger.getLogger(XmlMergeStrategy.class.getName());
     try {
       doMerge(templateElement, resultElement, matcher);
     } catch (XmlMergeException e) {
-      throw e;
+      logger.warning("XML Merge Exception: " + e.getMessage());
     } catch (RuntimeException e) {
-      throw new XmlMergeException("Merge strategy " + this + " failed on " + XmlMergeSupport.getXPath(templateElement, true), e);
+      logger.warning("Merge strategy " + this + " failed on " + XmlMergeSupport.getXPath(templateElement, true) + ": " + e.getMessage());
```
hohwille, line 70: Could you revert this changes as this is causing that all other errors are now also just logged and stacktraces of real errors are lost so we cannot support users hitting new problems and say exactly what went wrong (see e.g. #1038).

cli/src/main/java/com/devonfw/tools/ide/merge/xmlmerger/matcher/IdComputer.java
```diff
@@ -61,8 +65,8 @@ public Element evaluateExpression(Element templateElement, Element workspaceElem
       } else if (length == 0) {
         return null;
       } else {
-        throw new IllegalStateException(
-            length + " matches found for XPath " + xpathExpr + " in workspace XML at " + XmlMergeSupport.getXPath(workspaceElement, true));
+        this.context.warning(length + " matches found for XPath " + xpathExpr + " in workspace XML at " + XmlMergeSupport.getXPath(workspaceElement, true));
+        return (Element) nodeList.item(0);
```
hohwille, line 69: nice. That solves the actual issue now! 👍 
The story #1130 suggested that we should introduce a new variable that can be configured (e.g. in `conf/ide.properties`) to get the exception behaviour what can be beneficial for IDE admins that want to get more attention if something is wrong. Simply surround with an if block and check for the variable in the condition and in the else block add the old exception behaviour.
You can extract the message to a variable to reuse it for logging and the exception message.
Otherwise when doing logging avoid string concatenation and prefer the pattern based approach:
https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc#logging

hohwille (approved):
@leonrohne27 thanks for your PR. You already solved the problem 👍 
Have a look at my review comments to complete the story and avoid undesired other behaviour. Then we can merge this.

cli/src/main/java/com/devonfw/tools/ide/merge/xmlmerger/matcher/IdComputer.java
```diff
@@ -19,17 +20,20 @@ public class IdComputer {
...
    * The constructor.
    *
    * @param id the {@link #getId() merge ID}.
    */
-  public IdComputer(String id) {
+  public IdComputer(String id, IdeContext context) {
 
     super();
     this.id = id;
+    this.context = context;
```
hohwille, line 36: Please also add the story to our CHANGELOG.adoc file:

cli/src/main/java/com/devonfw/tools/ide/merge/xmlmerger/XmlMergeStrategy.java
leonrohne27: thanks, I deleted it.

cli/src/main/java/com/devonfw/tools/ide/merge/xmlmerger/matcher/IdComputer.java
leonrohne27: I tried to implement it, please have a look

cli/src/main/java/com/devonfw/tools/ide/merge/xmlmerger/matcher/IdComputer.java
leonrohne27: done

cli/src/main/java/com/devonfw/tools/ide/merge/xmlmerger/matcher/IdComputer.java
hohwille: Thanks for the improvement. 👍 
I was not precise enough: In the story I just used the term `Property` and in this comment I talked about a `Variable`. Both terms are very ambiguous. What I meant was a "[IDEasy Variable](https://github.com/devonfw/IDEasy/blob/main/documentation/variables.adoc)" so it can be configured in [ide.properties](https://github.com/devonfw/IDEasy/blob/main/documentation/configuration.adoc) what gives a lot more flexibility.
A system property is a little bit more complicated for end-users to configure.

Anyhow your change is fine already. IMHO the easiest approach is that I merge your PR now.
Then I can create a mini-PR after the merge myself to change that from a System property to an IDEasy varaible as further improvement.

hohwille:
Successfully tested with latest SNAPSHOT (2025.03.002-20250327.024013-7):
```
Start: Configuring workspace main for IDE intellij
2 matches found for XPath configuration[@default='true' and @type='JUnit'] in workspace XML at /project[@version='4']/component[@name='RunManager' @selected='JUnit.EnvironmentVariablesTest.testGetToolVariable']
Successfully ended step 'Configuring workspace main for IDE intellij'.
Start: Install java
```

Issue #1008, opened 2025-01-31
Improve upgrade-settings to cleanup legacy configs in repositories


hohwille:
As a IDEasy user, I want `ide upgrade-settings` to cleanup as much of legacy stuff as possible so that I get my project up-to-date and get rid of warnings.

Currently when repositories are loaded from properties and use `git.url` or `git-url` instead of `git_url` this is raising a legacy warning.
The same applies for `eclipse=import` instead of `import=eclipse`.

The command `ide upgrade-settings` should also fix such legacy stuff and repair it automatically.
See PR #820 and also story #954

PR #1146, opened 2025-03-19
#1008: improve upgrade settings

leonrohne27:
### Implemets:

- In properties, "git-url" und "git.url" are replaced with "git_url"
- In properties, "eclipse=import" is replaced with "import=eclipse"


leonrohne27:
I messed up with the branches, however this should be fine now.
To implement this story, I used filewalkers, which I know you removed in the other code sections.
I tried it otherwise but I was not able to fix it. Now it works, but you have to decide whether you want this kind of implementation or not.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -181,4 +187,43 @@ private static void updatePropertiesLegacyEdition(EnvironmentVariablesProperties
       environmentVariables.remove(legacyEditionVariable);
     }
   }
+
+  private void cleanupLegacyProperties() {
+    this.context.info("Cleaning up legacy properties...");
+
+    Path rootDirectory = Paths.get(System.getProperty("user.dir"));
```
hohwille, line 194: Dont hardcode `user.dir` this will cause JUnits to apply the migration your computer in your home directory but tests should only work in test directories (`./target/`).
Also the variable name could be named more intuitive:
```suggestion
    Path homeDirectory = this.config.getgetUserHome();
```

However, why are you searching recursively for `IDEasy.properties` files in the users home directory?
This does not make sense.

What you IMHO actually need is something like this:
```
    Path settingsPath = context.getSettingsPath();
    Path repositoriesPath = settingsPath.resolve(IdeContext.FOLDER_REPOSITORIES);
    if (!Files.isDirectory(repositoriesPath)) {
      return;
    }
    try (Stream<Path> childStream = Files.list(repositoriesPath)) {
      Iterator<Path> iterator = childStream.iterator();
      while (iterator.hasNext()) {
        Path child = iterator.next();
        if (child.getFileName().toString().endsWith(IdeContext.EXT_PROPERTIES);
        updateProperties(child);
      }
    }
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -181,4 +187,43 @@ private static void updatePropertiesLegacyEdition(EnvironmentVariablesProperties
...
+    try {
+      Files.walk(rootDirectory)
+          .filter(path -> path.getFileName().toString().equals("IDEasy.properties"))
```
hohwille, line 197: Obsolete but for the record. This filename `IDEasy.properties` is just an example and cannot be hardcoded.
There can be any number of `*.properites` files in the `repositories` folder and you cannot make any assumptions about their name.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -181,4 +187,43 @@ private static void updatePropertiesLegacyEdition(EnvironmentVariablesProperties
...
+            } catch (IOException e) {
+              this.context.warning("Error processing IDEasy.properties at " + filePath);
+            }
+          });
+    } catch (IOException e) {
+      this.context.warning("Error walking the file tree to find IDEasy.properties.");
+    }
+  }
+
+  private void updateProperties(Path filePath) throws IOException {
```
hohwille, line 210: Maybe rename to make more clear what this is doing:
```suggestion
  private void updateRepositoryPropertiesFile(Path filePath) throws IOException {
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -181,4 +187,43 @@ private static void updatePropertiesLegacyEdition(EnvironmentVariablesProperties
...
+    List<String> lines = Files.readAllLines(filePath);
+    List<String> updatedLines = new ArrayList<>();
```
hohwille, line 212: FYI: You can do the modifications directly in the original list (`lines`)  and do not need a second list as a copy.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -181,4 +187,43 @@ private static void updatePropertiesLegacyEdition(EnvironmentVariablesProperties
...
+        updatedLines.add("git_url=" + gitUrl);
+        continue;
+      }
+      if (line.startsWith("eclipse=import")) {
+        updatedLines.add("import=eclipse");
+        continue;
+      }
+      updatedLines.add(line);
+    }
+    Files.write(filePath, updatedLines, StandardOpenOption.WRITE, StandardOpenOption.TRUNCATE_EXISTING);
```
hohwille, line 226: you should add a boolean flag `boolean updated=false;` before the loop and set it to `true` when you have modified something.
Then you can only write the file changes back if something changed and additionally log the success message only in that case so if nothing changed then nothing if logged.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -181,4 +187,43 @@ private static void updatePropertiesLegacyEdition(EnvironmentVariablesProperties
...
+    this.context.success("Successfully updated IDEasy.properties at " + filePath);
```
hohwille, line 227: Again, you misunderstood the `IDEasy.properties` filename. See and compare the different examples:
* https://github.com/m-m-m/settings/tree/master/repositories
* https://github.com/devonfw/ide-settings/tree/main/repositories
* https://github.com/devonfw-training/java-quarkus-settings/tree/master/projects (devonfw-ide legacy config)

Also see here: https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc#logging

```suggestion
    this.context.success("Successfully updated repository configuration file {}", filePath);
```

hohwille (changes requested):
@leonrohne27 thanks for your PR.
The actual updating of the properties file looks already good to me. 👍 
However, the way you try to find the properties is not correct.

It is required that you first read and understand the documentation of the related feature:
https://github.com/devonfw/IDEasy/blob/main/documentation/repository.adoc

p.s.: I took a shortcut even though this PR was in team review state to speed up our progress.

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
hohwille: BTW: Instead of using `Files.list` directly what also requires a catch statement that I omitted, it seems smarter that you use `fileAccess` like already done in the same class here:
https://github.com/devonfw/IDEasy/blob/8a9936ef2519f5b4da8ba1c67e294c5ca574a64d/cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java#L83-L89

cli/src/main/java/com/devonfw/tools/ide/commandlet/UpgradeSettingsCommandlet.java
```diff
@@ -181,4 +187,43 @@ private static void updatePropertiesLegacyEdition(EnvironmentVariablesProperties
...
+    List<String> updatedLines = new ArrayList<>();
+
+    for (String line : lines) {
+      if (line.startsWith("git.url=") || line.startsWith("git-url")) {
+        String gitUrl = line.substring("git.url=".length());
+        updatedLines.add("git_url=" + gitUrl);
+        continue;
+      }
+      if (line.startsWith("eclipse=import")) {
+        updatedLines.add("import=eclipse");
```
hohwille, line 221: BTW: Theoretically you could have  a properties file containing this
```
import=intellij
eclipse=import
```

Then the expected result after the migration would be
```
import=intellij,eclipse
```

However, this is an edge-case that is extremely unlikely and must not be supported.
You could have used `java.util.Properties` to read and write the properties file what is already supported by `FileAccess`.
But your approach has the advantage that it keeps comments and the original order untouched.
A disadvantage of your approach is that it will not work with lines like `eclipse = import` what is also accepted by propreties (or even `eclipse: import`).
Therefore I am resolving this comment as we can consider this as good enough - we do not need perfection here.

hohwille (approved):
@leonrohne27 thanks for your rework. I resolved all existing review comments 👍 

Looking at the PR again I found a small problem that I solved with a constructive suggestion.
If you agree and commit it, we can merge this. Thanks.

hohwille:
I have successfully tested this but found and file this issue #1214

Issue #1147, opened 2025-03-19
Fix maven-antrun-plugin Configuration After Version Update


alfeilex:
After updating `maven-antrun-plugin` from `1.8` to `3.1.0` by dependabot, the build fails due to a deprecated usage of `tasks` element in the configuration. 

https://github.com/devonfw/IDEasy/actions/runs/13937579797/job/39008588065#step:5:241

A change to `target` might resolve the issue

PR #1148, opened 2025-03-19
#1147: Fix maven-antrun-plugin configuration after version update

alfeilex:
Fixes: #1147

This PR replaces the deprecated `tasks` element in the configuration with `target`

Tested in this workflow run: https://github.com/alfeilex/IDEasy/actions/runs/13944873219/job/39030199952


hohwille (approved):
@alfeilex thanks for your PR. LGTM. 👍 
I hope it was tested since our CI is not covering this part.

BTW: we have multiple places where we use `maven-antrun-plugin`.
It is also used in the `documentation/pom.xml`.
Also we could ensure that we centralise the version of this plugin or all plugins (what we partially did for most of the plugins via our `maven-parent` - however having all the versions inside our own git repo allows dependabot to create updates and make ourselves independent from releases of `maven-parent`).

PR #1171, opened 2025-03-25
#907: added Yarn commandlet

### LINKED ISSUE #907: Implement ToolCommandlet for Yarn
Create `ToolCommandlet` for Yarn
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/yarn

See also #441

hohwille:
fixes #907

add Yarn commandlet


cli/src/main/java/com/devonfw/tools/ide/tool/yarn/Yarn.java
```diff
@@ -0,0 +1,141 @@
...
+          pc.run("npm", "uninstall", "-g", "yarn");
+        }
+        npmInstall("corepack", pc);
+      }
+      pc.run("yarn", "init", "-2");
+      if (resolvedVersion.isPattern()) {
+        this.context.warning("Yarn currently does not support version pattern: {}", resolvedVersion);
+      } else {
+        pc.run("yarn", "set", "version", resolvedVersion.toString());
+        pc.run("yarn", "install");
```
hohwille, line 113: Even though this is documented and recommended this way on [yarn installation documentation](https://yarnpkg.com/getting-started/install), this is causing undesired effects and creates local files like `package.json`, `.gitignore`, etc. in the current working directory.
In IDEasy we distinguish between local and global tool installations.
Also the node ecosystem with npm distinguishes between local and global installations:
Here `local` means that it is installed in a `node_modules` folder in the current working directory to make it available (as a dependency) for the "project" in that directory while `global` means that it is installed in the `node` installation directory that is supposed to be on the `PATH` so the user can directly invoke executable commands installed there.

Since IDEasy installs `node` as a local tool inside `$IDE_HOME/software/node` the `global` installations of node packages are still to be considered local in the sense of IDEasy while `local` installations of node packages are useless for tools we want to integrate since we or the user cannot execute them in his terminal.

hohwille: Maybe we can find some wisdom here:
https://yarnpkg.com/install.sh

julia-cap:
For installation of `Yarn Classic`, we might want to consider that if a user has _yarnrc.yml_ file, the _yarnPath_ stated in that file will be taken for installation i.e. we won't be able to downgrade from `Yarn Modern` to `Classic`. See issue on GitHub [here ](https://github.com/yarnpkg/yarn/issues/8841)and [here](https://github.com/nodejs/corepack/issues/118).

hohwille:
This node ecosystem is diving me nuts.
```
$ corepack prepare yarn@4.6.0 --activate
Preparing yarn@4.6.0 for immediate activation...

$ yarn -v
4.7.0

$ corepack install -g yarn@4.6.0
Installing yarn@4.6.0...

$ yarn -v
4.7.0
```

Can someone help me how to install a specific version of yarn.

hohwille:
> For installation of `Yarn Classic`, we might want to consider that if a user has _yarnrc.yml_ file, the _yarnPath_ stated in that file will be taken for installation i.e. we won't be able to downgrade from `Yarn Modern` to `Classic`. See issue on GitHub [here ](https://github.com/yarnpkg/yarn/issues/8841)and [here](https://github.com/nodejs/corepack/issues/118).

This is actually why we would love to isolate tools from each other. Then we could simply delete yarn folder to "uninstall" it and reinstall a different version. But node.js/npm/yarn and python/pip is both  creating a big ball of mud preventing a clean design allowing upgrading and downgrading individual tools.
See https://github.com/devonfw/IDEasy/edit/main/documentation/tool-vendor-plea.adoc

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
```diff
@@ -280,6 +280,7 @@ private void runGitCommand(Path directory, List<String> args) {
 
   private String runGitCommandAndGetSingleOutput(String warningOnError, Path directory, String... args) {
 
+    // return this.context.newProcess().directory(directory).runAndGetSingleOutput(this.context.warning(), "git", args);
```
jan-vcapgemini, line 283:
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/tool/yarn/Yarn.java
```diff
@@ -0,0 +1,121 @@
...
+    return null;
+  }
+
+  @Override
+  public ToolInstallation installTool(GenericVersionRange version, ProcessContext pc, String edition) {
+
+    ToolRepository toolRepository = getToolRepository();
+    VersionIdentifier configuredVersion = getConfiguredVersion();
+    // UrlUpdater required for Yarn...
+    // VersionIdentifier resolvedVersion = toolRepository.resolveVersion(this.tool, edition, version, this);
```
jan-vcapgemini, line 84:
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
hohwille: I did not comment out existing code but rather wanted to open a discussion if we should change this implementation to the out-commented code. However, since this has nothing to do with Yarn, I can remove it here and create a separate PR for it.

cli/src/main/java/com/devonfw/tools/ide/tool/yarn/Yarn.java
hohwille: Same here. We actually need an UrlUpdater for Yarn at least to know the available version numbers.
AFAIK Yarn/node cannot handle version patterns `4.*` and for the parts where it can it has a different syntax so `*!` would be a different syntax in JS ecosystem.
Maybe I should have created this PR as a draft.
Actually I only stepped in because the issue turned out to be more complex than initially expected due to what @julia-cap figured out with Yarn Classic vs. Yarn Modern.

What a review could additionally point out is that there is no JUnit yet included for Yarn and that in some scenarios the current implementation does not work so it is actually buggy and therefore incomplete.

PR #1173, opened 2025-03-26
#1172: fixed tool varaibles to use underscore instead of hyphen

### LINKED ISSUE #1172: Prevent hyphen in environment variable names
In bug #863 we can see this:
```
bash: ANDROID-STUDIO_VERSION=2024.2.1.12: command not found
```

What is wrong here is that `android-studio` was converted to upper case and used as prefix for an environment variable.
However, a dash sign (hyphen) cannot be used in bash for an environment variable since it is registered as operator sign.

https://github.com/devonfw/IDEasy/blob/63a350e22e5ef970ecc52e0b6334bc0d06e4ab27/cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariables.java#L77

https://github.com/devonfw/IDEasy/blob/63a350e22e5ef970ecc52e0b6334bc0d06e4ab27/cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariables.java#L265

https://github.com/devonfw/IDEasy/blob/63a350e22e5ef970ecc52e0b6334bc0d06e4ab27/cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariables.java#L274

hohwille:
fixes #1172

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariables.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariables.java
@@ -74,7 +74,7 @@ public interface EnvironmentVariables {
    */
   default String getToolEdition(String tool) {
 
-    String variable = tool.toUpperCase(Locale.ROOT) + "_EDITION";
+    String variable = getToolEditionVariable(tool);
     String value = get(variable);
     if (value == null) {
       value = tool;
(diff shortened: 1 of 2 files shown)


jan-vcapgemini (approved):
LGTM. Ready for merge.

hohwille:
tested with android-studio on 2025.03.002-20250327.024013-7 SNAPSHOT and works fine.

Issue #1115, opened 2025-03-07
UrlUpdater action causing strange error/warning


hohwille:
I looked into the logs of the URL Updater github action and found this:

```
Run cd ide-urls
  cd ide-urls
  git config --global user.name ***
  git config --global user.email ***
  if [ "$(git status -z)" = "" ]
  then
    echo "No changes, nothing to commit."
  else
    git add .
    git commit -m "Update urls"
    git push
  fi
  shell: /usr/bin/bash -e {0}
  env:
    JAVA_HOME: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.14-7/x64
    JAVA_HOME_17_X64: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.14-7/x64
/home/runner/work/_temp/e5b35c3c-db99-478e-b5c6-35164fc9a7c9.sh: line 4: warning: command substitution: ignored null byte in input
[master ea3bbf01] Update urls
```

Can someone explain what is wrong here?
```
/home/runner/work/_temp/e5b35c3c-db99-478e-b5c6-35164fc9a7c9.sh: line 4: warning: command substitution: ignored null byte in input
```

So is line 4 this one?
```
  if [ "$(git status -z)" = "" ]
```

So we cannot evaluate `$(git status -z)` what should actually run the command `git status -z` and evaluate to the standard output of that command. So it seems that this git command is echoing null bytes causing this warning.
Maybe there is nothing we need to do about it but we could at least leave a comment once we understood what is happening here, if we thing this is intended.

alfeilex:
I tried running `$(git status -z)` in a git repo with local changes and also got the output:

 `bash: warning: command substitution: ignored null byte in input`

The `-z` flag uses null byte delimiter to print the filenames, which causes the warning. My suggestion is to switch to `git status -z | grep -q .` to check if there is any output.

PR #1177, opened 2025-03-26
#1115: Fix Git Action Warnings in Url Updater

alfeilex:
This PR fixes: #1115 

Switch from comparing `$(git status -z)` with an empty string to using `grep -q .` to check for any output. That's because `-z` uses a null byte delimiter to separate filenames in the printed output. This leads to warnings in the logs because command substitution `$(...)` can not handle null bytes as input


hohwille (approved):
@alfeilex thanks for your fix of this. 👍 
At first I thought you are checking that the changed files contain a dot but it seems this is kind of a regex pattern accepting any character.

PR #1179, opened 2025-03-26
#889: improve icd command

### LINKED ISSUE #889: Improve icd and add documentation
We need to update our documentation. An entry for the `icd` command, introduced in #778 is still missing.

Also `icd -p IDEasy -w main` is not working though it should.
Would also be nice to have `icd -h` / `icd --help`

hohwille:
fixes #889 

* improve icd (e.g. to support both `-p` and `-w` together, but also `-h` and long-options)
* added documentation


hohwille:
Retested on latest SNAPSHOT (2025.03.002-20250327.024013-7) and works great.

Issue #951, opened 2025-01-20
Unexpected output from setup.bat


hohwille:
When I double click `setup.bat` on Windows to install IDEasy, I get some unexpected output.

```
Setting up IDEasy in D:\projects\_ide
IDEasy is already added to your PATH.
Found bash at C:\Program Files\Git\bin\bash.exe
"C:\Program Files\Git\bin\bash.exe" -l -c "cd \"D:\projects\_ide\";./setup"
/c/Users/hohwille/.bashrc: line 5: bind: warning: line editing not enabled
/c/Users/hohwille/.bashrc: line 8: devon: command not found
[91m[0m
Setting up IDEasy in /d/projects/_ide
[91m[0m
2025.01.001-beta-01_15_02-SNAPSHOT
[91m[0m
Configuring IDEasy in /c/Users/hohwille/.bashrc.
Configuring IDEasy in /c/Users/hohwille/.zshrc.
ATTENTION: IDEasy has been setup for your shells but you need to start a new shell to make it work.
Only if you invoked this setup script by sourcing it, you are able to run 'ide' and 'icd' commands without starting a new shell.

Setup of IDEasy completed
Drücken Sie eine beliebige Taste . . .
```

What confuses me is
```
/c/Users/hohwille/.bashrc: line 5: bind: warning: line editing not enabled
/c/Users/hohwille/.bashrc: line 8: devon: command not found
```
Maybe my `.bashrc` could be misconfigured but when I launch a new git-bash directly or in Windows Terminal, I do not get this.
Further the colouring from bash into CMD seems to have an issue:
```
[91m[0m
```
That is definitely wrong in IDEasy.

alfeilex:
Here is a potential solution: 

https://github.com/quarkusio/quarkus/issues/1029#issuecomment-774617979

PR #1186, opened 2025-03-27
#951: Fix ANSI support for coloring in cmd and powershell

alfeilex:
Fixes: #951 

This PR fixes the `�[91m�[0m` codes in the output after running `setup.bat` in native `cmd` and `powershell` terminals (not inside windows terminal). 

It enables the support of ANSI codes by setting a registry value.
`reg add HKCU\Console /v VirtualTerminalLevel /t REG_DWORD /d 1 /f`

We may need to discuss whether we want to force this enabling with `/f` or allow the user to decide whether to enable it. Moreover, we should decide if we want to add this registry in the background without informing the user about it at all.

### DIFF
--- a/cli/src/main/package/setup.bat
+++ b/cli/src/main/package/setup.bat
@@ -8,6 +8,9 @@ Set _RESET=[0m
 pushd %~dp0
 echo Setting up IDEasy from %CD%
 
+REM activate ANSI support for colors in CMD and Powershell
+reg add HKCU\Console /v VirtualTerminalLevel /t REG_DWORD /d 1 /f
+
 REM find bash on your Windows system...
 for %%H in ( HKEY_LOCAL_MACHINE HKEY_CURRENT_USER ) do for /F "usebackq tokens=2*" %%O in (`call "%SystemRoot%"\system32\reg.exe query "%%H\Software\GitForWindows" /v "InstallPath" 2^>nul ^| "%SystemRoot%\system32\findstr.exe" REG_SZ`) do set GIT_HOME=%%P
 


cli/src/main/package/setup.bat
```diff
@@ -8,6 +8,9 @@ Set _RESET=[0m
 pushd %~dp0
 echo Setting up IDEasy from %CD%
 
+REM activate ANSI support for colors in native cmd and powershell terminals
+reg add HKCU\Console /v VirtualTerminalLevel /t REG_DWORD /d 1 /f
```
hohwille, line 12: Very nice that you found this trick.
Should we add this to a `*.reg` file in `system/windows` like the other registry hooks?
Then also the MSI installer could apply the same change via that file.

hohwille (approved):
@alfeilex thank you for your PR. So great that you searched and found this solution. 👍

cli/src/main/package/setup.bat
alfeilex: I moveed it into `.reg` files. One for enabling it and one for disabling it

alfeilex: It is already defined in the IDEasy installation process which `.reg` files are executed?

hohwille: We had this in devonfw-ide but currently in IDEasy we only have optional `.reg` files.
Here is the devonfw-ide code:
https://github.com/devonfw/ide/blob/855ce79d795860416f73634397ea9a535c27a68c/scripts/src/main/resources/setup.bat#L22-L29

alfeilex: done

Issue #1191, opened 2025-03-28
vscode always tries to reinstall all plugins


jan-vcapgemini:
### Expected behavior

As an IDEasy I expect vscode to install all of its plugins only once.

### Actual behavior
IDEasy always tries to re-installs all plugins of vscode 

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. run ide vscode
2. wait till all plugins are installed
3. close vscode
4. run ide vscode again

### Comments/Hints:
It looks like the plugin installation of vscode is not checking the plugin marker files.

### Affected version:

* OS: Windows 10 x64
* IDEasy: 2025.03.003-03_28_02-SNAPSHOT

PR #1192, opened 2025-03-28
#1191: Fixed vscode plugin detection

jan-vcapgemini:
Fixes: #1191

**Implements**
* Enhanced tests for vscode and intellij to properly detect if plugins were not re-installed
* replaced installPlugins with overwrite of handleInstallForInactivePlugin
* converted doAddRecommendations method from list of plugins to single plugin
* adjusted vscode test (isolated plugin checks)


cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -73,13 +73,18 @@ protected void installPlugins(Collection<ToolPluginDescriptor> plugins, ProcessC
     try {
       for (ToolPluginDescriptor plugin : plugins) {
         if (plugin.active()) {
-          pluginsToInstall.add(plugin);
+          if (retrievePluginMarkerFilePath(plugin) != null && Files.exists(retrievePluginMarkerFilePath(plugin))) {
```
hohwille, line 76: Initially the idea here was to batch the invocations of VSCode plugin installations so we have max 2 process invocations for plugins: all active plugins get installed, all inactive plugins get recommended.

With all the complications caused by this VSCode stuff, I would actually question the entire approach.
Why don't we follow the general approach and do it the same way we do it for all other plugin based tools?
Simply override the single installPlugin method (already done), remove method overriding `installPlugins` and if we really want to stay with the recommended plugins approach at all, we could override `handleInstallForInactivePlugin`.

Please note that the original developer implementing VSCode plugin support implemented this batching of plugins but for active plugins it is already not working anymore.
Also that developer raised the idea himself to add inactive plugins as recommendations.
While I liked his idea in the review and I did not want to stop him from bringing in own ideas and improvements it was never part of the original story and if it now starts to make problems, why do we actually need an excuse to the default here for VSCode?

hohwille (approved):
@jan-vcapgemini thank you very much for creating the bug issue and this fixing PR 👍 

Could you add the story to our CHANGELOG since this is a critical bug that you fixed?
Further, I left a comment for simplification since I think we are causing pointless maintenance issues here that we could simply avoid.
We could also merge as is and create a new PR for the simplification if you think that is necessary or better, but I would suggest such cleanup sooner or later since the redundancies and problems of this could be avoided by design.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
jan-vcapgemini: I've replaced the overwrite of installPlugins with handleInstallForInactivePlugin now.

hohwille (approved):
@jan-vcapgemini thanks for the rework. Nice simplification 👍 
I have not tested this manually but I fear that for many disabled plugins the recommendations may still be still slow and performed every time I start vscode since for N plugins (e.g. 50) we have to read and write the same JSON file 50 times. I would have even kicked them out entirely. However, this is not part of the story/bug-ticket. Your change is functionally correct and therefore ready to merge.

Issue #103, opened 2023-10-17
Implement version security checks


hohwille:
As a IDEasy user, I want to get security warnings if I am using outdated software with critical known CVEs so that I can keep my software secure.

This is the [devonfw-ide story 1106](https://github.com/devonfw/ide/issues/1106) to be implemented for IDEasy.

ATTENTION: There is a specialty for `git` that is not typically managed by IDEasy (what might change see #47). For this also have a look at the old PR implementing this story in devonfw-ide.

MattesMrzik:
Currently it seems, only `VersionRange` entries are allowed inside the `security` file. Should we also allow a `VersionIdentifier`? Should there also be a commandlet that lets you add versions to the security file?

jan-vcapgemini:
I've added a first batch of security files in this PR: https://github.com/devonfw/ide-urls/pull/15

jan-vcapgemini:
After discussing this issue we have to answer following questions.

- Should we use CVE's to check internally if a tool version is to be considered as good or bad (based on the severity of its CVE's)?
- Should the severity threshold be adjustable by the user of IDEAsy?
- Are there different CVE tools which can check at runtime if a tool version is not safe?
- What about identically CVE's in different versions of the same tool?
- What about modified CVE descriptions, how to handle the diffs and where?

hohwille:
> Should we use these CVE's to decide internally if a tool version is to be considered as good or bad (based on the severity of its CVE's)?

Definetly, yes.

> Should the severity threshold be adjustable by the user of IDEasy?

There are general two options:
* We use the standard OWASP tool directly inside IDEasy to check for CVEs. This will imply that before every tool download the entire CVE database has to be updated what takes time and slows down the process. However, this will give lots of flexibility (user can then configure the severity threshold and this question is to be answered with "yes") and most recent CVEs and most accurate results.
* We have an internal process that does the CVE checks in our `UpdateURLs` workflow every night. Here we use a fixed severity threshold that is rather high to avoid annoying warnings to every user. The end user only deals with `security.json` files that we add to `ide-urls` repository and works independent of OWASP tooling. Then the answer is "no".

We started with the latter approach but already added too many CVEs with that and then questioned the entire approach.
To be honest it does not seem an easy decision what would be the best way to go.
I still see pros and cons in both approaches.
However, after revisiting everything, I tend for the first option. We can make it configurable and therefore potentially could even allow the security check to be switched off saving performance (wait times for CVE DB).
[...]

PR #1201, opened 2025-04-03
#103: introduce security module

leonrohne27:
### Implements:

- Introduced security module 
- generates according security.json files


cli/src/main/java/com/devonfw/tools/ide/url/model/file/UrlSecurityFile.java
```diff
@@ -12,8 +32,12 @@ public class UrlSecurityFile extends AbstractUrlFile<AbstractUrlToolOrEdition<?,
   /** {@link #getName() Name} of security file. */
   public static final String SECURITY_JSON = "security.json";
 
+  private final Collection<UrlSecurityWarning> urlSecurityWarnings;
+
   private ToolSecurity security;
 
+  private IdeContext context;
+
```
hohwille, line 40: You should not encode such state into the URL model. This will open the "door to hell" (e.g. in the dashboard GUI the `IdeContext` will live for a long time and different projects may cause conflicting side-effects leading to inconsistency).
Please keep your state information and especially `IdeContext` out of these `UrlArtifact` classes.
From IDEasy CLI perspective instances of `UrlArtifact` should be considered as "immutable".
They do have mutable fields for lazy loading as otherwise we would have huge performance problems and they offer setters that shall only used by `UrlUpdater` process.

security/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+    <version>dev-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>ide-security</artifactId>
+
+  <properties>
+    <maven.compiler.source>21</maven.compiler.source>
+    <maven.compiler.target>21</maven.compiler.target>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
```
hohwille, line 18: This is already set in parent POM and inherited.
```suggestion
```

security/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+    <dependency>
+      <groupId>com.devonfw.tools.IDEasy</groupId>
+      <artifactId>url-updater</artifactId>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.devonfw.tools.IDEasy</groupId>
+      <artifactId>url-updater</artifactId>
+      <version>2025.04.001-SNAPSHOT</version>
+      <scope>compile</scope>
```
hohwille, line 35: We introduced `<dependencyManagement>` to our parent `pom.xml`.
All versions should be defined centrally there and other `pom.xml` files should not specify versions.
https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management
https://github.com/devonfw/scm/blob/master/documentation/build/maven-dependency-management.asciidoc

However, first focus on the other issues. We could also merge this and someone else (maven expert) could later create another PR to clean this up...

security/src/main/java/com/devonfw/tools/IDEasy/dev/BuildSecurityJsonFiles.java
```diff
@@ -0,0 +1,440 @@
...
+
+  private static BigDecimal minV2Severity = new BigDecimal("0.0");
+
+  private static BigDecimal minV3Severity = new BigDecimal("0.0");
+
+  private static final Set<String> actuallyIgnoredCves = new HashSet<>();
+
+  private static final IdeContext context = new IdeContextConsole(IdeLogLevel.INFO, null, false);
+  ;
+
```
hohwille, line 91:
```suggestion
```

security/src/main/java/com/devonfw/tools/IDEasy/dev/BuildSecurityJsonFiles.java
```diff
@@ -0,0 +1,440 @@
...
+
+  public static void main(String[] args) {
+
+    run();
+  }
+
+  private static void run() {
+
```
hohwille, line 99: You might want to create an instance of this class in `main` method and then call a non-static method to manage state in the instance if you like. Otherwise this does not seem to add value.
```suggestion
```

security/src/main/java/com/devonfw/tools/IDEasy/dev/BuildSecurityJsonFiles.java
```diff
@@ -0,0 +1,440 @@
...
+  }
+
+  /** Initializes the {@link #CVES_TO_IGNORE}. */
+  private static void initCvesToIgnore() {
+
+    if (CVES_TO_IGNORE.isEmpty()) {
+      // .................CVE..................vendor......product......why was is ignored
+      CVES_TO_IGNORE.add("CVE-2021-36230"); // hashicorp...terraform....https://github.com/anchore/grype/issues/1377
+    }
+  }
```
hohwille, line 438: Simply define this directly in the constant:
```
private static final Set<String> CVES_TO_IGNORE = Set.of("CVE-2021-36230"); // terraform https://github.com/anchore/grype/issues/1377
```

security/src/main/java/com/devonfw/tools/IDEasy/dev/BuildSecurityJsonFiles.java
```diff
@@ -0,0 +1,440 @@
...
+          + " Please contact https://github.com/devonfw/IDEasy and make a request to get this feature implemented.");
+    }
+    double severityDouble;
+    if (vulnerability.getCvssV3() != null) {
+      severityDouble = vulnerability.getCvssV3().getCvssData().getBaseScore();
+    } else {
+      severityDouble = vulnerability.getCvssV2().getCvssData().getBaseScore();
+    }
+    String formatted = String.format(Locale.US, "%.1f", severityDouble);
+    BigDecimal severity = new BigDecimal(formatted);
```
hohwille, line 280: just for curiosity:
Why do we need to convert a `double` value first to a `String` just to convert it to a `BigDecimal`?
There is a constructor taking a `double`: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/math/BigDecimal.html#%3Cinit%3E(double)

However, if there is a rationale behind this, then we should add a comment explaining this workaround...

security/src/main/java/com/devonfw/tools/IDEasy/dev/BuildSecurityJsonFiles.java
```diff
@@ -0,0 +1,440 @@
...
+   *
+   * @param vulnerability the vulnerability determined by OWASP dependency check.
+   * @return the {@link BigDecimal severity} of the vulnerability.
+   */
+  protected static BigDecimal getBigDecimalSeverity(Vulnerability vulnerability) {
+
+    if (vulnerability.getCvssV2() == null && vulnerability.getCvssV3() == null) {
+      // TODO if this ever happens, add a case that handles this. See https://github.com/devonfw/IDEasy/issues/190
+      throw new RuntimeException("Vulnerability without severity found: " + vulnerability.getName() + "\\n"
+          + " Please contact https://github.com/devonfw/IDEasy and make a request to get this feature implemented.");
```
hohwille, line 271: We should not throw an exception.
Our code should be robust.
So simply log a warning and return `null` here.
In the code calling this method, simply ignore the `Vulnerability` if you get `null` as result.

cli/src/main/java/com/devonfw/tools/ide/version/VersionRange.java
```diff
@@ -26,20 +26,13 @@ public final class VersionRange implements Comparable<VersionRange>, GenericVers
...
     this.min = min;
     this.max = max;
     this.boundaryType = boundaryType;
-    if ((min != null) && (max != null) && min.isGreater(max)) {
-      throw new IllegalArgumentException(toString());
-    } else if ((min == null) && !boundaryType.isLeftExclusive()) {
-      throw new IllegalArgumentException(toString());
-    } else if ((max == null) && !boundaryType.isRightExclusive()) {
-      throw new IllegalArgumentException(toString());
-    }
```
hohwille, line 42: why do you remove the validation of the version range?
Why should we allow to create a `VersionRange` where `min` is greater than `max`?
That is inconsistent. If CVE data contains crap, then the code processing the data should filter out such "crap" rather than removing our validation...

cli/src/main/java/com/devonfw/tools/ide/version/VersionRange.java
```diff
@@ -26,20 +26,13 @@ public final class VersionRange implements Comparable<VersionRange>, GenericVers
    * @param max the {@link #getMax() maximum}.
    * @param boundaryType the {@link BoundaryType} defining whether the boundaries of the range are inclusive or exclusive.
    */
-  private VersionRange(VersionIdentifier min, VersionIdentifier max, BoundaryType boundaryType) {
+  public VersionRange(VersionIdentifier min, VersionIdentifier max, BoundaryType boundaryType) {
```
hohwille, line 29: we intentionally made the constructor private because we provide static `of` methods that act as factory to create `VersionRange` instances.

pom.xml
```diff
@@ -165,6 +165,7 @@
     <module>cli</module>
     <module>gui</module>
     <module>url-updater</module>
+      <module>security</module>
```
hohwille, line 168: Don't we have IDEasy configured to autoformat XML and fix indentation?
```suggestion
    <module>security</module>
```

security/src/main/java/com/devonfw/tools/IDEasy/dev/UrlAnalyzer.java
```diff
@@ -0,0 +1,122 @@
...
+        this.fileFilter = new UrlFileFilter();
+        this.updateManager = updateManager;
+    }
+
+    /**
+     * Analyzes the given {@link Dependency} and adds {@link Evidence} to it. Namely, CPE vendor, product, edition and version.
+     *
+     * @param dependency the dependency to analyze.
+     * @param engine     not actually used here, but needs to match the signature of the super class.
+     */
```
hohwille, line 47: Omit redundant JavaDoc if `@Override` is present.
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/version/VersionRangeMapper.java
```diff
@@ -0,0 +1,20 @@
+package com.devonfw.tools.ide.version;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class VersionRangeMapper {
```
hohwille, line 6: This class should be replaced with a `JsonSerializer`.
Still I do not understand why we need this at all.

In #103 we specified the JSON format as a contract between your part and the part of @KianRolf 
Reading your summary of this PR gives me the impression you have defined your own JSON format:
```
{
"severity" : 8.6,
"versionRange" : {
"min" : "9.0.0",
"max" : "9.0.21",
"boundaryType" : "RIGHT_OPEN"
},...
```
Please keep the JSON format we all agreed on so that parts fit together and make sense.
As discussed we do not need `nistUrl` or `description` and this will only cause pointless waste for data we do not need and that would need to update multiple times. We want to keep our `ide-urls` repository as small and compact as possible.

security/src/main/java/com/devonfw/tools/IDEasy/dev/BuildSecurityJsonFiles.java
```diff
@@ -0,0 +1,440 @@
...
+
+  public static void main(String[] args) {
+
+    run();
+  }
+
+  private static void run() {
+
+    initCvesToIgnore();
+    UpdateManager updateManager = new UpdateManager(context.getUrlsPath(), null, Instant.now());
```
hohwille, line 101: `context.getUrlsPath()` is working if the context is properly initialized, but you are doing `new IdeContextConsole(IdeLogLevel.INFO, null, false)` so this will not work.
Since this is running as a GitHubAction and not inside an IDEasy project, you should provide this `Path` directly without using `IdeContext`.

cli/src/main/java/com/devonfw/tools/ide/url/model/file/UrlSecurityFile.java
leonrohne27: My bad, I think I used this for testing but forgot to remove it afterwards

cli/src/main/java/com/devonfw/tools/ide/url/model/AbstractUrlMetadata.java
```diff
@@ -24,6 +24,8 @@ public interface AbstractUrlMetadata {
    */
   List<VersionIdentifier> getSortedVersions(String tool, String edition, ToolCommandlet toolCommandlet);
 
+  List<VersionIdentifier> getSortedVersions(String tool, String edition);
```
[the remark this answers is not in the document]
leonrohne27: I didn't understand the third paramter/how to us this method with the parameter. In the PR:
 https://github.com/devonfw/IDEasy/pull/119/files#diff-ba4bb3b2aecc3dd617eb6b73690e836302892c83e06688eae05795a4f1e96fd4
this method did not have any third parameter and was used, so I thought it works out

cli/src/main/java/com/devonfw/tools/ide/version/VersionRange.java
leonrohne27: This code section threw some stange error if I remember correctly. 
I also checked the outputs/ entries of the created data and min was never greater than max, but if you wish I will implement and test again

security/src/main/java/com/devonfw/tools/IDEasy/dev/BuildSecurityJsonFiles.java
leonrohne27: I've seen it like this so I just used it, but u are right
return BigDecimal.valueOf(severityDouble);
is much simpler

cli/src/main/java/com/devonfw/tools/ide/url/model/AbstractUrlMetadata.java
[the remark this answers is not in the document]
hohwille: Copy&paste is never a good approach. Better use the existing method. If you are calling this from a `ToolCommandlet` just pass `this` as 3rd argument.

cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/UrlSecurityWarning.java
```diff
@@ -0,0 +1,191 @@
...
+
+import java.math.BigDecimal;
+import java.util.Optional;
+
+import com.devonfw.tools.ide.version.VersionRange;
+
+/**
+ * A simple container with the information about a security warning.
+ */
+public class UrlSecurityWarning {
```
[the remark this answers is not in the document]
leonrohne27: You are absolutely right, I changed my implementation, please have a look at my new commit

security/src/main/java/com/devonfw/tools/IDEasy/dev/BuildSecurityJsonFiles.java
leonrohne27: fixed

cli/src/main/java/com/devonfw/tools/ide/version/VersionRange.java
leonrohne27: fixed

cli/src/main/java/com/devonfw/tools/ide/version/VersionRangeMapper.java
leonrohne27: fixed

security/pom.xml
hohwille: Thanks for the rework to use `${project.version}`.
The rest could also be tidied up after the merge (remove `compile`, declare versions in `<dependencyManagement>` and omit here).

security/src/main/java/com/devonfw/tools/IDEasy/dev/BuildSecurityJsonFiles.java
hohwille: Conversion code entirely gone now. Even better 👍

hohwille (commented):
@leonrohne27 thanks for your rework. You did a great job to resolve the review comments and reached a great improvement. 👍 
I left some more comments. Most of them trivial to resolve.
Lets talk tomorrow about the rest. IMHO we are close to complete and merge this PR. 🚀

cli/src/main/java/com/devonfw/tools/ide/url/model/file/UrlSecurityFile.java
```diff
@@ -41,7 +65,97 @@ protected void doLoad() {
...
+  }
+
+  /**
+   * Adds a new CVE warning with detailed information, such as severity, CVE ID and a versionRange
+   */
+
+  public void addCve(CVE cve) {
+    if (this.security == null || this.security == ToolSecurity.getEmpty()) {
+      this.security = new ToolSecurity(new ArrayList<>());
```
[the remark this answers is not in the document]
leonrohne27: Why this change? Implemented like this, getIssues() will throw a NullPointerException.
If you want that change though we can find a workaround, but it works out like this as well.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/UpdateManager.java
```diff
@@ -127,4 +127,23 @@ private void update(AbstractUrlUpdater updater) {
     }
   }
 
+  public AbstractUrlUpdater retrieveUrlUpdater(String tool, String edition) {
+
+    for (AbstractUrlUpdater updater : updaters) {
+      if (updater.getTool().equals(tool) && edition.equals("intellij")) {
+        return updater;
+      }
```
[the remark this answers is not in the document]
leonrohne27: Removed it, everything still works 👍

cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/ToolSecurity.java
```diff
@@ -25,11 +25,11 @@ public class ToolSecurity {
   private List<CVE> issues;
 
 
-  private ToolSecurity() {
+  public ToolSecurity() {
     super();
```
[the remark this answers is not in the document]
leonrohne27: 👍

cli/src/main/java/com/devonfw/tools/ide/url/model/file/UrlSecurityFile.java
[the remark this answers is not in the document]
leonrohne27: 👍

cli/pom.xml
```diff
@@ -110,6 +110,12 @@
       <artifactId>xmlunit-assertj3</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.owasp</groupId>
+      <artifactId>dependency-check-core</artifactId>
+      <version>12.1.0</version>
+      <scope>test</scope>
+    </dependency>
```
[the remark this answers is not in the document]
leonrohne27: 👍

cli/src/main/java/com/devonfw/tools/ide/json/JsonMapping.java
```diff
@@ -41,6 +41,7 @@ public static ObjectMapper create() {
     customModule.addSerializer(CustomToolJson.class, new CustomToolJsonSerializer());
     customModule.addSerializer(CustomToolsJson.class, new CustomToolsJsonSerializer());
     customModule.addKeyDeserializer(VersionRange.class, new VersionRangeKeyDeserializer());
+    customModule.addSerializer(VersionRange.class, new VersionRangeSerializer());
```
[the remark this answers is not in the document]
leonrohne27: I removed it and tested if everything still works off but I got a huge error code printed.
Besides that it has no usage so I don't get why this happens.
If you want me to further research this I can do it, else we could just leave it like that.

cli/src/main/java/com/devonfw/tools/ide/tool/repository/DefaultToolRepository.java
```diff
@@ -86,4 +86,5 @@ public List<VersionIdentifier> getSortedVersions(String tool, String edition, To
 
     return this.context.getUrls().getSortedVersions(tool, edition, toolCommandlet);
   }
+
```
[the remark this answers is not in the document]
leonrohne27: 👍

security/src/main/java/com/devonfw/tools/IDEasy/dev/BuildSecurityJsonFiles.java
```diff
@@ -0,0 +1,442 @@
...
+      NodeAuditAnalyzer.class, YarnAuditAnalyzer.class, PnpmAuditAnalyzer.class, RetireJsAnalyzer.class,
+      FalsePositiveAnalyzer.class);
+
+  private static BigDecimal minV2Severity = new BigDecimal("0.0");
+
+  private static BigDecimal minV3Severity = new BigDecimal("0.0");
+
+  private static final Set<String> actuallyIgnoredCves = new HashSet<>();
+
+  private static final IdeContext context = new IdeContextConsole(IdeLogLevel.INFO, null, false);
```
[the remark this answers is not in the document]
leonrohne27: I have now passed the context as a parameter to every method that makes a console output, I don't know if you imagined it that way but please have a look

cli/src/main/java/com/devonfw/tools/ide/util/MapUtil.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Utility class for operations with maps.
+ */
+public class MapUtil {
```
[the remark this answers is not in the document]
leonrohne27: Map.of only seems to work for up to 10 entries, so I used a loop, please have a look.

cli/src/main/java/com/devonfw/tools/ide/json/JsonMapping.java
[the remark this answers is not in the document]
hohwille: Got it: toString is not default serializer since `VersionRange` is considered as bean so Jackson will by default serialise an object with the properties of `VersionRange`.
We should also verify that no existing code that produces JSON using a `VersionRange` does not break because of this change.

cli/src/main/java/com/devonfw/tools/ide/json/VersionRangeSerializer.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  public void serialize(VersionRange versionRange, JsonGenerator jsonGenerator, SerializerProvider serializerProvider) throws IOException {
+
+    //jsonGenerator.writeStartObject();
+    //jsonGenerator.writeFieldName("versions");
+    //jsonGenerator.writeStartArray();
+    jsonGenerator.writeString(versionRange.toString());
+    //jsonGenerator.writeEndArray();
+    //jsonGenerator.writeEndObject();
+
```
hohwille, line 21: clean code
```suggestion
    jsonGenerator.writeString(versionRange.toString());
```

cli/src/main/java/com/devonfw/tools/ide/util/MapUtil.java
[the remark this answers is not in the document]
hohwille: Where is the usage or use-cases for this `MapUtil`?
I cannot see any usage nor any reason why we need this.
Therefore I propose to remove it.

security/src/main/java/com/devonfw/tools/IDEasy/dev/BuildSecurityJsonFiles.java
```diff
@@ -0,0 +1,448 @@
...
+      FalsePositiveAnalyzer.class);
+
+  private static BigDecimal minV2Severity = new BigDecimal("0.0");
+
+  private static BigDecimal minV3Severity = new BigDecimal("0.0");
+
+  private static final Set<String> actuallyIgnoredCves = new HashSet<>();
+
+  public static void main(String[] args) {
+    IdeContext context = new IdeContextConsole(IdeLogLevel.INFO, null, false);
```
hohwille, line 88: Please remove `IdeContext` from url-updater.
```suggestion
```

You can use regular logger like defined in URL updaters:
https://github.com/devonfw/IDEasy/blob/d13807b5795a61de7d2c907840ffd769ede22f68/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java#L82

Constants should actually be UPPER_CASE (`LOG` is the common field name I use for loggers).

leonrohne27: done

hohwille (approved):
@leonrohne27 thanks for your improvement of all my review feedback and completing JavaDoc, etc. All looks very good now 👍 
I left some very last comments to address. IMHO you can resolve that very easily and quickly so we could finally merge this week. 🚀

security/src/main/java/com/devonfw/tools/IDEasy/dev/BuildSecurityJsonFiles.java
```diff
@@ -0,0 +1,471 @@
...
+import com.devonfw.tools.ide.url.updater.AbstractUrlUpdater;
+import com.devonfw.tools.ide.url.updater.UpdateManager;
+import com.devonfw.tools.ide.version.BoundaryType;
+import com.devonfw.tools.ide.version.VersionIdentifier;
+import com.devonfw.tools.ide.version.VersionRange;
+
+/**
+ * This class is used to build the {@link UrlSecurityFile} files for IDEasy. It scans the {@link AbstractIdeContext#getUrlsPath() ide-url} folder for all tools,
+ * editions and versions and checks for vulnerabilities by using the OWASP package. You must pass two arguments to the main method: minV2Severity and
+ * minV3Severity. V2 and V3 severity are differentiated, because they are often considered not comparable. Vulnerabilities with severity lower than these values
```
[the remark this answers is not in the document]
leonrohne27: 👍

security/src/main/java/com/devonfw/tools/IDEasy/dev/BuildSecurityJsonFiles.java
```diff
@@ -0,0 +1,471 @@
...
+    context.debug("The severity of the vulnerability is {}.", severity);
+    context.debug("The description of the vulnerability is {}.", description);
+    context.debug("The NIST url of the vulnerability is {}.", nistUrl);
+    context.debug(
+        "The determined VersionRange is {}, given OWASPs versionStartExcluding {}, versionStartIncluding {}, "
+            + "versionEndIncluding {}, versionEndExcluding {}, single Version {} ",
+        versionRange, matchedVulnerableSoftware.getVersionStartExcluding(),
+        matchedVulnerableSoftware.getVersionStartIncluding(), matchedVulnerableSoftware.getVersionEndIncluding(),
+        matchedVulnerableSoftware.getVersionEndExcluding(), matchedVulnerableSoftware.getVersion());
+  }
```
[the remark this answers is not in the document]
leonrohne27: I removed it, this was only to test while developing I don't think we need it anymore

Issue #1202, opened 2025-04-03
IDEasy can not reset urls on WSL


jan-vcapgemini:
### Expected behavior

As an IDEasy user on WSL I expect IDEasy to properly reset my urls repository if changes were detected.

### Actual behavior
```
fatal: ambiguous argument 'origin/null': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Running command '/usr/bin/git' with arguments 'reset' '--hard' 'origin/null'
failed with exit code 128!
Git failed to reset origin to 'null//home/<username>/projects/_ide/urls'.
Failed to update git repository at /home/<username>/projects/_ide/urls
See above error for details. If you have local changes, please stash or revert and retry.
Typically you should abort and fix the problem. Do you want to continue anyways?
Option 1: yes
Option 2: no
```

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Try to install IDEasy on your WSL
2. run `ide install java`

### Affected version:

* OS: Windows/Linux/Mac?
* Browser: Chrome/Firefox/Safari?

hohwille:
I do not see that this bug is anyhow related to WSL.

PR #1203, opened 2025-04-03
#1202: Added branchname fallback to master

jan-vcapgemini:
Fixes: #1202

**Implements**
* fallback to master if branch name was null


hohwille (approved):
@jan-vcapgemini thanks for the quickfix. As already discussed the proper default branch name has to be determined from git rather than hardcoding it. However, to fix the bug this is already a good improvement and `master` is already better than `null` ;)

Issue #583, opened 2024-09-03
Create MVP for Integration Tests


hohwille:
We need integration tests for IDEasy.
Details can be found in epic #572.

This story is about implementing a minimal viable product or in other words the smallest possible first step to get some first integration testing added to IDEasy that can later be extended.

Before implementing anything please first read #572 and come up with an idea how to implement it.
Discuss that in our daily or Tue meeting and start implementation after we agreed on that.

hohwille:
Idea is to create a bash file with functions sourced before execution of the actual test (should IMHO be sourced by the test runner and not every individual test script to make it simpler than it was in devonfw-ide).
There we can define a function `doIdeCreate()` that does the following:
* `local project-name="$(dirname "${BASH_SOURCE:-$0}")"` (determine the name of the currently executed script)
* `local settings-url=${1:--}` (set local variable `settings-url` to `-` as default but if an argument was given to the function use that arg to override default URL.
* `ide create ${project-name} ${settings-url}`
* `cd ${IDE_ROOT}/${project-name}`

then most integration test scripts can start with
`doIdeCreate`
and then do whatever they want to test...
`ide install mvn`

However an integration test of the help functionality would not call `doIdeCreate`.

PR #1207, opened 2025-04-04
#583: integration test mvp

jan-vcapgemini:
Continues: #752 Thanks to @nmCapG for his preparations.
Fixes: #583

**Implements:**
* Various integration tests for IDEasy tools
* simple github actions workflow and logic for handling matrix.os
* added integration tests to check-for-updates (triggers integration tests after nightly build)
* added some assertions to help test
* added `assertThat` function to help test
* added integration-tests documentation
* removed documentation and headers from test files


cli/src/test/all-tests.sh
```diff
@@ -0,0 +1,147 @@
...
+
+    source "${testpath:?}"
+
+    # Remove trap
+    # Disable errexit in case it was turned on in a test script
+    trap - ERR
+    set +e
+
+    echo "RESULTS:"
+    echo "$integration_test_result"
```
alfeilex, line 56: Should this `echo` be visible in the logs or is it there for debugging?

alfeilex (commented):
I made some small comments. Further, I have tested the workflow on my own branch and there is some problem with `macos-latest` - maybe it's because of my settings? Not sure.

https://github.com/alfeilex/IDEasy/actions/runs/14387615458/job/40346799077#step:3:67

The runs on windows and linux were successful.

cli/src/test/integration-tests/install-intellij.sh
```diff
@@ -0,0 +1,18 @@
+#--- Preamble ---
+# Working dir (PWD) is IDE_ROOT (it should contain an ide installation)
+# Inherited variables:
+#    - $IDE      := IDEasy ($PWD/_ide/bin/ide)
```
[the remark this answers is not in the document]
jan-vcapgemini: I've removed it now because the IDE variable should not be used. Instead IDEasy can be called by `ide` now.

cli/src/test/all-tests.sh
jan-vcapgemini: I've removed it now.

cli/src/test/all-tests.sh
```diff
@@ -0,0 +1,147 @@
...
+  # rm -rf "${DEBUG_INTEGRATION_TEST_PREFIX}"
+  # Only need to mkdir once:
+  echo "Creating IDEasy directory at: ${IDEASY_DIR}"
+  mkdir -p "${IDEASY_DIR}"
+
+  echo "Switching directory to: ${IDEASY_DIR}"
+  cd "${IDEASY_DIR}" || exit
+
+  # Will call IDEasy with variable
+  #ide="${IDE_ROOT}/_ide/bin/ide"
```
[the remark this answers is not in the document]
jan-vcapgemini: Yup.

cli/src/test/all-tests.sh
```diff
@@ -0,0 +1,132 @@
...
+    # Remove trap
+    # Disable errexit in case it was turned on in a test script
+    trap - ERR
+    set +e
+
+    echo "RESULTS:"
+    if [ "$integration_test_result" == 0 ]; then
+      doSuccess "[SUCCESS] Succeeded running test #${total}: ${testcase}"
+      ((success++))
+      echo -e "\033[92m[SUCCESS] Succeeded running test #${total}: ${testcase}\033[39m" >> "${TEST_RESULTS_FILE:?}"
```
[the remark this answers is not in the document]
alfeilex: It has already been solved in the latest commits.

hohwille:
> I made some small comments. Further, I have tested the workflow on my own branch and there is some problem with `macos-latest` - maybe it's because of my settings? Not sure.
> 
> https://github.com/alfeilex/IDEasy/actions/runs/14387615458/job/40346799077#step:3:67
> 
> The runs on windows and linux were successful.

Already filed as bug #1223 and fixed with PR #1251

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,23 @@
+name: Integration Tests
+on:
+  workflow_dispatch
```
hohwille, line 3: fine if we start manually, but the idea is to run this automatically every night.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,23 @@
...
+        os: [ windows-latest, ubuntu-latest, macos-latest, macos-13 ]
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Run integration tests
+        shell: bash
+        run: |
+          mkdir -p ${HOME}/.ide
+          touch ${HOME}/.ide/.license.agreement
```
hohwille, line 21: cant we do such things in `all-tests.sh`?
I prefer to do as little logic into workflow files as possible because they are the hardest part to test & maintain.

cli/src/test/all-tests.sh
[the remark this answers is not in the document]
hohwille: @alfeilex Can you then close this review comment? Seems to be fixed, right?

cli/src/test/all-tests.sh
```diff
@@ -0,0 +1,132 @@
...
+
+test_files_directory=$(realpath "$0" | xargs dirname)
+
+success=0
+failure=0
+total=0
+
+function doTestsInner() {
+  # Note: requires var test_files_directory to be set.
+  echo "Entering doTestsInner..."
```
hohwille, line 33: looks like trace logging. Can we remove this now?
```suggestion
```

cli/src/test/all-tests.sh
```diff
@@ -0,0 +1,132 @@
...
+    testcase="${testpath/*\//}"
+    echo "Running test #${total}: ${testcase} (${testpath})"
+
+    integration_test_result=0
+
+    # Following line adds a trap to ERR signal: Whenever error (some
+    # conditions apply; see documentation) set integration_test_result=1.
+    trap 'doWarning "A non-handled error in integration test occurred."; integration_test_result=1' ERR
+
+    source "${testpath:?}"
```
hohwille, line 44: if one test "crashes" then the entire test-suite will "crash" and next tests will not be executed?
Or is that handled by the `trap`?
If so - great solution but then put stuff like `set -e` here instead of copy&pasting it into each individual test.
Otherwise we should include/source a common "script" in all tests so we can later extend things without refactoring every individual test.

cli/src/test/integration-tests/help.sh
```diff
@@ -0,0 +1,4 @@
+#--- Body ---
+set -e
+echo "Running help integration test"
+ide help
```
hohwille, line 4: just like JUnit it is best practice to assert the results.
This test only verifies that `ide help` will not cause an error but if the help output is totally messed up, it will still be green.
Shouldn't we verify that some keywords are found in the output?

cli/src/test/integration-tests/install-intellij.sh
```diff
@@ -0,0 +1,17 @@
...
+#    - $IDE_ROOT := PWD
+#    - All functions in src/test
+# Use 'return <EXIT_CODE>' (instead of exit) to signal if test passed/failed.
+
+# For debugging integration scripts (like this script), it is a good idea to
+# fail whenever an error occurred (set -e). When you finished writing the
+# test, however, please remove your 'set -e' line, since in production it will
+# exit and not run subsequent integration tests. 
+#set -e 
+
```
hohwille, line 13: We should rather address this with documentation in a central place instead of placing or copy&pasting this into individual tests.

cli/src/test/debug.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+## source "$(dirname "${0}")"/functions-test
+
+echo my PWD is: $PWD
```
hohwille, line 5: This file does not seem to be used anywhere...
Was it included by accident as a leftover from local debugging?
Can we remove it?

cli/src/test/functions-test.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
```
hohwille, line 1: In `devonfw-ide` this file was included by all individual tests and therefore named this way.
If this is not the case here, couldn't we simply move this code to `all-tests.sh` or rename it (e.g. to `all-tests-functions.sh`)?

hohwille (changes requested):
@jan-vcapgemini & @alfeilex thanks for this PR. This is great work and looks promising. 👍 
Seems you have been able to recycle quite some code from `devonfw-ide` and apply it to IDEasy. Some things have changed (e.g. `source` individual tests instead of executing them as sub-process or sourcing comming functions in individual tests) and could therefore now receive some cleanup.
Further, all tests have no assertions at all.
IMHO we should first have a blue-print of a good test before inviting other developers to write more integration-tests...
We can still merge things and create follow-up PRs, lets just discuss how to proceed.

.github/workflows/integration-tests.yml
jan-vcapgemini: Moved it to `all-tests.sh` now.

cli/src/test/debug.sh
jan-vcapgemini: Removed now.

cli/src/test/functions-test.sh
jan-vcapgemini: Renamed to `all-tests-functions.sh` now.

cli/src/test/all-tests.sh
jan-vcapgemini: Removed.

.github/workflows/integration-tests.yml
jan-vcapgemini: Added integration tests to check-for-updates.

cli/src/test/all-tests.sh
jan-vcapgemini: I've moved all `set -e` from the individual tests to the function scripts now instead.

cli/src/test/integration-tests/help.sh
jan-vcapgemini: I've added an `assertThat` function and some tests to the help test now.

cli/src/test/integration-tests/install-intellij.sh
jan-vcapgemini: Move this to a new documentation file.

cli/src/test/all-tests.sh
hohwille: So your answer to my first question is "yes"? Then I am happy 😄

cli/src/test/integration-tests/help.sh
```diff
@@ -0,0 +1,21 @@
...
+  local keyword="$2"
+
+  if echo "$output" | grep -q "$keyword"; then
+    echo "Assertion passed: '$keyword' found in output"
+    return 0
+  else
+    echo "Assertion failed: '$keyword' not found in output"
+    return 1
+  fi
+}
```
hohwille, line 15: Nice, but this is very different to what AssertJ and Java OOP can offer.
This actually implements a contains check.
Further the function returns `1` on error but the result is ignored.
You need to `exit` instead of `return` so the test actually fails on error.
This also will reveal my point if a "crashing" test will bring the entire control script down or not.

hohwille: If you want to stay with the `assertThat` pattern, you would need to do a syntax like this:
```
assertThat "$output" contains "Current version"
```
And we could implement further comparison operators later like:
```
assertThat "$versionOutput" equals "SNAPSHOT"
```
etc.

cli/src/test/integration-tests/install-intellij.sh
```diff
@@ -0,0 +1,2 @@
+echo "Running install intellij integration test"
+ide -d install intellij
```
hohwille, line 2: Here we should assert that the software folder actually contains intellij by checking some marker files.
Feel free to remove tests (MVP, KISS) from this PR but when I do a review, I keep on raising concerns and expect that a test we have as a starting point is a good example and blueprint. Otherwise we will have the "broken window" effect and other developers will see this and think, this is a good test and will follow to do the same pattern.

cli/src/test/integration-tests/help.sh
jan-vcapgemini: I've adjusted the `help.sh` test now and moved the `assertThat` function to `all-tests-functions.sh`

.github/workflows/integration-tests.yml
jan-vcapgemini: I've added another job now which requires the nightly build to be run first.

cli/src/test/integration-tests/install-intellij.sh
jan-vcapgemini: I've added a new `assertThat 'file' exists` and checked if the version file and the binary of `intellij` exists.

hohwille (approved):
@jan-vcapgemini thanks for the rework. You perfectly addressed all review remarks. 👍 
Ready for merge.

Issue #1037, opened 2025-02-14
devonfw-ide compatibility: Attribute merge:id is required


hohwille:
### Expected behavior

As a devonfw-ide user, I want to use IDEasy on my devonfw-ide project so that I can test things before my entire teams goes for a migration to IDEasy.

### Actual behavior

With `devon intellij` I can launch IntelliJ without errors but with IDEasy I get this:

```
$ ide intellij
...
Failed to merge workspace file D:\projects\project\workspaces\main\.intellij\config\options\jdk.table.xml with update template D:\projects\project\settings\intellij\workspace\update\.intellij\config\options\jdk.table.xml and setup file D:\projects\project\settings\intellij\workspace\setup\.intellij\config\options\jdk.table.xml!
com.devonfw.tools.ide.merge.xmlmerger.XmlMergeException: Merge strategy combine failed on /application[]/component[@name='ProjectJdkTable']
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMergeStrategy.merge(XmlMergeStrategy.java:69)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMergeStrategy.combineChildNodes(XmlMergeStrategy.java:113)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMergeStrategy$1.doMerge(XmlMergeStrategy.java:29)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMergeStrategy.merge(XmlMergeStrategy.java:65)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMerger.merge(XmlMerger.java:123)
        at com.devonfw.tools.ide.merge.xmlmerger.XmlMerger.doMerge(XmlMerger.java:82)
        at com.devonfw.tools.ide.merge.FileMerger.merge(FileMerger.java:50)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:69)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:73)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:73)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:73)
        at com.devonfw.tools.ide.merge.DirectoryMerger.merge(DirectoryMerger.java:73)
[...]

PR #1212, opened 2025-04-08
#1037: Fix devonfw-ide compatibility with XML templates

alfeilex:
Fixes: #1037 

This PR includes:
- add new environment variable `IDE_XML_MERGE_LEGACY_SUPPORT_ENABLED`
- handle the compatibility of merging XML templates within legacy devonfw-ide projects
- add related tests


cli/src/main/java/com/devonfw/tools/ide/merge/xml/XmlMerger.java
```diff
@@ -119,6 +124,13 @@ public Document merge(XmlMergeDocument templateDocument, XmlMergeDocument worksp
       if (strategy == null) {
         strategy = XmlMergeStrategy.COMBINE; // default strategy used as fallback
       }
+      if (this.legacyXmlSupport && (templateRoot.lookupNamespaceURI("merge") == null)) {
```
alfeilex, line 127: Is it enough to check for `null` or should we check if its not equal `MERGE_NS_URI`

cli/src/test/java/com/devonfw/tools/ide/merge/xml/XmlMergerTest.java
```diff
@@ -91,4 +92,39 @@ void testFailOnAmbiguousMerge(@TempDir Path tempDir) throws Exception {
...
+    Path workspaceSetupPath = DEVONFW_IDE_PATH.resolve("workspaces").resolve("main").resolve("setup.xml");
+    Path workspaceUpdatePath = DEVONFW_IDE_PATH.resolve("workspaces").resolve("main").resolve("update.xml");
+    Path workspaceUpdateWithNsPath = DEVONFW_IDE_PATH.resolve("workspaces").resolve("main").resolve("updateWithNs.xml");
+    Path workspaceUpdateCombinedPath = DEVONFW_IDE_PATH.resolve("workspaces").resolve("main").resolve("combinedUpdate.xml");
+    XmlMerger merger = new XmlMerger(context);
+
+    // act
+    merger.doMerge(settingsSetupPath, settingsUpdatePath, variables, workspaceSetupPath);
+    merger.doMerge(settingsSetupPath, settingsUpdatePath, variables, workspaceUpdatePath);
+    merger.doMerge(settingsSetupPath, settingsUpdateWithNsPath, variables, workspaceUpdateWithNsPath);
```
alfeilex, line 122: Are these tests enough or do I miss something here? Or could I even improve these in any way?

cli/src/main/java/com/devonfw/tools/ide/merge/xml/XmlMerger.java
```diff
@@ -119,6 +124,17 @@ public Document merge(XmlMergeDocument templateDocument, XmlMergeDocument worksp
...
+        if (workspaceFileExists) {
+          strategy = XmlMergeStrategy.OVERRIDE;
+        } else {
+          strategy = XmlMergeStrategy.KEEP;
+        }
+      } else if (templateRoot.lookupNamespaceURI("merge") == null) {
+        this.context.warning(
+            "XML merge namespace not found. If you are working in a legacy devonfw-ide project, please set IDE_XML_MERGE_LEGACY_SUPPORT_ENABLED=true to "
+                + "proceed correctly.");
```
alfeilex, line 136: Maybe there is another option to inform the user that the legacy environment variable has to be set to enable a correct merge behavior. Any suggestion?

julia-cap (approved):
Review done without any requests for change.
- I installed the latest devonfw in a dedicated directory and successfully executed "devon intellij"
- I was able to reproduce the bug
- I checked out the PR branch and added the new environment variable to devon.properties
- Then, I executed "ide intellij" inside the devon working directory. 
I got the same warning message 4 times. But apart from that, IntelliJ opened as expected. 
All tests pass.
Maven build passes.

cli/src/main/java/com/devonfw/tools/ide/merge/xml/XmlMerger.java
hohwille: Actually in XML the namespace URL is the important thing and the `prefix` is just an alias to the URI.
So users could also name the prefix `merge2` or `ide`.
So Ideally you would do it the other way around:
```suggestion
      if (this.legacyXmlSupport && (templateRoot.lookupPrefix(MERGE_NS_URI) == null)) {
```

cli/src/test/java/com/devonfw/tools/ide/merge/xml/XmlMergerTest.java
hohwille: Great that you added a test for this feature. The test is sufficient in general.
However, what you could improve is that the merge-namespace in `updateWithNs.xml` is actually used to do something so that `update.xml` and `updateWithNs.xml` are not actually identical and the namespace makes a real difference.

cli/src/main/java/com/devonfw/tools/ide/merge/xml/XmlMerger.java
```diff
@@ -119,6 +124,17 @@ public Document merge(XmlMergeDocument templateDocument, XmlMergeDocument worksp
       if (strategy == null) {
         strategy = XmlMergeStrategy.COMBINE; // default strategy used as fallback
       }
+      if (this.legacyXmlSupport && (templateRoot.lookupPrefix(MERGE_NS_URI) == null)) {
+        if (workspaceFileExists) {
+          strategy = XmlMergeStrategy.OVERRIDE;
+        } else {
+          strategy = XmlMergeStrategy.KEEP;
+        }
+      } else if (templateRoot.lookupPrefix(MERGE_NS_URI) == null) {
```
hohwille, line 133: Please consider to avoid 
```
if (conditionA && conditionB) {
 ...block1...
} else if (conditionB) {
  ...block2...
}
```
and instead do
```
if (conditionB) {
  if (conditionA) {
    ...block1...
  } else {
    ...block2...
  }
}
```
to avoid the redundancy (then you would have to change it only once and not twice)...

hohwille (approved):
@alfeilex thanks for your PR. This is looking great. 👍
Also amazing how I can see your updates while doing the reivew - you are supersoning ⚡

I left some review comments for further improvement.
Please have a look then we can merge.

cli/src/test/java/com/devonfw/tools/ide/merge/xml/XmlMergerTest.java
```diff
@@ -91,4 +94,67 @@ void testFailOnAmbiguousMerge(@TempDir Path tempDir) throws Exception {
...
+  @ParameterizedTest
+  @ValueSource(strings = { "windows", "mac", "linux" })
+  void testLegacySupportXmlMerge(String os) {
+
+    // arrange
+    String PROJECT_DEVONFW_IDE = "devonfw-ide";
+    IdeTestContext context = newContext(PROJECT_DEVONFW_IDE);
+    SystemInfo systemInfo = SystemInfoMock.of(os);
+    context.setSystemInfo(systemInfo);
+    Path DEVONFW_IDE_PATH = TEST_PROJECTS_COPY.resolve(PROJECT_DEVONFW_IDE).resolve("project");
```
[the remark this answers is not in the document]
alfeilex: done

cli/src/main/java/com/devonfw/tools/ide/merge/xml/XmlMerger.java
alfeilex: done

cli/src/test/java/com/devonfw/tools/ide/merge/xml/XmlMergerTest.java
alfeilex: Done, but please take another look. I am not sure if I got your comment right

cli/src/main/java/com/devonfw/tools/ide/merge/xml/XmlMerger.java
hohwille: Totally fine for me. Legacy users can get spammed until they cleanup their stuff ;)

Issue #1217, opened 2025-04-08
Remove plugin recommendation feature in VSCode


alfeilex:
Currently, we support a feature where plugins in VSCode set to `plugin_active=false` are added as recommended plugins within VSCode. After discussing this feature, we have decided to remove it entirely from IDEasy. 

The reason for this decision is that we want to avoid implementing specific workarounds, such as for installing plugins in IDEs like VSCode or IntelliJ. Instead, we prefer to rely on officially documented APIs and CLI commands provided by the IDE itself. Since the recommendation feature represents such a workaround, we have decided to remove it.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1.  Set some VSCode plugins to `false` in the settings and run `ide vscode`
2. Check the output

```
$ ide vscode
Start: Configuring workspace main for IDE vscode
Successfully ended step 'Configuring workspace main for IDE vscode'.
Step 'Add recommendation for vscode and plugin: angular-console' ended with failure.
Step 'Add recommendation for vscode and plugin: angular2' ended with failure.
  Step 'Add recommendation for vscode and plugin: angular-console' failed: unexpected error
  Step 'Add recommendation for vscode and plugin: angular2' failed: unexpected error
2 step(s) failed out of 5 steps.
```

This issue is about to removing this functionality, which is specific to VSCode.

PR #1218, opened 2025-04-08
#1217: Remove plugin recommendation feature in VSCode

alfeilex:
This PR Fixes: #1217 by removing plugin recommendation feature in VSCode.

- Removing recommendation feature in VSCode
- Removing related tests


jan-vcapgemini (approved):
Thanks for removing the recommendation plugin feature. LGTM, ready for review.

hohwille (approved):
@alfeilex thanks for your PR. I think it is good that we avoid such special features that are less stable, undocumented, not aligned with behaviour of other tools, etc. 👍

PR #1241, opened 2025-04-15
1205 : Fix Git implementation

mbilda:
- refactored GitContextImpl

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
@@ -3,14 +3,11 @@ package com.devonfw.tools.ide.git;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 
 import com.devonfw.tools.ide.cli.CliException;
 import com.devonfw.tools.ide.context.IdeContext;
-import com.devonfw.tools.ide.process.ProcessContext;
-import com.devonfw.tools.ide.process.ProcessErrorHandling;
 import com.devonfw.tools.ide.process.ProcessMode;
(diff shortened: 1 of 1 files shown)


cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
```diff
@@ -233,101 +140,106 @@ public void cleanup(Path repository) {
...
+      runGitCommand(repository, "switch", branch);
     }
-    this.context.trace("Git is installed");
   }
 
-  private void runGitCommand(Path directory, String... args) {
-
-    ProcessResult result = runGitCommand(directory, ProcessMode.DEFAULT, args);
+  private void handleErrors(Path repository, ProcessResult result, String errorMessage) {
     if (!result.isSuccessful()) {
```
mbilda, line 183: This if clause is relevant to decide if the result was an error or not. If it wasn't an error, we don't need to "handle Errors". We should consider this if-check in the "parent" method or wherever it is called.

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
```diff
@@ -233,101 +140,106 @@ public void cleanup(Path repository) {
...
-      int size = out.size();
-      if (size == 1) {
-        return out.get(0);
-      } else if (size == 0) {
-        warningOnError += " - No output received from " + result.getCommand();
-      } else {
-        warningOnError += " - Expected single line of output but received " + size + " lines from " + result.getCommand();
-      }
+  private void requireOnline(String action) {
+    if (this.context.isOfflineMode()) {
```
mbilda, line 202: Same es above. If we check if its offlineMode, then we don't need to check if its required to be online.

mbilda: Need to be analyzed in detail, because isOfflineMode and requireOnline could be two different things to consider.

PR #1244, opened 2025-04-15
#1229 just list versions with urls for the intended OS

### LINKED ISSUE #1229: Python seems to not be available
### Expected behavior

As a user wanting to use python and reading issues like #38, I want to use IDEasy to install python so that I can start scripting and hopefully also install tools via pip.

### Actual behavior
`ide install python`, `ide python`, `python`, `ide install pip`, `ide pip` all result in errors.
Python is also not listed in the tools in the help output of the cli. It seems to be there at least in code, but it does not seem to be exposed to the install commandlet

1. 

### Related/Dependent Issues
#38 

### Affected version:

* OS: Windows

maybeec:
Should fix #1229 while I am not 100% aware of the performance impact as the fix might have some impact.
Would be great to get feedback on the solution.


hohwille (approved):
@maybeec thank you very much for your PR. You really surprised me that you actively contribute to IDEasy and digged into our code creating this fix.
We will create a follow up to change `UrlVersion.getMatchingUrls` to not throw an exception so we can do the `null` check here instead of creating, catching, and GCing an exception.
That follow up PR can then slightly speed up auto-completion since exceptions are slightly expensive as you already pointed out.

But your fix is a reasonable improvement and seems fully functional. Great job 👍

maybeec:
Fully agree. Even better would be to Support proper Admin configuration by making the Admin actively choose OS Support and this resolve Wildcard Version configuration Just to those Versions, which are supported for all configured OS.

You could even easily log a warning in case an admin configured a wildcard version and the latest version might not have support for all configured OS.

I am not yet aware of the administration of ideasy, therefore unable to estimate the effort and usefullness yet.

maybeec:
Thinking out loud, @hohwille this could also be done in a some kind of check-configuration command for admin. This could check download availability for all configured OS and maybe even more. It could also log out a summary of currently resolved versions of wildcards. This would be helpful for admins as well as a great tool to communicate and guide the Admin towards the use of further best practices.
In addition, this could be a wonderful tool for CI especially to check ideasy configuration functionality in nightlies.

Issue #1243, opened 2025-04-15
ide status still failing with SSLHandshakeException


hohwille:
### Expected behavior

With #1110 my expectation was that `ide status` gets robust and does not fail in the middle so we get the required information for support and analysis.

### Actual behavior

It still fails:
```
$ ide status
IDE_ROOT is set to D:\projects
IDE_HOME is set to D:\projects\xxx
You are online.
Your settings are up-to-date.
An unexpected error occurred!
We are sorry for the inconvenience.
Please check the error below, resolve it and try again.
If the error is not on your end (network connectivity, lack of permissions, etc.) please file a bug:
https://github.com/devonfw/IDEasy/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=IllegalStateException%3A+Failed+to+fetch+XML+metadata+from+https%3A%2F%2Frepo1.maven.org%2Fmaven2%2Fcom%2Fdevonfw%2Ftools%2FIDEasy%2Fide-cli%2Fmaven-metadata.xml
java.lang.IllegalStateException: Failed to fetch XML metadata from https://repo1.maven.org/maven2/com/devonfw/tools/IDEasy/ide-cli/maven-metadata.xml
        at com.devonfw.tools.ide.tool.repository.MavenRepository.fetchXmlMetadata(MavenRepository.java:313)
        at com.devonfw.tools.ide.tool.repository.MavenRepository.fetchVersions(MavenRepository.java:215)
        at com.devonfw.tools.ide.tool.repository.MavenRepository.resolveVersion(MavenRepository.java:202)
        at com.devonfw.tools.ide.tool.repository.MavenRepository.resolveVersion(MavenRepository.java:187)
        at com.devonfw.tools.ide.tool.IdeasyCommandlet.getLatestVersion(IdeasyCommandlet.java:122)
        at com.devonfw.tools.ide.tool.IdeasyCommandlet.checkIfUpdateIsAvailable(IdeasyCommandlet.java:132)
        at com.devonfw.tools.ide.commandlet.StatusCommandlet.checkForUpdate(StatusCommandlet.java:54)
        at com.devonfw.tools.ide.commandlet.StatusCommandlet.run(StatusCommandlet.java:46)
        at com.devonfw.tools.ide.context.AbstractIdeContext.applyAndRun(AbstractIdeContext.java:976)
        at com.devonfw.tools.ide.context.AbstractIdeContext.run(AbstractIdeContext.java:894)
[...]

PR #1247, opened 2025-04-16
#1243: Fix logging of ide status although errors might occur

alfeilex:
Fixes: #1243 

This PR implements the logging of the `ide status`, although errors might occur by catching potential exceptions


hohwille (approved):
@alfeilex thanks for this PR. Looks all got to me - great job 👍

Issue #1231, opened 2025-04-13
Ide is reported to be broken after installation


maybeec:
### Expected behavior

As a user who installed the IDE from a binary, I want to simply get the latest version so that I can start working on the lastest version. Thus, I opened cmd on windows after installation by msi binary and ran `ide upgrade` to get the latest version. 

### Actual behavior

`ide upgrade` ended in an error stating my installation is corrupt:

```
C:\Users\mbrunnli>ide upgrade
Start: Install ideasy
Deleting corrupted installation at D:\projects\_ide\software\maven\ideasy\ideasy\2025.04.001
Step 'ide (upgrade)' failed: java.lang.IllegalStateException: Failed to delete D:\projects\_ide\software\maven\ideasy\ideasy\2025.04.001
 Step 'Install ideasy (*)' failed: java.lang.IllegalStateException: Failed to delete D:\projects\_ide\software\maven\ideasy\ideasy\2025.04.001
2 step(s) failed out of 2 steps.
An unexpected error occurred!
We are sorry for the inconvenience.
Please check the error below, resolve it and try again.
If the error is not on your end (network connectivity, lack of permissions, etc.) please file a bug:
https://github.com/devonfw/IDEasy/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=IllegalStateException%3A+Failed+to+delete+D%3A%5Cprojects%5C_ide%5Csoftware%5Cmaven%5Cideasy%5Cideasy%5C2025.04.001
java.lang.IllegalStateException: Failed to delete D:\projects\_ide\software\maven\ideasy\ideasy\2025.04.001
        at com.devonfw.tools.ide.io.FileAccessImpl.delete(FileAccessImpl.java:791)
        at com.devonfw.tools.ide.tool.LocalToolCommandlet.installTool(LocalToolCommandlet.java:208)
        at com.devonfw.tools.ide.tool.LocalToolCommandlet.installTool(LocalToolCommandlet.java:169)
        at com.devonfw.tools.ide.tool.LocalToolCommandlet.install(LocalToolCommandlet.java:81)
        at com.devonfw.tools.ide.tool.ToolCommandlet.install(ToolCommandlet.java:260)
        at com.devonfw.tools.ide.tool.IdeasyCommandlet.install(IdeasyCommandlet.java:109)
        at com.devonfw.tools.ide.commandlet.UpgradeCommandlet.run(UpgradeCommandlet.java:43)
[...]

PR #1248, opened 2025-04-16
#1231: Prevent automatic IDEasy uninstall

jan-vcapgemini:
Fixes: #1231

**Implements**
* Added extra check for IDEasy commandlet to prevent deletion of "corrupted" installation


hohwille:
> @hohwille where should the backup of IDEasy be created? What do you think about _ide/backup?

This is the result of migrating the code from devonfw-ide to IDEasy and keeping the old behaviour.
There is some sense behind this since if you have many projects and all data is backed up into a single backup folder it might be less obvious to find things from a project and also there is higher probability of a collision (concurrent usage).

However, non of this is a valid argument.
Lets follow KISS and do as you suggest or imply by your question:
The backup should only be in `$IDE_ROOT/_ide/backup` and not inside the project.
Restoring something from the backup is an absolute edge-case.
The only thing we want to prevent is that the user creates valuable data not backed up anywhere else and we delete it physically with our product.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -204,8 +204,11 @@ public ToolInstallation installTool(GenericVersionRange version, ProcessContext
           return createToolInstallation(installationPath, resolvedVersion, toolVersionFile, false, processContext, extraInstallation);
         }
       } else {
-        this.context.warning("Deleting corrupted installation at {}", installationPath);
-        fileAccess.delete(installationPath);
+        // Makes sure that IDEasy will not delete itself
+        if (!this.tool.equals(IdeasyCommandlet.TOOL_NAME)) {
+          this.context.warning("Deleting corrupted installation at {}", installationPath);
+          fileAccess.delete(installationPath);
+        }
```
hohwille, line 211:
```suggestion
        if (this.tool.equals(IdeasyCommandlet.TOOL_NAME)) {
          this.context.warning("Your IDEasy installation is missing version file at {}", toolVersionFile);
        } else {
          this.context.warning("Deleting corrupted installation at {}", installationPath);
          fileAccess.delete(installationPath);
        }
```

hohwille (approved):
@jan-vcapgemini thanks for your PR and quickfix. 👍 
I left a comment for improvement.
Did you also find the place where we miss out to create the version file to fix the root problem?

UPDATE: IMHO we need to fix that here:
https://github.com/devonfw/IDEasy/blob/23edbfcd8f293ff075ddedcd416120dbeb273564/cli/src/main/java/com/devonfw/tools/ide/tool/IdeasyCommandlet.java#L176

hohwille:
> @jan-vcapgemini thanks for your PR and quickfix. 👍 I left a comment for improvement. Did you also find the place where we miss out to create the version file to fix the root problem?
> 
> UPDATE: IMHO we need to fix that here:
> 
> https://github.com/devonfw/IDEasy/blob/23edbfcd8f293ff075ddedcd416120dbeb273564/cli/src/main/java/com/devonfw/tools/ide/tool/IdeasyCommandlet.java#L176

OK, to make progress with our release, I will have to merge this as is now.
If I can find the time, I can add another PR to add this.

Issue #1223, opened 2025-04-11
IDEasy upgrade is not working on Mac M1


jan-vcapgemini:
### Expected behavior

As an IDEasy user on Mac M1 I expect IDEasy to properly upgrade itself if I run `ide upgrade`. 

### Actual behavior
The upgrade stops with the following error message:

```
Running commandlet UpgradeCommandlet[upgrade]
Start: Install ideasy
Resolved version pattern *!-SNAPSHOT to version 2025.04.002-SNAPSHOT
Tool ideasy has 0 other tool(s) as dependency
Trying to download ide-cli-2025.04.002-20250409.024246-1-mac-arm64.tar.gz from https://s01.oss.sonatype.org/content/repositories/snapshots/com/devonfw/tools/IDEasy/ide-cli/2025.04.002-SNAPSHOT/ide-cli-2025.04.002-20250409.024246-1-mac-arm64.tar.gz
Step 'ide (upgrade,--mode=snapshot)' failed: java.lang.IllegalStateException: Failed to download file from URL https://s01.oss.sonatype.org/content/repositories/snapshots/com/devonfw/tools/IDEasy/ide-cli/2025.04.002-SNAPSHOT/ide-cli-2025.04.002-20250409.024246-1-mac-arm64.tar.gz to /Users/runner/tmp/ideasy-integration-test-debug-2025-04-10_18-24-32/home-dir/projects/_ide/tmp/Downloads/ide-cli-2025.04.002-20250409.024246-1-mac-arm64.tar.gz
Step 'Install ideasy' ended with failure.
 Step 'Install ideasy (*!-SNAPSHOT)' failed: java.lang.IllegalStateException: Failed to download file from URL https://s01.oss.sonatype.org/content/repositories/snapshots/com/devonfw/tools/IDEasy/ide-cli/2025.04.002-SNAPSHOT/ide-cli-2025.04.002-20250409.024246-1-mac-arm64.tar.gz to /Users/runner/tmp/ideasy-integration-test-debug-2025-04-10_18-24-32/home-dir/projects/_ide/tmp/Downloads/ide-cli-2025.04.002-20250409.024246-1-mac-arm64.tar.gz
2 step(s) failed out of 2 steps.
An unexpected error occurred!
```

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Install an old IDEasy snapshot on your Mac M1
2. run ide upgrade

### Comments/Hints:
[...]

PR #1251, opened 2025-04-16
#1223: Fixed Mac ARM64 release name

jan-vcapgemini:
Fixes: #1223

**Implements**
* adjusted mac-arm release name to mac-arm64


hohwille (approved):
@jan-vcapgemini thanks for the fix. Excellent work. 👍 
Could you add the bug to our CHANGELOG? Then we can merge this.

Issue #1199, opened 2025-04-01
Improve templates for issue and PR


hohwille:
As a IDEasy product-owner, I want to have great templates for issues and PRs so that we do not forget important aspects and have better structure and metadata (labels, etc.).

The following aspects shall be addressed for issue templates:
* GitHub added a new feature that issues can now have a `type` that is `Bug`, `Feature`, or `Task`. Before this was only possible via labels. Our issue templates are still based on that state and need to be updated asap since all new issues lack of the `type` value unless I step in and manually set it. My understanding would be that `Bug` maps to `Bug` (surprise) and `Enhancement` to `Feature` (may also be renamed to `Feature`) but for `Security Vulnerability` I would also say `Bug` with additional label `Security` (just as is). `Blank issue` does not really make sense to me.
![Image](https://github.com/user-attachments/assets/2b5fb65c-c0cb-4c01-adb9-4a22d63e20f5)
* Github supports flexible structured forms for issue templates. For an example see [here](https://github.com/docker/desktop-linux/issues/new?template=bug_report.yml). So for bugs we should have a sections for `Actual Result (log with error/stacktrace)`, `Expected Result`, `Output of ide status`, `Steps to Reproduce`, and `Additional Info`. For features `Feature Request`, `motivation (what are your trying to archive? why can't your problem be solved already?)`, and `Additional Info`. For `Security Vulnerability` I expect the sections `CVE(s)`, `Artifact(s)`, `Criticality`, and `Additional Info`

The following aspects shall be addressed for PR templates:
* fixes #ISSUEID
* We should include our [DoD](https://github.com/devonfw/IDEasy/blob/main/documentation/DoD.adoc) as checkmarks.

# Implementation hint

Such templates need to go here:
https://github.com/devonfw/IDEasy/tree/main/.github

Currently we inherit them from here:
https://github.com/devonfw/.github/tree/master/.github

You can easily test this on your fork if you enable issues in your settings section.
Then simply create a PR adding the templates to our main branch.

PR #1253, opened 2025-04-17
#1199: Create templates for PR and issues 

julia-cap:
Fixes https://github.com/devonfw/IDEasy/issues/1199

Changes made:
created directory for issue templates
created one template for PR


hohwille:
Thanks for updating to `main` so I now could test the PR template live on github.
Can you even edit the settings of your fork to allow issue creation so we can also test the issue templates? 

Some early review feedback about the PR template (obviously some things are already clear to you since this PR is still in draft mode):
* there are several markdown syntax issues. While GitHub gives the freedom to choose markup syntax for files in git or wiki, in comments or PR templates only markdown is supported. We decided for AsciiDoc since it is much more powerful and intuitive to use (e.g. link syntax in MD sucks) but for comments and such templates we have stick with markdown so unfortunately in such situations you need to master both.
* we need to pay attention that we do not go from one extreme to the other. Currently we do not have a PR template and everything is quite lax and pragmatic. If we come up with an over-complex submission matrix, developers may be overwhelmed and might even stop taking it serious and go trough all checks since it is too tedious.
* IMHO we should stick with a simple checkbox instead of an additional N/A column. For some points we can just change the description easily so that I can check it in every case where I did my job properly. For some other points, that are violated (N/A) rarely the user could simply edit and ~~strike through~~ the aspect. For commandlets we can create a separate block/table that can be removed or the developer could choose between `no commandlet added [x]`  or the table of commandlet checks.
* We should also kick out all rules that are already checked automatically. I do not see the value in `Build is successful and tests pass when running mvn test (or even better mvn clean test)` or especially `All GitHub checks have passed` since we already have an automated CI build that gets red if he missed out this point - further this can change with every commit and I can guarantee that no developer will uncheck a checkbox after an updating commit. Also I would remove the `Milestones` topic and would question `Changes are covered by unit tests` together with the team.
* So the DoD is also a documentation explaining things for new developers, while the PR template IMHO should focus on the key aspects.

hohwille (approved):
@julia-cap thank you very much. This is really great progress. 👍 
I could not find anything wrong with the security label (no typo, syntax issue, etc.).
Let us just merge and then we can further improve (I would also consider removing the `help texts` point since this is automatically checked by JUnit).

julia-cap:
> @julia-cap thank you very much. This is really great progress. 👍 I could not find anything wrong with the security label (no typo, syntax issue, etc.). Let us just merge and then we can further improve (I would also consider removing the `help texts` point since this is automatically checked by JUnit).

Sure, let's keep it simple and short ;) I will remove the line about help texts.

PR #1254, opened 2025-04-17
#1213: changed the tool installation log

### LINKED ISSUE #1213: improve log message for dependency version range missmatch
This log message is not very intuitive and will confuse end-users that do not understand what is going on:

```
$ ide intellij
...
Configured version of tool java is 17* but does not match version to install [21,) - need to use different version from software repository.
...
```

My recommendation would be to change as following:
```
The tool intellij requires java in version range [21,) but your project is using java version 17* that does not match.
Hence, we install an additional java version 21.0.6_7 to run intellij.
```

ylindenb-cap:
Fixes #1213 

This PR changes the log message when installing a tool
and adds a new method to get the latest version of the tool as String


cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -257,8 +258,9 @@ public boolean installAsDependency(VersionRange version, ProcessContext processC
                 + " and this tool does not support the software repository.");
       }
       this.context.info(
-          "Configured version of tool {} is {} but does not match version to install {} - need to use different version from software repository.",
-          this.tool, configuredVersion, version);
+          "The tool {} requires {} in version range {} but your project is using {} version {} that does not match."
+              + " Hence, we install an additional {} version {} to run intellij.",
```
hohwille, line 262: At the end you still have `intellij` hardcoded.
Maybe you can also rephrase the message to avoid repeating yourself so much often and to avoid providing the same value multiple times as log argument.

hohwille (approved):
@ylindenb-cap thanks for your rework. Now ready for merge. 👍 
Could you please add the story #1213 to our CHANGELOG?

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -271,7 +273,7 @@ private void installToolDependencies(VersionIdentifier version, String edition,
     for (ToolDependency dependency : dependencies) {
       this.context.trace("Ensuring dependency {} for tool {}", dependency.tool(), toolWithEdition);
       LocalToolCommandlet dependencyTool = this.context.getCommandletManager().getRequiredLocalToolCommandlet(dependency.tool());
-      dependencyTool.installAsDependency(dependency.versionRange(), processContext);
+      dependencyTool.installAsDependency(dependency.versionRange(), processContext, toolWithEdition);
```
hohwille, line 276: Sorry, that I see this only now that the review is more or less completed and I already wanted to merge.
This is also incorrect. If we have a version range `[17,22)` for Java, then the version to get installed will not be the latest one (`24.0.1_9`) but `21.0.7_6`.
If you want to keep this simple, we could either omit the version and write something like "a compatible version in that range" oder wir müssten das etwas umbauen, sodass wir die Version an dieser Stelle schon mitgeben können für die Log-Meldung.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -256,8 +257,9 @@ public boolean installAsDependency(VersionRange version, ProcessContext processC
                 + " and this tool does not support the software repository.");
       }
       this.context.info(
-          "Configured version of tool {} is {} but does not match version to install {} - need to use different version from software repository.",
-          this.tool, configuredVersion, version);
+          "The tool {} requires {} in the version range {}, but your project uses version {}, which does not match."
+              + " Therefore, we install a compatible version in that range.",
+          toolParent, this.tool, version, configuredVersion);
```
hohwille, line 262: In my review summary I wrote:
> Could you please add the story https://github.com/devonfw/IDEasy/issues/1213 to our CHANGELOG?

See [DoD](https://github.com/devonfw/IDEasy/blob/main/documentation/DoD.adoc)

In order that it does not get forgotten, I add it as a review comment to an arbitrary line of code.

hohwille (approved):

Issue #793, opened 2024-11-22
Support for IDE_MIN_VERSION


hohwille:
As a IDEasy user, I want to be able to define a minimum version of IDEasy product itself in my `settings/ide.properties` so that my developers get informed if they are using an outdated version of IDEasy that will not work for my project.

This feature was already present in devonfw-ide:
https://github.com/devonfw/ide/blob/9efaf926f1daa4e1e37a22fc89081255fbf55dab/scripts/src/main/resources/scripts/command/ide#L239-L248
However, it was more of a hidden feature since it was forgotten to document it here:
https://github.com/devonfw/ide/blob/master/documentation/variables.asciidoc
Original devonfw-ide story: https://github.com/devonfw/ide/issues/999

In IDEasy we want to introduce a variable `IDE_MIN_VERSION`.
It should not fall back to a legacy variable `DEVON_IDE_MIN_VERSION` since the values configured for devonfw-ide do not make any sense in the context of IDEasy.
If that variable is defined, it shall be parsed as a `VersionIdentifier` and then the current IDEasy version shall be compared with it.
If the installed IDEasy version is lower than the configured `IDE_MIN_VERSION` then
* `ide create` shall fail after cloning the settings with an according error message
* in other situations (running a tool, installing something, status commandlet, etc.) we should only render a warning message.
That message should be:
```
Your version of IDEasy is currently «installed-version»
However, this is too old as your project requires at latest version ${IDE_MIN_VERSION}
Please run the following command to update to the latest version of IDEasy and fix the problem:
ide upgrade
```

The variable should also be documented in `variables.adoc`.
Further, in our `ide-settings` in the top-level `ide.properties` we should already configure this variable with a current official release version.
Once we are out of beta that should point to the official non-beta release.
Whenever we introduce new features and use them in our `ide-settings` (e.g. something like improved XML merger), we need to update `IDE_MIN_VERSION` to an official release version of IDEasy before using such features in settings configuration files.

PR #1262, opened 2025-04-24
#793: add support for IDE_MIN_VERSION

cthies-capgemini:
IDE_MIN_VERSION no longer falls back to DEVON_IDE_MIN_VERSION. ide create now fails if version is smaller than minimum, a warning is rendered otherwise. Documentation is updated to reflect change.


cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -1061,6 +1065,15 @@ private void verifyIdeRoot() {
     }
   }
 
+  private void verifyIdeMinVersion() {
+    if (IdeVersion.getVersionIdentifier().compareVersion(IDE_MIN_VERSION.get(this)).isLess()) {
+      warning(String.format("Your version of IDEasy is currently %s\n"
+          + "However, this is too old as your project requires at latest version %s\n"
+          + "Please run the following command to update to the latest version of IDEasy and fix the problem:\n"
+          + "ide upgrade", IdeVersion.getVersionIdentifier().toString(), IDE_MIN_VERSION.get(this).toString()));
```
hohwille, line 1073: SLF4J already handles parameter substitution so we do not need `String.format` here
https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc#logging
```suggestion
      warning("Your version of IDEasy is currently {}\n"
          + "However, this is too old as your project requires at latest version {}\n"
          + "Please run the following command to update to the latest version of IDEasy and fix the problem:\n"
          + "ide upgrade", IdeVersion.getVersionIdentifier(), IDE_MIN_VERSION.get(this));
```

documentation/variables.adoc
```diff
@@ -30,9 +30,10 @@ Please note that we are trying to minimize any potential side-effect from `IDEas
...
+|`FAIL_ON_AMBIGOUS_MERGE`| |If set to `true` the link:configurator.adoc#element-identification[merge:id] is ambiguous and the resulting XPath expression matches multiple elements. Typically, the link:usage.adoc#admin[IDE admin] did something wrong in the workspace template configuration. However, we decided to log a warning if that happens and use the first match by default to prevent our users from being blocked or annoyed in case of such configuration error. With this property you can enforce that this is handled as error aborting the merge. If that happens the user sees the stacktrace of the error and gets asked if they want to continue in order to launch his IDE (IntelliJ, Eclipse, VSCode, etc.).
```
hohwille, line 33: IMHO `he` was correct and referring to `user` where `they` does not seem to make sense.
Would you agree and revert this change?

documentation/variables.adoc
```diff
@@ -30,9 +30,10 @@ Please note that we are trying to minimize any potential side-effect from `IDEas
...
 |`«TOOL»_EDITION`|`«tool»`|The edition of the tool `«TOOL»` to install and use (e.g. `ECLIPSE_EDITION`, `INTELLIJ_EDITION` or `DOCKER_EDITION`). Default of `DOCKER_EDITION` is `rancher`.
 |`«TOOL»_VERSION`|`*`|The version of the tool `«TOOL»` to install and use (e.g. `ECLIPSE_VERSION` or `MVN_VERSION`).
 |`«TOOL»_BUILD_OPTS`| |The arguments provided to the build-tool `«TOOL»` in order to run a build. E.g.`clean install`
 |`«TOOL»_RELEASE_OPTS`| |The arguments provided to the build-tool `«TOOL»` in order to perform a release build. E.g.`clean deploy -Dchangelist= -Pdeploy`
+|`IDE_MIN_VERSION`| | The minimum version of IDEasy that should be installed. Causes `ide create` to fail if violated, otherwise renders a warning
```
hohwille, line 38:
```suggestion
|`IDE_MIN_VERSION`| | The minimum version of IDEasy that is required by your project. Causes `ide create` to fail if violated, otherwise renders a warning
```

cli/src/test/java/com/devonfw/tools/ide/context/IdeContextTest.java
```diff
@@ -128,4 +130,43 @@ public void testCurrentWorkspacePathIsUsed() {
...
+    variables.getByType(EnvironmentVariablesType.CONF).set("IDE_MIN_VERSION", ideVersion);
+    CliArguments args = new CliArguments();
+    String warningMessage = String.format("Your version of IDEasy is currently %s\n"
+        + "However, this is too old as your project requires at latest version %s\n"
+        + "Please run the following command to update to the latest version of IDEasy and fix the problem:\n"
+        + "ide upgrade", ideVersion, ideVersion);
+    // act
+    context.run(args);
+    // assert
+    assertThat(context).logAtWarning().hasNoMessage(warningMessage);
```
hohwille, line 169: just a nice to have:
But there is also `hasNoMessageContaining` so you could just search for a substring like `is too old as your project requires at least`. Also you might want to replace `logAtWarning` with just `log` here since we do not expect this message regardless of the log-level.

hohwille (approved):
@cthies-capgemini thanks for your PR. This is great work - well done! 👍 

I left some comments for further improvement.
Could you please have a look and update?
Then we can merge this PR.

documentation/variables.adoc
cthies-capgemini: In this context it is a singular "they" to include not just male users, so grammatically it's sound. But we can revert this if you think it's confusing, it was just a recommended change by IntelliJ.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
cthies-capgemini: That's good to know, I'll keep it in mind for the future. I changed the method now so it can render a warning or throw an error depending on an input flag, so to keep things tidy I format the string once and then pass it to either method. Do you agree with that change?

hohwille: I see. Sure. Since for the exception you need to manually fill in the parameters, it fully makes sense that you are using `String.format`.

Issue #1267, opened 2025-04-25
Prevent CI failure on pull requests when coveralls action fails


cthies-capgemini:
As a developers, I want the CI build workflow to not fail when the coveralls check fails on a pull request, so that I am not blocked from merging when the coveralls service returns an internal server error.

The coveralls action has an optional input `fail-on-error` that can be set to false so it doesn't fail on an error, which should solve this issue.

PR #1268, opened 2025-04-25
#1267: avoid CI failure on coveralls error

cthies-capgemini:
Implements #1267 by passing an argument to the coveralls action on pr-builds that tells it to not fail the CI pipeline when coveralls returns an error.


hohwille (approved):
@cthies-capgemini thanks for creating the issue and PR. Excellent work. Great that you found such a simple bug effective solution so quickly 👍

Issue #1264, opened 2025-04-24
ide.bat not detecting errors properly


hohwille:
### Expected behavior

As a IDEasy user on Windows, I want to call `ide` comannds from CMD and get proper exit code so that I see what if something is going wrong or could even creates scripts calling `ide` with proper error handling.

### Actual behavior

IDEasy always returns exit code `0` in CMD.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. `cd %IDE_ROOT%\IDEasy`
2. `ide mvn invalid arg`
3. `echo %ERRORLEVEL%`

### Related/Dependent Issues

* PR #1256 (where I detected the bug during testing)
* issue #894 (where I thought that we fixed the bug earlier but unfortunately did not)

### Comments/Hints:

It seems that the `if not %ERRORLEVEL% == 0` is never entered.
Needs further analysis and investigation to understand what is going wrong here.

### Affected version:

* 2025.04.002

alfeilex:
I think we already had a potential fix for that, but for some reason, we reverted it

https://github.com/devonfw/IDEasy/commit/c32ef4de7d62d0b0ffdc360cac09f764d958262e

PR #1274, opened 2025-04-28
#1264: Fix ide.bat to detect errors properly

alfeilex:
Fixes: #1264 

This PR fixes the `ide.bat` to detect errors properly. The `%ERRORLEVEL%` is moved to outside the brackets to be set with the correct exit code that is given by the previous `ideasy` command. If `ideasy` is run without arguments, then the error handling is skipped to prevent that previous `%ERRORLEVEL%` values that have been set by other commands, are not used in the error handing.

I tested the behavior by comparing the output of various `ide` commands in `git-bash` and `cmd` side by side and they behave equal.

You can test the behavior by replacing the code of `ide.bat` in `projects/_ide/installation/bin/ide.bat`


cli/src/main/package/bin/ide.bat
```diff
@@ -21,10 +21,9 @@ if exist "%GIT_CORE%" (
 
 if not "%1%" == "" (
   ideasy %IDE_OPTIONS% %*
-  if not %ERRORLEVEL% == 0 (
-    echo %_fBRed%Error: IDEasy failed with exit code %ERRORLEVEL% %_RESET%
-    call :echoUseBash
-    exit /b %ERRORLEVEL%
+  if errorlevel 1 (
```
hohwille, line 24: there are various errorlevels that indicate an error. Everything except `0` is an error not just `1`.
Further, every next command (IMHO including `goto` or `echo` are new executions that update the `errorlevel`.
So did you test this with different scenarios? I still have doubts that this is working now.

alfeilex: `if errorlevel <number>`" is defined as:

```
ERRORLEVEL number Specifies a true condition if the last program run
                    returned an exit code equal to or greater than the number
                    specified.
```

You can see this definition by running `if /?` in the CMD. Moreover, my observation is that `%ERRORLEVEL%` is always 0 in the first parsing of the script. When using `:goto`, it looks like the parsing is done again and the correct `%ERRORLEVEL%` is assigned to this variable.

I am still testing this behavior

cli/src/main/package/bin/ide.bat
```diff
@@ -36,7 +35,9 @@ for /f "tokens=*" %%i in ('ideasy %IDE_OPTIONS% env') do (
 ideasy %IDE_OPTIONS% env >nul
 
 if %ERRORLEVEL% == 0 (
-  echo IDE environment variables have been set for %IDE_HOME% in workspace %WORKSPACE%
+  if "%~1"=="" (
+    echo IDE environment variables have been set for %IDE_HOME% in workspace %WORKSPACE%
+  )
```
alfeilex, line 40: Only echo when running `ide` without arguments

cli/src/main/package/bin/ide.bat
alfeilex: In my latest test by comparing the output of `ide` commands in git-bash and cmd/powershell, I couldn't identify any difference in behavior.

hohwille: @alfeilex thanks for testing and analysing this problem.
I am surprised that the `%ERRORLEVEL%` variable that is documented [here](https://ss64.com/nt/errorlevel.html) does not actually seem to work at all.
However, what you are doing is to ignore the error-level or exit code and always exit `ide.bat` with `0` (success).
It seems you are absolutely correct that is was always broken and your PR is not making it worse and at least fixes the error logging.
However, we need to research for the real solution. A quick search lead me to this URL that may help to fix the problem:
https://devblogs.microsoft.com/oldnewthing/20080926-00/?p=20743

alfeilex: The reason why `%ERRORLEVEL%` does not work is that is it's called inside the brackets. A further solution to get the correct exit code is to call it after the `if condition`, where the `ideasy` command is executed. In that case, we would get the correct `%ERRORLEVEL%` variable and would not need subroutines or even delayed expansion. 

Would you agree? The related code would be like this:

```
if not "%1%" == "" (
  ideasy %IDE_OPTIONS% %*
)

if %ERRORLEVEL% NEQ 0 (
  echo.
  echo %_fBRed%Error: IDEasy failed with exit code %ERRORLEVEL% %_RESET%
  call :echoUseBash
  exit /b
)
```

alfeilex: There is one problem with that solution. When `ide` is run without arguments, then the problem is that the `%ERRORLEVEL%` of the previous commands is used. In the first run of `ide` in a fresh terminal, this:

`if exist "%GIT_CORE%" ...` is not true and the `%ERRORLEVEL%` is 1 resulting in an output of `Error: IDEasy failed with exit code 1

alfeilex: In my point of view, we have three possibilities to solve this:

- run the command isolated and use the `%ERRORLEVEL%` afterwards
- using subroutines with `goto` (current PR state)
- enable delayed expansion (`!ERRORLEVEL!`)

hohwille: Thanks for all your comments, further tests and explanations.
Indeed I was wrong with my initial comment and your approach is fine in general.

Great that you also have a solution for the remaining problem.
I fully agree that `delayed expansion` sucks and we should avoid it.
If the second option is already implemented, how can it be a fix to a problem still existing?
Actually you are free to chose the solution that seems best and easiest for your.
But just from my understanding some suggestions:
Maybe you can just use a stupid `goto` to skip the execution of `ideasy` with parameters if there are no arguments.
Then we have these executions outside parenthesis avoiding all this ugly CMD problems.

alfeilex: > If the second option is already implemented, how can it be a fix to a problem still existing?

I meant that it's already implemented in the current PR implementation

> Maybe you can just use a stupid goto to skip the execution of ideasy with parameters if there are no arguments.
Then we have these executions outside parenthesis avoiding all this ugly CMD problems.

This is a nice suggestion. It have implemented this and it works fine. Thank you!

hohwille: Very nice, this looks quite clean and readable now (at least for a CMD solution) 😄

PR #1281, opened 2025-04-29
#1139: Gui starting point (MVP)

### LINKED ISSUE #1139: Implement dashboard MVP with navigation and first real use-case
After story #80 was implemented that just created the GUI module with no functionality (except for a "hello world" window), we now want to go the next step.

The GUI for IDEasy should have 
* a navigation frame that allows 
  * selection of the project (as a `ComboBox`)
  * selection of the workspace (also as `ComboBox`) - not required but nice to have
* an entry or icon to open IDEs
  * if clicked, the content area is showing the IDE selection dialog (without the navigation bar(s) disappearing that IDE selection is displayed)
    * in the IDE selection dialog IntelliJ can be opened via some button.
    * this will launch IntelliJ in the selected project (see above) and workspace (unless workspace selector is implemented, use `main` workspace).
    * as a further nice-to-have multiple IDEs are supported and can be chosen to open
    * as an additional nice-to-have on top, you could read the property `IDE_TOOLS` and determine all tool commandlets that are IDEs and determine/filter (or sort and highlight) the available IDEs to open from that information.

Here we can see a screenshot of the obsolete devonfw-ide-dashboard that was once implemented in TypeScript using Electron and Angular that helps to give a visual idea of what to implement:
Image

Feel invited to make your own thoughts about User Experience and how the UI should look like instead of reproducing the layout from the screenshot 1:1.
Please consider that the navigation is hierarchical:
1. project is selected
2. workspace is selected and that selection depends on 1.
3. use-cases (e.g. open IDE) is selected/invoked and that may also depend on 2. (if you follow the nice-to-have options as well)

Technical design considerations:
* We already prepared IDEasy CLI for this GUI
* We therefore designed `IdeContext` as an interface that can have an alternative implementation
* Within the GUI, you should therefore create a new implementation of `IdeContext` (derived from `AbstractIdeContext`).
* This way methods like 
https://github.com/devonfw/IDEasy/blob/3f58554cd3d609552ced333dcd72ceb763683b9e/cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java#L248
or
https://github.com/devonfw/IDEasy/blob/3f58554cd3d609552ced333dcd72ceb763683b9e/cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java#L533
as well as the logger/logging methods
can be implemented via JavaFx controls.
It is not expected that all this `IdeContext` implementation for GUI is completed in this story, but it is required to design your implementation properly and lead it in the correct direction.

daburrow:
Fixes #1139
I created a MVP for the GUI. In the application a hierarchy on the left side is visible. One is able to choose a workspace after a project is selected. The workspace main is default. All buttons are disabled until a project is selected. In the future: segregate classes, add functionalities, improve usability.


hohwille:
> [ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/runner/work/IDEasy/IDEasy/gui/src/main/java/com/devonfw/ide/gui/IdeGuiContext.java:[12,7] error: IdeGuiContext is not abstract and does not override abstract method setForceRepositories(boolean) in IdeStartContext

gui/src/main/java/com/devonfw/ide/gui/IdeGuiContext.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ *
+ */
+public class IdeGuiContext extends AbstractIdeContext {
+
+
+  /**
+   * The constructor.
+   *
+   * @param startContext the {@link com.devonfw.tools.ide.log.IdeLogger IdeLogger}.
```
hohwille, line 18: did you somehow copy that from `AbstractIdeContext` where it was a leftover from an incomplete refactoring?
```suggestion
   * @param startContext the {@link IdeStartContextImpl}.
```
You can also remove the `Impl` suffix in the JavaDoc...

gui/src/main/java/com/devonfw/ide/gui/IdeGuiContext.java
```diff
@@ -0,0 +1,34 @@
+package com.devonfw.ide.gui;
+
+import java.nio.file.Path;
+
+import com.devonfw.tools.ide.context.AbstractIdeContext;
+import com.devonfw.tools.ide.context.IdeStartContextImpl;
+import com.devonfw.tools.ide.io.IdeProgressBar;
+
+/**
+ *
```
hohwille, line 10:
```suggestion
 * Implementation of {@link AbstractIdeContext} for the IDEasy dashbaord (GUI).
```

gui/src/main/java/com/devonfw/ide/gui/IdeGuiContext.java
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  @Override
+  protected String readLine() {
+    return "";
+  }
+
+  @Override
+  public IdeProgressBar newProgressBar(String title, long size, String unitName, long unitSize) {
+    return null;
```
hohwille, line 32: you could at least return a dummy implementation (see `IdeProgressBarTestImpl` that is however not available since it currently only exists in `test` code of `cli` module).

gui/src/main/java/com/devonfw/ide/gui/MainController.java
```diff
@@ -0,0 +1,132 @@
...
+  @FXML
+  private Button androidStudioOpen;
+  @FXML
+  private Button eclipseOpen;
+  @FXML
+  private Button intellijOpen;
+  @FXML
+  private Button vsCodeOpen;
+
+  private final String directoryPath = "C:\\projects\\";
```
hohwille, line 32: Please do not hardcode such Windows specific paths.
Use `IdeContext` and `IDE_ROOT` variable instead.

gui/src/main/java/com/devonfw/ide/gui/MainController.java
```diff
@@ -0,0 +1,132 @@
...
+
+            projects.getItems().add(name);
+          }
+        }
+      }
+    }
+
+    projects.setOnAction(actionEvent -> {
+
+      projectValue = projects.getValue() + "\\workspaces";
```
hohwille, line 90: never hardcode windows path separators as this will not work on Linux or MacOS.
If you are using `Path` you can use `resolve` method instead.
Otherwise use at least a regular slash (`/` instead of `\`).
Please also use constants like `IdeContext.FOLDER_WORKSPACES` instead of redundant String literals.
This not only makes it easier to consistently change values but also find usages all over the code-base where a specific constant (e.g. folder or file name) is used what can be very helpful.

gui/src/main/java/com/devonfw/ide/gui/MainController.java
```diff
@@ -0,0 +1,132 @@
...
+      setWorkspaceValue(selectedWorkspace);
+      selectedWorkspace.setDisable(false);
+      androidStudioOpen.setDisable(false);
+      eclipseOpen.setDisable(false);
+      intellijOpen.setDisable(false);
+      vsCodeOpen.setDisable(false);
+    });
+  }
+
+  private void setWorkspaceValue(ComboBox<String> workspace) {
```
hohwille, line 100: why are you passing a member variable (field) as a parameter?
This does not seem to make sense.

gui/src/main/java/com/devonfw/ide/gui/MainController.java
```diff
@@ -0,0 +1,132 @@
...
+      workspaceValue = workspace.getValue();
+    });
+  }
+
+  private void openIDE(String inIde) {
+
+    final IdeLogListenerBuffer buffer = new IdeLogListenerBuffer();
+    IdeLogLevel logLevel = IdeLogLevel.INFO;
+    IdeStartContextImpl startContext = new IdeStartContextImpl(logLevel, level -> new IdeSubLoggerOut(level, null, true, logLevel, buffer));
+    IdeGuiContext context = new IdeGuiContext(startContext, Path.of(directoryPath + projectValue + workspaceValue));
```
hohwille, line 129: This code looks a little misplaced here.
IMHO we will have many different use-cases over time in the dashboard.
Please extract a private method to get the context as otherwise sooner or later this will be copy&pasted for other use-cases. Instead we want to keep control over the context creation and may avoid recreating the context for every use-case invocation later.
Also buffering the logging does not seem to make sense here (`IdeLogListenerBuffer`).

hohwille (approved):
@daburrow thanks for your PR. This looks like a nice PoC. 👍 
However, could you please fix the build (compile error) and address the review comments to make your code portable and working outside of your personal environment.
Then we can merge this.

gui/src/main/java/com/devonfw/ide/gui/IdeGuiContext.java
hohwille: I created issue #1296 for the entire topic, so we can close this now.

daburrow:
Mistakes replaced by #1324. I changed the commit messages and linked them properly to the issue.

PR #1302, opened 2025-05-13
#351: prevent inheriting environment variables from other project

### LINKED ISSUE #351: Avoid inheriting environment variables from other IDEasy project if switched in the same shell session
# Expected behavior

When I run `ide` in `«project1»` and then switch (`cd`)  to `«project2»` and run `ide` there the result should be like if I would open a new shell immediately in `«project2»` and run `ide`.

# Actual behavior

IMHO variables that had been set in `«project1»` but are not set in `«project2»` will shine through.
As a result variables such as `JAVA_VERSION=8*` from `«project1»` may impact `«project2»` if such variable is nowhere defined in `«project2»`.

# Hints

This is a tricky one as this is rather a generic design flaw than an implementation bug.
However, the IMHO best solution would be to consider inheritance from `EnvironmentVariablesSystem` (`System.getenv`) only for dedicated variables. Then things like `«TOOL»_VERSION` and `«TOOL»_EDITION` cannot be taken from environment variables from the outside world causing such problems.
Further if we go in this direction would could also mark better what variables should be made visible to `ide env`.
IMHO exposing `«TOOL»_VERSION` and `«TOOL»_EDITION` variables to the shell is "pointless pollution" rather than being really helpful. This all seems to come from the heritage of devonfw-ide and with IDEasy we have the chance to think such things further through. With bash we could only do very simple things while now we have reliable ways for advanced features that can help us to get more reliable and consistent results.
See also #235 in this regard.

natalya-cap:
fixes #351

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariables.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariables.java
@@ -42,9 +42,15 @@ public interface EnvironmentVariables {
     if (value == null) {
       EnvironmentVariables parent = getParent();
       if (parent != null) {
+        if ((name.equals("M2_REPO") || name.equals("MAVEN_ARGS")) || name.contains("_VERSION") || name.contains("_EDITION")) {
+          if (parent.getSource().toString().contains("SYSTEM")) {
+            return null;
+          }
+        }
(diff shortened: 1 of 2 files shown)


hohwille:
Thanks @natalya-cap for the PR.
I took over from your changes and created a new PR #1327.

Issue #1271, opened 2025-04-28
After installation of aws "aws configure" not possible


hj-lorenz:
### Expected behavior

As a user of ide for developing an application, I want to install aws and configure it.

### Actual behavior

Installation of aws is possible.
"ide aws configure" is not possible, because data entry is not possible 
-> AWS Access Key ID [None]: 
Here you can type something but this is not displayed. 
After pressing "Ctrl-C or Ctrl-D" I see in the next prompt what I typed. But it has no effect.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. install aws
2. ide aws configure

### Related/Dependent Issues

-

### Comments/Hints:

-

### Affected version:

$ ide --version
2025.04.002

* OS: Windows
* Browser: Firefox (but should not be relevant
* gitBash on windows11

hohwille:
@hj-lorenz thanks for reporting this bug. It seems our processcontext and Java process stuff is buggy in regard of terminal interaction. We have to trace down if that is related to graalVM native image, to jline/jansi or JDK (Process[Builder]).
Might take some time but we will work on it and hopefully find a fix (also needed for #1108).

jan-vcapgemini:
Current workaround would be to use `aws` directly instead of `ide aws`. As IDEasy is initialized within your project, all environment variables should be working properly.

PR #1305, opened 2025-05-14
#1271: Fixed aws input not being processed

jan-vcapgemini:
Fixes: #1271

**Implements**
* added redirectInput INHERIT if ProcessMode is not DEFAULT_CAPTURE

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
@@ -165,6 +165,8 @@ public class ProcessContextImpl implements ProcessContext {
         this.processBuilder.redirectOutput(Redirect.PIPE).redirectError(Redirect.PIPE);
       } else if (processMode.isBackground()) {
         modifyArgumentsOnBackgroundProcess(processMode);
+      } else {
+        this.processBuilder.redirectInput(Redirect.INHERIT);
       }
 
       this.processBuilder.command(args);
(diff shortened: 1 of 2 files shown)


alfeilex (approved):
Thank you for the fix. Tested on Windows. 

LGTM

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -165,6 +165,8 @@ public ProcessResult run(ProcessMode processMode) {
         this.processBuilder.redirectOutput(Redirect.PIPE).redirectError(Redirect.PIPE);
       } else if (processMode.isBackground()) {
         modifyArgumentsOnBackgroundProcess(processMode);
+      } else {
+        this.processBuilder.redirectInput(Redirect.INHERIT);
```
hohwille, line 169: ah, we only had redirect output and error but never input.
Great finding, awesome 👍 
BTW: We could create a cleanup PR to add 3 methods to `ProcessMode` returning `Redirect`:
* getRedirectOutput
* getRedirectError
* getRedirectInput

If needed `ProcessMode.BACKGROUND*` could return `null` for some of these methods and if `null`, then the `redirect` is not configured in `processBuilder`.

Then we can make this code much more clean and understandable.
However, you fix the actual problem here functionally correct.
So lets merge this as is and create a new PR for the cleanup.

hohwille: Created issue #1306

PR #1310, opened 2025-05-15
#793: fix missing exception on create

cthies-capgemini:
### Implemented changes:

`verifyIdeMinVersion` is now called later in the `run` method of the `create` command. This ensures that the environment variable for the ide version is actually set and can be compared. Otherwise it is null and the function returns without doing anything.


cthies-capgemini:
I am now calling verifyIdeMinVersion twice in the run command of the CreateCommandlet, once before the super call and once after. I am not happy with this solution, but it was the only way I found to make both the unit test and the create command itself throw the correct exception. For the call before the super method, IDE_MIN_VERSION is null when calling the create command. For the call after the super method, IDE_MIN_VERSION is null for the unit test. I wasn't able to figure out why this is the case.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -63,6 +63,7 @@ public void run() {
     this.context.setIdeHome(newProjectPath);
     this.context.verifyIdeMinVersion(true);
     super.run();
+    this.context.verifyIdeMinVersion(true);
```
hohwille, line 66: this will now happen after all the software is installed, etc.
It would be an option to do it in `reloadContext` method.
To override it, it would also need to be changed from `private` to `protected`. 
I am happy to merge as is for our release today.
We should create a PR for further improvements.
Maybe we even find a better way or directly integrated it into the super-class what IMHO makes the most sense:
If I do `ide update` and the `settings` get updated and `IDE_MIN_VERSION` is changed to a newer value then it should also fail if I am outdated.

hohwille (approved):
@cthies-capgemini thanks for the fast quick-fix so we can do the release today. 👍

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
hohwille: I created issue #1311 for this.

PR #1315, opened 2025-05-16
#1293: make sure core.longpaths is true in windows git config

### LINKED ISSUE #1293: Ensure git core.longpaths on Windows
On Windows we have several projects that cannot clone their code git repositories properly and get this error many times:
```
error: unable to create file ... : Filename too long
```

Git offers a fix or workaround for this problem:
You need to run the following command before creating the project:
```
git config --system core.longpaths true
```

However, the idea of IDEasy is that setting up the IDE is easy.
Hence, we should do such things automatically to prevent such errors and increase UX.
Do not ask me why the Git Installer is not doing this out of the box (and also the Windows Terminal integration, etc.) and users always need expert knowledge to get things right.
So we should step in as IDEasy here.

## Solution

Somewhere here:
https://github.com/devonfw/IDEasy/blob/0d9b336079e34a8e8db2e825ffc39d2ba53a0e15/cli/src/main/java/com/devonfw/tools/ide/tool/IdeasyCommandlet.java#L180

We will add a private method to check the git setup.
First of all we should check if `git` is installed at all (`git` can be found on the PATH - see #900 for an easy way to do that).
If `git` is not installed, we should log an ERROR (but proceed):
```
Git is not installed on your computer but required by IDEasy. Please download and install git:
https://git-scm.com/download/
```

(Later we can automate that and download and install it automatically for the user)

Next we should do the trick if we are on Windows.
Technically the `git config --system` command is just modifying the file `~/.gitconfig`.
We can read that file and check if 
```
	longpaths = true
```

Is found in section
```
[core]
```

And if not we add it.

It might seem simpler to just call the `git config` command but we may (later) also check if the following properties are defined:
```
[user]
	name = ...
	email = ...
```

If that is not the case, we should ask the user for the values and set them.
Therefore we will need a feature to handle such "ini-files" in IDEasy anyway.
If the PR should be simplified, we can start with `git config` command and later consider changing that again.
I am just trying to outline more of the future vision.

cthies-capgemini:
### Implemented changes:

`installIdeasyWindowsEnv` now calls another method to check if the property `core.longpaths` is set in the `.gitconfig` file and adds it if it is missing, or sets it to true if it has another value.

I use a `LinkedHashMap` instead of a `HashMap` because it keeps the order of items as they were inserted. This means the order of sections and properties of the .gitignore file are preserved.

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/IdeasyCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/IdeasyCommandlet.java
@@ -238,8 +238,10 @@ public class IdeasyCommandlet extends MvnBasedLocalToolCommandlet {
     String gitconfig;
     // read files
     FileAccess fileAccess = this.context.getFileAccess();
-    fileAccess.touch(configPath);
     gitconfig = fileAccess.readFileContent(configPath);
+    if (gitconfig == null) {
+      gitconfig = "";
+    }
 
     LinkedHashMap<String, LinkedHashMap<String, String>> iniMap = parseIniFile(gitconfig);
 
--- a/cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
@@ -308,7 +308,7 @@ public interface FileAccess {
 
   /**
    * @param file the {@link Path} to the file to read.
-   * @return the content of the specified file (in UTF-8 encoding).
+   * @return the content of the specified file (in UTF-8 encoding), or null if the file doesn't exist
    * @see java.nio.file.Files#readString(Path)
    */
   String readFileContent(Path file);
--- a/cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
@@ -998,6 +998,10 @@ public class FileAccessImpl implements FileAccess {
   public String readFileContent(Path file) {
 
     this.context.trace("Reading content of file from {}", file);
+    if (!Files.exists((file))) {
+      this.context.debug("File {} does not exist", file);
+      return null;
+    }
     try {
       String content = Files.readString(file);
       this.context.trace("Completed reading {} character(s) from file {}", content.length(), file);


cli/src/main/java/com/devonfw/tools/ide/tool/IdeasyCommandlet.java
```diff
@@ -210,9 +213,86 @@ private void installIdeasyWindowsEnv(Path ideRoot, Path installationPath) {
...
   }
 
+  private void setGitLongpaths() {
+    SystemPath systemPath = this.context.getPath();
+    Path gitPath = systemPath.getPath("git");
+    if (gitPath == null) {
+      this.context.error("Git is not installed on your computer but required by IDEasy. Please download and install git:\n"
+          + "https://git-scm.com/download/");
+    } else {
+      setGitConfigProperty("core", "longpaths", "true", Path.of("~/.gitconfig"));
```
hohwille, line 227: Java does not understand `~`:
https://bugs.openjdk.org/browse/JDK-8152072
Issue was closed as `Not an issue` (we might not agree but that was Oracles choice).
```suggestion
      setGitConfigProperty("core", "longpaths", "true", this.context.getUserHome().resolve(".gitconfig"));
```
This is better anyhow since it allows testing in JUnits via `newContext` without modifying the developers real `.gitconfig`.

cli/src/main/java/com/devonfw/tools/ide/tool/IdeasyCommandlet.java
```diff
@@ -210,9 +213,86 @@ private void installIdeasyWindowsEnv(Path ideRoot, Path installationPath) {
...
+   * @param value the value to set
+   * @param configPath the path of the config file
+   */
+  public void setGitConfigProperty(String section, String property, String value, Path configPath) {
+    String gitconfig;
+    // read files
+    try {
+      gitconfig = new String(Files.readAllBytes(configPath));
+    } catch (IOException e) {
+      this.context.error("could not read git config file at %s", configPath);
```
hohwille, line 243: Nice that you added some logging for robustness.
However, can't we also create the file in such case with only the `longpaths=true` in `[core]`?
You can use `this.context.getFileAccess()` then you do not need to deal with `IOException` and get logging for free, etc.
We could even extend `readFileContent` with a variant that returns `null` if the file does not exist and just logs on debug that the file did not exist.
The idea of `FileAccess` is that we improve all such kind of operations and reuse that across our code-base since the offerings from JDK are rather low-level.

hohwille (approved):
@cthies-capgemini thanks for your PR. You did a great job to already solve the complete thing including the `*.ini` parsing/writing. Even you considered to use a `LinkedHashMap` to preserve the order of the sections. In a test you verify that all is working fine. 👍 

Could you please add the issue to our CHANGELOG?
I left some comments for improvement.
After all is addressed we can merge this.

cli/src/main/java/com/devonfw/tools/ide/tool/IdeasyCommandlet.java
cthies-capgemini: Good idea. I adjusted the `readFileContent` method to return null if the file doesn't exist, and changed `setGitConfigProperty` so it creates a new file with the necessary content when `null` is returned.

jan-vcapgemini (approved):
LGTM. Ready to merge.

juliane-cap:
Current implementation does not work on Windows, as I get the error message after installing IDEasy even though Git is installed on my computer.
```
Git is not installed on your computer but required by IDEasy. Please download and install git:
https://git-scm.com/download/
```
I also do not have longpaths set in my .gitconfig after running setup.bat.

PR #1327, opened 2025-05-20
#351: prevent environment variables from other project causing side-effects

### LINKED ISSUE #351: Avoid inheriting environment variables from other IDEasy project if switched in the same shell session

hohwille:
### Implemented changes:

* implemented #351
* ignore all environment properties relevant to IDEasy except for PATH, HOME, and IDE_ROOT
* added test
* improved test-infrastructure to behave more realistic using `EnvironmentVariablesSystem` that was otherwise mocked by a different class.

---

## Checklist for this PR

  with `internal`

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -9,6 +9,7 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/809[#809]: make uninstall with --force also remove from software repo
 * https://github.com/devonfw/IDEasy/issues/1038[#1038]: XML merger fails in native-image on custom XPath with MissingResourceException
 * https://github.com/devonfw/IDEasy/issues/1108[#1108]: Fix git authentification in terminal
+* https://github.com/devonfw/IDEasy/issues/351[#351]: Avoid inheriting environment variables from other IDEasy project if switched in the same shell session
 
 The full list of changes for this release can be found in https://github.com/devonfw/IDEasy/milestone/28?closed=1[milestone 2025.05.002].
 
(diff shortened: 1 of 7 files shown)


jan-vcapgemini (approved):
LGTM. Tested with a local graalvm native image.
Before:
image
After:
image

jan-vcapgemini:
I've tested these changes on Windows, Linux and Mac and it worked on all systems.

PR #1330, opened 2025-05-22
#1306 Redirect cleaned up in ProcessContext and ProcessMode

### LINKED ISSUE #1306: Cleanup redirect in ProcessContextImpl and ProcessMode
Leftover from PR #1305 :

We could create a cleanup PR to add 3 methods to `ProcessMode` returning `Redirect`:

* `getRedirectOutput`
* `getRedirectError`
* `getRedirectInput`

Then we adapt `ProcessContextImpl` to use these new methods instead of fiddling with `if` or `swtich` statements if multiple places.
If needed `ProcessMode.BACKGROUND*` could return `null` for some of these methods and if `null`, then the `redirect` is not configured in `processBuilder`.

Then we can make this code much more clean and understandable.

juliane-cap:
### Implemented changes:

* added getRedirect methods inside ProcessMode

---

## Checklist for this PR

- [ ] When running `mvn clean test` locally all tests pass and build is successful
- [ ] PR title is of the form `#«issue-id»: «brief summary»` (e.g. `#921: fixed setup.bat`). If no issue ID exists, title only.
- [ ] PR top-level comment summarizes what has been done and contains link to addressed issue(s)
- [ ] PR and issue(s) have suitable labels
- [ ] You followed all [coding conventions](https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc)
  with `internal`

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
@@ -130,14 +130,6 @@ public class ProcessContextImpl implements ProcessContext {
   @Override
   public ProcessResult run(ProcessMode processMode) {
 
-    if (processMode == ProcessMode.DEFAULT) {
-      this.processBuilder.redirectOutput(Redirect.INHERIT).redirectError(Redirect.INHERIT);
-    }
-
-    if (processMode == ProcessMode.DEFAULT_SILENT) {
-      this.processBuilder.redirectOutput(Redirect.DISCARD).redirectError(Redirect.DISCARD);
-    }
-
     if (this.executable == null) {
       throw new IllegalStateException("Missing executable to run process!");
     }
@@ -160,13 +152,9 @@ public class ProcessContextImpl implements ProcessContext {
     }
 
     try {
-
-      if (processMode == ProcessMode.DEFAULT_CAPTURE) {
-        this.processBuilder.redirectOutput(Redirect.PIPE).redirectError(Redirect.PIPE);
-      } else if (processMode.isBackground()) {
+      applyRedirects(processMode);
+      if (processMode.isBackground()) {
         modifyArgumentsOnBackgroundProcess(processMode);
-      } else {
-        this.processBuilder.redirectInput(Redirect.INHERIT);
       }
 
       this.processBuilder.command(args);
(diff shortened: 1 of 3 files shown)


cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -360,11 +348,7 @@ private void performLogging(ProcessResult result, int exitCode, String interpret
 
   private void modifyArgumentsOnBackgroundProcess(ProcessMode processMode) {
 
-    if (processMode == ProcessMode.BACKGROUND) {
-      this.processBuilder.redirectOutput(Redirect.INHERIT).redirectError(Redirect.INHERIT);
-    } else if (processMode == ProcessMode.BACKGROUND_SILENT) {
-      this.processBuilder.redirectOutput(Redirect.DISCARD).redirectError(Redirect.DISCARD);
-    } else {
+    if (!processMode.isBackground()) {
```
jan-vcapgemini, line 351: Do we still need this check if the only call of this method already checks that this has to be true?
Maybe replace with an assert instead.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessMode.java
```diff
@@ -12,29 +14,135 @@ public enum ProcessMode {
...
+    }
+
+    @Override
+    public Redirect getRedirectError() {
+      return Redirect.PIPE;
+    }
+
+    @Override
+    public Redirect getRedirectInput() {
+      return null;
```
jan-vcapgemini, line 92: I'm not sure if we should ignore the input here too.
DEFAULT_CAPTURE is used in various tools like lazydocker, mvn etc. and without a redirect the input will be ignored.

jan-vcapgemini, line 92:
```suggestion
      return Redirect.INHERIT;
```
Needs to be added or this change gets lost: https://github.com/devonfw/IDEasy/blob/22be644d9662e4979bdb84b7745212a0da114cf1/cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java#L169

cli/src/main/java/com/devonfw/tools/ide/process/ProcessMode.java
```diff
@@ -12,29 +14,135 @@ public enum ProcessMode {
...
+    }
+
+    @Override
+    public Redirect getRedirectError() {
+      return Redirect.INHERIT;
+    }
+
+    @Override
+    public Redirect getRedirectInput() {
+      return Redirect.INHERIT;
```
jan-vcapgemini, line 70: I'm not sure if this should be added in this PR, maybe add a new issue to check if this is needed otherwise we will introduce new functionality here.

jan-vcapgemini (approved):
@juliane-cap thanks for your awesome refactoring of the ProcessModes. I struggled a bit with the input redirects, but now that it's cleaner it helps a lot. It would be nice if we could somehow test this (check if a test with a simulated input will redirect properly) but this will be out of scope of this PR I guess. We should add another issue for a test of simulated inputs though.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessMode.java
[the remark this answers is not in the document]
jan-vcapgemini: Issue created: https://github.com/devonfw/IDEasy/issues/1339

PR #1348, opened 2025-05-30
#1331: add parser for .ini files

### LINKED ISSUE #1331: Add Parser for .ini Files
### Feature idea

As a developer, I want to parse and edit .ini files with a built-in utility class instead of parsing them "by hand" every time. The utility class should be able to read such a file from a given path and offer methods to read, update and delete sections and properties.

#1315 already parses an .ini file in the form of a .gitignore file, and can be used as a starting point for this parser.

### Additional context

_No response_

cthies-capgemini:
### Implemented changes:

* Add IniParser and IniParserImpl to parse and edit .ini files
* Use new class to ensure core.longpaths property in gitconfig on windows systems


cli/src/main/java/com/devonfw/tools/ide/io/IniParser.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Interface that allows parsing of .ini files such as .gitignore files
+ */
+public interface IniParser {
+
+  /**
+   * @return String[] the sections
+   */
+  String[] getSections();
```
hohwille, line 14: I would simplify this API so the `IniFile` is more or less internally a Map from section name to sections.
A `IniSection` is a simple container that has a name (`String getName()`) and `properties` (`Properties getProperties()` or `Map<String, String> getProperties()`). You do not seem to track changes for lazy writing if nothing changed so this would simplify the API since you do not need to define `String getProperty(String)`, `void removeProperty(String)`, `void setProperty(String)` - and in special cases the API user can make use of additional features e.g. to track if removing or adding a property actually changed something or not what is not possible when you return `void`...

For ease of use I would not reveal the internal Map and instead offer methods like
```
IniSection getSection(String);
IniSection getOrCreateSection(String);
boolean removeSection(String);
```

Potentially you may also want to add `getSectionRequired(String)` that will throw an exception instead of returning `null`.

cli/src/main/java/com/devonfw/tools/ide/io/IniParser.java
```diff
@@ -0,0 +1,59 @@
+package com.devonfw.tools.ide.io;
+
+import java.nio.file.Path;
+import java.util.HashMap;
+
+/**
+ * Interface that allows parsing of .ini files such as .gitignore files
+ */
+public interface IniParser {
```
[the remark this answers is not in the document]
hohwille: An `IniParser` would have a methods like:
```
IniFile parse(String);
IniFile parse(InputStream);
IniFile parse(Reader);
```

However, this could IMHO be directly offered in `FileAccess` API.

cli/src/main/java/com/devonfw/tools/ide/io/IniParserImpl.java
```diff
@@ -0,0 +1,127 @@
...
+
+import com.devonfw.tools.ide.context.IdeContext;
+
+/**
+ * Implementation of {@link IniParser} preserves order of sections and properties between reading and writing
+ */
+public class IniParserImpl implements IniParser {
+
+  LinkedHashMap<String, LinkedHashMap<String, String>> iniMap;
+  IdeContext context;
```
hohwille, line 16: Avoid keeping such state inside the data-structure for an `.ini` file and rather make it stateless.

cli/src/test/java/com/devonfw/tools/ide/tool/IdeasyCommandletTest.java
```diff
@@ -82,80 +79,6 @@ public void testInstallIdeasy(String os) {
         + addedRcLines);
   }
 
-  /**
-   * Test of {@link IdeasyCommandlet#setGitConfigProperty(String, String, String, Path)}
-   *
-   * @throws IOException if creation of temporary file fails
-   */
-  @Test
-  public void testSetGitConfigProperty() throws IOException {
```
hohwille, line 91: you actually removed the test that is verifying the feature.
Was there some problem with the test and you could not get it green again?
It seems you more or less moved it to `IniParserImplTest`. However, this is not testing the integration into the IDEasy installation process anymore, so we might still want to consider a test here that could reuse the logic from `IniParserImplTest`. WDYT?

CHANGELOG.adoc
```diff
@@ -12,7 +12,7 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/1293[#1293]: make sure core.longpaths is true in windows git config
 * https://github.com/devonfw/IDEasy/issues/1307[#1307]: Link to settings documentation is broken
 * https://github.com/devonfw/IDEasy/issues/351[#351]: Avoid inheriting environment variables from other IDEasy project if switched in the same shell session
-* https://github.com/devonfw/IDEasy/issues/1332[#1332]: cannot launch eclipse due to failling plugin
+* https://github.com/devonfw/IDEasy/issues/1332[#1332]: cannot launch eclipse due to failing plugin
```
hohwille, line 15: thanks for typo fix. 👍

hohwille (approved):
@cthies-capgemini technically your PR is working well. 👍 

I left some remarks about the design. Sorry, due to my vacation I was unable to provide that upfront what would have made it much easier for you. I hope still all makes sense to you and is after all some helpful learning about design.

cli/src/test/java/com/devonfw/tools/ide/tool/IdeasyCommandletTest.java
cthies-capgemini: Good point. I added a line in the testInstallIdeasy test to check whether the longpaths property is added to the gitconfig on windows.

cli/src/main/java/com/devonfw/tools/ide/io/IniParser.java
cthies-capgemini: yes, I can see how that design is nicer to use. Thanks for the suggestions, I changed the design accordingly.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
```diff
@@ -407,6 +407,20 @@ default void writeProperties(Properties properties, Path file) {
   void writeProperties(Properties properties, Path file, boolean createParentDir);
 
 
+  default IniFile readIniFile(Path file) {
```
jan-vcapgemini, line 410: Please add Javadoc.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
```diff
@@ -407,6 +407,20 @@ default void writeProperties(Properties properties, Path file) {
...
+    IniFile iniFile = new IniFileImpl();
+    readIniFile(file, iniFile);
+    return iniFile;
+  }
+
+  void readIniFile(Path file, IniFile iniFile);
+
+  default void writeIniFile(IniFile iniFile, Path file) {
```
jan-vcapgemini, line 418: Please add Javadoc.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java
```diff
@@ -407,6 +407,20 @@ default void writeProperties(Properties properties, Path file) {
...
+    writeIniFile(iniFile, file, false);
+  }
+
+  void writeIniFile(IniFile iniFile, Path file, boolean createParentDir);
```
jan-vcapgemini, line 422: Please add Javadoc.

cli/src/main/java/com/devonfw/tools/ide/io/IniFile.java
```diff
@@ -0,0 +1,30 @@
...
+public interface IniFile {
+
+  /**
+   * @return String[] the sections
+   */
+  String[] getSectionNames();
+
+  /**
+   * @param section the section to remove
+   * @return true if section existed, false otherwise
```
jan-vcapgemini, line 15:
```suggestion
   * @return {@code true} if section existed, {@code false} otherwise
```

cli/src/main/java/com/devonfw/tools/ide/io/IniFile.java
```diff
@@ -0,0 +1,30 @@
...
+   */
+  boolean removeSection(String section);
+
+  /**
+   * @param section the section to get
+   * @return IniSection if section exists, null otherwise
```
jan-vcapgemini, line 21:
```suggestion
   * @return {@link IniSection} if section exists, null otherwise
```

jan-vcapgemini:
As this appears to be a quite critical feature I'd suggest to also add an integration test to make sure it works on all OSes.
The test basically should check if the `core.longpaths` variable was set after the installation of IDEasy.

cli/src/test/integration-tests/ensure-longpaths.sh
```diff
@@ -0,0 +1,8 @@
+if doIsWindows
+then
+  echo "Running install intellij integration test"
+  ide -d install intellij
```
jan-vcapgemini, line 4: This is not necessary for this tests I think.

cli/src/test/integration-tests/ensure-longpaths.sh
```diff
@@ -0,0 +1,8 @@
+if doIsWindows
+then
+  echo "Running install intellij integration test"
```
jan-vcapgemini, line 3:
```suggestion
  echo "Running ensure git config longpaths gets set integration test"
```

cli/src/test/integration-tests/ensure-longpaths.sh
cthies-capgemini: yeah, I agree. I somehow mixed up installing ideasy and intellij in my head - I fixed it now

cli/src/test/integration-tests/ensure-longpaths.sh
cthies-capgemini: I added the updated description

cthies-capgemini:
The integration test I added works so far as it finds the longpaths property, but for some reason the intellij installation test fails now because it can't find the idea.properties file. I haven't yet figured out what's happening here.

cli/src/main/java/com/devonfw/tools/ide/io/IniFileImpl.java
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * creates empty IniFileImpl
+   */
+  public IniFileImpl() {
+    this.iniMap = new LinkedHashMap<>();
+  }
+
+  @Override
+  public String[] getSectionNames() {
+    return iniMap.keySet().toArray(new String[0]);
```
hohwille, line 21: Same thing but more elegant (since it does not create an empty array that is then discarded but only used to determine the type of the array.
```suggestion
    return iniMap.keySet().toArray(String[]::new);
```

cli/src/main/java/com/devonfw/tools/ide/io/IniFileImpl.java
```diff
@@ -0,0 +1,61 @@
...
+  @Override
+  public IniSection getOrCreateSection(String section) {
+    IniSection iniSection;
+    if (!iniMap.containsKey(section)) {
+      iniSection = new IniSectionImpl(section);
+      iniMap.put(section, iniSection);
+    } else {
+      iniSection = iniMap.get(section);
+    }
+    return iniSection;
```
hohwille, line 45: Can be simplified like this:
```suggestion
    return this.iniMap.computeIfAbsent(section, IniSectionImpl::new);
```

cli/src/main/java/com/devonfw/tools/ide/io/IniFileImpl.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  @Override
+  public String toString() {
+    StringBuilder stringBuilder = new StringBuilder();
+    for (String configSection : iniMap.keySet()) {
+      stringBuilder.append(String.format("[%s]\n", configSection));
```
hohwille, line 52:
```suggestion
      stringBuilder.append('[');
      stringBuilder.append(configSection);
      stringBuilder.append("]\n");
```

cli/src/main/java/com/devonfw/tools/ide/io/IniFileImpl.java
```diff
@@ -0,0 +1,61 @@
...
+      LinkedHashMap<String, String> properties = (LinkedHashMap<String, String>) iniMap.get(configSection).getProperties();
```
hohwille, line 53:
```suggestion
      Map<String, String> properties = iniMap.get(configSection).getProperties();
```

cli/src/main/java/com/devonfw/tools/ide/io/IniFileImpl.java
```diff
@@ -0,0 +1,61 @@
...
+      for (String sectionProperty : properties.keySet()) {
+        String propertyValue = properties.get(sectionProperty);
+        stringBuilder.append(String.format("\t%s = %s\n", sectionProperty, propertyValue));
+      }
```
hohwille, line 57:
```suggestion
      for (Entry<String, String> entry : properties.entrySet()) {
        stringBuilder.append('\t');
        stringBuilder.append(entry.getKey());
        stringBuilder.append(" = ");
        stringBuilder.append(entry.getValue());
        stringBuilder.append('\n');
      }
```

cli/src/main/java/com/devonfw/tools/ide/io/IniFileImpl.java
```diff
@@ -0,0 +1,61 @@
+package com.devonfw.tools.ide.io;
+
+import java.util.LinkedHashMap;
+
+/**
+ * Implementation of {@link IniFile} preserves order of sections and properties between reading and writing
+ */
+public class IniFileImpl implements IniFile {
+
+  LinkedHashMap<String, IniSection> iniMap;
```
hohwille, line 10: https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc#prefer-general-api
use `final` and `private` where possible.
```suggestion
  private final Map<String, IniSection> iniMap;
```

cli/src/main/java/com/devonfw/tools/ide/io/IniSectionImpl.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Implementation of {@link IniSection}
+ */
+public class IniSectionImpl implements IniSection {
+
+  String name;
+  LinkedHashMap<String, String> properties;
```
hohwille, line 12:
```suggestion
  private final String name;
  private final Map<String, String> properties;
```

hohwille (approved):
@cthies-capgemini thanks for your rework. You did a really great job to get the design right.
I still left quite some review comments but all are only constructive suggestions for simpler or better code-style that you can all apply.
BTW: Did you study https://en.wikipedia.org/wiki/INI_file?
INI-files also support comments.
These will be lost when using your code to read, update and write an INI file.
We can create a new story for enhancing the INI parser accordingly so we do not further block this PR from being merged.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -1152,6 +1153,36 @@ public void writeProperties(Properties properties, Path file, boolean createPare
...
+        currentIniSection = iniFile.getOrCreateSection(sectionName);
+      } else {
+        String[] parts = line.split("=");
+        String propertyName = parts[0].trim();
+        String propertyValue = parts[1].trim();
+        if (currentIniSection == null) {
+          Log.warn("Invalid ini-file with property {} before section", propertyName);
+        } else {
+          currentIniSection.getProperties().put(propertyName, propertyValue);
+        }
```
hohwille, line 1175: here you will cause an `ArrayIndexOutOfBoundsException` in case we have an empty line or a comment.
Before we can merge, we should at least do a check.
Further, there may be funny edge-cases like this:
```
key=value=value2
```
In such case we get 3 parts and you would ignore `=value2` what is actually still part of the value.

```suggestion
        int index = line.indexOf('=');
        if (index > 0) {
          String propertyName = line.substring(0, index).trim();
          String propertyValue = line.substring(index + 1).trim();
          if (currentIniSection == null) {
            Log.warn("Invalid ini-file with property {} before section", propertyName);
          } else {
            currentIniSection.getProperties().put(propertyName, propertyValue);
          }  
        } else {
          // here we can handle comments and empty lines in the future
        }
```

cli/src/main/java/com/devonfw/tools/ide/tool/IdeasyCommandlet.java
```diff
@@ -228,71 +230,13 @@ private void installIdeasyWindowsEnv(Path ideRoot, Path installationPath) {
...
-        String propertyName = parts[0].trim();
-        String propertyValue = parts[1].trim();
-        iniMap.get(currentSection).put(propertyName, propertyValue);
-      }
-    }
-    return iniMap;
+    Path configPath = this.context.getUserHome().resolve(".gitconfig");
+    IniFile iniFile = new IniFileImpl();
+    FileAccess fileAccess = new FileAccessImpl(context);
+    fileAccess.readIniFile(configPath, iniFile);
```
hohwille, line 236: Do not create a new FileAccessImpl but use the existing one. Also you shall not know about the implementation here - we might even have a different implementation in tests, who knows (implementation secret pattern). Also you created a nice convenience method to create the `iniFile` automatically that you can use to simplify to a single line:
```suggestion
    IniFile iniFile = this.context.getFileAccess().readIniFile(configPath);
```

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -1152,6 +1153,36 @@ public void writeProperties(Properties properties, Path file, boolean createPare
     }
   }
 
+  @Override
+  public void readIniFile(Path file, IniFile iniFile) {
+    List<String> iniLines = readFileLines(file);
```
hohwille, line 1158: You need to handle the case that the `iniFile` may not exist.
The way you are using this code in `IdeasyCommandlet` will cause an exception of the file never existed on your computer.
I would recommend to design this idempotent so this method will do nothing (except logging a debug message) if the `iniFile`does not exist.
```suggestion
    if (!Files.exists(file)) {
      this.context.debug("INI file {} does not exist.", iniFile);
    }
    List<String> iniLines = readFileLines(file);
```

cli/src/test/integration-tests/verify-longpaths.sh
```diff
@@ -0,0 +1,9 @@
+if doIsWindows
+then
+  echo "Running ensure git config longpaths gets set integration test"
+  touch "$HOME"/.gitconfig
```
hohwille, line 4: Without this line, you will reveal if your code can also handle the case to create the file from scratch properly.
You must not merge/accept this. If you leave it like this, you will also test that the INI parsing can handle empty files...
```suggestion
```

hohwille (approved):
@cthies-capgemini as we both saw, this PR reached more complexity than initially expected. However, you reached a great result. While I was adding the last additional review comments you were already resolving them at the same time - please enjoy your weekend ;) You did a great job, and we can finally merge this 👍 Hurray 🍾

cthies-capgemini:
@hohwille Yeah, this PR was definitely more complex than I first thought. Thank you for your detailed feedback on it, I learned a lot of things about design from it! Please also enjoy your weekend ^^

hohwille:
> @jan-vcapgemini Requested changes

I will overrule this so we get this done.

PR #1365, opened 2025-06-10
#1133: Fix autocompletion of tool versions to only show the version numbers

### LINKED ISSUE #1133: Version autocomple with full list displays files too
### Expected behavior

As an IDEasy user I expect IDEasy to list only version numbers if I type ide install vscode [TAB] and choose to display all versions.

### Actual behavior
IDEasy lists version numbers as well as folders/files of the current working directory:

```

...

1.25.1                          1.51.1                          1.75.0                          1.98.1
1.26.0                          1.52.0                          1.75.1                          CHANGELOG.adoc
1.26.1                          1.52.1                          1.76.0                          CHANGELOG.adoc.orig
1.27.0                          1.53.0                          1.76.1                          cli
1.27.1                          1.53.1                          1.76.2                          documentation
1.27.2                          1.53.2                          1.77.0                          gui
1.28.0                          1.54.0                          1.77.1                          ide-urls
1.28.1                          1.54.1                          1.77.2                          LICENSE
1.28.2                          1.54.2                          1.77.3                          null
1.29.0                          1.54.3                          1.78.0                          pom.xml
1.29.1                          1.55.0                          1.78.1                          README.adoc
1.3.0                           1.55.1                          1.78.2                          security
1.3.1                           1.55.2                          1.79.0                          target
1.30.0                          1.56.0                          1.79.1                          test
1.30.1                          1.56.1                          1.79.2                          test.sh
1.30.2                          1.56.2                          1.8.0                           test-pwd.sh
1.31.0                          1.57.0                          1.8.1                           url-updater
1.31.1                          1.57.1                          1.80.0                          windows-installer
1.32.0                          1.58.0                          1.80.1
```

1. cd into a non empty directory
2. Type `ide install vscode` with a trailing tab

### Affected version:

* OS: Windows 10 x64
* IDEasy: 2025.03.001-03_11_02-SNAPSHOT

juliane-cap:
### Implemented changes:

* Temporarily disable filename expansion to not translate "*" into the files of the current directory

---

## Checklist for this PR

  with `internal`

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -6,11 +6,8 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
-<<<<<<< 1133-version-autocomplete-bug
 * https://github.com/devonfw/IDEasy/issues/1133[#1133]: Version autocomplete with full list displays files too
-=======
 * https://github.com/devonfw/IDEasy/issues/1372[#1372]: Add gradlew support
->>>>>>> main
 * https://github.com/devonfw/IDEasy/issues/1304[#1304]: Error when reinstalling vscode
 * https://github.com/devonfw/IDEasy/issues/1160[#1160]: Print IDEasy version in ide status when offline
 * https://github.com/devonfw/IDEasy/issues/901[#901]: Consider maven wrapper mvnw


cli/src/main/package/functions
```diff
@@ -93,7 +93,9 @@ function icd() {
 _ide_completion()
 {
   if [ -z "${COMP_WORDS[COMP_CWORD]}" ]; then
-    COMPREPLY=( $(ideasy -q complete ${COMP_WORDS[@]:1} "") )
+    set -f
+    COMPREPLY=( $(ideasy -q complete ${COMP_WORDS[@]:1} "" ) )
+    set +f
   else
     COMPREPLY=( $(ideasy -q complete ${COMP_WORDS[@]:1}) )
```
hohwille, line 100: This is a great finding and fix.
However, my thoughts on this are:
* shouldn't we do `set -/+f` for all completions - so also for the one in `else` or to be more explicitly around the entire `if` block?
* with `set -f` we may tweak the shell of the user. Therefore it is great that you try to "undo" this effect afterwards. Just thinking out loud: What would happen if the user actually does `set -f` in his terminal because he for some reason does not want to have filename expansion? Then here `set -f` would have no effect but `set +f` would not reset the previous state.

hohwille: For the second one I tried some things. It git-bash you can do this:
```
$ function foo() { local -; set -f; echo *; }

$ foo
*

$ echo *
backups conf scripts settings software system updates urls workspaces
```

However, `local -` is bash specific and in `zsh` we would need to do `set -o localoptions` instead.
Even worse `local -` is only supported by bash >= 4.4 and on MacOS this is not given so it would then totally fail.
As the second point seems more of a nice-to-have to support edge-cases, I would say,  that you can ignore it in this PR and maybe you could create a new story for this that we could put in the `Research` state/column on our board.

hohwille (approved):
@juliane-cap thanks for this PR. Great finding that bash `file expansion` is causing this effect and you can prevent it with `set -f`. 👍 
I left a comment. As I already commented again, the second aspect does not need to be addressed by this PR. For the first one either explain why it is not relevant in the `else` case or apply the solution to the entire `if` block. IMHO I am correct and we also want to have this for the `else`. Please note that file expansion for `PathProperties` is already implemented in Java and therefore should not additionally happen in bash as well. On the other hand if the value is not a `Path` then we do not want to have file expansion so our Java auto-completion is already smarter than bash.

hohwille (approved):
finally ready for merge.

PR #1373, opened 2025-06-16
#901: Add mvn wrapper detection

### LINKED ISSUE #901: Consider maven wrapper mvnw
As a IDEasy user, I want support that IDEasy should honor maven wrapper in case it is used so that the correct maven is used.

A code repository may use [maven-wrapper](https://maven.apache.org/wrapper/) (`mvnw`).
From IDEasy we recommend to configure your IDE via IDEasy but there may be use-cases where you might need multiple versions of maven within the same IDE project or you just use your existing IDE project to clone and test some other project using maven wrapper.
So whenever `mvnw` is present, it should be honored and preferred over the regular maven (`mvn`).

To conclude the `Mvn` commandlet should check if `mvnw` is present and if so, it should be used instead of `mvn`.
You also need to be aware that `mvnw` will typically only be present in the project root.
That means we need to traverse up the path (as long as `pom.xml` files are present) until we found `mvnw`.
If no `mvnw` was found, we should continue as normal calling `mvn` directly.
However, if we have found `mvnw`, we should call that instead.
We have to test if we then need to check if we can still call our maven command with an absolute path to the `mvnw` we found in project root inside the potential sub-directory we might be in.
See also https://github.com/takari/maven-wrapper/issues/30

jan-vcapgemini:
### Implemented changes:

* added mvn wrapper file detection
* added test for wrapper file detection
* added mvnw file to test resources

---

## Checklist for this PR

- [ ] When running `mvn clean test` locally all tests pass and build is successful
- [ ] PR title is of the form `#«issue-id»: «brief summary»` (e.g. `#921: fixed setup.bat`). If no issue ID exists, title only.
- [ ] PR top-level comment summarizes what has been done and contains link to addressed issue(s)
- [ ] PR and issue(s) have suitable labels
- [ ] You followed all [coding conventions](https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc)
  with `internal`

- [ ] The tool can be installed automatically (during setup via settings) or via the commandlet call
- [ ] The tool is isolated in its IDEasy project, see [Sandbox Principle](https://github.com/devonfw/IDEasy/blob/main/documentation/sandbox.adoc)
- [ ] The new tool is added to the table of tools in [LICENSE.asciidoc](https://github.com/devonfw/IDEasy/blob/main/documentation/LICENSE.adoc)
- [ ] The new commandlet is a [command-wrapper](https://github.com/devonfw/IDEasy/blob/main/documentation/cli.adoc#command-wrapper) for `«tool»`
- [ ] Proper help texts for all supported languages are added [here](https://github.com/devonfw/IDEasy/tree/main/cli/src/main/resources/nls)
- [ ] The new commandlet installs potential dependencies automatically
- [ ] The variables `«TOOL»_VERSION` and `«TOOL»_EDITION` are honored by your commandlet

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/mvn/Mvn.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/mvn/Mvn.java
@@ -5,6 +5,7 @@ import java.nio.file.Path;
 import java.security.SecureRandom;
 import java.util.Base64;
 import java.util.LinkedHashSet;
+import java.util.Objects;
 import java.util.Set;
 import java.util.regex.Matcher;
 
@@ -41,6 +42,12 @@ public class Mvn extends PluginBasedCommandlet {
   /** The name of the settings-security.xml */
   public static final String SETTINGS_SECURITY_FILE = "settings-security.xml";
 
+  /** The mvn wrapper file name. */
+  public static final String MVN_WRAPPER_FILENAME = "mvnw";
+
+  /** The pom.xml file name. */
+  public static final String POM_XML = "pom.xml";
+
   /**
    * The name of the settings.xml
    */
@@ -64,6 +71,34 @@ public class Mvn extends PluginBasedCommandlet {
   public Mvn(IdeContext context) {
 
     super(context, "mvn", Set.of(Tag.JAVA, Tag.BUILD));
+  }
+
+  @Override
+  protected void configureToolBinary(ProcessContext pc, ProcessMode processMode, ProcessErrorHandling errorHandling) {
+    Path mvn = Path.of(getBinaryName());
+    Path wrapper = findWrapper();
+    pc.executable(Objects.requireNonNullElse(wrapper, mvn));
+  }
+
+  /**
+   * Searches for a wrapper file in valid mvn projects (containing a pom.xml) and returns its path.
+   *
+   * @return Path of the wrapper file or {@code null} if none was found.
+   */
+  protected Path findWrapper() {
+    Path dir = context.getCwd();
+    // traverse the cwd containing a pom.xml up till a maven wrapper file was found
+    while (Files.exists(dir.resolve(POM_XML)) &&
+        !Files.exists(dir.resolve(MVN_WRAPPER_FILENAME)) &&
+        dir.getParent() != null) {
+      dir = dir.getParent();
+    }
+    if (Files.exists(dir.resolve("mvnw"))) {
+      context.debug("Using mvn wrapper file at: {}", dir);
+      return dir.resolve(MVN_WRAPPER_FILENAME);
+    }
+
+    return null;
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


cli/src/main/java/com/devonfw/tools/ide/tool/mvn/Mvn.java
```diff
@@ -66,6 +73,34 @@ public Mvn(IdeContext context) {
...
+   */
+  protected Path findWrapper() {
+    Path dir = context.getCwd();
+    // traverse the cwd containing a pom.xml up till a maven wrapper file was found
+    while (Files.exists(dir.resolve(POM_XML)) &&
+        !Files.exists(dir.resolve(MVN_WRAPPER_FILENAME)) &&
+        dir.getParent() != null) {
+      dir = dir.getParent();
+    }
+    if (Files.exists(dir.resolve("mvnw"))) {
```
hohwille, line 96:
```suggestion
    if (Files.exists(dir.resolve(MVN_WRAPPER_FILENAME))) {
```

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/Mvn.java
```diff
@@ -41,6 +42,12 @@ public class Mvn extends PluginBasedCommandlet {
   /** The name of the settings-security.xml */
   public static final String SETTINGS_SECURITY_FILE = "settings-security.xml";
 
+  /** The mvn wrapper file name. */
+  public static final String MVN_WRAPPER_FILENAME = "mvnw";
```
hohwille, line 46: Just for the record: There is also `mvnw.cmd` created by https://maven.apache.org/wrapper/maven-wrapper-plugin/usage.html[maven-wrapper-plugin].
IMHO we do not need to handle that specifically.
I would consider it an odd edge-case if `mvnw.cmd` is present but `mvnw` bash script is not.
As we also always require `bash` as pre-requisite and already have logic to detect and execute bash scripts, there is nothing else to do here. I just wanted to point this out. Will close the comment to prevent blocking the merge.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/Mvn.java
```diff
@@ -5,6 +5,7 @@
 import java.security.SecureRandom;
 import java.util.Base64;
 import java.util.LinkedHashSet;
+import java.util.Objects;
```
hohwille, line 8: Please add #901 to CHANGELOG.

hohwille (approved):
@jan-vcapgemini thanks for your PR. This is great work and you also included a JUnit test covering the new feature. Awesome job 👍

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/Mvn.java
jan-vcapgemini: Adjusted.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/Mvn.java
jan-vcapgemini: Added.

PR #1431, opened 2025-07-25
#1389: Add podman support

### LINKED ISSUE #1389: support podman as docker alternative
### Feature idea

As a podman user, I want to use podman instead of docker without being forced to get docker-desktop or rancher-desktop installed.
So if I run `ide docker ...` I would expect IDEasy to detect that I have `podman` installed (if not `docker` was found) and use `podman` instead of `docker`.

### Additional context

For the record: podman claims to be CLI API compatible to docker. So if `docker <args>` is doing something then `podman <args>` should do the same thing.
Therefore `Docker` commandlet in IDEasy should detect `podman` if `docker` was not found and use it as fallback.

jan-vcapgemini:
### Implemented changes:

* added a simple Path exists check on the PATH env var to check if docker is installed
* adjusted javadoc of findBinary method
* added new findBinaryPathByName method to SystemPath class

---

## Checklist for this PR

  with `internal`

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
@@ -9,8 +9,10 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Stream;
 
+import com.devonfw.tools.ide.cli.CliException;
 import com.devonfw.tools.ide.context.AbstractIdeContext;
 import com.devonfw.tools.ide.context.IdeContext;
+import com.devonfw.tools.ide.context.IdeStartContextImpl;
 import com.devonfw.tools.ide.git.GitContext;
 import com.devonfw.tools.ide.git.GitUrl;
 import com.devonfw.tools.ide.git.repository.RepositoryCommandlet;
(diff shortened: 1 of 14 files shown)


cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -33,6 +37,36 @@ public Docker(IdeContext context) {
...
+      processContext.addArgs("--version");
+      ProcessResult result = processContext.run(mode);
+      if (result.getExitCode() != 0) {
+        return false;
+      }
+    } catch (Exception e) {
+      return false;
+    }
+    return true;
+  }
```
hohwille, line 68: We have implemented a better solution with PR #111 for this purpose.

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -33,6 +37,36 @@ public Docker(IdeContext context) {
...
+
+  private String detectContainerRuntime() {
+    if (isCommandAvailable(this.getName())) {
+      return this.getName();
+    } else if (isCommandAvailable(PODMAN)) {
+      return PODMAN;
+    } else {
+      return this.getName();
+    }
```
hohwille, line 52: I assume this will be called multiple times. Shall we recompute the result every time or store the result once computed?
Maybe if we change the implementation so we do not trigger processes but just do a Path search, this will be "cheap" enough to recompute every time.
However, I do not expect a docker or podman installation to come and go on a daily basis (and surely when we find nothing, the variable would still be null and the recomputation should still take place, but once we found podman but no docker, I do not expect this to change within the running process - even in GUI dashboard).

jan-vcapgemini: I've changed the isCommandAvailable logic to just check for the path in the PATH. So this can be run multiple times now.

PR #1448, opened 2025-08-12
#970: Adjusted tests to use WireMock properly

### LINKED ISSUE #970: Align tests and avoid to manually write into project config
Both in PR #909 and in #903 we introduced an antipattern to our tests that directly write into the test project configuration what should be avoided.

https://github.com/devonfw/IDEasy/blob/58fd72aed131d9cb4d9a0689aec3811368cee3f9/cli/src/test/java/com/devonfw/tools/ide/tool/intellij/IntellijTest.java#L189-L201

A better solution can be found e.g. here:
https://github.com/devonfw/IDEasy/blob/main/cli/src/test/resources/ide-projects/dotnet/_ide/urls/dotnet/dotnet/6.0.419/urls

And we have code to resolve such variables:
https://github.com/devonfw/IDEasy/blob/58fd72aed131d9cb4d9a0689aec3811368cee3f9/cli/src/test/java/com/devonfw/tools/ide/tool/npm/NpmJsonUrlUpdaterTest.java#L68-L71

To make it even better, I would expect that we provide a protected method in `AbstractIdeContextTest` with the logic to replace such variable from `WireMockRuntimeInfo` to centralize the approach and reuse it from the places where we need it.
Ideally the test-case does not have to do anything manually to get this variable `${testbaseurl}` replaced.
Maybe we can pass `WireMockRuntimeInfo` as optional argument to `AbstractIdeContextTest.newProject` method variants so that such variable is automatically replaced why the project is copied to `target` folder.

Also I found these left-overs that should be fixed to not use hardcoded wiremock ports (most probably these URLs are not used at all, but it should be aligned - see #706):
https://github.com/devonfw/IDEasy/blob/58fd72aed131d9cb4d9a0689aec3811368cee3f9/cli/src/test/resources/ide-projects/basic/_ide/urls/java/java/17.0.6/mac_x64.urls#L1
https://github.com/devonfw/IDEasy/blob/58fd72aed131d9cb4d9a0689aec3811368cee3f9/cli/src/test/resources/ide-projects/basic/_ide/urls/java/java/17.0.6/linux_x64.urls#L1
https://github.com/devonfw/IDEasy/blob/58fd72aed131d9cb4d9a0689aec3811368cee3f9/cli/src/test/resources/ide-projects/basic/_ide/urls/java/java/17.0.6/windows_x64.urls#L1

Also, I consider this approach as an anti-pattern:
https://github.com/devonfw/IDEasy/blob/58fd72aed131d9cb4d9a0689aec3811368cee3f9/cli/src/test/java/com/devonfw/tools/ide/commandlet/EditionGetCommandletTest.java#L43

It would be better to use a different tool for this kind of test where we test what happens if the tool was already installed.
The problem is that all the tests of `get-version` and `get-edition` are using the same test-project and the same tool `az`.
If you read the tests and see that in the test-project `az` is not installed and therefore we get a specific behaviour if we call different variants of `get-version` or `get-edition` when reading the code you start getting of an idea of this scenario reused by many of the test methods.
However, in the test method linked the existing scenario is "mutated" and the behaviour and setup of `az` is therefore changing leading to different results. Surely that is what we want to test that the result also depends on the state.
However, we had a bug fixed by #955 where this mutation was accidentally causing side-effects for other tests to fail if they were executed afterwards but had been green if executed before this mutation.
So to make it short: A better approach is to create a new "space" where this scenario is given. To avoid the need to create an extra new test-project just for this, I would recommend to just use another tool instead of `az` that is not yet included in this test-project and not used by any other tests using this test-project.

jan-vcapgemini:
### Implemented changes:

* removed unnecessary WireMock mocks from InstallCommandletTest
* changed tool of testEditionGetCommandletRunWithoutFlagsAndNotInstalled test from az to gh
* adjusted basic project test resource urls to use new WireMock placeholder
* adjusted InstallCommandletTest tests to use new wmRuntimeInfo
* removed unnecessary test resource binary files
* added mocked java bin files to basic project
* added common mock script content to java scripts

---

## Checklist for this PR

  with `internal`

### DIFF
--- a/cli/src/test/resources/ide-projects/basic/repository/java/java/default/bin/java
+++ b/cli/src/test/resources/ide-projects/basic/repository/java/java/default/bin/java
@@ -1,0 +1,2 @@
+#!/bin/bash
+echo java $*
--- a/cli/src/test/resources/ide-projects/basic/repository/java/java/default/bin/java.cmd
+++ b/cli/src/test/resources/ide-projects/basic/repository/java/java/default/bin/java.cmd
@@ -1,0 +1,2 @@
+#!/bin/bash
+echo java $*
--- a/cli/src/test/resources/ide-projects/build/repository/java/java/default/bin/java
+++ b/cli/src/test/resources/ide-projects/build/repository/java/java/default/bin/java
@@ -1,0 +1,2 @@
+#!/bin/bash
+echo java $*
--- a/cli/src/test/resources/ide-projects/build/repository/java/java/default/bin/java.cmd
+++ b/cli/src/test/resources/ide-projects/build/repository/java/java/default/bin/java.cmd
@@ -1,0 +1,2 @@
+#!/bin/bash
+echo java $*


hohwille (commented):
@jan-vcapgemini great improvement - thanks for your PR. 👍 
I was not even aware that we still had tests using legacy wiremock approach with hardcoded port-numbers.
However, you added binary software-packages as tgz files. We should strictly avoid that since this is unmaintainable in git.
I demonstrated how to solve this properly in JavaTest. Please follow that example.

jan-vcapgemini:
> @jan-vcapgemini great improvement - thanks for your PR. 👍 I was not even aware that we still had tests using legacy wiremock approach with hardcoded port-numbers. ~However, you added binary software-packages as tgz files. We should strictly avoid that since this is unmaintainable in git. I demonstrated how to solve this properly in JavaTest. Please follow that example.~ Sorry, my fault: You removed such files. I was also not aware that we ever added such stuff. Seems I missed this out in earlier reviews or was not involved before merge... Great that you removed this anti-pattern. Awesome job 🥇
> 
> BTW:
> 
> > changed tool of testEditionGetCommandletRunWithoutFlagsAndNotInstalled test from az to gh
> 
> Why did you have to do that? I am not getting the point here...

Because of this comment in the orginal issue:

> It would be better to use a different tool for this kind of test where we test what happens if the tool was already installed.
> The problem is that all the tests of get-version and get-edition are using the same test-project and the same tool az.
> If you read the tests and see that in the test-project az is not installed and therefore we get a specific behaviour if we call different variants of get-version or get-edition when reading the code you start getting of an idea of this scenario reused by many of the test methods.
> However, in the test method linked the existing scenario is "mutated" and the behaviour and setup of az is therefore changing leading to different results. Surely that is what we want to test that the result also depends on the state.
> However, we had a bug fixed by https://github.com/devonfw/IDEasy/pull/955 where this mutation was accidentally causing side-effects for other tests to fail if they were executed afterwards but had been green if executed before this mutation.
> So to make it short: A better approach is to create a new "space" where this scenario is given. To avoid the need to create an extra new test-project just for this, I would recommend to just use another tool instead of az that is not yet included in this test-project and not used by any other tests using this test-project.

PR #1569, opened 2025-11-04
#1596: Add support for comma-separated workspace values in repository configuration

maybeec:
Fixes: #1596 

## Description
This PR adds support for comma-separated workspace values in repository configuration, allowing repositories to be cloned into multiple workspaces during project setup.
Currently, you can workaround this by simply having duplicate repositories configurations with different names pointing to different workspaces. This proposed feature is for convenience only.

## Changes
- Modified `RepositoryConfig` to use `Set<String>` for workspaces instead of single String
- Added `getWorkspaces()` method in `RepositoryProperties` to parse comma-separated values  
- Refactored `RepositoryCommandlet` to clone repositories into multiple workspaces
- Updated workspace status files to include workspace name for per-workspace tracking
- Added 3 comprehensive test cases covering multiple scenarios
- Updated documentation in `repository.adoc`
- Updated CHANGELOG.adoc

## Checklist
- [x] `mvn clean test` passes locally
- [x] Followed coding conventions from `coding-conventions.adoc`
- [x] Added to CHANGELOG.adoc
- [x] Code is properly tested with unit tests
- [x] Documentation updated

## Testing
All 9 tests in `RepositoryCommandletTest` pass successfully, including:
- Multiple workspaces with comma separation
- Handling of whitespace around commas
- Empty workspace defaulting to "main"

## Related Issues
This implements a feature request to support cloning repositories into multiple workspaces via comma-separated configuration values.


hohwille:
If we want to support this feature, is redundant git cloning really the answer?
How about copying or linking the already cloned project to additional workspaces?
Maybe even [git worktrees](https://git-scm.com/docs/git-worktree) would make much more sense here?

The latter (worktrees) is also a great git-feature that many developers could benefit from but are not even aware it exists. Maybe we can also help to "spread the word" this way.

maybeec:
Agree. There are lot of such solutions.
Nonetheless, I am focusing on KISS. This supports me in very big teams and also with AI. The easier to work with, the easier it is to guide people and AI. Thus, a hard copy is currently the most valuable.
As said, the feature is not deciding about good and bad. It's basically convenience. I have set it up with the workaround mentioned.


Btw. I realized by the other tickets, that lists are implemented by bash array syntax. Is there a general support for comma separated lists at all? Otherwise, it might be proper (if of interest) to refactor it to have same bash array syntax and not it in settings, which will be confusing

PR #1572, opened 2025-11-05
#1338: Replace IllegalStateException with assert

### LINKED ISSUE #1338: Check ProcessContext modifyArgumentsOnBackgroundProcess
### Feature idea

The `modifyArgumentsOnBackgroundProcess` method is currently throwing an `IllegalStateException` if the method was called with a non background `processMode`. The only call of this method and the name suggests that the method should only be called if `isBackground` is true. We should check if this exception is still needed and maybe can be replaced with an assert instead.

### Additional context

Mentioned in: https://github.com/devonfw/IDEasy/pull/1330#discussion_r2107087054

maybeec:
### Implemented changes:

* Replaced `IllegalStateException` with `assert` statement in `modifyArgumentsOnBackgroundProcess` method
* Added test `testModifyArgumentsOnBackgroundProcessWithBackgroundMode` to verify the method works correctly with background process modes

### Rationale:

The `modifyArgumentsOnBackgroundProcess` method was throwing an `IllegalStateException` when called with a non-background process mode. However, analysis of the code shows that this method is only called from one location in the `run` method, and that call is already guarded by a check for `processMode.isBackground()`.

Since the method is private and the only caller already ensures the precondition is met, the exception check is redundant. This change replaces the `IllegalStateException` with an `assert` statement, which is more appropriate for internal consistency checks that should never fail in normal operation.

This makes the code more efficient by removing unnecessary exception handling for a condition that should never occur in normal operation, while still maintaining the safety check through assertions when enabled.

---

## Checklist for this PR

  with `internal`

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -6,10 +6,12 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/536[#536]: IDEasy complete tries to match commandlets twice
 * https://github.com/devonfw/IDEasy/issues/1510[#1510]: Make icd more usable
 * https://github.com/devonfw/IDEasy/issues/1169[#1169]: mklink fails if link already exists
 * https://github.com/devonfw/IDEasy/issues/1553[#1553]: IDEasy is slow since 2025.09.001
 * https://github.com/devonfw/IDEasy/issues/1555[#1555]: Handle npm_config_prefix environment variable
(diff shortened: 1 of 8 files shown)


hohwille (approved):
@maybeec thanks for implementing #1338. 👍 
All done as specified. However, IMHO nice-to-have cosmetics.
If we ever run into this situation we have done a severe programming mistake and need to reveal and fix it. If that is then reported as IllegalStateException or assertion will not really matter. The assertion may help that and end-user does not get entirely blocked but most likely then crazy things will happen anyway...

PR #1597, opened 2025-11-18
#39: implement pip commandlet

### LINKED ISSUE #39: Implement ToolCommandlet for pip
Create `ToolCommandlet` for pip
Old commandlet:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/pip

Duplicate of #353 

ATTENTION: Blocked by #38

maybeec:
### Implemented changes:

* Created `Pip` commandlet that delegates to `Uv` commandlet by prepending "pip" to arguments
* Added help texts in English and German explaining that pip delegates to `uv pip` internally
* Created `PipTest` to verify basic functionality
* Updated CHANGELOG.adoc with issue #39

### Technical Details:

The pip commandlet extends `DelegatingToolCommandlet<Uv>` and overrides `runTool()` to prepend "pip" to the arguments before delegating to UV. This means when users run `ide pip install <package>`, it internally executes `uv pip install <package>`.

This implementation follows the hint provided in the issue that pip should be implemented as a delegate to `uv pip`.

---

## Checklist for this PR

- [x] When running `mvn clean test` locally all tests pass and build is successful (test errors are pre-existing)
- [x] You have added the issue implemented by your PR in [CHANGELOG.adoc](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc) unless issue is labeled with `internal`

- [x] The tool can be installed automatically (during setup via settings) or via the commandlet call (delegates to UV)
- [x] The tool is isolated in its IDEasy project, see [Sandbox Principle](https://github.com/devonfw/IDEasy/blob/main/documentation/sandbox.adoc) (handled by UV)
- [x] The new tool is added to the table of tools in [LICENSE.asciidoc](https://github.com/devonfw/IDEasy/blob/main/documentation/LICENSE.adoc) (pip already present)
- [x] The new commandlet is a [command-wrapper](https://github.com/devonfw/IDEasy/blob/main/documentation/cli.adoc#command-wrapper) for pip
- [x] The new commandlet installs potential dependencies automatically (delegates to UV which handles dependencies)
- [x] The variables `«TOOL»_VERSION` and `«TOOL»_EDITION` are honored by your commandlet (handled by delegation to UV)

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
@@ -1,29 +1,27 @@
 package com.devonfw.tools.ide.commandlet;
 
-import static com.devonfw.tools.ide.variable.IdeVariables.GRADLE_BUILD_OPTS;
-import static com.devonfw.tools.ide.variable.IdeVariables.MVN_BUILD_OPTS;
-import static com.devonfw.tools.ide.variable.IdeVariables.NPM_BUILD_OPTS;
-
-import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.List;
+import java.util.Locale;
 
 import com.devonfw.tools.ide.cli.CliException;
 import com.devonfw.tools.ide.context.IdeContext;
 import com.devonfw.tools.ide.property.StringProperty;
-import com.devonfw.tools.ide.tool.ToolCommandlet;
+import com.devonfw.tools.ide.tool.LocalToolCommandlet;
 import com.devonfw.tools.ide.tool.gradle.Gradle;
 import com.devonfw.tools.ide.tool.mvn.Mvn;
 import com.devonfw.tools.ide.tool.npm.Npm;
+import com.devonfw.tools.ide.tool.yarn.Yarn;
 
 /**
  * Build tool {@link Commandlet} for automatically detecting build configuration files and running the respective tool.
  */
 public class BuildCommandlet extends Commandlet {
 
-  /**
-   * The arguments to build with.
-   */
-  public StringProperty arguments;
+  private static final List<Class<? extends LocalToolCommandlet>> BUILD_TOOLS = List.of(Mvn.class, Gradle.class, Yarn.class, Npm.class);
+
+  /** The explicit build options to use (if empty use defaults). */
+  public final StringProperty arguments;
 
   /**
    * The constructor.
@@ -47,45 +45,33 @@ public class BuildCommandlet extends Commandlet {
   public void run() {
 
     Path buildPath = this.context.getCwd();
-    String[] defaultToolOptions = new String[0];
 
     if (buildPath == null) {
       throw new CliException("Missing current working directory!");
     }
 
-    ToolCommandlet commandlet = null;
-    if (Files.exists(buildPath.resolve("pom.xml"))) {
-      commandlet = this.context.getCommandletManager().getCommandlet(Mvn.class);
-      defaultToolOptions = getDefaultToolOptions(MVN_BUILD_OPTS.getName());
-    } else if (Files.exists(buildPath.resolve("build.gradle"))) {
-      commandlet = this.context.getCommandletManager().getCommandlet(Gradle.class);
-      defaultToolOptions = getDefaultToolOptions(GRADLE_BUILD_OPTS.getName());
-    } else if (Files.exists(buildPath.resolve("package.json"))) {
-      if (Files.exists(buildPath.resolve("yarn.lock"))) {
-        // TODO: add yarn here
-      } else {
-        commandlet = this.context.getCommandletManager().getCommandlet(Npm.class);
-
-        defaultToolOptions = getDefaultToolOptions(NPM_BUILD_OPTS.getName());
+    List<String> args = this.arguments.asList();
+    LocalToolCommandlet commandlet = null;
+    for (Class<? extends LocalToolCommandlet> toolClass : BUILD_TOOLS) {
+      LocalToolCommandlet toolCommandlet = this.context.getCommandletManager().getCommandlet(toolClass);
+      Path buildDescriptor = toolCommandlet.findBuildDescriptor(buildPath);
+      if (buildDescriptor != null) {
+        commandlet = toolCommandlet;
+        if (args.isEmpty()) {
+          String variableName = commandlet.getName().toUpperCase(Locale.ROOT) + "_BUILD_OPTS";
+          args = getDefaultToolOptions(variableName);
+        }
       }
-    } else {
+    }
+    if (commandlet == null) {
       throw new CliException("Could not find build descriptor - no pom.xml, build.gradle, or package.json found!");
     }
-
-    if (this.arguments.asArray().length != 0) {
-      defaultToolOptions = this.arguments.asArray();
-    }
-
-    if (commandlet != null) {
-      commandlet.runTool(defaultToolOptions);
-    }
-
+    commandlet.runTool(args);
   }
 
-  private String[] getDefaultToolOptions(String buildOptionName) {
+  private List<String> getDefaultToolOptions(String buildOptionName) {
 
-    String[] defaultToolOptions;
-    defaultToolOptions = this.context.getVariables().get(buildOptionName).split(" ");
-    return defaultToolOptions;
+    String[] defaultToolOptions = this.context.getVariables().get(buildOptionName).split(" ");
+    return List.of(defaultToolOptions);
   }
 }
--- a/cli/src/main/java/com/devonfw/tools/ide/process/ProcessResult.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/process/ProcessResult.java
@@ -68,18 +68,24 @@ public interface ProcessResult {
   }
 
   /**
-   * @param logger the {@link IdeSubLogger logger} to use.
    * @return the first captured standard out. Will be {@code null} if not captured but redirected.
-   * @throws IllegalStateException if more than one output was captured and the {@link IdeSubLogger logger} was null.
    */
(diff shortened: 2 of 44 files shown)


maybeec:
1. I upgraded the github actions setup java commandlet as there were downloading issues of java, which meanwhile seem to be a reducable to a global issue. So the upgrade did not make any improvement but it also does not harm, so I will keep it here as a small by-product improvement
2. I am not 100% sure about the solution. I implemented it as pure delegate as you can directly use `uv pip` although you could especially also install it to virtual environments by `uv pip install pip`. But I think `uv pip` should do the job already as needed.
3. I do not have test capabilities for all environments, seeking support from anyone to do the testing on isolated and different environments. Thanks!

cli/src/test/resources/ide-projects/pip/repository/uv/uv/default/uv
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+echo "uv $*"
```
jan-vcapgemini: These test files are not being used by the tests. So I would either add a run test too (not sure how to simulate that though) or remove them.
For the run test you could simply add a check for the pip argument in the uv binary and echo uv pip install f.e..

jan-vcapgemini (approved):
@maybeec thanks for adding this missing commandlet. Looks good to me, ready for review.

cli/src/main/java/com/devonfw/tools/ide/tool/pip/Pip.java
```diff
@@ -0,0 +1,40 @@
...
+import com.devonfw.tools.ide.process.ProcessMode;
+import com.devonfw.tools.ide.process.ProcessResult;
+import com.devonfw.tools.ide.tool.DelegatingToolCommandlet;
+import com.devonfw.tools.ide.tool.uv.Uv;
+import com.devonfw.tools.ide.version.GenericVersionRange;
+
+/**
+ * {@link DelegatingToolCommandlet} for <a href="https://pip.pypa.io/">pip</a>.
+ */
+public class Pip extends DelegatingToolCommandlet<Uv> {
```
hohwille: Sorry, I should have been more precisely in the story description #39:
When we integrate a tool `x`, our users expect that they can do the following:
```
$ ide install x
$ x --version
```
The option `--version` is just an example here and ideally the `install` commandlet also takes an optional argument for the version of the tool to install where we have auto-completion. To configure the tool `x`, we have variables `X_VERSION` and `X_EDITION`, etc.

A learning from `devonfw-ide` was that we strictly follow this rule and name `x` for our IDEasy commandlet name exactly as the binary the regular end-users invoke. We initially started with `devon maven setup` and had `MAVEN_VERSION`. I will update our documentation to better reflect this philosophy so it will become transparent to everybody in the future.

With this approach the above will not be possible.
Users will still not have a direct `pip` command and if they call scripts, that expect such command, they will not work.

Do not get me wrong: Your idea is great and `uv` is a great tool implemented in rust that can emulate `pip` and is way more efficient. I was already thinking about the same idea to at least have an option/config to let `ide pip ...` work this way.
However, the "normal" behaviour that we should IMHO have here is that `ide install pip [«version»]` will install the "real" `pip` so it is available as a direct command.
Also I was already thinking about just setting an alias `pip=ide pip` or `pip=uv pip`, however, in edge-cases there can be minimal differences and therefore I think it is reasonable that by default we install the "real" `pip`.

In order to support this, we would need to implement `Pip` as a regular `LocalToolCommandlet`.
We then override the `install` method to delegate to `uv pip install pip` (see my comment in #39).
Next we would need to implement a `PipRepository` or `UvRepository` (similar to `MvnRepository` or `NpmRepository`), that would delegate to `uv` commands to determine available versions, etc.
IMHO the situation is very similar to the `npm` world so looking at those commandlets and its repository should be a good example.

cli/src/test/resources/ide-projects/pip/_ide/urls/pip/pip/dependencies.json
```diff
@@ -0,0 +1 @@
+{}
```
hohwille: if we do not have any dependency, we typically skip this file.
However, in this case, `pip` actually does have dependencies:
1. pyhton
2. uv (however, this is already transitive from python, so I would omit it).

We also need a change in `ide-urls` to add this file there with the productive version ranges.
Figuring that out is not always easy since projects hardly document this well.
I am happy to assist if needed.

cli/src/test/resources/ide-projects/pip/_ide/urls/uv/uv/0.5.11/mac_x64.urls
```diff
@@ -0,0 +1 @@
+${testbaseurl}/download/uv/uv/0.5.11/uv-0.5.11.tar.gz
```
hohwille: if all the files are identical, then we should not do a separation by OS.
This is a pattern I keep repeating in reviews so I also need to write some documentation for it:
We have some edge-cases, were we have JUnits that test a tool for different OS variants because the code of that tool commandlet is having code like `if (this.context.getSystemInfo().isWindows() { foo(); } else { bar(); }`.
For these tests we represent the productive structural differences of the folders and files of the different OS and then have the test parameterized by OS.

However, the regular case is that our tests mock the tool in a simple way and if the commandlet itself does not have any OS dependent behaviour then this is perfectly fine and we should not distinguish OS in the test-data nor in the test execution itself.

But if you find such OS specific tests as example, you easily copy the pattern as a developer without thinking what is appropriate for the situation.

To make it short: 
Instead of having 3 `*.urls` files here, simply only have one `urls` file.

hohwille (changes requested):
@maybeec thank you very much for your PR. 👍 
I had a long discussion already about our plans in the dailies with @jan-vcapgemini that we never documented in the story. Sorry for that and hence my review comments requesting some rework.

For my curiosity: I am happy that you upgraded `setup-java` action lib. Is this required for this PR and especially why did you change the JDK distribution to `zulu`?
I do not mind this change, I just want to understand if this is related to the change or if it is a general best practice you learned that I might be missing.
BTW: On our long learning with GitHub we discovered that we often end up with PRs that grow too big and are hard to review (I am not talking about this PR, but PRs changing 100+ files, what we have seen several times). Still I sometimes tend to include refactorings in a bugfix and maybe also include unrelated cleanups that I hit on the way. However, on the long run, we learned that ideally we create small PRs that only focus on a single bug or feature. If we want to also update an action or cleanup some code we hit, it is best, to just create another PR for it.
Thanks for all your input: I clearly see that I need to take some time to write way more documentation about all these things. Ideally in reviews, I only need to paste links to existing documentation if something is not as desired (most likely not in your case but more for junior developers contributing PRs).

cli/src/main/java/com/devonfw/tools/ide/tool/pip/Pip.java
hohwille: I just added this to the documentation and hope it helps for the future:
https://github.com/devonfw/IDEasy/blob/main/documentation/cli.adoc#tool-commandlets

cli/src/test/resources/ide-projects/pip/_ide/urls/pip/pip/dependencies.json
hohwille: regarding 1. my initial intent was exactly to remove the file if it is not needed by the test. A non existing `dependencies.json` is better than an empty one.

However, I additionally wanted to say that we will need the file for production in the `ide-urls` repository.
Why? Because there are specific versions if `pip` and `python` that are incompatible with each other.
The `dependencies.json` is the only way to express this.
However, if you remove the file, we can resolve this thread, merge it and I can add a separate PR addressing the `dependencies.json`.

regarding 2. I was more lax in earlier reviews and we ended up with test-projects that are building up an unrealistic or even inconsistent state. As a result tests were not finding the real problems or fixing real bugs causes the test to be red.
If we add a `dependency.json` as test-data it does not have to reflect the productive state but instead of building an artificial test-screnario based on `foo` and `bar` stuff is definitely worse than just taking the current production state or a relevant fraction of it. It is actually less effort if you already have the production config when creating the test-data.

Just my 50 cents ;) Feel free to remove the file and resolve the conversation.

cli/src/main/java/com/devonfw/tools/ide/tool/pip/Pip.java
```diff
@@ -0,0 +1,141 @@
...
+
+    return true;
+  }
+
+  @Override
+  protected void performToolInstallation(ToolInstallRequest request, Path installationPath) {
+
+    VersionIdentifier resolvedVersion = request.getRequested().getResolvedVersion();
+
+    // Ensure uv is installed first
```
hohwille:
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/tool/pip/Pip.java
```diff
@@ -0,0 +1,141 @@
...
+    ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.THROW_ERR);
+    ProcessResult result = uv.runTool(pc, ProcessMode.DEFAULT, "pip", "install", pipPackageSpec);
+
+    if (!result.isSuccessful()) {
+      throw new IllegalStateException("Failed to install pip version " + resolvedVersion + " via uv");
+    }
+
+    // Create installation path and write version file
+    this.context.getFileAccess().mkdirs(installationPath);
+    this.context.writeVersionFile(resolvedVersion, installationPath);
```
hohwille: is the `installationPath` then `software/pip` that will contain nothing but the `.ide.software.version` file for `pip` then? Not wrong but maybe confusing. For `npm` based stuff we do not write these `.ide.software.version` and ask `npm` for the installed version via CLI. We could also do that here via `pip list`.
However, to keep it constructive and efficient, I would conclude that you have done a great job with this PR and this is ready for merge now. I can add further perfection with separate PRs once I find some time.
There is nothing wrong here.

PR #1638, opened 2025-12-07
#39: refactoring to extract package-manager logic out of node/npm

hohwille:
allowing reuse for pip

### Implemented changes:

* complex refactoring to extract `PackageManagerBasedLocalToolCommandlet`
* renamed `PackageManager` to `NativePackageManager` to avoid confusion
* refactored all CLI args from vararg array to `List<String>` to simplify usage (prepending or appending extra arguments)
* fixed inconsistencies in `ide-projects` (causing `VersionIdentifier` to be instantiated with values like `npm --version` as version)
* added assertion to prevent such nonsense in the future
* debugged infinity loops and fixed cyclic dependencies, adding comments to make that transparent
* reworked `BuildCommandlet` and improved code-style by adding `LocalToolCommandlet.findBuildDescriptor` method (SoC)
* improved `LocalToolCommandlet.findWrapper` to reuse `findBuildDescriptor`

---

## Checklist for this PR

Make sure everything is checked before merging this PR. For further info please also see
our [DoD](https://github.com/devonfw/IDEasy/blob/main/documentation/DoD.adoc).

- [x] When running `mvn clean test` locally all tests pass and build is successful
- [x] PR title is of the form `#«issue-id»: «brief summary»` (e.g. `#921: fixed setup.bat`). If no issue ID exists, title only.
- [x] PR top-level comment summarizes what has been done and contains link to addressed issue(s)
- [x] PR and issue(s) have suitable labels
- [x] Issue is set to `In Progress` and assigned to you *or* there is no issue (might happen for very small PRs)
- [x] You followed all [coding conventions](https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc)
- [x] You have added the issue implemented by your PR in [CHANGELOG.adoc](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc) unless issue is labeled
  with `internal`

PR #1639, opened 2025-12-07
#39: pip commandlet

### LINKED ISSUE #39: Implement ToolCommandlet for pip

hohwille:
### Implemented changes:

* This is a fork of PR #1597 so credits go to @maybeec since I took over his work
* However, I implemented PR #1638 and reused this to share common logic between node/npm and uv/python/pip since both worlds use a package manager that we use instead of our default installation routines.
* I was able to cause so problems bugs in some edge-cases with the current state of #1597 (e.g. install python, install pip, uninstall uv, run pip, bang)
* Also I now give the option to the user to decide if he wants to use the "real pip" or the emulated "uv pip". IMHO end users will typically expect that `ide pip «args»` is doing the same thing as `pip «args»` as we promised in our documentation. Therefore I am still thinking if the default should actually be the "real pip" - currently I made `uv` be the default.

---

## Checklist for this PR

  with `internal`

- [x] The new commandlet is tested on all platforms it is available for or tested on all platforms that are in scope of the linked issue

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
@@ -99,6 +99,8 @@ public abstract class LocalToolCommandlet extends ToolCommandlet {
     VersionIdentifier resolvedVersion = installation.resolvedVersion();
     if (request.isAlreadyInstalled()) {
       return installation;
+    } else {
+      this.context.debug("Installation from {} to {}.", request.getInstalled(), request.getRequested());
     }
     FileAccess fileAccess = this.context.getFileAccess();
     boolean ignoreSoftwareRepo = isIgnoreSoftwareRepo();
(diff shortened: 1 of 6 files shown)


hohwille:
> Is there any option to stay in control as a settings admin over all installations to not end up in different behaviors on the team's machines as some chose so, some others ?

First of all the project settings can define this via
```
PIP_EDITION=uv
```
vs.
```
PIP_EDITION=pip
```

The general [configuration philosophy of IDEasy](https://github.com/devonfw/IDEasy/blob/main/documentation/configuration.adoc) is that users can override things in their `conf/ide.properties`.
By doing so you take control but also responsibility ("With great power comes great responsibility").
Typically the average IDEasy user never does things like this unless he was told so by others.
So to what I observed the users only override stuff there when there is a real purpose to do so.
The bigger problem is IMHO if users do override something, they easily forget about this and months or years later this may become a problem.
We support the command `ide -d env` to get a quick overview of your configurations and where they originate from.

IMHO it is not a good idea to give 100% control to an "IDE admin".
This is from a command and conquer thinking.
User should have the last word and "manual override" like the pilot of a plane can overrule the auto-pilot.

Issue #1658, opened 2025-12-15
Make Jackson waterproof and avoid it failing in native image


hohwille:
### Task summary

Issue #1644 was again one of many issues where we implemented some feature using Jackson to deserialize JSON to Java and all was green in Java but failed in GraalVM native image.
My initial plan was to find a way to disable (deep) reflection support in Jackson somehow entirely and force ourselves to always write JsonDeserializers already to make it work in Java.
However, @maybeec had a great idea that seems even better:
In PR #1655 he wrote a [JUnit test](https://github.com/maybeec/IDEasy/blob/bec640685b3b475ebdef4c6f1e54bb827602a3f8/cli/src/test/java/com/devonfw/tools/ide/nativeimage/GraalVmReflectionConfigTest.java) that tries to find all classes that we deserialize from JSON and checks that they are all listed in `reflect-config.json`. Currently this test only finds classes that implement `JsonObject` or `JsonVersionItem`.
The problem is that there are many other classes affected not found by this condition:
* `Cve`
* `ToolSecurity`
* `ToolDependency`
* `VersionRange`
* `VersionIdentifier`
* `CustomToolsJson`
* `CustomToolJson`

Side note: `JsonVersionItem` is not used by IDEasy itself but only by our UrlUpdaters that always run in JVM.
But since these classes are not inside the `cli` module, the test would not find them anyhow so it does not harm.
Since `JsonObject` is only a marker interface, we should derive `JsonVersionItem` from `JsonObject`.

My ideas on this:
* We already have `JsonMapping` to somehow centralize and wrap our Jackson configuration.
* We could even wrap Jackson entirely creating our own API to deserialize JSON that only accepts `Class<? extends JsonObject>`.
* BTW: Do we need to create separate instances of ObjectMapper or can we share a singleton instance? IMHO it is not stateful and therefore we could simply do that. Then we could even simplify our own code with the "wrapper".
* Our test can then find at least all root objects that *must*  implement `JsonObject`.
* We could find all child objects via refelection by decomposing the root objects.

If that approach turns out way too complex, we could still try to find out if we can disable the generic reflection based deserializer in Jackson what should IMHO be possible or we raise a feature request for that.

PR #1675, opened 2026-01-14
#1298: support ide-extra-tools.json #1658: prevent Jackson reflection

### LINKED ISSUE #1298: Support for extra tool
As a IDEasy, I want the ability to have second version of the same tool in my project so that I can run two different parts of my project with different version of that tool.
An example is Java where some projects may need a different version for the frontend than for the backend part because a migration is ongoing or one part is currently blocked from also upgrading to the newer java version.

As a workaround custom tools can be used but that requires to maintain a custom software repository and upload/provide all the tools there. For some vanilla tool already supported by IDEasy this overhead does not seem to make sense.

## Solution
As a solution variables like `EXTRA_«TOOL»_VERSION` and potentially `EXTRA_«TOOL»_EDITION` should be supported.
If `«tool»` gets installed (`ide install «tool»`) then we should check for `EXTRA_«TOOL»_VERSION` and if that is defined, do an extra handling:
* we create an extra installation to `$IDE_HOME/software/extra/«tool»` configured by `EXTRA_«TOOL»_VERSION`. The edition is configured by `EXTRA_«TOOL»_EDITION` but if undefined falls back to the regular edition mechanism (``«TOOL»_EDITION` with default of `«tool»` and additional existing special excuses). This installation must not be added to `PATH` or `«TOOL»_HOME` (maybe `EXTRA_«TOOL»_HOME` would be a good idea).
* if the resolved extra version and edition are identical to the regular one, we log a WARNING since then the extra tool will be identical to the regular tool and the log message should indicate that something is questionable.
* depending on the tool this feature may be disabled. E.g. for GraalVM this feature will not make sense since the default installation already resides in `extra`. Further, for IDEs I also do not see the use-case to install 2 different versions in parallel. But especially for SDKs like `java` or `dotnet` this can make sense and this is the typical use-case this story is about.

hohwille:
### Implemented changes:

* added support for ide-extra-tools.json
* added according data objects (Model for JSON)
* added serializers and deserializers
* disable bean relfection support in Jackson
* added and registered all missing serializers and deserializers
* consistently aligned all JUnits (default visiblity, JavaDoc, clean-code)

---

## Checklist for this PR

  with `internal`

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/custom/CustomToolJsonDeserializer.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/custom/CustomToolJsonDeserializer.java
@@ -7,10 +7,9 @@ import com.devonfw.tools.ide.json.JsonBuilder;
 import com.devonfw.tools.ide.json.JsonObjectDeserializer;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.DeserializationContext;
-import com.fasterxml.jackson.databind.JsonDeserializer;
 
 /**
- * {@link JsonDeserializer} for {@link CustomTool}.
+ * {@link JsonObjectDeserializer} for {@link CustomTool}.
  */
 public class CustomToolJsonDeserializer extends JsonObjectDeserializer<CustomTool> {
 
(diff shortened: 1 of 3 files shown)


hohwille:
I am always surprised how I can create so hard git merge conflicts with myself when I branched from the same feature-branch that was already merged... Seems this is caused by squashing so next time I really have to first complete the one PR and then pull main after its squashed merged and then create my new feature branch from that...
However, I would still consider it a bug in git if during the merge files that have been deleted/renamed suddenly show up again and have merge conflicts...

PR #1794, opened 2026-04-01
#451: mac gatekeeper quarantine removal

### LINKED ISSUE #451: Make IDEasy usable on MacOS with active Gatekeeper
On modern MacOS devices gatekeeper is most likely activated to increase security.
As a result apps downloaded from the Internet, cannot be opened without strange quirks.
However, the entire purpose of IDEasy is to download apps from the internet and allow you to start them.
Therefore, with the MacOS gatekeeper active currently IDEasy is almost not usable since every app will open a popup like this:
macos-app-will-damage-your-computer-move-to-trash
Normal users will not know how to proceed since `Cancel` will prevent that the app is opened and `Move to trash` is even worse and will delete the  app and break the installation that was just created by IDEasy.
There are some security mechanisms implemented in IDEasy that do checksum verification.

Surely the vendors like the apache software foundation could potentially publish a new release of their tool with evil code that will erase your disc or encrypt all your files and ask for ransom. However, open-source software is based on trust and transparency. If a prominent OSS tool would ever do that it will kill its entire reputation for the tool and the foundation behind it - forever.
Also since we test releases upfront, we hopefully notice and would ban the release from our urls repository so that it does not reach our users.

I just researched the topic and studied [this article](https://www.idownloadblog.com/2017/04/20/fix-application-from-internet-gatekeeper/).

So after reading all the options, I come to conclusion that we should try to run something like this on the macos app after the installation:
```
xattr -d com.apple.quarantine /path-to-app
```
I am unsure if that operation will work for the entire app or has to be called recursively for every file in the app.

shodiBoy1:
### This PR fixes #451 

### Implemented changes:

 On modern macOS (15.1+, Apple Silicon), just removing com.apple.quarantine didn't work - unsigned apps still showed the "is damaged" popup. Tested this on a real M1 Pro with IntelliJ CE and Android Studio.       
                                                                                                                                                                                                                      
  The fix does two things after extraction:                                                                                                                                                                           
  - xattr -cr to clear all extended attributes (quarantine, resource forks, etc.)                                                                                                                                   
  - codesign --force --deep --sign - to ad-hoc sign the .app bundle - but only if it's not already properly signed (so we don't break notarized apps like Eclipse)                                                    
                                                                                                                                                                                                                    
  Other changes:                                                                                                                                                                                                      
- Version file (.ide.software.version) stays at rootDir instead of being copied inside the .app, because codesigning seals the bundle and any writes after that break it                                                         
- linkDir now correctly points inside the .app bundle (e.g. Contents/MacOS) where the binary lives. The symlink target changed, so getInstalledVersion() was updated to resolve the symlink back to rootDir via getValidInstalledSoftwareRepoPath to find the version file                                                                                                                                                                       
- Added ProcessErrorHandling to codesign calls - NONE for the verification step (failure is expected for unsigned apps) and LOG_WARNING for the signing step                                                            
- Removed findBinDir() from MacOsHelper and the getToolBinPath() macOS workaround - both were redundant after fixing the linkDir assignment                                                                                      
- Updated tests to use getInstalledVersion() instead of checking the version file through the symlink   
---

## Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -706,6 +706,7 @@ public void extract(Path archiveFile, Path targetDir, Consumer<Path> postExtract
     postExtractHook(postExtractHook, properInstallDir);
     move(properInstallDir, targetDir);
     delete(tmpDir);
+    removeQuarantineAttribute(targetDir);
```
hohwille: This should only happen for an application. The `extract` function is generic and can also be used to extract a ZIP file with pure data or config that is not an application. Please note that the executed `xattr` command will typically trigger a popup with password request so this should only appear for applications (ideally only if they are not signed).
Please check PR #453 where this was already done in the right place.

hohwille (changes requested):
@shodiBoy1 thanks for your PR. Great that you are working on this nasty problem that is currently blocking Mac users and renders the UX of IDEasy on Mac void. 👍 
Currently, I do not see the difference in the `xattr` compared to PR #453. Can you explain why this should fix the problem now? Have you tested this approach on MacOS and it worked for unsigned apps like e.g. IntelliJ?

FYI: Did you also see this comment? https://github.com/devonfw/IDEasy/issues/451#issuecomment-2211455935
When we remove the `quarantine` attribute, MacOS protected the app directory and we cannot make any modifications to it after that. This IMHO implies that we cannot keep the current solution with the `.ide.software.version` file that we simply copy to the `linkDir` as a workaround. Maybe it could work if we do that before we remove the `quarantine` attribute?
Further, in PR #453 @jan-vcapgemini made a review comment that this `xattr` execution should be moved to `MacOsHelper` what makes sense to me and should be followed.

shodiBoy1:
previous fix was wrong  just doing xattr -r -d com.apple.quarantine isn't enough on macOS 15.1+ with Apple Silicon. Unsigned apps still get blocked even without the quarantine attribute.

new fix xattr -cr to clear all extended attributes, then codesign --force --deep --sign - to ad-hoc sign the .app bundle. The codesign part is what actually makes it work. I skip signing if codesign -v passes so we don't break already notarized apps like Eclipse.

Moved the logic into MacOsHelper.removeQuarantineAttribute() and it's called from LocalToolCommandlet.installTool() now, not from extract.

.ide.software.version stays at rootDir can't write inside the .app after codesigning or you get "Operation not permitted". Version file is written before signing so no issue there.

Tested with IntelliJ CE and Android Studio on M1 Pro, works fine.

cli/src/main/java/com/devonfw/tools/ide/os/MacOsHelper.java
```diff
@@ -48,10 +55,54 @@ public MacOsHelper(IdeContext context) {
...
+      return;
+    }
+    if (this.context == null) {
+      LOG.debug("Cannot fix Gatekeeper for {} - no context available", path);
+      return;
+    }
+    // clear all extended attributes (quarantine, resource forks, etc.)
+    LOG.debug("Clearing extended attributes from {}", path);
+    try {
+      this.context.newProcess().executable("xattr").addArgs("-cr", path).run(ProcessMode.DEFAULT_SILENT);
```
hohwille, line 83: isn't this causing a popup where the user has to confirm admin operation on MacOS (maybe even with password)?
If not, just comment and resolve this thread.
If yes, you should log before on interaction that we need to remove quarantine what is a restricted operation (see e.g. #415).

cli/src/main/java/com/devonfw/tools/ide/os/MacOsHelper.java
```diff
@@ -48,10 +55,54 @@ public MacOsHelper(IdeContext context) {
...
+    } catch (Exception e) {
+      LOG.trace("Could not clear extended attributes from {}: {}", path, e.getMessage());
```
hohwille: IMHO we should log errors on warn or at least info.
```suggestion
      LOG.warn("Could not clear extended attributes from {}: {}", path, e.getMessage(), e);
```

cli/src/main/java/com/devonfw/tools/ide/os/MacOsHelper.java
```diff
@@ -48,10 +55,54 @@ public MacOsHelper(IdeContext context) {
...
+        if (!verifyResult.isSuccessful()) {
+          LOG.debug("Ad-hoc codesigning {}", appDir);
+          ProcessResult signResult = this.context.newProcess().executable("codesign")
+              .addArgs("--force", "--deep", "--sign", "-", appDir).run(ProcessMode.DEFAULT_SILENT);
+          if (!signResult.isSuccessful()) {
+            LOG.warn("Could not codesign {} - app may be blocked by Gatekeeper", appDir);
+          }
+        }
+      } catch (Exception e) {
+        LOG.trace("Codesign not available for {}: {}", appDir, e.getMessage());
```
hohwille: same here? Raise to info or warn, always include the exception itself so stracktrace can be seen.
BTW: If I am not mistaken, we only print stacktraces if log-level is error or otherwise if debug or trace is activated.

cli/src/main/java/com/devonfw/tools/ide/os/MacOsHelper.java
hohwille: Do we always have to do this? There are apps for MacOS where all is working without this.
Should we have a black or whitelist?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -520,13 +518,6 @@ protected ToolInstallation createToolInstallation(Path rootDir, VersionIdentifie
   protected ToolInstallation createToolInstallation(Path rootDir, Path linkDir, Path binDir, VersionIdentifier version, boolean newInstallation,
       EnvironmentContext environmentContext, boolean additionalInstallation) {
 
-    if (linkDir != rootDir) {
-      assert (!linkDir.equals(rootDir));
-      Path toolVersionFile = rootDir.resolve(IdeContext.FILE_SOFTWARE_VERSION);
-      if (Files.exists(toolVersionFile)) {
-        this.context.getFileAccess().copy(toolVersionFile, linkDir, FileCopyMode.COPY_FILE_OVERRIDE);
-      }
-    }
```
hohwille, line 529: This might be problematic on macos, but if you remove it, you also need to fix the code to determine the tool version that you broke by just removing this code-block.

hohwille (approved):
@shodiBoy1 thanks for your update. Now this looks way more promising. 👍 
However, there are still things to clarify before we can merge.

CHANGELOG.adoc
```diff
@@ -18,6 +18,7 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/1702[#1702]: UI Buttons do not scale properly with window resize
 * https://github.com/devonfw/IDEasy/issues/1751[#1751]: Add go commandlet (go-lang support)
 * https://github.com/devonfw/IDEasy/issues/1732[#1732]: Add stash support for git-pull
+* https://github.com/devonfw/IDEasy/issues/451[#451]: Automatically remove macOS quarantine attribute after tool extraction
```
hohwille: I resolved the merge conflict.
But this line still needs to be moved up.
Maybe even to 2026.05.001 that is not yet added to the CHANGELOG.

cli/src/main/java/com/devonfw/tools/ide/os/MacOsHelper.java
shodiBoy1: xattr -cr doesnt cause any popup or password request  at least while testing on my mac I never got any popup. It just works on files the current user owns, no admin needed.                                                 
                                                                                                                                                                                                                                 
about the whitelist/blacklist right now xattr -cr runs for every tool but its fast and harmless for tools. codesigning part only runs for unsigned .app bundles (we check with codesign -v first, so signed apps like Eclipse are left alone). I can add a check to skip xattr -cr entirely when theres no .app bundle if you want?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
shodiBoy1: version detection still works just a different approach. before, linkDir pointed inside the .app bundle so the symlink landed there, and the version file had to be copied from rootDir into linkDir. 
in my fix linkDir == rootDir, so the symlink points to the top level directory where the version file is already written no copy needed. 
also copying into the .app bundle would break after codesigning since macOS seals it. binaries are still found correctly through findBinDir() in getToolBinPath() which navigates into the .app bundle to find executables.

cli/src/main/java/com/devonfw/tools/ide/os/MacOsHelper.java
shodiBoy1: yep, changed both to LOG.warn with the exception included so it shows the exception info at warn level and full stacktrace when running in debug mode.

CHANGELOG.adoc
shodiBoy1: moved changelog entry up to 2026.04.002

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
shodiBoy1: @hohwille you were right, I didn't fully understand how linkDir and rootDir work together. i broke the linkDir assignment by collapsing two lines into one. I restored it so linkDir points inside the  
  .app bundle again. instead of copying the version file into the bundle (which would break codesigning), I changed getInstalledVersion to follow the symlink back to rootDir where the version file already lives.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -416,13 +422,17 @@ private String getEdition(Path toolRepoFolder, Path toolInstallFolder) {
   }
 
   private Path getInstalledSoftwareRepoPath(Path toolPath) {
+    return getInstalledSoftwareRepoPath(toolPath, false);
```
hohwille, line 425: Shouldn't the default be `true` to keep existing behaviour or did you check all existing usages and change to explicit call with `true` as argument where the logging still makes sense?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
hohwille: Great. Via `getInstalledSoftwareRepoPath` you are now resolving the top-level directory where the version file is located also in case of a mac app.
On windows and linux this should not cause any harm. It surely will cost some extra millis to resolve the link but that should not matter.

cli/src/test/java/com/devonfw/tools/ide/tool/aws/AwsTest.java
```diff
@@ -103,7 +103,8 @@ void testAwsPrintHelp() {
 
   private void checkInstallation(IdeTestContext context) {
 
-    assertThat(context.getSoftwarePath().resolve("aws/.ide.software.version")).exists().hasContent("2.24.15");
+    Aws commandlet = new Aws(context);
```
hohwille:
```suggestion
    Aws commandlet = context.getCommandletManager().getCommandlet(Aws.class);
```

hohwille (approved):
@shodiBoy1 thank you so much for your improvements. You fixed a severe blocker bug causing big progress for IDEasy users on MacOS! 🥇 
I only added cosmetic review comments but this PR is now about ready for merge.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
hohwille: OK, lets not block this great feature any further because of an edge-case log statement being not made in some case. If relevant we can also fix that after the merge. The MacOS progress is so much more important and pending for ~1 year now.

PR #1803, opened 2026-04-07
#1760: accept empty input for one option

### LINKED ISSUE #1760: Accept empty input for single option
### Feature idea

Currently, the user is queried on cve issues to select an option of a version to be installed. But if there is just one option to go, we should better just log a warning an proceed instead stopping and asking the user strange questions:

Image

### Additional context

_No response_

ducminh02:
### Implemented changes:

* Updated displayOptionsAndGetAnswer in AbstractIdeContext.java to automatically select a single available option only if it is not a vulnerable tool version, ensuring users must still explicitly accept version-related risks.

* Adjusted the unit test in IdeContextTest.java to use a real Cve (vulnerability) object, verifying that the system correctly stops to prompt the user when a single but "unsafe" version is offered.
---

## Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -1018,6 +1018,17 @@ public <O> O question(O[] options, String question, Object... args) {
   }
 
   private <O> O displayOptionsAndGetAnswer(O[] options) {
+    if (options.length == 1) {
```
jakozian: I reproduced a scenario where one safe option is prompted and now the process proceeds without me giving an answer. Wasn't the issue to accept empty input for a single (safe) choice? I could be wrong, but if this is the case maybe there should not be a question but instead an info log and then proceed.
Imo it would be fine to not await an answer if there is only one safe option, but then the user has to be informed about that. Now it is weird: A question gets asked but the process proceeds immediately.

image

My scenario was simply achieved by commenting out cve options in the source code to get only one option.

ducminh02: okay I understand your point. However I think that the issue were that when the user tries to download a vulnerable version of a tool, they should be prompted asking if they are sure if they want to install the tool, even if there are only 1 version available. If the version of the tool is not vulnerable then they should not be prompted.

Image

you can try to leave the mvn 3.0.0 here, comment out the nearest and the latest options like you did before and they will still be prompted even if that is the only option

The idea of the issue as I understand is for all of the prompting and asking for user input, if there is only 1 option available, it should be automatically chosen without the user having to type in 1

jakozian: I understand what you mean, but
1. As the title of the issue/pr suggests an empty input (by pressing enter) does at the moment not get accepted:
image

2. I still argue that asking the user a question (on interaction log level in blue) and than just proceeding without any hint is a bit misleading from a ux perspective.

ducminh02: ahh okay I get what you mean, let me change that. Thank you for the review and the hint

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -1033,6 +1033,9 @@ private <O> O displayOptionsAndGetAnswer(O[] options) {
       }
       IdeLogLevel.INTERACTION.log(LOG, "Option {}: {}", numericKey, title);
     }
+    if (options.length == 1) {
+      mapping.put("", options[0]);
+    }
     O option = null;
```
hohwille, line 1039: Nice and elegant solution with also nice test coverage.
Can you please also add the issue to our CHANGELOG?

hohwille (approved):
@ducminh02 excellent KISS fix and great test improvement to cover the functionality. 👍 
Please add to CHANGELOG, then ready for merge.

PR #1806, opened 2026-04-09
#1337: add tests to simulate user input

### LINKED ISSUE #1337: Add tests to simulate user input
### Feature idea

As an IDEasy developer I need to check if my `ProcessContext` capture types work properly. It would be nice if we could somehow simulate a user input within a test.

### Additional context

Mentioned in: https://github.com/devonfw/IDEasy/pull/1330#pullrequestreview-2868136141

laert-ll:
### Implemented changes:

* Added integration test simulating user input during IDE installation to verify that the license agreement is created when accepted by user input as well as otherwise when declined.

---

## Checklist for this PR


jakozian (approved):
Nice work, i tried out the test and it succeeded 👍 
I think this issues does not concern the user so it should be removed out of the CHANGELOG.

hohwille (changes requested):
@laert-ll sorry but it seems there was a misunderstanding to what I suggested in the daily.
I checked the issue and it seems I only explained my expectation in daily but should have written it also in the issue to make it clear.

The story #1337 was about the problem that the ticket creator did not know how to provide user input in integration tests. Therefore we should not create a integration test called `user-input-during-installation.sh` but I suggested that you test the bug we had with our license agreement not being printed.
Therefore you should call the test e.g. `install-requires-license-agreement.sh`.
The test itself should do the following:

1. `rm -f ~/.ide/.license.agreement`
2. run the installation so it will ask for the license agreement - for that you need to provide `-f` or `--force` mode as otherwise the user is asked two questions. But here provide `no` as answer
3. assert that the license agreement was printed 
4. assert that the installation was aborted and the license agreement file (from 1.) was not recreated
5. rerun step 2 with `yes` as anwser
6. same as 3.
7. assert that the installation was performed and the license agreement file is now present.

BTW: You can also include a rerun of steps 2.-4. with `2` instead of `no` as input.

For 1. we also have to be aware that we are creating more undesired side-effects in integration tests. Therefore we should also create a PR that first "saves" the users state (backup `~/.ide/.license.agreement`, `~/.bashrc`, and `~/.zshrc` if present, on Windows backup env variables `IDE_HOME` and `IDE_ROOT` from registry...
After test - both if failed and if successful - "restore" state from backup).

laert-ll:
Apologies for the misunderstanding and I really appreciate the thorough reply.

I applied the changes you described, let me know if there's any further changes needed you see fit.

As for avoiding the side effects, as discussed yesterday in the daily, I will wait for a separate story to implement those changes, and I'd be glad to work on that as well. Let me know.

cli/src/test/integration-tests/install-requires-license-agreement.sh
```diff
@@ -0,0 +1,54 @@
...
+# by user input as well as otherwise when declined
+
+LICENSE_FILE="$HOME/.ide/.license.agreement"
+
+# Usage: run_test <input> <expected_behavior>
+run_test() {
+    local input=$1
+    local behavior=$2
+
+    echo "Running case: input '$input' should $behavior"
```
hohwille: Great, this makes debugging more easy. Lets even make it more readable if someone finds it in the logs and does not know the details of this specific integration-test.

```suggestion
    echo "Running case: input '$input' should $behavior the license"
```

hohwille (approved):
@laert-ll thanks for your rework.
Now this reads like a perfect integration-test:
It has a clear goal to test an end-user functionality, properly asserts the expected results and gives perfect output on success or error to see what happened.

cli/src/test/integration-tests/install-requires-license-agreement.sh
```diff
@@ -0,0 +1,54 @@
...
+    fi
+}
+
+echo "Running license agreement required during installation integration test"
+
+run_test "yes" "accept"
+run_test "1"   "accept"
+run_test "no"  "decline"
+run_test "2"   "decline"
+
```
hohwille: Oh, I missed something in the review.
Without #1821 being implemented, this test will and the end leave the side-effect that the license file is deleted breaking following integration tests.
Before we can merge this we need to invert the order or it will break the integration-tests:
```suggestion
run_test "no"  "decline"
run_test "2"   "decline"
run_test "yes" "accept"
run_test "1"   "accept"
```

laert-ll:
@hohwille thank you for your feedback!
I implemented the suggested changes, the branch should be ready for merge.

PR #1807, opened 2026-04-09
#1802 GUI State management implementation

### LINKED ISSUE #1802: Enable state management for the IdeContext in the GUI.
### Task summary

We need to adapt a state management for managing the different project contexts in the GUI.

### Additional context

-

laim2003:
### Implemented changes:

* Implemented basic State Management for the IdeContext in the GUI
* State management should be thread safe: [Thread safe singleton implementation](https://www.baeldung.com/java-bill-pugh-singleton-implementation)
* Added custom adaption of JavaFXs alert class for IDEasy

Main points:

- [x] Basic state management
- [x] Start-context <-> context seperation 
- [x] Ensure thread safety
- [x] Callback pattern for any dependent compontents

---

## Checklist for this PR

  with `internal`


cli/pom.xml
```diff
@@ -138,8 +138,6 @@
               </archive>
             </configuration>
           </execution>
-
-          <!-- jar for test classes -->
```
hohwille, line 142: why removing this comment that could be helpful to explain the execution?

gui/src/main/java/com/devonfw/ide/gui/context/ProjectManager.java
```diff
@@ -0,0 +1,120 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.stream.Stream;
+
+import org.jline.utils.Log;
```
hohwille: This is IMHO the wrong logger

gui/src/main/java/com/devonfw/ide/gui/context/ProjectManager.java
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Service class that allows to access the list of projects
+ */
+public class ProjectManager {
+
+  private final Path ideRootDirectory;
+
+  private final ArrayList<String> projectNames = new ArrayList<>();
+  private final HashMap<String, ArrayList<String>> workspaces = new HashMap<>();
```
hohwille: https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc#prefer-general-api
```suggestion
  private final List<String> projectNames = new ArrayList<>();
  private final Map<String, List<String>> workspaces = new HashMap<>();
```
You can even create custom Java classes to add more semantics since structures like `Map<String, List<String>` are hard to understand. If it was `Map<String, Workspaces>` it would be easier to understand esp. when `Workspaces` has reasonable JavaDoc.
However, I would question why we should store and cache the workspaces for all projects...
Your current code will never refresh the caches (surely you could be calling `refreshProjects()`) and therefore never shows updated information when a new workspace is created after the GUI was launched. When a project is changed, we should IMHO always refresh the newly selected project's workspaces and afford the 20ms to read them from the disc (again).

gui/src/test/resources/ide-projects/testProject/project-0/home/.ide/ide.properties
```diff
@@ -0,0 +1,13 @@
...
+DOCKER_EDITION=docker
+FOO=foo-${BAR}
+TEST_ARGS1=${TEST_ARGS1} user1
+TEST_ARGS2=${TEST_ARGS2} user2
+TEST_ARGS3=${TEST_ARGS3} user3
+TEST_ARGS7=user7
+TEST_ARGS10=user10
+TEST_ARGSb=userb
+TEST_ARGSc=${TEST_ARGS1} userc
+TEST_ARGSd=${TEST_ARGS1} userd
```
hohwille: the `basic` project was the first `ide-project` that we created.
It was misused later by many testers who were too lazy to create a new `ide-project` for their specific purpose.
Therefore all crazy things ended up in this `basic` project.
Please not not copy & paste all this mess into the `gui` module.
Nobody will ever understand why we have this here in `gui` and the true answer will be: "for nothing"...

gui/src/test/resources/ide-projects/testProject/project-3/home/.ide/ide.properties
```diff
@@ -0,0 +1,13 @@
...
+DOCKER_EDITION=docker
+FOO=foo-${BAR}
+TEST_ARGS1=${TEST_ARGS1} user1
+TEST_ARGS2=${TEST_ARGS2} user2
+TEST_ARGS3=${TEST_ARGS3} user3
+TEST_ARGS7=user7
+TEST_ARGS10=user10
+TEST_ARGSb=userb
+TEST_ARGSc=${TEST_ARGS1} userc
+TEST_ARGSd=${TEST_ARGS1} userd
```
hohwille: :)

gui/src/test/resources/ide-projects/testProject/project-3/settings/ide.properties
```diff
@@ -0,0 +1,9 @@
+#********************************************************************************
+# This file contains project specific environment variables
+#********************************************************************************
+JAVA_VERSION=17*
+MVN_VERSION=3.9.0
+ECLIPSE_VERSION=2023-03
+INTELLIJ_EDITION=ultimate
+IDE_TOOLS=mvn,eclipse
+BAR=bar-${SOME}
```
hohwille: :)

gui/src/test/resources/ide-projects/testProject/project-4/home/.ide/ide.properties
```diff
@@ -0,0 +1,13 @@
...
+DOCKER_EDITION=docker
+FOO=foo-${BAR}
+TEST_ARGS1=${TEST_ARGS1} user1
+TEST_ARGS2=${TEST_ARGS2} user2
+TEST_ARGS3=${TEST_ARGS3} user3
+TEST_ARGS7=user7
+TEST_ARGS10=user10
+TEST_ARGSb=userb
+TEST_ARGSc=${TEST_ARGS1} userc
+TEST_ARGSd=${TEST_ARGS1} userd
```
hohwille: :)

gui/src/test/resources/ide-projects/testProject/project-5/home/.ide/ide.properties
```diff
@@ -0,0 +1,13 @@
...
+DOCKER_EDITION=docker
+FOO=foo-${BAR}
+TEST_ARGS1=${TEST_ARGS1} user1
+TEST_ARGS2=${TEST_ARGS2} user2
+TEST_ARGS3=${TEST_ARGS3} user3
+TEST_ARGS7=user7
+TEST_ARGS10=user10
+TEST_ARGSb=userb
+TEST_ARGSc=${TEST_ARGS1} userc
+TEST_ARGSd=${TEST_ARGS1} userd
```
hohwille: :)

hohwille (commented):
@laim2003 thanks for your PR.
I like the general design and you added many valuable fixes and improvements for the GUI.
I still have some doubts about the singleton and static member access approach.
Typically this will sooner or later cause trouble.
Maybe we can avoid this where possible.

Please avoid this extensive copy&paste of pointless and confusing properties, plugin settings that are never used, etc.
The golden rule for `ide-projects` for test is to keep them as minimalistic as possible.
Every line from such file and every entire file that can be removed without the test failing should be removed.

laim2003:
> @laim2003 thanks for your PR. I like the general design and you added many valuable fixes and improvements for the GUI. I still have some doubts about the singleton and static member access approach. Typically this will sooner or later cause trouble. Maybe we can avoid this where possible.
> 
> Please avoid this extensive copy&paste of pointless and confusing properties, plugin settings that are never used, etc. The golden rule for `ide-projects` for test is to keep them as minimalistic as possible. Every line from such file and every entire file that can be removed without the test failing should be removed.

regarding the singleton design, i created a new ticket to document this #1975

gui/src/main/java/com/devonfw/ide/gui/context/ProjectManager.java
laim2003: Thank you IntelliJ auto complete!

gui/src/main/java/com/devonfw/ide/gui/context/ProjectManager.java
laim2003: Good idea, will change this so that we read the projects/workspaces everytime. Originally, I had the idea of observing changes to the project/workspace folder structure using a [WatchService](https://docs.oracle.com/javase/tutorial/essential/io/notification.html), but I felt that that would be overengineered for this quite simple use case...

cli/pom.xml
laim2003: this should show up as outdated. I have undone this change.

gui/src/main/java/com/devonfw/ide/gui/context/ProjectManager.java
```diff
@@ -0,0 +1,79 @@
...
+   * @return the list of project (names) in the project directory
+   */
+  public List<String> getProjectNames() {
+
+    try (Stream<Path> subPaths = Files.list(ideRootDirectory)) {
+      return subPaths
+          .filter(Files::isDirectory)
+          .map(Path::getFileName)
+          .map(Path::toString)
+          .filter(name -> !name.startsWith("_") && Files.exists(ideRootDirectory.resolve(name).resolve("workspaces")))
```
hohwille: We should make this method reusable here:
https://github.com/devonfw/IDEasy/blob/6080ee79bcab0b1292121e35b99b3157558bb23d/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L428

We could even make it public if required.
BTW: The only forbidden or reserved project name is `_ide` and for that we have the constant `IdeContext.FOLDER_UNDERSCORE_IDE`.
I do not think users should do `ide create _project -` but I guess it is technically possible.

gui/src/test/resources/ide-projects/testProject/project-0/workspaces/foo-test/placeholder
```diff
@@ -0,0 +1 @@
+this is the foo-bar workspace
```
hohwille: Everywhere else we just use empty `.gitkeep` files to add a folder to git. Shouldn't we follow this pattern here as well?

gui/src/test/resources/ide-projects/testProject/project-1/workspaces/foo-test/placeholder
```diff
@@ -0,0 +1 @@
+this is the foo-bar workspace
```
hohwille: does it make sense to create `project-0`,..., `project-5` that are all identical and have the same workspaces?
Actually if all this should make sense it would be more reasonable to have some differences between the workspaces so your test can really check if the workspaces found for some project really fit to that project and are not from a different one...

gui/src/test/java/com/devonfw/ide/gui/context/ProjectManagerTest.java
```diff
@@ -0,0 +1,132 @@
...
+    ideRoot = context.getIdeRoot();
+  }
+
+  @Test
+  void testProjectManagerFull() throws NotDirectoryException {
+
+    projectManager = new ProjectManager(ideRoot);
+
+    assertThat(projectManager).isNotNull();
+    assertThat(projectManager.getProjectNames()).containsExactlyInAnyOrder("project-0", "project-1", "project-2", "project-3", "project-4", "project-5");
```
hohwille: In `cli` we have added our test-infrastructure such that we can continuously add new code with new tests and potentially new test projects.
This will break as soon as we add a new test-project.
IMHO this does not really work out well for maintenance.
Maybe only assert that the project names contain these projects or filter the project names by a pattern `project-[0-9]` so other developers can later add `foo` and `bar` projects without breaking this test.

gui/src/test/java/com/devonfw/ide/gui/context/ProjectManagerTest.java
```diff
@@ -0,0 +1,132 @@
...
+    } finally {
+      Files.deleteIfExists(ideRoot.resolve("testFile"));
+    }
+  }
+
+  @Test
+  void testRefreshProjects() throws IOException {
+
+    projectManager = new ProjectManager(ideRoot);
+    assertThat(projectManager.getProjectNames()).containsExactlyInAnyOrder("project-0", "project-1", "project-2", "project-3", "project-4", "project-5");
```
hohwille: Also in JUnits we can consider reuse instead of copy&paste.
Consider adding `List.of("project-0", "project-1", "project-2", "project-3", "project-4", "project-5")` as a constant.
Then you can create a new list adding the constant + `project-6`.

gui/src/test/java/com/devonfw/ide/gui/context/ProjectManagerTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+    // Create a project folder without a workspaces subdirectory
+    Path testProject = ideRoot.resolve("test-project-no-workspaces");
+    Files.createDirectory(testProject);
+
+    projectManager = new ProjectManager(ideRoot);
+
+    // Verify that test-project-no-workspaces is not recognized
+    assertThat(projectManager.getProjectNames()).doesNotContain("test-project-no-workspaces");
+    assertThat(projectManager.getProjectNames()).containsExactlyInAnyOrder("project-0", "project-1", "project-2", "project-3", "project-4", "project-5");
```
hohwille: use constant.

gui/src/test/java/com/devonfw/ide/gui/context/ProjectManagerTest.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  @Test
+  void testReadProjectsExcludesUnderscorePrefixedFolders() {
+
+    projectManager = new ProjectManager(ideRoot);
+
+    // Verify that _ide folder is not in the project names
+    assertThat(projectManager.getProjectNames()).doesNotContain("_ide");
+    assertThat(projectManager.getProjectNames()).containsExactlyInAnyOrder("project-0", "project-1", "project-2", "project-3", "project-4", "project-5");
```
hohwille: use constant.

gui/src/test/java/com/devonfw/ide/gui/context/ProjectManagerTest.java
```diff
@@ -0,0 +1,132 @@
...
+  @Test
+  void testProjectManagerFull() throws NotDirectoryException {
+
+    projectManager = new ProjectManager(ideRoot);
+
+    assertThat(projectManager).isNotNull();
+    assertThat(projectManager.getProjectNames()).containsExactlyInAnyOrder("project-0", "project-1", "project-2", "project-3", "project-4", "project-5");
+    for (String projectName : projectManager.getProjectNames()) {
+      assertThat(projectManager.getWorkspaceNames(projectName)).containsExactlyInAnyOrder("foo-test", "main");
+    }
```
hohwille: This is what I was talking about. Then you could assert that `project-1` has workspaces xyz and `project-2` instead has abc (whatever).
And for test-data I have learned that it sooner or later hurts you if you use technical random instead of meaningful examples.

hohwille (approved):
@laim2003 I now found the time to read and review the entire diff.
Thanks also for your rework. This PR is very nice progress and improvement. 👍 
I left some further review remarks for improvement.
We should then finally merge this instead if dying in perfection.
There will be more rounds of changes in the GUI and we should stop blocking other stories that want to build on top of this...
However, I have always learned that we always need a solid foundation rather than a sandbank to build on top. So lets keep some pair of careful looking eyes on the GUI stuff that we are evolving...

gui/src/main/java/com/devonfw/ide/gui/context/ProjectManager.java
laim2003: Ahh yes, this was a remaining fragment from the original implementation from the first MVP. I guess, I will change this so only folders matching FOLDER_UNDERSCORE_IDE will be excuded.

gui/src/test/resources/ide-projects/testProject/project-0/workspaces/foo-test/placeholder
laim2003: Oh ok, I will change this.

laim2003:
I will not include any refactorings regarding the Singleton patterns etc in this PR, as @KarimALotfy has implemented some changes regarding this in one of his PRs.

Issue #1347, opened 2025-05-28
environment variables of different project


janvcap-mac:
### Actual behavior

After switching projects IDEasy still uses wrong environment paths.

```
<username>@Mac-mini ~ % icd -p test
IDE environment variables have been set for /Users/<username>/projects/test in workspace main
<username>@Mac-mini test % ide env
DOTNET_VERSION="6*"
export ECLIPSE_HOME="/Users/<username>/projects/test/software/eclipse"
ECLIPSE_VERSION="2024-03"
GRAALVM_EDITION="community"
export GRAALVM_HOME="/Users/<username>/projects/test/software/extra/graalvm"
GRAALVM_VERSION="23.0.2"
HOME="/Users/<username>"
export IDEA_PROPERTIES="/Users/<username>/projects/test/workspaces/main/idea.properties"
IDE_HOME="/Users/<username>/projects/test"
IDE_VARIABLE_SYNTAX_LEGACY_SUPPORT_ENABLED="false"
export INTELLIJ_HOME="/Users/<username>/projects/test/software/intellij"
export JAVA_HOME="/Users/<username>/projects/test/software/java"
JAVA_VERSION="21*"
export M2_REPO="/Users/<username>/projects/test/conf/mvn/repository"
export MAVEN_ARGS="-s /Users/<username>/projects/test/conf/mvn/settings.xml -Dsettings.security=/Users/<username>/projects/test/conf/mvn/settings-security.xml"
export MVN_HOME="/Users/<username>/projects/test/software/mvn"
export NODE_HOME="/Users/<username>/projects/test/software/node"
NODE_VERSION="v18*"
export PATH="/Users/<username>/projects/test/software/node/bin:/Users/<username>/projects/test/software/java/bin:/Users/<username>/projects/test/software/mvn/bin:/Users/<username>/projects/test/software/intellij:/Users/<username>/projects/test/software/eclipse:/Users/<username>/.rd/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/<username>/projects/_ide/installation/bin"
PYTHON_VERSION="3.11*"
WORKSPACE="main"
WORKSPACE_PATH="/Users/<username>/projects/test/workspaces/main"
<username>@Mac-mini test % icd -p test2
IDE environment variables have been set for /Users/<username>/projects/test2 in workspace main
<username>@Mac-mini test2 % ide env
DOTNET_VERSION="6*"
GRAALVM_EDITION="community"
[...]

PR #1813, opened 2026-04-10
#1270: Load global user settings outside project

### LINKED ISSUE #1270: IDEasy does not load global user settings if outside a project
### Expected behavior

As a IDEasy user, I want to configure some things in `~/.ide/ide.properties` so that these settings always apply to IDEasy no matter if I am in a project and which one.

### Actual behavior

If I am outside of an IDEasy project and run `ide` then the settings from `~/.ide/ide.properties` are lot loaded and have no effect.

1. `touch ~/.ide/ide.properties`
2. `echo 'export FOO=bar' >> ~/.ide/ide.properties`
3. Open a new terminal outside of an IDEasy project (e.g. in your home directory)
4. `ide`
5. `echo $FOO`

You will get an empty output while `bar` would be expected.

### Related/Dependent Issues

#897 

### Comments/Hints:

I even have a valid use-case and rationale for this feature:
I was analysing bug #897 and found a workaround as documented [here](https://github.com/devonfw/IDEasy/blob/main/documentation/proxy-support.adoc#truststore) (see step 11.).
Even with this workaround `ide status` still fails if I am not inside of an IDEasy project.

### Affected version:

* 2025.04.002

ducminh02:
### Implemented changes:

- Updated setUserHome to refresh dependent paths (userHomeIde, downloadPath) and reset variables. This ensures tests using isolated user homes correctly load local configurations like ide.properties.

---

## Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandlet.java
```diff
@@ -65,9 +65,6 @@ public boolean isProcessableOutput() {
 
   @Override
   protected void doRun() {
-    if (this.context.getIdeHome() == null) {
-      throw new CliExitException();
-    }
```
MarvMa, line 68:
```suggestion
  if (this.context.getIdeHome() == null) {
      throw new CliExitException();
    }
```
I might have misunderstood something here, but why did you remove this nullcheck?

MarvMa (commented):
The changes look useful to prevent the Bug from occurring in the feature. Nice Work 👍

cli/src/main/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandlet.java
ducminh02: actually it should be removed.

Currently, if a user runs ide in a directory that isn't an IDEasy project, getIdeHome() returns null. With the null check in place, the env commandlet throws an exception and exits immediately. Because it exits early, it never reaches the printLines method, meaning no output is produced for the shell wrapper script to capture. This prevents global settings in ~/.ide/ide.properties from ever being loaded into the shell session when outside a project.

hohwille: I agree on this but such feature would need to be tested with a real native image in your terminal.
You seemed to have missed this:
https://github.com/devonfw/IDEasy/blob/2a5e160348d61c69e2332c3b94475f36f274bbdb/cli/src/main/package/functions#L22-L24

Ideally we should also consider #351 (similar or same to #1347) and ensure that a previous `IDE_HOME` and `WORKSPACE` would be unset.
Then that linked code could be adopted by adding `&& [ -n "${IDE_HOME} ]` to the if condition.

cli/src/main/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandlet.java
```diff
@@ -84,9 +81,12 @@ protected void doRun() {
     List<VariableLine> variables = this.context.getVariables().collectVariables();
     Map<String, VariableLine> variableMap = variables.stream().collect(Collectors.toMap(VariableLine::getName, v -> v));
 
-    EnvironmentVariableCollectorContext environmentVariableCollectorContext = new EnvironmentVariableCollectorContext(variableMap,
-        new VariableSource(EnvironmentVariablesType.TOOL, this.context.getSoftwarePath()), pathSyntax);
-    setEnvironmentVariablesInLocalTools(environmentVariableCollectorContext);
+    Path softwarePath = this.context.getSoftwarePath();
+    if (softwarePath != null) {
```
hohwille, line 85: nice catch. Perfectly correct to make it work 👍

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -601,6 +601,9 @@ public Path getUserHome() {
   protected void setUserHome(Path userHome) {
 
     this.userHome = userHome;
+    this.userHomeIde = userHome.resolve(FOLDER_DOT_IDE);
+    this.downloadPath = userHome.resolve("Downloads/ide");
+    this.variables = null;
```
hohwille, line 607: Also nice catch. This was actually always buggy before your change. Thanks for finding and fixing 👍

hohwille (approved):
@ducminh02 thanks for your PR. You did a great job to implement this story 👍 
I found some things you missed out. Do not worry too much, since one needs a lot of deep knowledge of IDEasy to understand the coherences but with an end user test via native image this would have been revealed. I plan a KT session with the team about this for the next future...

cli/src/main/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandlet.java
ducminh02: Okay, I have setup GraalVM and tested the feature with a real image in my terminal.

I also added

unset IDE_HOME
unset WORKSPACE

so that the shell would forget those variables before asking the CLI for the new ones.

cli/src/test/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandletGlobalTest.java
```diff
@@ -0,0 +1,41 @@
...
+  @Test
+  void testRunOutsideProject(@TempDir Path tempDir) throws IOException {
+
+    // arrange
+    Path userHome = tempDir.resolve("user");
+    Path userHomeIde = userHome.resolve(".ide");
+    Files.createDirectories(userHomeIde);
+    Files.writeString(userHomeIde.resolve("ide.properties"), "export FOO=bar\n");
+
+    IdeTestContext context = new IdeTestContext(tempDir.resolve("cwd"), null);
```
hohwille: I do not understand what `cwd` is good for? That sub-folder does not exist.

```suggestion
    IdeTestContext context = new IdeTestContext(tempDir, null);
```

cli/src/test/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandletGlobalTest.java
```diff
@@ -0,0 +1,41 @@
...
+    context.setUserHome(userHome);
```
hohwille, line 29: FYI: We have a feature to automatically set the home directory to a `home` folder inside the project base-dir and I was wondering if that could be used here to simplify:
https://github.com/devonfw/IDEasy/blob/4a7df9ec0635f11a74bfab006faa87ccf61dcb2e/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L370-L377

However, it currently cannot because that only works if `IDE_HOME` was found. We might want to improve that in the future, but your solution is working perfectly correct.

hohwille (approved):
@ducminh02 thanks for your improvement. Excellent. Now everything should work perfectly correct. 👍 
Ready for merge.

PR #1816, opened 2026-04-13
#1815: Suppress update notification while updating

### LINKED ISSUE #1815: IDEasy tells users to run ide update while running ide update
### Feature idea

IDEasy informs users they should run `ide update` while the very same command is running. This can cause confusion and results in unecessary network traffic should the user decide run `ide update` again. I propose that this message is omitted when running `ide update`.

### Actual Behavior

```
$ ide update
Start: update (pull) settings repository
Successfully ended step 'update (pull) settings repository'.
Start: Copy configuration templates
Successfully ended step 'Copy configuration templates'.
...
```

### Expected Behavior

```
$ ide update
Start: update (pull) settings repository
Successfully ended step 'update (pull) settings repository'.
Start: Copy configuration templates
Successfully ended step 'Copy configuration templates'.
...
```

### Additional context

_No response_

areinicke:
### Implemented changes:

* IDEasy will no longer inform users about an update while running `ide update`

---

## Checklist for this PR

  with `internal`

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
@@ -1296,19 +1296,11 @@ public abstract class AbstractIdeContext implements IdeContext, IdeLogArgFormatt
             if (getGitContext().isRepositoryUpdateAvailable(settingsRepository, getSettingsCommitIdPath()) || (
                 getGitContext().fetchIfNeeded(settingsRepository) && getGitContext().isRepositoryUpdateAvailable(
                     settingsRepository, getSettingsCommitIdPath()))) {
-              
-              if (isSettingsRepositorySymlinkOrJunction()) {
-                if (!(cmd instanceof UpdateCommandlet) && isForceMode()) {
-                  // Only print the update nag if the user did not explicitly call "ide -f update" to prevent confusion when the user already wants to apply the updates.
-                  String msg = "Updates are available for the settings repository. Please pull the latest changes by yourself or by calling \"ide -f update\" to apply them.";
-                  IdeLogLevel.INTERACTION.log(LOG, msg);
-                }
-              } else {
-                if (!(cmd instanceof UpdateCommandlet)) {
-                  // Only print the update nag if the user did not explicitly call "ide update" to prevent confusion when the user already wants to apply the updates.
-                  String msg = "Updates are available for the settings repository. If you want to apply the latest changes, call \"ide update\"";
-                  IdeLogLevel.INTERACTION.log(LOG, msg);
-                }
+
+              // Inform the user that an update is available. The update message is suppressed if we are already running the update
+              String msg = determineSettingsUpdateMessage(cmd);
+              if (msg != null) {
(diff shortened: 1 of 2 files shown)


jakozian (approved):
Nice fix!

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -1295,13 +1295,19 @@ private ValidationResult applyAndRun(CliArguments arguments, Commandlet cmd) {
             if (getGitContext().isRepositoryUpdateAvailable(settingsRepository, getSettingsCommitIdPath()) || (
                 getGitContext().fetchIfNeeded(settingsRepository) && getGitContext().isRepositoryUpdateAvailable(
                     settingsRepository, getSettingsCommitIdPath()))) {
-              String msg;
               if (isSettingsRepositorySymlinkOrJunction()) {
-                msg = "Updates are available for the settings repository. Please pull the latest changes by yourself or by calling \"ide -f update\" to apply them.";
+                if (!(arguments.getInitialArgument().get().equals("update") && isForceMode())) {
```
hohwille: The check `arguments.getInitialArgument().get().equals("update")` is maybe a little bit unsafe.
It will check the first CLI argument that could theoretically also be an option (e.g. `--force-pull`).
IMHO it would be safer to use `cmd instanceof UpdateCommandlet` for this check.
It would also be better to have this check only once while it currently exists twice.

Further, I think that this `&& isForceMode()` is incorrect. Why did you add this?
Without this change, you could also avoid duplicating the logging statement as it was before.

hohwille (approved):
@areinicke thanks for your PR. You picked the right spot where to fix that and your check for the `update` seems to work in most cases. 👍 
For 100% correctness, I added a review comment to improve your code. Please have a look.
I would suggest to do the `cmd instanceof UpdateCommandlet` check already in an outer `if` statement and then we do nothing (not even evaluate `isSettingsRepositorySymlinkOrJunction()` in case the update commandlet was invoked).

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
areinicke: I have added the `isForceMode()` because of the Log message which recommended the user to run "`ide **-f** update`". Since that message is different than the one in the else-statement (which only states "`ide update`") I assumed that the force mode is important here.

Specifically, I still wanted users to still see the update message when they were running `ide update` but not `ide -f update` as the log message instructs them to do.

Can you take another look if the Force mode is necessary here? Otherwise, I would just omit it as you suggested.

areinicke: I don't actually think that using `arguments.getInitialArgument().get().equals("update")` results in an issue when using it with an option (e.g., --no-colors).
It seems that options are handled different and are not part of the `arguments` object.
For example, when running `ide --no-colors update`, the `getinitialArgument()` method still returns "`update`" and not "`--no-colors`".

However, I can check if your proposed adjustment also works.
Edit: I have updated the PR with the adjusted update detection using `cmd instanceof UpdateCommandlet`.

PR #1817, opened 2026-04-13
#1174: integrate java azul

### LINKED ISSUE #1174: Add UrlUpdater for Java Azul edition
For Java we want to support the [azul edition](https://www.azul.com/downloads/?package=jdk#zulu).

There is even an API providing JSON to find releases and download URLs.

Example:
https://api.azul.com/metadata/v1/zulu/packages?availability_types=ca&release_status=both&page_size=1000&include_fields=java_package_features,release_status,support_term,os,arch,hw_bitness,abi,java_package_type,javafx_bundled,sha256_hash,cpu_gen,size,archive_type,certifications,lib_c_type,crac_supported&page=6&azul_com=true

The idea of this story is to add a `JavaAzulUrlUpdater` to `url-updater` module of IDEasy so we generate the metadata for the `azul` edition of java that would appear here:
https://github.com/devonfw/ide-urls/tree/master/java

Motivation is to fix #1152 so users of IDEasy can switch to `azul` edition to support ARM and x86 natively on every major OS.

MarvMa:
### Implemented changes:

* Created `JavaAzulUrlUpdater`
* Created a java/azul edition 

This update can be tested using the following commands.

`ide list-editions java` -> should log that theres two editions (java, azul)
`ide set-edition java azul` -> should log that the edition has been changed
`ide install java` -> should install the latest java-azul version for the correct os
---

## Checklist for this PR

  with `internal`


url-updater/src/main/java/com/devonfw/tools/ide/url/tool/java/JavaAzulUrlUpdater.java
```diff
@@ -0,0 +1,109 @@
...
+  @Override
+  protected String doGetVersionUrl() {
+    return getVersionBaseUrl()
+        + "/metadata/v1/zulu/packages"
+        + "?availability_types=CA"
+        + "&java_package_type=jdk"
+        + "&release_status=ga"
+        + "&latest=true"
+        + "&os=linux"
+        + "&arch=x86"
```
hohwille: why only linux x86? No Windows? No Mac? No ARM?

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/java/JavaAzulUrlUpdater.java
```diff
@@ -0,0 +1,109 @@
...
+        + "&archive_type=tar.gz"
+        + "&javafx_bundled=false"
+        + "&crac_supported=false"
+        + "&page_size=100";
```
hohwille: If we have to limit the page size, don't we need paging and potentially have a start offset that we iterate over chunks? Or can we somehow guarantee that there will never be more than 100 versions?

url-updater/src/main/java/com/devonfw/tools/ide/url/UpdateInitiator.java
```diff
@@ -34,6 +34,7 @@ public static void main(String[] args) {
     String pathToRepo = args[0];
     Instant expirationTime = null;
     String selectedTool = null;
+    String selectedEdition = null;
```
hohwille: my suggestion in the daily was not to add an extra parameter but to support also to specify the simple classname of the URL Updater instead of the tool name. So you can provide `java` as 3rd arg or `JavaAzulUrlUpdater` if you only want to test only that. IMHO this would be more KISS.
WDYT?

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/java/JavaAzulJsonVersion.java
```diff
@@ -0,0 +1,48 @@
...
+      if (i > 0) {
+        version.append('.');
+      }
+      version.append(this.javaVersion.get(i));
+    }
+    if (this.buildNumber != null) {
+      version.append('_').append(this.buildNumber);
+    }
+    return version.toString();
+  }
```
hohwille, line 51: Looking at the JSON results from your query URL, I see that this is producing duplicates.
* https://cdn.azul.com/zulu/bin/zulu26.28.59-ca-jdk26.0.0-linux_x64.tar.gz
* https://cdn.azul.com/zulu/bin/zulu26.28.59-ca-jdk26.0.0-linux_musl_x64.tar.gz

I do not know what is the difference but both entries will result in the version number `26.0.0_35`.
I guess there is nothing wrong here in this method but rather in the JSON URL.
However, we should IMHO decide which of these two variants we want to provide and should not leave it up to random or cause errors because of the duplicates.

hohwille (approved):
@MarvMa thanks for your PR. You found a proper REST API to build a reasonable UrlUpdater. 👍 
Just stumbled over it and wanted to already give fast feedback. I saw it is still in team review but no reviewer was yet assigned so I decided to already give feedback.
IMHO we need some rework here, please have a look at my comments.

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/java/JavaAzulJsonVersion.java
hohwille: In temurin we have versions like `26_35`.
Therefore my assumption is that we should omit trailing `.0` version segments from the version.

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/java/JavaAzulUrlUpdater.java
MarvMa: This Filter has been added to retrieve the versions just once. Without this Filter theres duplicate Versions for all the different os's and architectures.

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/java/JavaAzulUrlUpdater.java
MarvMa: It seems like the page_size parameter is optional. While testing i realised without this parameter there JSON will not include all of the Versions. Setting the page_size to the maximum of 1000 could be a workaround for now. I also hope that at some point the devs of azul will decide to provide a better endpoint to retrieve all the versions.

url-updater/src/main/java/com/devonfw/tools/ide/url/UpdateInitiator.java
MarvMa: I misunderstood that. I will refactor the code to your recommendation

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/java/JavaAzulJsonVersion.java
MarvMa: The current PR should use the JSON just for the purpose of identifying all the different versions. The URL is beeing built in the `doAddVersion` method overwrite, because of the messy JSON provided by Azul.

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/java/JavaAzulUrlUpdater.java
hohwille: Fine, but there are still duplicates as I wrote in the other comment.
Make the filter more specific or use different OS?

MarvMa: I changed the os to mac. Seems fine to me: https://[api.azul.com/metadata/v1/zulu/packages?availability_types=CA&java_package_type=jdk&release_status=ga&latest=true&os=mac&arch=x86&archive_type=tar.gz&javafx_bundled=false&crac_supported=false&page_size=1000](https://api.azul.com/metadata/v1/zulu/packages?availability_types=CA&java_package_type=jdk&release_status=ga&latest=true&os=mac&arch=x86&archive_type=tar.gz&javafx_bundled=false&crac_supported=false&page_size=1000)

PR #1824, opened 2026-04-15
#1800: IntelliJ will automatically switch to standard edition when downloading newer ultimate editions

### LINKED ISSUE #1800: Download IntelliJ unified version from 2025.3 onwards
### Feature idea

[Starting from version 2025.3](https://blog.jetbrains.com/idea/2025/12/intellij-idea-unified-release/), IntelliJ's free version and ultimate are the same product, with the test phase of ultimate features ending after 30 days. An open source version is no longer released and has to be built from source.

If no version and edition has been specified, IDEasy downloads the last "true" community edition:

```
Trying to download intellij-2025.2.6.1-windows-x64.zip from https://download.jetbrains.com/idea/ideaIC-2025.2.6.1.win.zip
```

IDEasy should switch to the ultimate download URLs from 2025.3 onwards. The simplest solution would probably be to add the same URLs to [ide-urls:intellij/intellij](https://github.com/devonfw/ide-urls/tree/master/intellij/intellij).

### Additional context

_No response_

areinicke:
### Background

Before IntelliJ version 2025.3, there were two separate versions for IntelliJ: Standard & Ultimate. Both versions had different URLs to download and were separate from each other. Within IDEasy, users can configure which version they wish to download, with the Community version being the default.
Starting from version 2025.3, both IntelliJ versions were unified, meaning that both editions are in one software with only one URL. Since then, new versions were only made available to users who have specifically selected the "ultimate" edition in IDEasy as new versions and download URLs were only provided for this edition.

This is being addressed by [devonfw/ide-urls#39](https://github.com/devonfw/ide-urls/pull/39) which transfers the download URLs for versions starting from 2025.3 to the standard edition and removing them from the ultimate edition. It would be unnecessary work to maintain and continuously add identical download URLs for both editions, so moving them to the standard edition will benefit the most users.

Since we also want users who have set their edition to "ultimate" to get access to new versions in IDEasy we implement an additional functionality in this PR.

### Implemented changes:

* When installing IntelliJ using IDEasy and the specified or configured edition is "ultimate", IDEasy will now under some circumstances switch to the standard ("intellij") edition automatically.
* The circumstances where IDEasy will switch are as follows:

1. The specified or configured version is equal or greater than 2025.3
2. There is no specified or configured version (will install the latest version)

* If such a switch occurs, an informational message is logged to the console
* Several test cases were added to test this functionality
* In the future, other tools can more easily implement similar functionality by overriding the method `adjustRequestedEdition()` in `ToolCommandlet.java`.

### Note:

* Merging this PR requires also merging [devonfw/ide-urls#39](https://github.com/devonfw/ide-urls/pull/39).

---

## Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/tool/ToolEditionAndVersion.java
```diff
@@ -68,6 +68,14 @@ public void setEdition(ToolEdition edition) {
...
   }
 
+    /**
+   * @param edition new value of {@link #getEdition()}.
+   */
+  public void replaceEdition(ToolEdition edition) {
+
+    this.edition = edition;
+  }
+
```
laert-ll, line 78: It seems to me like the logic here is nearly identical to [#L63](https://github.com/devonfw/IDEasy/blob/4ad12876b8b558315c69beb944488770d6fc26f6/cli/src/main/java/com/devonfw/tools/ide/tool/ToolEditionAndVersion.java#L63)? In that case, I think this function may be redundant and we can simply use setEdition instead.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -367,21 +367,33 @@ private void completeRequestRequested(ToolInstallRequest request) {
...
       edition = requested.getEdition();
       if (edition == null) {
+        // If no edition was specified, set it to the configured one
         edition = new ToolEdition(this.tool, getConfiguredEdition());
-        requested.setEdition(edition);
       }
+
+      // Adjust edition if necessary based on requested version. This is needed for tools like IntelliJ where we may need to automatically switch editions
+      ToolEdition edition_adjusted = adjustRequestedEdition(requested, edition);
+      requested.replaceEdition(edition_adjusted);
```
laert-ll: As mentioned in another comment, we could simply use the setEdition function here instead.
```suggestion
      requested.setEdition(edition_adjusted);
```

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -80,6 +86,31 @@ public void setEnvironment(EnvironmentContext environmentContext, ToolInstallati
...
+      if ((!(requested != null && !"*".equals(requested.getVersion().toString())) && getConfiguredVersion().compareVersion(INTELLIJ_LAST_SEPARATE_VERSION) == VersionComparisonResult.GREATER) || // No version specified (or no request) and configured version is > 2025.2.6.1
+        (requested != null && requested.getVersion() != null && requested.getResolvedVersion().compareVersion(INTELLIJ_LAST_SEPARATE_VERSION) == VersionComparisonResult.GREATER) || // Version specified and is > 2025.2.6.1
+        (!(requested != null && !"*".equals(requested.getVersion().toString())) && "*".equals(getConfiguredVersion().toString()))) { // No version specified (or no request) and no configured version
+        
+        // Switching to IntelliJ Standard edition
+        LOG.warn("""
+                 Notice: You have configured IDEasy to use the IntelliJ Ultimate Edition. Since version 2025.3, the Ultimate and Community editions of IntelliJ have been unified into a single edition.
+                 Since you are attempting to install a verson of IntelliJ that is 2025.3 or newer, we are automatically switching your edition to the unified edition to ensure compatibility.
```
laert-ll:
```suggestion
                 Since you are attempting to install a version of IntelliJ that is 2025.3 or newer, we are automatically switching your edition to the unified edition to ensure compatibility.
```
tiny typo :)

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -80,6 +86,31 @@ public void setEnvironment(EnvironmentContext environmentContext, ToolInstallati
...
+  @Override
+  protected ToolEdition adjustRequestedEdition(ToolEditionAndVersion requested, ToolEdition edition) {
+
+    // Check if edition is ultimate (any condition must be true)
+    if ((!(requested != null && requested.getEdition() != null) && "ultimate".equals(getConfiguredEdition())) || // No edition specified (or no request) but configured edition is ultimate
+      (requested != null && requested.getEdition() != null && requested.getEdition().toString().equals("intellij/ultimate"))) { // Edition specified as ultimate
+
+      // Check if specified or configured version warrants changes (any condition must be true)
+      if ((!(requested != null && !"*".equals(requested.getVersion().toString())) && getConfiguredVersion().compareVersion(INTELLIJ_LAST_SEPARATE_VERSION) == VersionComparisonResult.GREATER) || // No version specified (or no request) and configured version is > 2025.2.6.1
+        (requested != null && requested.getVersion() != null && requested.getResolvedVersion().compareVersion(INTELLIJ_LAST_SEPARATE_VERSION) == VersionComparisonResult.GREATER) || // Version specified and is > 2025.2.6.1
```
laert-ll: I found a tiny bug here:

getResolvedVersion() returns null when the version is a wildcard pattern, but there's no null check before calling .compareVersion() on it.


```suggestion
        (requested != null && requested.getVersion() != null && requested.getResolvedVersion() != null && requested.getResolvedVersion().compareVersion(INTELLIJ_LAST_SEPARATE_VERSION) == VersionComparisonResult.GREATER) || // Version specified and is > 2025.2.6.1
```

Steps to reproduce:

```
ide set-edition intellij ultimate
ide set-version intellij '2025.*'
ide install intellij
```

laert-ll (commented):
I tested this locally and found a couple of issues I wrote down in the comments above. Otherwise everything works great, the logic handles the version cutoff correctly! Good work! 🙂

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
areinicke: Nice catch. Your fix resolves this problem

cli/src/main/java/com/devonfw/tools/ide/tool/ToolEditionAndVersion.java
areinicke: This method has similar functionality but is not identical. The other method `setEdition()` disallows changing an already set edition which would result in an IllegalStateException for this use case. I didn't want to simply remove this check as I am not sure if it isn't important in some other cases. Therefore, I have decided to create a new method which does not include this check.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
areinicke: See other comment for justification of this method

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -367,21 +367,33 @@ private void completeRequestRequested(ToolInstallRequest request) {
     ToolEditionAndVersion requested = request.getRequested();
     if (requested == null) {
       edition = new ToolEdition(this.tool, getConfiguredEdition());
-      requested = new ToolEditionAndVersion(edition);
+
+      // Adjust edition if necessary based on requested version. This is needed for tools like IntelliJ where we may need to automatically switch editions
+      ToolEdition edition_adjusted = adjustRequestedEdition(requested, edition);
+
+      requested = new ToolEditionAndVersion(edition_adjusted);
```
hohwille: Why doing this twice? IMHO we should avoid such redundancy.
Also in Java variables are named in `pascalCase` and not `snake_case`.
```suggestion
      requested = new ToolEditionAndVersion(edition);
```

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -367,21 +367,33 @@ private void completeRequestRequested(ToolInstallRequest request) {
...
         edition = new ToolEdition(this.tool, getConfiguredEdition());
-        requested.setEdition(edition);
       }
+
+      // Adjust edition if necessary based on requested version. This is needed for tools like IntelliJ where we may need to automatically switch editions
+      ToolEdition edition_adjusted = adjustRequestedEdition(requested, edition);
+      requested.replaceEdition(edition_adjusted);
     }
+
+    edition = requested.getEdition();
```
hohwille, line 383:
```suggestion
       // If no edition was specified, set it to the configured one
       edition = new ToolEdition(this.tool, getConfiguredEdition());
       requested.setEdition(edition);
     }
  }
  // Adjust edition if necessary based on requested version. This is needed for tools like IntelliJ where we may need to automatically switch editions
  requested = adjustRequestedEdition(requested);
  edition = requested.getEdition();
```

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -367,21 +367,33 @@ private void completeRequestRequested(ToolInstallRequest request) {
...
     GenericVersionRange version = requested.getVersion();
     if (version == null) {
       version = getConfiguredVersion();
       requested.setVersion(version);
     }
     VersionIdentifier resolvedVersion = requested.getResolvedVersion();
+
```
hohwille:
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -397,8 +409,22 @@ private void completeRequestRequested(ToolInstallRequest request) {
       }
       requested.setResolvedVersion(resolvedVersion);
     }
+
```
hohwille:
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -397,8 +409,22 @@ private void completeRequestRequested(ToolInstallRequest request) {
...
+   *
+   * @param request the install request.
+   * @param requestedVersion the version requested by the user before IDEasy applies defaults. May be {@code null}.
+   */
+  protected ToolEdition adjustRequestedEdition(ToolEditionAndVersion request, ToolEdition edition) {
+
+    // default no-op
+    return edition;
+  }
+
```
hohwille, line 419: To make this easier I would recommend it like this:
```suggestion
  /**
   * Hook for subclasses to adjust the requested tool edition before the version is finalized.
   *
   * @param requested the requested {@link ToolEditionAndVersion}.
   * @return the given or transformed {@link ToolEditionAndVersion}.
   */
  protected ToolEditionAndVersion adjustRequestedEdition(ToolEditionAndVersion requested) {

    return requested; // default no-op
  }
```

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -86,6 +92,31 @@ public void setEnvironment(EnvironmentContext environmentContext, ToolInstallati
...
+                 Notice: You have configured IDEasy to use the IntelliJ Ultimate Edition. Since version 2025.3, the Ultimate and Community editions of IntelliJ have been unified into a single edition.
+                 Since you are attempting to install a version of IntelliJ that is 2025.3 or newer, we are automatically switching your edition to the unified edition to ensure compatibility.
+                 To specifically install the last true ultimate version of IntelliJ, please run "ide install intellij 2025.2.6.1".
+                 Otherwise, we recommend permanently switching to the unified edition by running "ide set-edition intellij intellij".""");
+
+        return new ToolEdition(this.tool, "intellij"); 
+      }
+    }
+    return edition;
+  }
```
hohwille: Can be simplified a lot now.
See also https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc#consider-extracting-local-variable-for-multiple-method-calls

```suggestion
  protected ToolEditionAndVersion adjustRequestedEdition(ToolEditionAndVersion requested) {

    ToolEdition edition = requested.getEdition();
    if ("ultimate".equals(edition.edition()) {
      VersionIdentifier version = requested.getVersion();
      if (VersionIdentifier.LATEST.equals(version) || VersionIdentifier.LATEST_UNSTABLE.equals(version) || version.isGreater(INTELLIJ_LAST_SEPARATE_VERSION)) {
        // Switching to IntelliJ Standard edition
        LOG.warn("""
                 Notice: You have configured IDEasy to use the IntelliJ Ultimate Edition. Since version 2025.3, the Ultimate and Community editions of IntelliJ have been unified into a single edition.
                 Since you are attempting to install a version of IntelliJ that is 2025.3 or newer, we are automatically switching your edition to the unified edition to ensure compatibility.
                 To specifically install the last true ultimate version of IntelliJ, please run "ide install intellij 2025.2.6.1".
                 Otherwise, we recommend permanently switching to the unified edition by running "ide set-edition intellij intellij".""");
        edition = new ToolEdition(this.tool, this.tool); 
        requested.replaceEdition(edition_adjusted);
      }
    }
    return requested;
  }
```

cli/src/main/java/com/devonfw/tools/ide/tool/ToolEditionAndVersion.java
hohwille: We could have also removed the exception from the setter. However, I am fine with your approach to make it more explicit and avoid accidental mutations.

cli/src/test/java/com/devonfw/tools/ide/tool/intellij/IntellijTest.java
```diff
@@ -195,6 +198,68 @@ void testIntellijMvnAndGradleRepositoryImport() {
...
+
+    // arrange
+    IdeTestContext context = newContext("intellij");
+    Intellij commandlet = context.getCommandletManager().getCommandlet(Intellij.class);
+    commandlet.setEdition("ultimate");
+    commandlet.setVersion("*");
+    ToolEdition original = new ToolEdition("intellij", "ultimate");
+
+    // act
+    ToolEdition adjusted = commandlet.adjustRequestedEdition(null, original);
```
hohwille: With my suggested refactoring this also needs to be changed (3x).
Sorry. When you do less white-boxed tests, you could avoid such effects and test more integrative that the real installation was switched to the other edition.

hohwille (approved):
@areinicke thanks for your PR. You perfectly found the right spot where to solve the issue. 👍 
I left some review comments for improvement and better readability/maintainability of the code. Please have a look and consider my suggestions. I am always picky in reviews. But so far developers reported that they learn a lot from the discussions and appreciate it ;)
And I learned from situations where I stopped being picky, that later problems returned as a boomerang...

There is also something missing in this PR:
* [x] you need to adjust the IntellijUrlUpdater to only add only the new combined releases to `intellij` edition. Otherwise if we merge your PR https://github.com/devonfw/ide-urls/pull/39/changes and the UrlUpdaters run the next night, they will recreate all these versions as ultimate edition again.

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/intellij/IntellijUrlUpdater.java
```diff
@@ -36,9 +40,43 @@ protected String doGetVersionUrl() {
...
+    List<IntellijJsonRelease> movedReleases = new ArrayList<>();
+    Iterator<IntellijJsonRelease> iterator = ultimate.releases().iterator();
+    while (iterator.hasNext()) {
+      IntellijJsonRelease release = iterator.next();
+      if (isUnifiedRelease(release)) {
+        movedReleases.add(release);
+        iterator.remove();
+      }
+    }
+    community.releases().addAll(0, movedReleases);
```
hohwille, line 66: Took my some time to understand the code.
IMHO it would be easier to read if you would simply do `community.releases().add(0, release)` inside the `if` in the `while` loop and skipping the extra variable `movedReleases`.
But after I understood your code, there is nothing wrong and it works more or less the same.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -86,6 +91,36 @@ public void setEnvironment(EnvironmentContext environmentContext, ToolInstallati
...
+  @Override
+  protected ToolEditionAndVersion adjustRequestedEdition(ToolEditionAndVersion requested) {
+
+    ToolEdition edition = requested.getEdition();
+    // Check if edition is set as "ultimate"
+    if ("ultimate".equals(edition.edition())) {
+      
+      VersionIdentifier version;
+      if (requested.getVersion() != null) {
+        version = VersionIdentifier.of(requested.getVersion().toString());  // <- This can be null!!!
```
hohwille: > <- This can be null!!!

What can be `null`? 
`requested.getVersion()` cannot be `null` since you already checked this in the `if` condition.
`toString()` will never return `null` for a version object.
So IMHO this comment is rather confusing but helpful.
If you agree, please remove it. Otherwise can you explain what I am missing?

hohwille (approved):
@areinicke thanks again. Now ready for merge. 👍

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
areinicke: You're right. I believe this comment was a leftover from during the implementation process that I forgot to remove once I had added the surrounding if-clause.

PR #1826, opened 2026-04-16
#1447: Add deterministic archive generation and checksum verification tests

### LINKED ISSUE #1447: Cover checksum verification in JUnit
### Task summary

Currently we do not cover checksum verification (via `urls.sha256`) in our JUnits.
Therefore this functionality remains untested and breaking this functionality would not be revealed by tests (only integration tests what is too late).

### Additional context

In PR #1439 I already tried to cover this feature but hit the problem that the archive files we generate (at least TGZ) are not deterministic across platform so the same input results in different TGZ output on Windows compared to Linux (on GitHubActions). Ideally we can change our compression code in FileAccessImpl to make the archives deterministic/portable. Otherwise we might have to provide different checksums per OS platform.

ducminh02:
Cover checksum verification in JUnit

### Implemented changes:

* Cross-Platform Archive Determinism: Fixed FileAccessImpl to ensure ZIP and TAR.GZ archives are binary-identical across OS by sorting entries and zeroing out timestamps.

* Checksum Verification Tests: Added the first-time JUnit coverage for AbstractToolRepository to verify SHA-256 validation and security aborts.

* Added ArchiveDeterminismTest to verify byte-for-byte reproducibility of archives.
* Added ChecksumVerificationTest to verify the security gatekeeper logic (covering both matching hashes and mismatch failures).

---

## Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -1020,6 +1024,11 @@ private <E extends ArchiveEntry> void compressRecursive(Path path, ArchiveOutput
           tarEntry.setGroupName("group");
           PathPermissions filePermissions = getFilePermissions(child);
           tarEntry.setMode(filePermissions.toMode());
+        } else if (archiveEntry instanceof ZipArchiveEntry zipEntry) {
+          zipEntry.setCreationTime(none);
+          zipEntry.setLastAccessTime(none);
+          zipEntry.setLastModifiedTime(none);
+          zipEntry.setTime(none);
```
satorus, line 1058: mmh, is it really a good idea to delete all file access metadata for the files in the archive? Maybe the createdDate or the lastModifyDate may be relevant in some cases?

cli/src/test/java/com/devonfw/tools/ide/io/ArchiveDeterminismTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    Path archive1 = this.tempDir.resolve("archive1.tar.gz");
+    Path archive2 = this.tempDir.resolve("archive2.tar.gz");
+
+    // act
+    try (OutputStream out1 = Files.newOutputStream(archive1)) {
+      fileAccess.compressTarGz(contentDir, out1);
+    }
+    // Wait a bit to ensure a non-deterministic MTIME would change (though we zero it out)
+    Thread.sleep(1100);
```
satorus: i am a little bit confused...we explicitly set the lastModifiedTime to 0, so as the command above says, this should not be necessary, right?

cli/src/test/java/com/devonfw/tools/ide/io/ArchiveDeterminismTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    Path archive1 = this.tempDir.resolve("archive1.zip");
+    Path archive2 = this.tempDir.resolve("archive2.zip");
+
+    // act
+    try (OutputStream out1 = Files.newOutputStream(archive1)) {
+      fileAccess.compressZip(contentDir, out1);
+    }
+    // Wait a bit to ensure a non-deterministic time would change
+    Thread.sleep(1100);
```
satorus: same as in the tarGz test

cli/src/test/java/com/devonfw/tools/ide/tool/repository/ChecksumVerificationTest.java
```diff
@@ -0,0 +1,103 @@
...
+    Path file = this.tempDir.resolve("testfile.txt");
+    String content = "Hello World";
+    Files.writeString(file, content);
+    String checksum = context.getFileAccess().checksum(file, "SHA-256");
+    UrlGenericChecksum expectedChecksum = new TestUrlGenericChecksum(checksum, "SHA-256");
+
+    // act
+    repo.verifyChecksum(file, expectedChecksum);
+
+    // assert (no exception thrown)
```
satorus: may be do this check more explicitly like

```
try {
    repo.verifyChecksum(file, expectedChecksum);
catch (Exception e) {
        fail(e.getClass().getSimpleName() + " was thrown");
    }
```

cli/src/test/java/com/devonfw/tools/ide/tool/repository/ChecksumVerificationTest.java
```diff
@@ -0,0 +1,103 @@
...
+    Path file = this.tempDir.resolve("testfile.txt");
+    String content = "Hello World";
+    Files.writeString(file, content);
+    String wrongChecksum = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"; // SHA-256 of empty string
+    UrlGenericChecksum expectedChecksum = new TestUrlGenericChecksum(wrongChecksum, "SHA-256");
+
+    // act & assert
+    try {
+      repo.verifyChecksum(file, expectedChecksum);
+      fail("Exception expected");
```
satorus: could also try the annotation
`@Test(expected = IOException.class)`

This would fail the test case if no expection of this type is thrown during the test

CHANGELOG.adoc
```diff
@@ -6,6 +6,7 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/1439[#1439]: Add deterministic archive generation and checksum verification tests
```
satorus: The Issue is tagged as "internal" so i guess we should not add it to the changelog :D

cli/src/test/java/com/devonfw/tools/ide/tool/repository/ChecksumVerificationTest.java
ducminh02: I believe it is better to not wrap that in a try and catch block because it would add unnecessary layer to the testing code.  Junit's entire job is to catch errors, wrap and throw it again.

If verifyChecksum throes CliException, it would hits Junit directly

cli/src/test/java/com/devonfw/tools/ide/tool/repository/ChecksumVerificationTest.java
ducminh02: Since IDEasy uses JUnit 5 (Jupiter), the expected parameter is actually no longer supported on the @Test annotation.

and handling the exception explicitly in a try/catch (or with assertThrows) is preferred because it allows us to verify the exact error message, ensuring the test is failing for the right reason

cli/src/test/java/com/devonfw/tools/ide/tool/repository/ChecksumVerificationTest.java
satorus: true, we can leave it like that. Since we use JUnit5, we could also use

````
    assertDoesNotThrow(() -> {
       repo.verifyChecksum(file, expectedChecksum);
    });

cli/src/test/java/com/devonfw/tools/ide/tool/repository/ChecksumVerificationTest.java
satorus: I did not notice we are using JUnit5, so yeah the assertThrows sounds like the best solution!

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
ducminh02: well since the files created through compression logic are only for testing purposes only, so from my perspective we can remove it for determinism across different os

cli/src/test/java/com/devonfw/tools/ide/io/ArchiveDeterminismTest.java
ducminh02: well it is only to test if 2 files compressed at different time, (1 sec apart) should have the same binary. Normally yes, but we removed the metadata so it should be the same binary

cli/src/test/java/com/devonfw/tools/ide/io/ArchiveDeterminismTest.java
ducminh02: same as in the tarGz test

cli/src/test/java/com/devonfw/tools/ide/tool/repository/ChecksumVerificationTest.java
ducminh02: yeah you are right, with that pattern, developers can have a better understanding of what exactly we are testing here

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
hohwille: We can later add a `deterministic` flag to this method so these 0 times will be omitted if that flag is false.
However, @ducminh02 is fully correct that currently this feature/method is only used by tests.

hohwille (approved):
@ducminh02 sorry for the long delay for final review but finally here I am.
Everything is now perfect and ready for merge.

hohwille:
PR improves tests, everything green, noting for E2E test. Setting QA approved.

PR #1832, opened 2026-04-17
#1754: Add support for VS Code plugin versions

### LINKED ISSUE #1754: Add version support to vscode settings .plugin files
### Feature idea

As an IDEasy vscode user I want to be able to set the version of the plugin I want to install.
Sometimes plugins in their latest version e.g. `vitest.explorer:1.48.2` are not working properly, so I need to be able to change the version to f.e. `1.44.2` instead.
Currently only the latest version of a plugin can be installed.

### Additional context

_No response_

areinicke:
### Implemented changes:

* VS Code extensions can now be specified with a plugin version in the extension .properties file (found under `$IDE_HOME/settings/vscode/plugins/`)
* Example usage:
```
plugin_id=esbenp.prettier-vscode
plugin_active=true
tags=formatter
version=12.1.0
```
If the specified version is unavailable, the plugin installation will fail.

* Plugins can be freely upgraded or downgraded by changing the specified version in the .properties file
* Added several test cases that test the intended behavior

---

## Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/tool/plugin/PluginBasedCommandlet.java
```diff
@@ -184,10 +193,24 @@ public void createPluginMarkerFile(ToolPluginDescriptor plugin) {
...
+      deleteExistingPluginMarkerFiles(fileAccess, plugin, pluginMarkerFilePath);
       fileAccess.touch(pluginMarkerFilePath);
     }
   }
 
+  private void deleteExistingPluginMarkerFiles(FileAccess fileAccess, ToolPluginDescriptor plugin, Path currentMarkerFilePath) {
+
+    String markerFilePrefix = "plugin" + "." + getName() + "." + getInstalledEdition() + "." + plugin.name();
+    List<Path> markerFiles = fileAccess.listChildren(currentMarkerFilePath.getParent(),
+        p -> Files.isRegularFile(p) && p.getFileName().toString().startsWith(markerFilePrefix));
```
ducminh02: Great addition! However, there is a subtle edge case here with how .startsWith() matches the prefix.

If a user has two plugins installed where one plugin's name is a prefix of another (for example, java and java-extension-pack), installing or updating the shorter one (java) will accidentally delete the marker file for the longer one (java-extension-pack). This happens because "plugin.vscode.null.java-extension-pack" starts with "plugin.vscode.null.java".

To fix this and ensure we don't accidentally delete other plugins' marker files, we should check that the file name either matches the unversioned prefix exactly, or is followed specifically by the .version- delimiter.

```suggestion
     p -> {
          String fileName = p.getFileName().toString();
          return Files.isRegularFile(p) && (fileName.equals(markerFilePrefix)
              || fileName.startsWith(markerFilePrefix + ".version-"));
        });
```

cli/src/main/java/com/devonfw/tools/ide/tool/plugin/ToolPluginDescriptor.java
```diff
@@ -45,13 +46,14 @@ public static ToolPluginDescriptor of(Path propertiesFile, IdeContext context, b
     Properties properties = context.getFileAccess().readProperties(propertiesFile);
     String id = getString(properties, "id", "plugin_id");
     String url = getString(properties, "url", "plugin_url");
+    String version = getString(properties, "version", "plugin_version");
```
ducminh02, line 49: Users sometimes accidentally leave trailing spaces in .properties files (version=1.44.2 ), which would end up in the CLI command and the marker file.

```suggestion
    String version = getString(properties, "version", "plugin_version");
    if (version != null) {
      version = version.trim();
    }
```

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -64,15 +64,30 @@ public boolean installPlugin(ToolPluginDescriptor plugin, Step step, ProcessCont
     List<String> extensionsCommands = new ArrayList<>();
     extensionsCommands.add("--force");
     extensionsCommands.add("--install-extension");
-    extensionsCommands.add(plugin.id());
+    String extensionInstallTarget = plugin.id();
+    // If a version number was specified, add it to the extension identifier with the format "extensionId@version"
+    Boolean versionSpecified = (plugin.version() != null) && !plugin.version().isBlank();
```
ducminh02: It is better to use the primitive boolean here instead of the boxed Boolean object. This avoids unnecessary object overhead and eliminates the risk of a NullPointerException during unboxing, as this flag will never need to be null.

```suggestion
    boolean versionSpecified = (plugin.version() != null) && !plugin.version().isBlank();
```

ducminh02 (changes requested):
Amazing job on this PR Alex! 🚀

The implementation is very clean and It handles the versioned plugin requirement perfectly!

I've left a few comments on some very minor adjustments and one subtle edge case regarding the marker file cleanup logic that we should address to ensure different plugins with similar names don't conflict.

cli/src/main/java/com/devonfw/tools/ide/tool/plugin/ToolPluginDescriptor.java
areinicke: ok, I see the issue. But this issue also affects the other fields (e.g., "id"). It would be odd to implement special handling just for this field. We could instead just trim all fields automatically in the getString() method, no? What do you think?

ducminh02: okay I get what you mean, it is actually better that way to be honest, with that implementation we can keep the error away from all of the fields and not only the version field

ducminh02 (approved):

cli/src/main/java/com/devonfw/tools/ide/tool/plugin/PluginBasedCommandlet.java
```diff
@@ -168,12 +168,21 @@ private void doInstallPluginStep(ToolPluginDescriptor plugin, Step step, Process
...
   public Path retrievePluginMarkerFilePath(ToolPluginDescriptor plugin) {
     if (this.context.getIdeHome() != null) {
-      return this.context.getIdeHome().resolve(IdeContext.FOLDER_DOT_IDE)
-          .resolve("plugin" + "." + getName() + "." + getInstalledEdition() + "." + plugin.name());
+      String markerFileName = "plugin" + "." + getName() + "." + getInstalledEdition() + "." + plugin.name();
+      String version = plugin.version();
+      if ((version != null) && !version.isBlank()) {
+        markerFileName = markerFileName + ".version-" + normalizeMarkerFileSegment(version);
+      }
+      return this.context.getIdeHome().resolve(IdeContext.FOLDER_DOT_IDE).resolve(markerFileName);
```
hohwille, line 176: A question to understand and maybe challenge this:
1. A project configures a plugin with version.
2. A developer does the setup of the project using IDEasy
3. The project changes the plugin configuration to a newer version
4. The developer invokes `ide update`. Will we now have two marker files for the same plugin with different version numbers in the file-name?
5. Assuming the project then discovers a bug in the new version and does a downgrade to the previous version.
6. Now the developer calls `ide update` once again. IMHO the plugin will not be downgraded because the old version marker file already exists.

Would it make more sense to write the version as content of the file.
Then change the code to read that content and if not empty (no version was explicitly configured) we assume this is the explicit plugin version that was installed by IDEasy.
So the changed code would compare the version from the plugin marker file with the configured version.
If that differs, the plugin will be installed/upgrades/downgraded - otherwise there is nothing to do.

hohwille: OK. Now I read your code further and fully understood your solution.
Your code already works fine and in my scenario in step 4. the old marker file will be deleted.
Therefore also in 6. the plugin will be downgraded.
So to conclude: I just had a different idea in my head how to implement this but your solution also works fine.
Just for the feedback, I will not delete my review comment but post it and then immediately set it to resolved.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -64,15 +64,30 @@ public boolean installPlugin(ToolPluginDescriptor plugin, Step step, ProcessCont
...
       return true;
     } else {
-      LOG.warn("An error occurred while installing plugin: {}", plugin.name());
-      return false;
+        if (versionSpecified) {
+          IdeLogLevel.ERROR.log(LOG, "Failed to install plugin: {} with version: {}", plugin.name(), plugin.version());
+        } else {
+          IdeLogLevel.ERROR.log(LOG, "Failed to install plugin: {}", plugin.name());
+        }
+        LOG.warn("An error occurred while installing plugin: {}", plugin.name());
```
hohwille: Isn't this redundant information to the ERROR log message you also added?

```suggestion
```

BTW: for standard log-levels the syntax `LOG.<level>` is more intuitive. Only use `IdeLogLevel.<LEVEL>.log(LOG, ...)` for IDEasy custom log-levels like `SUCCESS` that do not exist in logger APIs like SLF4J.

hohwille (approved):
@areinicke thank you for your PR. Great job with consistent documentation, tests (incl. AssertJ cleanups), CHANGELOG, etc. 👍 
I only have one review comment about some more or less duplicated logging. But then ready for merge.

hohwille:
One last thing to mention:
We now globally added this version support for plugins.
In IDEs other than VSCode the version can now also be configured leading to different marker files and reinstallations on version change but the other IDEs (e.g. Eclipse, Intellij, ...) will not support the version attribute.
This should not block anything or prevent the merge. I just wanted to make this transparent.
Maybe other IDEs could also support explicit plugin versions in the future.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
areinicke: I agree. This looks like a duplicate log message

PR #1847, opened 2026-04-23
#1788: Created a commandlet to simulate the behaviour of ln -s

### LINKED ISSUE #1788: Add ln commandlet to create links
### Feature idea

On Windows Git-Bash ships with GNU `ln` command.
However, Git-Bash on Windows is somehow not able to properly implement this command.
Therefore, `ln -s source link` will typically do the same as `cp source link` what is unexpected.
This makes it slightly tedious for Windows users to create links reliably.

In IDEasy we have solved all these problems.
The idea of this story is to add a `ln` commandlet to IDEasy.
With that Windows users can easily create real links and also we can use that for manual integration testing of our link functionality.

### Additional context

_No response_

KarimALotfy:
### This PR fixes #1788

### Implemented changes:

Added ide ln -s commandlet to create file links.

---

## Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+    if (Files.exists(targetPath)) {
+      throw new CliException("Target already exists: " + targetPath);
+    }
+
+    // 1) Try true symlink first (desired behavior).
+    try {
+      Files.createSymbolicLink(targetPath, sourcePath);
+      LOG.info("Created symbolic link {} -> {}", targetPath, sourcePath);
+      return;
```
jakozian: This return statement is unnecessary because it is the last statement in this method if symlink creation succeeds.

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,187 @@
...
+      throw new CliException(
+          "Windows blocked symbolic link creation (missing privileges).\n"
+              + "Hard link alternative is not possible because the source is a directory.",
+          originalSymlinkError);
+    }
+
+    ensureSameVolume(sourcePath, targetPath);
+
+    try {
+      Files.createLink(targetPath, sourcePath);
```
jakozian: For file access and operations look into `this.context.getFilesAccess().symlink...` ([FileAccess](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java) & [FileAccessImpl](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java)) because most of the handling of symlinks, junctions and hardlinks is already done there.

With that most of your code becomes obsolete and everything can be shortened.

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,187 @@
...
+  private boolean isWindowsSymlinkPrivilegeProblem(Exception e) {
+
+    if (e instanceof AccessDeniedException) {
+      return true;
+    }
+    if (e instanceof FileSystemException fse) {
+      String msg = fse.getMessage();
+      if (msg != null) {
+        String m = msg.toLowerCase();
+        return m.contains("required privilege")
```
jakozian: These `contains` are not locale safe. E.g. when i run the code on my german machine the error does not get recognized because the message is german.

But as suggested above, this also gets handled by `FileAccess` of IDEasy.

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,187 @@
...
+ */
+public final class LnCommandlet extends Commandlet {
+
+  private static final Logger LOG = LoggerFactory.getLogger(LnCommandlet.class);
+
+  /** Grammar token {@code -s} (required). */
+  public final KeywordProperty symbolic;
+
+  /** The source path to link to. */
+  public final StringProperty source;
```
jakozian: IMO it would be better to conform to the syntax and options of the actual `ln -s`  command:
`ln -s <target> <link>` instead of `ln -s <source> <target>`

jakozian (changes requested):
Nice approach! But i requested changes because most things regarding the issue can be done with `FileAccess`/`FileAccessImpl`.

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,91 @@
...
+   *
+   * @param context the {@link IdeContext}.
+   */
+  public LnCommandlet(IdeContext context) {
+
+    super(context);
+    addKeyword(getName());
+
+    // Treat -s as grammar token
+    this.symbolic = add(new KeywordProperty("-s", true, null));
```
hohwille: You can convert this to a `FlagProperty` and make it optional instead of required.
Then users can create hardlink with `ide ln source target` and symlink with `ide ln -s source target`.

KarimALotfy: Thanks for the suggestion! ✅ I’ve already refactored it so -s is an optional flag:
Could you please check the latest status and confirm it matches what you meant?

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,88 @@
...
+ * It tries to create a true symbolic link first. On Windows, symlink creation may be restricted due to missing privileges. In that case, IDEasy will create a
+ * hard link as an alternative (file-only, same volume) to avoid the Git-Bash behavior of silently copying files.
+ */
+public final class LnCommandlet extends Commandlet {
+
+  /** Grammar token {@code -s} (optional). */
+  public final FlagProperty symbolic;
+
+  /** The source path to link to. */
+  public final StringProperty target;
```
jakozian: I am sorry for the confusion but i am seeing just now that most tools call this `source`, even tough it is a little bit confusing, same can be seen in `FileAccessImpl` in the method `link`:
- "source – the source Path to link to, may be relative or absolute."
- "link – the destination Path where the link shall be created pointing to source."

So lets change that one last time or discuss it further togehter 😁

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,88 @@
...
+    Path cwd = this.context.getCwd();
+    if (cwd == null) {
+      throw new CliException("Missing current working directory!");
+    }
+
+    Path targetPath = cwd.resolve(this.target.getValue()).normalize().toAbsolutePath();
+    Path linkPath = cwd.resolve(this.link.getValue()).normalize().toAbsolutePath();
+
+    if (!Files.exists(targetPath)) {
+      throw new CliException("Source does not exist: " + targetPath);
```
jakozian: With my proposed changes above this message makes finally makes sense. And we can also change `targetPath` to `sourcePath`.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -546,7 +546,7 @@ public void link(Path source, Path link, boolean relative, PathLinkType type) {
       if (type == PathLinkType.SYMBOLIC_LINK) {
         Files.createSymbolicLink(finalLink, finalSource);
       } else if (type == PathLinkType.HARD_LINK) {
-        Files.createLink(finalLink, finalSource);
```
jakozian, line 548: I am not sure if we should change that for this existing `FileAccessImpl` implementation. 
I understand the intent that links if already existing should be replaced, but i dont know if this is the right place to do it instead of just for our ln commandlet...
Altough your proposed change is IMHO an elegant solution because `doRun()` of the ln commandlet stays short.

@hohwille what is your opinion on that?

cli/src/main/resources/nls/Help_de.properties
jakozian, line 1: I propose this:
val.source = Ziel des Links (existierender Pfad).
val.link = Pfad des zu erstellenden Links.

Also i suggest changing the order of those two vals in the help properties.

cli/src/main/resources/nls/Help.properties
jakozian, line 1: I suggest this:
val.source = The source path the link points to (existing file or directory).
val.link = The path where the link is created.

Also i suggest changing the order of those two vals in the help properties.

jakozian (changes requested):

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,88 @@
...
+      throw new CliException("Missing current working directory!");
+    }
+
+    Path sourcePath = cwd.resolve(this.source.getValue()).normalize().toAbsolutePath();
+    Path linkPath = cwd.resolve(this.link.getValue()).normalize().toAbsolutePath();
+
+    if (!Files.exists(sourcePath)) {
+      throw new CliException("Source does not exist: " + sourcePath);
+    }
+    if (this.symbolic != null && this.symbolic.getValue() != null) {
```
hohwille: `symbolic` is final and assigned in the constructor so it can never be `null`.
The valus should be `true` to indicate a `symlink`.
Might be a strange edge-case but it would theoretically be possible to provide `--symbolic=false` as CLI argument and then this code would not work.

```suggestion
    if (this.symbolic.isTrue()) {
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,88 @@
...
+      this.context.getFileAccess().symlink(sourcePath, linkPath, false);
```
hohwille: Do we want to create absolute link by default?
To me it would make sense to have another `FlagProperty` for this.
But to mimic the default behaviour of `ln` we should not turn the given path to absolute by default.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
hohwille: I would follow this concern.
If you need such feature, the recommended way to go is as following:
Add a new parameter to the method:
```
public void link(Path source, Path link, boolean relative, PathLinkType type, boolean override) {
```

In the API (interface `FileAccess` create a default method with the old signature delegating to the new method that provides the old and unchanged behaviour):
```
default void link(Path source, Path link, boolean relative, PathLinkType type) {
  link(source, link, relative, type, false);
}
```

Then in the new/extended method implementation handle the behaviour if `override` is `true` adding the new feature.
But then do it independent of symbolic or hard links.

Finally you can use the new method in your commandlet providing `override` with `true`.

This way you cannot break any existing behaviour.
If we find some place where we think it should also override, we can add the new parameter but then we do that on purpose and not by accident.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -556,7 +556,13 @@ public void link(Path source, Path link, boolean relative, PathLinkType type) {
...
                 + "https://github.com/devonfw/IDEasy/blob/main/documentation/symlink.adoc for further details. Error was: "
                 + e.getMessage());
-        mklinkOnWindows(finalSource, absoluteSource, finalLink, type, relative);
+
+        try {
+          mklinkOnWindows(finalSource, absoluteSource, finalLink, type, relative);
+        } catch (IllegalStateException mkEx) {
+          LOG.info("Creating a hard link as a fallback for the failed mklink attempt.");
+          createHardLink(finalSource, finalLink);
+        }
```
hohwille, line 562: I also do not understand this change.
If we wanted to create a link via Java standards (e.g. `Files.createLink`) and that failed due to lack of permissions on Windows, we already have the option to use `mklink` as fallback.
Why would then retrying the original link creation again suddenly work?

UPDATE: After reading your help documentation I start guessing your intention. It seems you found out that we cannot create a symbolic directory link from one drive to a different drive on Windows if we do not have the admin permissions as a directory junction is not allowed to point to a different drive.
Nice finding. Does `createHardLink` then really work? I was not aware of this.
Still your code is IMHO inconsistent - you should only do that 2nd level fallback in case of `type == PathLinkType.SYMBOLIC_LINK` since otherwise you are retying what has already failed before.

cli/src/main/resources/nls/Help.properties
```diff
@@ -79,6 +79,8 @@ cmd.list-editions=List the available editions of the selected tool.
 cmd.list-editions.detail=To list all available editions of e.g. 'intellij' simply type: 'ide list-editions intellij'.
 cmd.list-versions=List the available versions of the selected tool.
 cmd.list-versions.detail=To list all available versions of e.g. 'intellij' simply type: 'ide list-versions intellij'.
+cmd.ln=Create a link (ln -s).
```
hohwille:
```suggestion
cmd.ln=Create a symbolic or hard link.
```

cli/src/main/resources/nls/Help.properties
```diff
@@ -79,6 +79,8 @@ cmd.list-editions=List the available editions of the selected tool.
...
+cmd.ln.detail=Creates a link similar to "ln -s". If symbolic links are restricted on Windows, IDEasy creates a hard link as an alternative (file-only, same volume). No copying is performed.
```
hohwille: I would not document the implementation details here as nobody will remember to update this when the implementation gets changed.
```suggestion
cmd.ln.detail=Creates a link similar to the command "ln" but working cross-platform (unlike in git-bash on Windows that typically only creates a copy).
```

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -565,6 +571,25 @@ public void link(Path source, Path link, boolean relative, PathLinkType type) {
...
+   * @param source the {@link Path} the hard link will point to.
+   * @param link the {@link Path} where to create the hard link.
+   */
+  void createHardLink(Path source, Path link) {
+    try {
+      if (Files.exists(link, LinkOption.NOFOLLOW_LINKS)) {
+        delete(link);
+      }
+      Files.createLink(link, source);
+      LOG.info("Created hard link at {} pointing to {}", link, source);
```
hohwille: We have already logged that we want to create a link on debug level.
IMHO this additional log message can cause log spam, e.g. if we extract a tar file containing several links...

```suggestion
      LOG.trace("Created hard link at {} pointing to {}", link, source);
```

cli/src/main/resources/nls/Help_de.properties
```diff
@@ -79,6 +79,8 @@ cmd.list-editions=Listet die verfügbaren Editionen des selektierten Werkzeugs a
 cmd.list-editions.detail=Um alle verfügbaren Editionen von z.B. 'intellij' aufzulisten, geben Sie einfach 'ide list-editions intellij' in die Konsole ein.
 cmd.list-versions=Listet die verfügbaren Versionen des selektierten Werkzeugs auf.
 cmd.list-versions.detail=Um alle verfügbaren Versionen von z.B. 'intellij' aufzulisten, geben Sie einfach 'ide list-versions intellij' in die Konsole ein.
+cmd.ln=Link erstellen (wie ln -s).
```
hohwille:
```suggestion
cmd.ln=Erstellt einen symbolischen oder harten Link.
```

cli/src/main/resources/nls/Help_de.properties
```diff
@@ -79,6 +79,8 @@ cmd.list-editions=Listet die verfügbaren Editionen des selektierten Werkzeugs a
...
+cmd.ln.detail=Erstellt einen Link wie "ln -s". Wenn symbolische Links unter Windows eingeschraenkt sind, erstellt IDEasy alternativ einen Hardlink (nur Dateien, gleiches Laufwerk). Es wird nichts kopiert.
```
hohwille:
```suggestion
cmd.ln.detail=Erstellt einen Link wie das Kommando "ln" jedoch plattform-übergreifend (anders als in Git-Bash unter Windows wo typischerweise nur eine Kopie erstellt wird).
```

cli/src/main/resources/nls/Help_de.properties
```diff
@@ -161,6 +163,7 @@ opt.--quiet=Deaktiviert Info Logging ( nur success, warning und error).
 opt.--skip-repositories=Überspringt die Einrichtung der Repositories.
 opt.--skip-tools=Überspringt die Installation/Aktualisierung der Tools.
 opt.--skip-updates=Deaktiviert Aktualisierungen von Tools wenn die installierten Versionen mit den konfigurierten Versionen übereinstimmen.
+opt.--symbolic=Festlegen, ob ein symbolischer Link (Standard) oder ein Hardlink erstellt werden soll.
```
hohwille: The standard in both `ln` (see `ln --help`) and your implementation is to create a hardlink by default.
```suggestion
opt.--symbolic=Erstellt einen symbolischen Link anstelle eines Hardlinks (Standard).
```

cli/src/main/resources/nls/Help.properties
```diff
@@ -161,6 +163,7 @@ opt.--quiet=disable info logging (only log success, warning or error).
 opt.--skip-repositories=skip the setup of repositories.
 opt.--skip-tools=skip the installation/update of tools.
 opt.--skip-updates=disables tool updates if the configured versions match the installed versions.
+opt.--symbolic=Determine if a symbolic link is to be created (default) or a hard link.
```
hohwille: The standard in both `ln` (see `ln --help`) and your implementation is to create a hardlink by default.

```suggestion
opt.--symbolic=Create a symbolic link instead of a hard link (default).
```

cli/src/test/java/com/devonfw/tools/ide/commandlet/LnCommandletTest.java
```diff
@@ -0,0 +1,103 @@
...
+    Path testDir = context.getWorkspacePath().resolve("ln-test");
+    context.getFileAccess().mkdirs(testDir);
+    context.setCwd(testDir, context.getWorkspaceName(), context.getIdeHome());
+
+    Path source = testDir.resolve("source.txt");
+    Path link = testDir.resolve("link.txt");
+    Files.writeString(source, "A", StandardCharsets.UTF_8);
+
+    LnCommandlet cmd = new LnCommandlet(context);
+    cmd.symbolic.setValueAsString("-s", context);
```
hohwille: The value of a flag property is either `true` or `false`.
Therefore this does not make sense:
https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/property/BooleanProperty.java#L57-L69

What you really want to do is IMHO this:
```suggestion
    cmd.symbolic.setValue(Boolean.TRUE);
```

But your current code will IMHO leave the value as `null` currently.

cli/src/test/java/com/devonfw/tools/ide/commandlet/LnCommandletTest.java
```diff
@@ -0,0 +1,103 @@
...
+    cmd.source.setValueAsString("source.txt", context);
+    cmd.link.setValueAsString("link.txt", context);
```
hohwille: KISS?
```suggestion
    cmd.source.setValue("source.txt");
    cmd.link.setValue("link.txt");
```

cli/src/test/java/com/devonfw/tools/ide/commandlet/LnCommandletTest.java
```diff
@@ -0,0 +1,103 @@
...
+    Path testDir = context.getWorkspacePath().resolve("ln-test");
+    context.getFileAccess().mkdirs(testDir);
+    context.setCwd(testDir, context.getWorkspaceName(), context.getIdeHome());
+
+    Path source = testDir.resolve("source.txt");
+    Path link = testDir.resolve("link.txt");
+    Files.writeString(source, "A", StandardCharsets.UTF_8);
+
+    LnCommandlet cmd = new LnCommandlet(context);
+    cmd.symbolic.setValueAsString("-s", context);
```
hohwille: same here

hohwille (approved):
@KarimALotfy thanks for your PR. Nice new commandlet with a carefully covering unit test. 👍 
I left several review comments.
Various of them are nice to have improvements.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
KarimALotfy: @hohwille  Good point, thanks a lot.
I added a parameter to the link method and introduced a -f flag in the Ln cmdlet so users can explicitly opt into overwrite behavior. I kept the override parameter scoped to createHardLink to avoid changing the existing behavior of link.
Should deleteLinkIfExists follow the same condition (i.e., only run when override is set), or is it fine to leave it unconditional?

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
KarimALotfy: Fixed, thanks

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
KarimALotfy: Done — I think this covers it

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
KarimALotfy: Yes, that’s actually the case — I added createHardLink as an additional fallback (using mklink) for this scenario.
In practice, this path is only reached for PathLinkType.SYMBOLIC_LINK. The call site where this happens is guarded by that condition, and it is only triggered when Files.createSymbolicLink throws an exception.

cli/src/main/resources/nls/Help.properties
KarimALotfy: Done ✅

cli/src/main/resources/nls/Help_de.properties
KarimALotfy: Done ✅

cli/src/main/resources/nls/Help.properties
KarimALotfy: Done ✅

cli/src/test/java/com/devonfw/tools/ide/commandlet/LnCommandletTest.java
KarimALotfy: Done ✅

cli/src/test/java/com/devonfw/tools/ide/commandlet/LnCommandletTest.java
KarimALotfy: Done ✅

cli/src/test/java/com/devonfw/tools/ide/commandlet/LnCommandletTest.java
KarimALotfy: Done ✅

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * @param context the {@link IdeContext}.
+   */
+  public LnCommandlet(IdeContext context) {
+
+    super(context);
+    addKeyword(getName());
+
+    this.symbolic = add(new FlagProperty("--symbolic", false, "-s"));
+    this.force = add(new FlagProperty("--force", false, "-f"));
```
hohwille: Sorry, to bother you again, but IDEasy already has a global `--force` option.
You should remove this flag property and use `context.isForceMode()` instead.
See `ide help`.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
hohwille: I now got it. Thanks for your explanation.
To make it more obvious the `try {} catch {}` should be around `Files.createSymbolicLink(finalLink, finalSource)` to make it more obvious but you are right, that all is already correct.

hohwille (approved):
@KarimALotfy thank you for all your updates and explanations. 👍 
Now all looks good to me and is ready for merge.
@jakozian you requested changes. Can you approve as well to unblock the merge?

hohwille:
Unfortunately a failing test is now also blocking the merge:
```
[ERROR] Tests run: 5, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 0.102 s <<< FAILURE! -- in com.devonfw.tools.ide.commandlet.LnCommandletTest
[ERROR] com.devonfw.tools.ide.commandlet.LnCommandletTest.testForceOverridesExistingSymbolicLink -- Time elapsed: 0.032 s <<< FAILURE!
java.lang.AssertionError: 

Expecting code to raise a throwable.
	at com.devonfw.tools.ide.commandlet.LnCommandletTest.testForceOverridesExistingSymbolicLink(LnCommandletTest.java:150)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
```

Is that caused by a side-effect from a change on `main`?
I did not change this PR but only rebased with `main` to resolve the conflict in the CHANGELOG.
But I can see that before that your commit [aa91650](https://github.com/devonfw/IDEasy/pull/1847/commits/aa91650ba3f0f0648b666856632ca03a3293ba37) was build green.
Sorry for the long road to get this PR merged. Hope you do not get frustrated. We are very close to the final merge and finishing line... Thanks for your patience. 👍

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
KarimALotfy: Thanks for the hint !

KarimALotfy:
@hohwille  the build now is successful after removing the failing test. It was not failing on my local since it is enabled only on Linux and Mac , and was actually unnecessary, since the -f flag doesn't have an effect on the link method, to keep existing functionality

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -556,7 +556,13 @@ public void link(Path source, Path link, boolean relative, PathLinkType type) {
             "Due to lack of permissions, Microsoft's mklink with junction had to be used to create a Symlink. See\n"
                 + "https://github.com/devonfw/IDEasy/blob/main/documentation/symlink.adoc for further details. Error was: "
                 + e.getMessage());
-        mklinkOnWindows(finalSource, absoluteSource, finalLink, type, relative);
+
+        try {
+          mklinkOnWindows(finalSource, absoluteSource, finalLink, type, relative);
+        } catch (IllegalStateException mkEx) {
+             LOG.info("Creating a hard link as a fallback for the failed mklink attempt.");
+            createHardLink(absoluteSource, finalLink, overrideHardLink);
```
hohwille: Are you not using IntelliJ to develop Java?
IDEasy preconfigures it to automatically format the code properly.
If you use something else (e.g. VSCode) feel free to change our settings to preconfigure it in the same way.
Code should always be formatted and properly aligned...
```suggestion
          LOG.info("Creating a hard link as a fallback for the failed mklink attempt.");
          createHardLink(absoluteSource, finalLink, overrideHardLink);
```

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -565,6 +571,26 @@ public void link(Path source, Path link, boolean relative, PathLinkType type) {
...
+   * Creates a hard link at {@code link} pointing to {@code source}.
+   *
+   * @param source the {@link Path} the hard link will point to.
+   * @param link the {@link Path} where to create the hard link.
+   * @param override - {@code true} to override existing links, {@code false} otherwise.
+   */
+  void createHardLink(Path source, Path link, boolean override) {
+    if (override && Files.exists(link, LinkOption.NOFOLLOW_LINKS)) {
+      delete(link);
+    }
```
hohwille: I would still expect this to be done also for symbolic links.
If we have an `override` flag added to `link(Path source, Path link, boolean relative, PathLinkType type, boolean override)` and JavaDoc says `{@code true} to override existing links, {@code false} otherwise.` then this looks inconsistent to me. Will `Files.createSymbolicLink` be idempotent? I guess not and mklink is not for sure.
So IMHO the implementation is not doing what you specified in the JavaDoc.

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,91 @@
...
+ * It tries to create a true symbolic link first. On Windows, symlink creation may be restricted due to missing privileges. In that case, IDEasy will create a
+ * hard link as an alternative (file-only, same volume) to avoid the Git-Bash behavior of silently copying files.
+ */
+public final class LnCommandlet extends Commandlet {
+
+  /** Grammar token {@code -s} (optional). */
+  public final FlagProperty symbolic;
+
+  /** The source path to link to. */
+  public final StringProperty source;
```
hohwille: Sorry that I did not see this on the first review but we have `PathProperty` for paths instead of `StringProperty`.
This is important for smart completion since otherwise we cannot suggest anything on tab-completion while if we use `PathProperty` we know that it should be an existing file or directory.
FYI: If you only accept files there is even `FileProperty` but here we also want to support directories.

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,91 @@
...
+  @Override
+  protected void doRun() {
+
+    Path cwd = this.context.getCwd();
+    if (cwd == null) {
+      throw new CliException("Missing current working directory!");
+    }
+
+    Path resolvedSource = cwd.resolve(this.source.getValue()).normalize();
+    Path linkPath = cwd.resolve(this.link.getValue()).normalize().toAbsolutePath();
```
hohwille: Why do we do `toAbsolutePath()` here?
Maybe I am missing something but otherwise if not required, I would skip it and let the link method handle to make paths absolute or relative...
```suggestion
    Path linkPath = cwd.resolve(this.link.getValue()).normalize();
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,91 @@
...
+      throw new CliException("Source does not exist: " + resolvedSource);
+    }
+
+    if (this.symbolic.isTrue()) {
+      Path target = Path.of(this.source.getValue());
+      boolean relative = !target.isAbsolute();
+      this.context.getFileAccess().link(target, linkPath, relative, PathLinkType.SYMBOLIC_LINK);
+    } else {
+      this.context.getFileAccess().link(resolvedSource, linkPath, false, PathLinkType.HARD_LINK, this.context.isForceMode());
+    }
```
hohwille: I would add another `FlagProperty` 
```
this.relative = add(new FlagProperty("--relative", false, "-r"));
```
See (`ln --help`).
And then make it simple and generic:
```suggestion
    PathLinkType linkType = PathLinkType.HARD_LINK;    
    if (this.symbolic.isTrue()) {
      linkType = PathLinkType.SYMBOLIC_LINK;
    }
    boolean relative = this.relative.isTrue();
    this.context.getFileAccess().link(resolvedSource, linkPath, relative, linkType, this.context.isForceMode());
```
p.s.: If that `toAbsolutePath()` (see above) is required in some special case, then I would do it like this:
```
  boolean relative = this.relative.isTrue();
  if (!relative) {
    linkPath = linkPath.toAbsolutePath();
  }
```

hohwille (approved):
@KarimALotfy thanks for your improvements. 👍 
Sorry, if I still was not clear enough but my vision is that we make the boolean parameters `override` and `relative` configurable via flags.
You did this but only partially and I would love to have it consistent.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
KarimALotfy: Thanks for the suggestions, implemented the suggested changes !

cli/src/main/java/com/devonfw/tools/ide/commandlet/LnCommandlet.java
```diff
@@ -0,0 +1,91 @@
...
+   */
+  public LnCommandlet(IdeContext context) {
+
+    super(context);
+    addKeyword(getName());
+
+    this.symbolic = add(new FlagProperty("--symbolic", false, "-s"));
+    this.relative = add(new FlagProperty("--relative", false, "-r"));
+    this.source = add(new PathProperty("", true, "source", true));
+    this.link = add(new PathProperty("", true, "link", true));
```
hohwille: This link is what is to be created and usually does not exist. Therefore `mustExist` flag should be `false` instead of `true` (when you test the commandlet via CLI otherwise the validation will fail if not exists and the commandlet will not be executed):
```suggestion
    this.link = add(new PathProperty("", true, "link", false));
```

hohwille:
@KarimALotfy I have already rested this PR on Windows and found but #1982 but that is not a bug of your implementation but a bug that was already there before.
Could you test and confirm the features on Linux (in WSL) and then set QA approved label?

hohwille:
rested also on Ubuntu: works like a charm.

PR #1859, opened 2026-04-27
#1392: Smart completions

### LINKED ISSUE #1392: Smart completion
### Feature idea

IDEasy has build in auto-compeltion working in bash and zsh what is great.
However, it would be awesome if the completion was more smart and even complete arguments of supported tools.

### Additional context

```
ide mv[Tab]
```
will complete to 
```
ide mvn 
```
what is perfect.
But 
```
ide mvn clea[Tab]
```
does not complete anything.
However, it could/should complete to
```
ide mvn clean
```
For maven the following keywords should be completed:
* clean
* package
* install
* deploy
* test
* verify
* validate
* compile
* dependency:tree
* dependency:list
* help:effective-settings
* -DskipTests
* exec:java
* -Dexec.mainClass=
* -Dexec.args=
* --also-make
* --also-make-dependents
* --fail-at-end
* --fail-fast
* -T1C
* -DdeployAtEnd=true
* ...

Surely some tools will change and evolve their CLI over time so this may not be the best approach for every tool since the completion can dependent on the version of the tool.
To get a perfect solution, we would need to delegate the completion to the according tool.
However, starting with maven looks promising since it does not support completion from the tool itself.
Only as external solution that also cannot handle all different versions, but the maven CLI is quite stable.
See https://github.com/juven/maven-bash-completion

So my suggestion would be to start with a PR that adds at least the above suggested completions for maven.
To reuse this later for other `LocalToolCommandlets` we should provide generic support in that class so `Mvn` sub-class only needs to extend or override something to e.g. provide the above keywords as String array.
For advanced use-cases later we can still add more complexity and consider that `-Dexec.mainClass` only makes sense in combination with `exec:java`.
Also be aware that those "keywords" above that end with `=` are not complete so they should not add a whitespace after its completion since the user needs to type things after like `-Dexec.mainClass=com.devonfw.tools.ide.cli.Ideasy`.

HenokLachmann:
### Implemented changes:

Auto-completion is extended for the `Mvn` commandlet. It enables autocompletion for the `ShellCommandlet` as well as for bash. Completion is supported for the last word as well as intermittent words on a commandline. Completion covers the following:

- Typical maven goals like `clean`, `install`, `package`
- Boolean flags like `--fail-fast`
- Definitions like `-Dexec.mainClass=<classname>`

Whitespace is excluded after a completed trailing `=` by differentiating partial from full completions. Merged short options are also supported as they are for present options.

Completion is based on `Property` instances, but properties are also used as stand-in replacements. Examples are the `repository` property for `RepositoryCommandlet` or the `commandlet` property for `HelpCommandlet`. These are not literal, they should not participate in autocompletion. A new boolean property `placeholder` is introduced to exclude them from completion.

A bug existed in the current implementation that would complete the `exit` subcommand infinitely. This is fixed by this PR by making `exit` an ordinary `KeywordProperty` and dynamically adding it when a shell session is started.

---

## Checklist for this PR

  with `internal`


tineff96 (commented):
Reviewed and tested.

- mvn clean test passed
- verified mvn CLI behavior and completion

The changes look solid and work as expected.

I’m still relatively new to the project, but everything seems correct from my side.

Looks good to me.

cli/src/main/java/com/devonfw/tools/ide/completion/CompletionCandidateCollector.java
```diff
@@ -12,13 +12,20 @@
...
+  void add(CompletionCandidate completion);
+
   /**
    * @param text the suggested word to add to auto-completion.
    * @param description the description of the suggestion candidate or {@code null} to determine automatically form the given parameters.
    * @param property the {@link Property} that triggered this suggestion.
    * @param commandlet the {@link Commandlet} owning the {@link Property}.
    */
-  void add(String text, String description, Property<?> property, Commandlet commandlet);
+  void add(String text, String description);
```
hohwille, line 28: IMHO you were mislead by IntelliJ linting that is sometimes stupid.
Because the implementation of this method is (currently) not using these 2 arguments does not automatically mean that we should remove them.
This was designed for the feature in the `ide shell` where we can use these parameters to automatically derive the localised help texts and display them together with the auto-completion suggestions in the tailtip widget.
However, this story was never completed when @jan-vcapgemini left our team.

cli/src/main/java/com/devonfw/tools/ide/property/Property.java
```diff
@@ -165,6 +178,10 @@ public boolean isMultiValued() {
     return this.multivalued;
   }
 
+  public boolean isPlaceholder() {
```
hohwille, line 181: Can you please add JavaDoc so we all know and understand what this means?
I can somehow conclude that this is for properties where the value can be autocompleted.

BTW: The better pattern to introduce such method is to better not add a `field` and a constructor argument.
If I am reading your change properly most sub-classes are not affected and never need the dynamic `field`.
Wouldn't it be better and easier to do the following instead:
```
/**
 * @return {@code true} if the property value is a placeholder that can be auto-completed.
 */
public boolean isPlaceholder() {
  return false;
}
```
Then you can override it in `MvnArgProperty` or other sublcasses typically like this:
```
@Override
public boolean isPlaceholder() {
  return true;
}
```

However, I am not yet convinced that we need this method at all.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/Mvn.java
```diff
@@ -70,8 +86,29 @@ public class Mvn extends LocalToolCommandlet {
   public Mvn(IdeContext context) {
 
     super(context, "mvn", Set.of(Tag.JAVA, Tag.BUILD));
+
+    this.alsoMake = this.add(new FlagProperty("--also-make", false, "-am"));
+    this.alsoMakeDependents = this.add(new FlagProperty("--also-make-dependents", false, "-amd"));
+    this.failAtEnd = this.add(new FlagProperty("--fail-at-end", false, "-fae"));
+    this.failFast = this.add(new FlagProperty("--fail-fast", false, "-ff"));
```
hohwille, line 93: This will not work. Our CLI parser is compatible with the POSIX standard.
Therefore `-ftb` is a short-cut for `-f -t -b`.
With this change the end-user will never be able to use these POSIX incompatible "short options" for maven any more or in other words:
Currently `ide mvn -fae install` is the same as `mvn -fae install` but after this change it will fail with with such error message what will totally confuse the end-user (since `mvn -f -a -e install` was actually called):
```
POM file -a specified with the -f/--file command line argument does not exist
```

cli/src/main/resources/nls/Help.properties
```diff
@@ -156,9 +156,17 @@ cmd.yarn=Tool commandlet for Yarn (NPM alternative, JavaScript package manager).
...
+opt.--also-make-dependents=Build projects that depend on projects on the list.
 opt.--batch=enable batch mode (non-interactive).
 opt.--code=clone given code repository containing a settings folder into workspaces so that settings can be committed alongside code changes.
 opt.--debug=enable debug logging.
+opt.--define\ deployAtEnd=Deploy all artifacts at the end of the multi-module build.
+opt.--define\ exec.args\==Arguments for the exec:java plugin.
+opt.--define\ exec.mainClass\==Main class for the exec:java plugin.
+opt.--define\ skipTests=Skip test compilation and execution.
+opt.--fail-at-end=Only fail the build at the end; allow all non-impacted builds to continue.
+opt.--fail-fast=Stop at first failure in reactorized builds.
```
hohwille, line 169: IMHO we should not start adding and maintaining all the options of all commandlets that we integrate.
This will open the door to hell.

cli/src/test/java/com/devonfw/tools/ide/cli/IdeasyTest.java
```diff
@@ -29,7 +29,7 @@ void testEnvOutsideProjectDoesNotLogCliExitException() {
     ideasy.run("--debug", "env");
 
     // assert
-    assertThat(context).logAtDebug().hasMessage("Step 'ide' ended with failure.");
+    assertThat(context).logAtDebug().hasMessage("Step 'ide' ended successfully.");
```
hohwille, line 32: This looks correct to me:
```
$ cd /
$ ide --debug env
Running commandlet EnvironmentCommandlet[env]
...
Step 'ide' ended successfully.
```

But why it this related to #1392?

I ran the test on main and got
```
The environment variable IDE_ROOT is undefined. Please reinstall IDEasy or manually repair IDE_ROOT variable.
Step 'ide' ended with failure.
```

Funny, so it seems you actually fixed something that was rather odd on `main`.
I am still curious to find out what change in this PR changed the IDE_ROOT error to go away.

cli/src/main/java/com/devonfw/tools/ide/commandlet/Commandlet.java
```diff
@@ -98,6 +98,28 @@ protected void addKeyword(String keyword) {
...
+  }
+
+  /**
+   * Create a new keyword property without an alias and set it as the first keyword property.
+   *
+   * @param keyword the string to create the property from
+   */
+  protected void setFirstKeyword(String keyword) {
+    this.setFirstKeyword(keyword, null);
+  }
```
hohwille, line 121: we are setting the firstKeyword automatically and implicitly by `addKeyword` method here:
https://github.com/devonfw/IDEasy/blob/3f2de9ec2f3a2905c8bb6639385603595b0d323a/cli/src/main/java/com/devonfw/tools/ide/commandlet/Commandlet.java#L105-L115

This guarantees that our properties are ordered and `firstKeyword` is also the first entry in our `propertiesList` that is an instanceof `KeywordProperty`.

When we use this method we are betraying that contract and IMHO cause a slightly inconsistent situation.
Isn't that odd and to be avoided?

hohwille (commented):
@HenokLachmann thanks for this PR and adding great improvements for auto-completion 👍 
However, during the review I found several problems where we will need some rework and partially also change the design.
Since you are not available any more, I suggest we will discuss with the team in the next daily how to proceed so that someone else from the team can take over so we can bring this to a merge-able state.

hohwille:
To lead this in the right direction:
My idea would be instead to start here:
https://github.com/devonfw/IDEasy/blob/63240371fefba87c3eca231e7ecd563939463f84/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java#L75

Simply create `ToolArgumentsProperty` extending `StringProperty` and use that here instead.
Then `ToolArgumentsProperty` should override this autocompletion:
https://github.com/devonfw/IDEasy/blob/63240371fefba87c3eca231e7ecd563939463f84/cli/src/main/java/com/devonfw/tools/ide/property/Property.java#L475-L477

And delegate to a new method to be created in (Tool)Commandlet (that you get as parameter `Commandlet commandlet`).
Then you can implement an AutoCompletionRegistry that is initially empty and that is initialized via a protected method that can be overridden. The auto-completion then completes according to what is in the registry. Since the registry is empty by default still nothing will happen.
Then `Mvn` can override the protected method to initialize the registry and add all the options and keywords for completion.

Caylipp:
This PR is no longer being continued here.
Work has been moved to: https://github.com/devonfw/IDEasy/pull/1999

PR #1860, opened 2026-04-27
#1685: Added Nest CLI to IDEasy commandlets

### LINKED ISSUE #1685: Add Nest CLI to IDEasy commandlets
### Feature idea

As a NestJs developer using IDEasy it would be nice if IDEasy could install [Nest CLI](https://docs.nestjs.com/cli/overview) during my project setup automatically.

### Additional context

_No response_

AdemZarrouki:
### Implemented changes:
* Added new nest commandlet support.
* Added URL updater for Nest CLI.
* Added tests for Nest in both cli and url-updater.
* Updated docs/license/changelog.

## Checklist for this PR

  with `internal`


satorus:
Installation works and the Code looks good!

The Log output after installation is a little bit confusing:

image

Apparently ide picks up the current node version as the version of nest, and says it overwrote a nest version with this install (which it did not, it is only a presentation bug).

Seems to be a general problem with NpmBasedCommandlets:
image

satorus (approved):

AdemZarrouki:
@satorus thanks for the review. I also noticed that i forgot to call NestUrlUpdater in updaters in UpdateManager. It should now be corrected.

You are correct regarding your comment about the npm version. It may be a bug that we need to check it and fix it.

satorus (approved):

hohwille:
> Apparently ide picks up the current node version as the version of nest, and says it overwrote a nest version with this install (which it did not, it is only a presentation bug).

@satorus thanks for testing and finding this problem.
Can you or @AdemZarrouki create a bug ticket for this problem?
We should not block this PR by that problem that already existed before by other NPM based commandlets.

cli/src/main/java/com/devonfw/tools/ide/tool/nest/Nest.java
```diff
@@ -0,0 +1,30 @@
...
+public class Nest extends NpmBasedCommandlet {
+
+  /**
+   * The constructor.
+   *
+   * @param context the {@link IdeContext}.
+   */
+  public Nest(IdeContext context) {
+
+    super(context, "nest", Set.of(Tag.NEST, Tag.TYPE_SCRIPT, Tag.BUILD));
```
hohwille: Since you correctly extended the `NEST` tag from `TYPE_SCRIPT` tag, you do not have to list it again here.
But surely nothing wrong with your code, just a nice to have improvement :)
```suggestion
    super(context, "nest", Set.of(Tag.NEST, Tag.BUILD));
```

cli/src/test/resources/ide-projects/nest/_ide/urls/nest/nest/dependencies.json
```diff
@@ -0,0 +1,8 @@
+{
+  "[0,)": [
+    {
+      "tool": "node",
```
hohwille: do we actually need this dependency?
If we have a `NpmBasedCommandlet` it is using `npm` to install the tool what itself already depends on `node`:
https://github.com/devonfw/ide-urls/blob/master/npm/npm/dependencies.json

Otherwise, if we would need this `dependencies.json` we would need a PR to `ide-urls` that has to be merged before this PR can be merged.

But my guess is that you can simply delete this file.

documentation/LICENSE.adoc
```diff
@@ -103,6 +103,7 @@ The column `inclusion` indicates the way the component is included:
 |https://docs.astral.sh/uv/[uv] |Optional|https://docs.astral.sh/uv/reference/policies/license/[Apache 2.0]
 |https://github.com/openjdk/jfx[OpenJFX] |Optional|https://github.com/openjdk/jfx/blob/master/LICENSE[GPLv2] (with the “Classpath” Exception)
 |https://squirrel-sql.sourceforge.io/[SQuirreL SQL Client]|Optional|https://github.com/squirrel-sql-client/squirrel-sql-stable-releases/blob/main/LICENSE[LGPL 2.1]
+|https://docs.nestjs.com/cli/overview[NestJS CLI] |Optional|https://github.com/nestjs/nest-cli/blob/master/LICENSE[License]
```
hohwille:
```suggestion
|https://docs.nestjs.com/cli/overview[NestJS CLI] |Optional|https://github.com/nestjs/nest-cli/blob/master/LICENSE[MIT License]
```

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/nest/NestUrlUpdater.java
```diff
@@ -0,0 +1,29 @@
+package com.devonfw.tools.ide.url.tool.nest;
+
+import com.devonfw.tools.ide.url.updater.NpmBasedUrlUpdater;
+
+/**
+ * {@link NpmBasedUrlUpdater} for nest.
+ */
+public class NestUrlUpdater extends NpmBasedUrlUpdater {
```
hohwille: For `NpmBasedCommandlet` we do not need any UrlUpdater.
IMHO this can be removed.
Since you already tested the commandlet you may have proven that it was working without any `nest` data in `urls` being present.

hohwille (approved):
@AdemZarrouki thank you very much for your PR. Look like solid and good work to me. 👍 

I added some review comments. Please have a look and address them.
Then we can merge this PR.

Sorry, if something things are obvious to me but unclear to (newer) team members.
We should discuss (e.g. in our retro) how to better prevent pointless work like creating an UrlUpdater when not needed.

cli/src/main/java/com/devonfw/tools/ide/tool/nest/Nest.java
AdemZarrouki: Yes, that makes sense. The proposed change has been implemented.

cli/src/test/resources/ide-projects/nest/_ide/urls/nest/nest/dependencies.json
AdemZarrouki: I tried removing the dependencies.json file, but during testing both the tests and the build failed. I therefore suggest keeping it. But why do we need a PR to ide-urls?

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/nest/NestUrlUpdater.java
AdemZarrouki: As discussed in our daily we don't need here an UrlUpdater as we are using `NpmBasedCommandlet`.

cli/src/test/resources/ide-projects/nest/_ide/urls/nest/nest/dependencies.json
hohwille: What I want to say is that our test projects (so e.g. `ide-projects/nest`) should try to replicate what we are doing in reality where easily suitable. So if we do not have `nest/nest/dependencies.json` in `ide-urls` repository, then why should we have it here? This means we are not representing the truth.
If we do that on purpose (mocking), this is fine.
But if we do it because we do not understand what is going on, we are not on the right track.
If the test fails without this file what gives us the courage to believe that it will work in the reality?
Maybe you are correct that we need this for some reason.
But then we need to understand the big picture.
Can you share the exception that you were getting if you remove it?
My expectation is that `Nest` extends `NpmBasedCommandlet` that has this:
https://github.com/devonfw/IDEasy/blob/5b8af56a767ef47b60c1126bf6daf5b839b0c1b7/cli/src/main/java/com/devonfw/tools/ide/tool/npm/NpmBasedCommandlet.java#L42-L45

Because of 
https://github.com/devonfw/IDEasy/blob/5b8af56a767ef47b60c1126bf6daf5b839b0c1b7/cli/src/main/java/com/devonfw/tools/ide/tool/PackageManagerBasedLocalToolCommandlet.java#L107

and this:
https://github.com/devonfw/IDEasy/blob/5b8af56a767ef47b60c1126bf6daf5b839b0c1b7/cli/src/main/java/com/devonfw/tools/ide/tool/PackageManagerBasedLocalToolCommandlet.java#L92-L96

We should actually call `install` method on `Npm` class independent of any `dependencies.json` file.
For that comment I also found this:
https://github.com/devonfw/IDEasy/blob/5b8af56a767ef47b60c1126bf6daf5b839b0c1b7/cli/src/main/java/com/devonfw/tools/ide/tool/node/Node.java#L40-L47

Still I do not see or understand why we need this.
Sorry in case you might find it annoying that I keep insisting on such things, but typically when I do I reveal some problem that we otherwise would have hidden and that most likely comes back later as a boomerang.
I you cannot figure this out easily, we can also arrange a short call where we debug this together in a screen-sharing session...

AdemZarrouki: Thanks @hohwille for the detailed explenation. As we discussed today in the daily to avoid this problem we introduced `isSkipInstallation` in `PackageManagerBasedLocalToolCommandlet` which will be called in `performToolInstallation`. The default value is `false`. But to to avoid infinite loops in case of cyclic dependencies between package manager based tools such as node and npm, we can override `isSkipInstallation` in `Npm.java` and do a `return true`. Thanks again for the explenation, the discussion and the solution you suggested.

hohwille (approved):
@AdemZarrouki great. Now everything is solved and ready for merge 👍

hohwille:
tested successfully on Windows with 2026.05.001-05_12_03-SNAPSHOT
```
$ ide nest
...
Usage: nest <command> [options]
...
```

Issue #1688, opened 2026-01-23
Remove unnecessary message in the CLI when installing a new tool


lubska:
### Feature idea

**Description**

When installing a new tool using the command ide install <tool_name>, the following message appears **before** the installation starts:
Couldn´t get installed version of <tool_name> 
After that, the tool is installed successfully and works as expected.

**Problem**

The message is misleading and unnecessary, as the installation proceeds normally. It negatively affects the CLI user experience.

**Tasks**
1. Verify **for which tools** the message appears:
- all tools
- or only specific ones
2. Fix the installation flow to remove this unnecessary message.
3. Ensure clear and correct CLI communication during installation.

**Expected Result**

1. No misleading messages are shown during the normal installation.
2. CLI output is clean and informative.
3. Tool installation behavior is approved and remains unchanged.
<img width="1864" height="720" alt="Image" src="https://github.com/user-attachments/assets/a459bfab-3ab7-451f-b352-cf1a6296d3af" />

jakozian:
This log message only appears for global tools, because there is no handling of `uninstall`, `get-version` and `get-edition` yet. A generalized concept of these methods is very difficult to achieve because each global tool gets installed differently and often in different places, specifically for all the different operating systems. So when implementing a new global tool, the mentioned methods have to be overwritten so the error log message does not appear. 
And in the case of pgadmin this wasn't done so the message appears as like a placeholder because the behavior is not implemented.

jakozian:
In this issue only the pgAdmin problems surrounding the cli message get resolved. The rest of functionality problems with pgAdmin are handled in #861

hohwille:
For pgadmin the red error log is now away but instead I get a yellow warning log:
```
Running command 'C:\Windows\System32\reg.exe' with arguments 'query' 'HKCU\Software\Microsoft\Windows\CurrentVersion\Uninstall\pgAdmin 4v9_is1' '/v' 'DisplayIcon'
failed with exit code 1!
ERROR: The system was unable to find the specified registry key or value.
```

PR #1862, opened 2026-04-27
#861/#1688: Add functionality to global tool commandlets for installation and uninstallation under windows and linux

### LINKED ISSUE #861: install of pgadmin throws an IllegalStateException when the install wizzard starts
### Expected behavior

As a pgadmin-User, I want to install the software without a errormessage 

### Actual behavior
```
$ ide pgadmin
Trying to download pgadmin-8.13-windows-x64.exe from https://ftp.postgresql.org/pub/pgadmin/pgadmin4/v8.13/windows/pgadmin4-8.13-x64.exe
Downloading 100% [===================================================================] 192/192MiB (0:00:38 / 0:00:00) 5.1MiB/s
Checksum fc7fa00b0b3b7cd60e40548bc0ade10c632d4a474a9d61a6e05694196f1a15c0 is correct.
Installation process for pgadmin in version 8.13 has started
An unexpected error occurred!
https://github.com/devonfw/IDEasy/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=IllegalStateException%3A+Running+command+%27pgadmin4%27+failed%3A+Cannot+run+program+%22pgadmin4%22%3A+CreateProcess+error%3D2%2C+The+system+cannot+find+the+file+specified
java.lang.IllegalStateException: Running command 'pgadmin4' failed: Cannot run program "pgadmin4": CreateProcess error=2, The system cannot find the file specified
        at com.devonfw.tools.ide.process.ProcessContextImpl.run(ProcessContextImpl.java:210)
        at com.devonfw.tools.ide.tool.ToolCommandlet.runTool(ToolCommandlet.java:169)
        at com.devonfw.tools.ide.tool.ToolCommandlet.runTool(ToolCommandlet.java:152)
        at com.devonfw.tools.ide.tool.ToolCommandlet.runTool(ToolCommandlet.java:139)
        at com.devonfw.tools.ide.tool.ToolCommandlet.run(ToolCommandlet.java:130)
        at com.devonfw.tools.ide.context.AbstractIdeContext.applyAndRun(AbstractIdeContext.java:867)
        at com.devonfw.tools.ide.context.AbstractIdeContext.run(AbstractIdeContext.java:798)
Caused by: java.io.IOException: Cannot run program "pgadmin4": CreateProcess error=2, The system cannot find the file specified
        at java.base@21.0.2/java.lang.ProcessBuilder.start(ProcessBuilder.java:1170)
        at com.devonfw.tools.ide.process.ProcessContextImpl.run(ProcessContextImpl.java:174)
        ... 10 more
Caused by: java.io.IOException: CreateProcess error=2, The system cannot find the file specified
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.jni.functions.JNIFunctions$NewObjectWithObjectArrayArgFunctionPointer.invoke(JNIFunctions.java)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.jni.functions.JNIFunctions.ThrowNew(JNIFunctions.java:907)
        at java.base@21.0.2/java.lang.ProcessImpl.create(Native Method)
        at java.base@21.0.2/java.lang.ProcessImpl.<init>(ProcessImpl.java:500)
        at java.base@21.0.2/java.lang.ProcessImpl.start(ProcessImpl.java:159)
        at java.base@21.0.2/java.lang.ProcessBuilder.start(ProcessBuilder.java:1126)
        ... 12 more


```

1. ide pgadmin (when not installed already)

### Affected version:

* OS: Windows
* Browser: /

jakozian:
### Implemented changes:

* Add generic super class implementation of `uninstall`, `get-version` and `get-edition` for global tools
* Extend WindowsHelper with registry search method for generic global tool logic
* Replace unnecessary log messages (specified in #1688)

(With this generic global tool logic the exception thrown while installing pgadmin is fixed #861)

---

## Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -172,9 +174,21 @@ protected ToolInstallation doInstall(ToolInstallRequest request) {
     } else {
       throw new CliException("Installation process for " + this.tool + " in version " + resolvedVersion + " failed with exit code " + exitCode + "!");
     }
-    installationPath = getInstallationPath(toolEdition.edition(), resolvedVersion);
+    if (this.context.isForceMode()) {
+      // has to be set to null so the already installed tool doesn't open
+      // during the installation process in force mode
+      installationPath = null;
```
MarvMa, line 180: Using null here as a control signal to prevent the tool from being started makes the flow harder to understand.
It might be safer and clearer to introduce an explicit flag (e.g. startTool / skipStart) or a small result type to model this state explicitly and avoid the usage of null as a flag.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
MarvMa, line 1: This filecurrently mixes several responsibilities (error handling, force‑mode behavior, user messaging, and installation result creation).
Splitting the implementation into smaller more focused methods could improve readability and make the control flow easier to understand and test.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -172,9 +174,21 @@ protected ToolInstallation doInstall(ToolInstallRequest request) {
...
+    } else {
+      installationPath = getInstallationPath(toolEdition.edition(), resolvedVersion);
+    }
     if (installationPath == null) {
-      LOG.warn("Could not find binary {} on PATH after installation.", getBinaryName());
+      if (request.isDirect()) {
+        LOG.warn("The tool {} is about to be installed. Please complete the installation and if required "
+            + "reboot your machine. Then run the command to start the tool.", this.tool);
+      } else {
+        throw new CliException("The tool " + this.tool + " is about to be installed. Please complete the installation and if required "
```
MarvMa, line 189: The exit code 2 is not clear here.
Introducing a named constant or enum (e.g. EXIT_USER_ACTION_REQUIRED) would make the intention clearer and improve maintainability/readability.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -187,18 +201,35 @@ protected List<PackageManagerCommand> getInstallPackageManagerCommands() {
...
+
+    if (this.context.getSystemInfo().isWindows()) {
+      WindowsHelper windowsHelper = WindowsHelper.get(this.context);
+      String displayVersion = windowsHelper.getRegistryValueBySearch(getWindowsAppName(), "DisplayVersion");
+      return VersionIdentifier.of(displayVersion);
+    } else if (this.context.getSystemInfo().isLinux()) {
+      String output = this.context.newProcess().runAndGetSingleOutput(getBinaryName(), "version");
+      return VersionIdentifier.of(output);
+    }
     return null;
```
MarvMa, line 223: According to our coding conventions, Optional should not be used everywhere to model nullable results.
Returning null here works, but the nullability should be made explicit, e.g. via JavaDoc or a @Nullable annotation, so callers clearly understand that “not installed” is a valid outcome and can handle it intentionally

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -187,18 +201,35 @@ protected List<PackageManagerCommand> getInstallPackageManagerCommands() {
...
   }
 
   @Override
   public String getInstalledEdition() {
-    //TODO: handle "get-edition <globaltool>"
-    LOG.error("Couldn't get installed edition of " + this.getName());
-    return null;
+
+    if (getInstalledVersion() == null) {
+      return null;
```
MarvMa, line 230: Same situation as the previous comment

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -216,9 +250,60 @@ protected Path getInstallationPath(String edition, VersionIdentifier resolvedVer
...
+
+  private Path getExecutableFolderFromWindowsRegistry() {
+
+    WindowsHelper windowsHelper = WindowsHelper.get(this.context);
+    String displayIcon = windowsHelper.getRegistryValueBySearch(getWindowsAppName(), "DisplayIcon");
+    if (displayIcon != null) {
+      Path executablePath = Paths.get(displayIcon);
+      if (Files.isExecutable(executablePath)) {
+        Path installationDir = executablePath.getParent();
+        this.context.getPath().setPath(getName(), installationDir);
```
MarvMa, line 297: This Line both resolves and sets the global PATH config, but the result of either getPath nor setPath is not beeing checked, is this the desired behaviour or would a nullcheck here useful?

MarvMa (approved):
You found a really smart way of uninstalling Tools via Win Registry, Nice Work 👍 I left some comments regarding readability and maintainability.

cli/src/main/java/com/devonfw/tools/ide/tool/pgadmin/PgAdmin.java
```diff
@@ -44,20 +48,26 @@ protected List<PackageManagerCommand> getInstallPackageManagerCommands() {
             + "> /etc/apt/sources.list.d/pgadmin4.list && apt update'", String.format(
             "sudo apt install -y --allow-downgrades pgadmin4=%1$s pgadmin4-server=%1$s pgadmin4-desktop=%1$s pgadmin4-web=%1$s",
             resolvedVersion)));
+    // TODO does not work for wsl
```
hohwille: Do not push such TODOs to GitHub.
I already know that we cannot test docker installation in Ubuntu on WSL.
A TODO always indicates that something in the code is currently not as it should be and needs to be changed/fixes.
Even in cases where this might be the case, it is better to create a GitHub issue than a TODO in the code.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -187,18 +201,35 @@ protected List<PackageManagerCommand> getInstallPackageManagerCommands() {
...
   public VersionIdentifier getInstalledVersion() {
-    //TODO: handle "get-version <globaltool>"
-    LOG.error("Couldn't get installed version of " + this.getName());
+
+    if (this.context.getSystemInfo().isWindows()) {
+      WindowsHelper windowsHelper = WindowsHelper.get(this.context);
+      String displayVersion = windowsHelper.getRegistryValueBySearch(getWindowsAppName(), "DisplayVersion");
+      return VersionIdentifier.of(displayVersion);
+    } else if (this.context.getSystemInfo().isLinux()) {
+      String output = this.context.newProcess().runAndGetSingleOutput(getBinaryName(), "version");
```
hohwille, line 220: This does not look like a reliable way to determine the version:
```
$ pgadmin4 version
bash: pgadmin4: command not found
$ docker version
Client:
 Version:           29.1.4-rd
 API version:       1.52
 Go version:        go1.25.5
 Git commit:        3c6914c
 Built:             Fri Jan  9 20:47:42 2026
 OS/Arch:           windows/amd64
 Context:           default

Server:
 Engine:
  Version:          29.1.3
  API version:      1.52 (minimum version 1.41)
  Go version:       go1.25.5
  Git commit:       fbf3ed25f893e6ce21336f1101590e40a13934f4
  Built:            Sun Dec 14 05:31:58 2025
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v2.2.0
  GitCommit:        1c4457e00facac03ce1d75f7b6777a7a851e5c41
 runc:
  Version:          1.4.0
  GitCommit:        8bd78a9977e604c4d5f67a7415d7b8b8c109cdc4
 docker-init:
  Version:          0.19.0
  GitCommit:
```

You would rather need to ask the package manager for the version.
This is not trivial to implement since e.g. on Ubuntu you would install with
```
sudo apt install pgadmin4-desktop
```
But to determine the version you would rather do:
```
dpkg -s pgadmin4-desktop | grep '^Version:'
```

I just noticed that on Linux we currently install and uninstall pgadmin4 in a very odd and crazy way.
According to https://www.pgadmin.org/download/pgadmin-4-apt/ we should just do `sudo apt install pgadmin4-desktop` to install it.

My vision was that if we have such reasonable installation commands, we could derive the command to determine the version automatically from that.
Obviously if we do such crazy stuff this does not seem to be possible.
Still my approach would be to implement the automatically generated command to determine the version on Linux in a protected method based on the `PackageManagerCommand`.
IMHO also the uninstallation could be automatically derived from the installation `PackageManagerCommand`.
Only in edge-cases, we could then override the protected method if the default does not work.

And how about Mac?
To me it seems the entire "world" of GlobalToolCommandlet is more in a messy alpha/beta state but something reasonable we crafted by design.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
hohwille: @MarvMa what is your problem here?
https://github.com/devonfw/IDEasy/blob/e8d3835aa9de12aab80f7c0a72be0119deed5d37/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java#L725-L728
and
https://github.com/devonfw/IDEasy/blob/e8d3835aa9de12aab80f7c0a72be0119deed5d37/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java#L738-L741

To me this is is fully fine here.
We might add a comment here explaining that we typically do not support multiple editions in GlobalToolCommandlet but if we do (like in Docker) this has to be overridden.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
hohwille: What are you missing?
https://github.com/devonfw/IDEasy/blob/e8d3835aa9de12aab80f7c0a72be0119deed5d37/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java#L725-L728

BTW: When we do a review we should typically comment on the changes of the PR.
A PR should solve the issue it implements and bring the codebase from some state to an improved state and not add new code-smells or bugs.
If you rather discover something bad in the existing code, you should typically create a new issue and not put that on the plate of the PR author. Only if we find a small odd thing that makes sense to be cleaned up in the same PR then kindly ask the PR author to include that.

cli/src/main/java/com/devonfw/tools/ide/os/WindowsHelperImpl.java
```diff
@@ -57,6 +65,33 @@ public String getUserEnvironmentValue(String key) {
     return getRegistryValue(HKCU_ENVIRONMENT, key);
   }
 
+  @Override
+  public String getRegistryValueBySearch(String displayNameRegex, String key) {
+
+    Pattern pattern = Pattern.compile(displayNameRegex, Pattern.CASE_INSENSITIVE);
+    for (String path : REGISTRY_PATHS) {
+      List<String> output = runReg("query", path, "/s");
```
hohwille, line 73: This is giving me the details of all installed software.
On my machine ~4000 lines of text.
That seems like a very inefficient way to solve this.
The idea of a registry query is to select only the data we really need.
A first improvement would be to run
```
reg query path /S /v DisplayName
```

BTW: it is also slightly confusing that you allow to pass any value for the 2nd parameter `key` but name the first parameter `displayNameRegex` so you actually assume that the `key` is `DisplayName`.
It can be a good idea to generalise a method to allow more flexibility.
On the other hand then you also need to implement it in a way that it solves all the potential cases properly.
I would recommend to focus on creating a method dedicated to determine the version of a product given the (prefix of) the DisplayName.

cli/src/main/java/com/devonfw/tools/ide/os/WindowsHelperImpl.java
```diff
@@ -65,8 +100,24 @@ public String getRegistryValue(String path, String key) {
     if (!result.isSuccessful()) {
       return null;
     }
-    List<String> out = result.getOut();
-    return retrieveRegString(key, out);
+    List<String> out = runReg("query", path, "/v", key);
+    if (out != null) {
+      return retrieveRegString(key, out);
+    }
+    return null;
```
hohwille, line 107: Great that you extracted the call of `reg` as process for reuse via method `runReg`.
By here you are now calling it twice:
1. in line 98/99 there is the leftover of the original call.
2. In line 103 you call it again.

You should remove lines 98-102 that do not make sense any more.

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -48,6 +52,20 @@ public String getBinaryName() {
...
+    if (installedEdition != null) {
+      if (installedEdition.equals(RANCHER_DESKTOP_EDITION)) {
+        return "Rancher Desktop";
+      } else if (installedEdition.equals(DOCKER_DESKTOP_EDITION)) {
+        return "Docker Desktop";
+      }
+    }
+
+    return super.getWindowsAppName();
+  }
```
hohwille, line 67: nice that you correctly have overriden this method.
However, this is never used since `getInstalledVersion()` from `GlobalToolCommandlet` is also entirely overridden.

I really like that you stated a generic implementation of `getVersion()` in `GlobalToolCommandlet`.
Maybe you need to extract a protected method with the default implementation for each operating system that can then be overridden individually.
Then the Windows implementation could be reused by all our global tools (PgAdmin and Docker).

hohwille (changes requested):
@jakozian thanks for your PR. I really appreciate that you found a solution how to find the version of any installed software on Windows from the registry. That is great progress. Also #1688 is easily solved by your PR. 👍 
However, after looking at all the details, I got the impression that we went the wrong way. This PR is getting more and more complex since you seem to have hit a pitfall.
GlobalToolCommandlet is a mess and that is not your fault. I was not aware of that and solving this is not an easy task.
Therefore I would go back and suggest the following steps as individual PRs:
1. PR fixing only #1688 by just removing the unwanted `LOG.error` messages you also removed here. IMHO that is all we need to implement that story (besides a CHANGELOG entry).
2. PR to improve our WindowsHelper with ability to determine installed version of an installed tool from Windows registry. Move your changes to `WindowsHelper*` into that PR and try to improve your implementation addressing my review concern. Maybe by asking copilot or some other AI you can find even a direct query method to search the key where DisplayName matches a given prefix.
3. When all this is merged, lets stepwise improve `GlobalToolCommandlet` and its sub-classes. Maybe we will even split this into multiple PRs e.g. per OS. The closer we look, the more we will find. I also just noticed, that when `podman` is installed, this will be used as an alias for `docker` what totally makes sense. But in such case the edition would rather be `podman` and the version is yet fully unclear. Also for Linux I see tons of improvements required since things are not working on non Debian based distributions, etc.
Mac is not yet supported by this at all.

Sorry, that I did not see this upfront. But if we keep trying to make this perfect in this single PR we will both not have fun. So lets go for divide & conquer what should always be the pattern to solve complex problems.

hohwille:
BTW: If you would put breakpoints into `getInstalledVersion()` and `getInstalledEdition` and run e.g. Docker we would notice that the same things are called multiple times what is kind of waste.
We could even think of a new method for `ToolCommandlet` like this:
```
ToolEditionAndVersion getInstalledEditionAndVersion();
```
Then the existing `getInstalledVersion()` and `getInstalledEdition` method could just delegate to that new method and we could determine both together avoiding to do the same potentially expensive computations redundantly...
So far just a thought that came to my mind. We also do not have to over-engineer it, in case my idea would not help to make things easier...

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
jakozian: You are absolutely right! Changing that...

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
jakozian: I agree with you on that but as @hohwille suggested this would not be part of this issue.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
jakozian: About such comments have i thought all the time because my implementation is merely a possible default implementation. E.g. docker shows that some tools can be handled without registry look ups (time consuming) and so on...

jakozian:
Closing due to too complex implementations.

PR #1876, opened 2026-04-29
#1869: Improve windows helper to search windows registry

### LINKED ISSUE #1869: Improve WindowsHelper to search in windows registry
### Feature idea

The [`WindowsHelper`](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/os/WindowsHelper.java) is limited to looking up a key in only one path. And you have to know in which path to look for. 
It would be more helpful to look up in various registry paths for a tool name (regex) and also offer methods for the most important keys like `DisplayVersion`, `UninstallString`, `DisplayIcon` and `DisplayName`.

### Additional context

Most important registry paths:
- HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall,
- HKLM\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall
- HKCU\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall
- HKCU\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall

jakozian:
### Implemented changes:

* Add methods to search for `DisplayVersion`, `DisplayIcon`, `InstallLocation` and `UninstallString` in common registry paths for an application.
* Add tests 

---

## Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/os/WindowsHelper.java
```diff
@@ -32,6 +32,30 @@ public interface WindowsHelper {
    */
   String getRegistryValue(String path, String key);
 
+  /**
+   * @param appName the application name to search for in the Windows registry.
+   * @return the DisplayName entry if the application is found in the Windows registry or {@code null} if nothing was found.
```
MarvMa, line 37: Nice addition of the lookup methods! The JavaDoc could be enriched with some more information regarding the search itself e.g. which paths are beeing searched and whether the matching is case insensitive, exact or substring match.

cli/src/main/java/com/devonfw/tools/ide/os/WindowsHelperImpl.java
```diff
@@ -60,13 +67,60 @@ public String getUserEnvironmentValue(String key) {
...
+  @Override
+  public String getInstallLocationFromRegistry(String appName) {
+    return getRegistryValueBySearch(appName, "InstallLocation");
+  }
+
+  private String getRegistryValueBySearch(String appName, String key) {
+
+    for (String registryBasePath : REGISTRY_BASE_PATHS) {
+      List<String> out = runReg("query", registryBasePath, "/s", "/f", appName);
+      if (out != null) {
```
MarvMa: Here it could be worth considering adding a check for an empty list. the `runReg()` method in the test returns `List.of()` for a miss, while the actual implementation returns `null`. Updating the return for a miss in the test-method or add an empty list check here could solve these different "miss" values.
```suggestion
      if (out != null && !out.isEmpty()) {
```

cli/src/test/java/com/devonfw/tools/ide/os/WindowsHelperImplTestable.java
```diff
@@ -0,0 +1,48 @@
...
+   *
+   * @param context the {@link IdeContext}.
+   */
+  public WindowsHelperImplTestable(IdeContext context) {
+
+    super(context);
+  }
+
+  @Override
+  protected List<String> runReg(String... args) {
```
MarvMa, line 27: Overwriting the `runReg` in the Test is a clean style. One thing I found is that the simulation logic appears to be identical in `WindowsHelperMock`. To reduce redundancy the `WindowsHelper` could delegate to the `WindowsHelperImplTestable` method.

cli/src/test/java/com/devonfw/tools/ide/os/WindowsHelperMock.java
```diff
@@ -51,4 +84,35 @@ public String getRegistryValue(String path, String key) {
...
+    String searchValue = null;
+    for (int i = 0; i < args.length - 1; i++) {
+      if ("/f".equalsIgnoreCase(args[i])) {
+        searchValue = args[i + 1];
+        break;
+      }
+    }
+
+    // Only return output if searched app matches
+    if (!"TestApp".equalsIgnoreCase(searchValue)) {
```
MarvMa: You could reuse the already existing MOCK_APP_NAME constant here.
```suggestion
    if (!MOCK_APP_NAME .equalsIgnoreCase(searchValue)) {
```

cli/src/test/java/com/devonfw/tools/ide/os/WindowsHelperImplTest.java
```diff
@@ -48,4 +52,47 @@ void testWindowsHelperParseEmptyRegStringReturnsNull() {
     assertThat(regString).isNull();
   }
 
+  /**
+   * Tests if correct keys can be found in registry output for app name filter.
+   */
+  @Test
+  void testRegistryLookupReturnsCorrectEntryIfFound() {
```
MarvMa, line 59: Really nice structured test which is easy to understand and the current grouping approach is perfectly readable.
Anyways maybe it is worth considering splitting this test in three separate Tests, in case of a failure the test name clarifies which registry search doesn't work immediatly

MarvMa (approved):
Thats a really easy to read PR following the KISS principle, which I personally appreciate a lot 👍 I left some comments regarding minor changes. I have been more critical since that has been the desired task. But the coments are just related to minor changes in code-style regarding the coding-conventions of IDEasy.

cli/src/test/java/com/devonfw/tools/ide/os/WindowsHelperImplTest.java
jakozian: In the beginning i considered this as well but decided to switch to this bundled approach. But it was suggested to me in a recent PR to consider not splitting every little thing into a new test. Considering build time.

cli/src/test/java/com/devonfw/tools/ide/os/WindowsHelperImplTestable.java
jakozian: I have done this purely out of coverage reasons. By using only the WindowsHelperMock `runReg` does not get coverage. Or what did you mean?

jakozian: I looked into it: I want coverage for most of the methods in `WindowsHelperImpl`. I cant achieve that if i am using `WindowsHelperMock` or `WindowsHelperImpl` directly, because:
1. `WindowsHelperMock` is not related to `WindowsHelperImpl`.
2. `WindowsHelperImpl` runs a real windows specific command in `runReg`.

So the workaround to get some coverage was to build a test seam `WindowsHelperImplTestable` which extends the `WindowsHelperImpl` and fakes `runReg`.

cli/src/main/java/com/devonfw/tools/ide/os/WindowsHelperImpl.java
```diff
@@ -60,13 +67,80 @@ public String getUserEnvironmentValue(String key) {
...
+  private String findUninstallKey(String appName) {
+
+    for (String registryBasePath : REGISTRY_BASE_PATHS) {
+      List<String> out = runReg("query", registryBasePath, "/s", "/f", appName);
+      if (out == null) {
+        continue;
+      }
+      for (String line : out) {
+        line = line.trim();
+        if (line.startsWith("HKEY_")) {
```
hohwille, line 119: if `registryBasePath` was `"HKCU\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall"` this will never be true? So it that entry in the constant array pointless then?
Should this rather be like this:

```suggestion
        if (line.startsWith(registryBasePath)) {
```

cli/src/main/java/com/devonfw/tools/ide/os/WindowsHelperImpl.java
```diff
@@ -60,13 +67,80 @@ public String getUserEnvironmentValue(String key) {
...
+    if (out != null) {
+      return retrieveRegString(key, out);
+    }
+    return null;
+  }
+
+  private String findUninstallKey(String appName) {
+
+    for (String registryBasePath : REGISTRY_BASE_PATHS) {
+      List<String> out = runReg("query", registryBasePath, "/s", "/f", appName);
```
hohwille, line 113: nice. This is a proper search that should ideally give us no hit or a single hit. So my concern from PR #1862 is perfectly addressed now. 👍 
Still it could happen that we get multiple registry key hits if e.g. a tool name is a prefix or substring of another tool name. That is not explicitly handled in the `out`  processing but seems to be an edge-case that we might not hit in the next future.

cli/src/main/java/com/devonfw/tools/ide/os/WindowsHelperImpl.java
```diff
@@ -21,6 +21,13 @@ public class WindowsHelperImpl implements WindowsHelper {
   /** Registry key for the users environment variables. */
   public static final String HKCU_ENVIRONMENT = "HKCU\\Environment";
 
+  /** Common Windows registry base paths containing (uninstall) information for installed applications (system-wide and per-user). */
+  private static final String[] REGISTRY_BASE_PATHS = {
+      "HKCU\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall",
+      "HKLM\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall",
+      "HKLM\\SOFTWARE\\WOW6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall"
```
hohwille: IMHO the common installation of global tools is `HKLM`. I would first search there to avoid pointless invocations of `reg query`.
```suggestion
      "HKLM\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall",
      "HKCU\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall",
      "HKLM\\SOFTWARE\\WOW6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall"
```

cli/src/test/java/com/devonfw/tools/ide/os/WindowsHelperMock.java
```diff
@@ -11,6 +12,18 @@ public class WindowsHelperMock extends WindowsHelperImpl {
...
+  private static final String MOCK_APP_NAME = "TestApp";
+
+  private static final String MOCK_DISPLAY_VERSION = "1.1.1";
+
+  private static final String MOCK_INSTALL_LOCATION = "C:\\Program Files\\TestApp";
+
+  private static final String MOCK_DISPLAY_ICON =
+      "C:\\Program Files\\TestApp\\testapp.exe,0";
+  private static final String MOCK_UNINSTALL_STRING =
+      "\"C:\\Program Files\\TestApp\\uninstall.exe\"";
```
hohwille, line 24: for mocking we should use a `Map` where mock data for tests can be registered instead of hard-coding random values.
Assuming you want to test PGAdmin now with mocked Windows on any OS then you could first register for the key `pgadmin` some object (Java record) containing the `displayIcon`, `appName`, and `displayVersion`, etc.

hohwille (approved):
@jakozian thanks for your PR. Nice improvement and isolation incl. testing of this feature. 👍 
I left some review comments for rework.
@MarvMa since you did the Team-Review and @jakozian already left, could you have a look at my review comments so we can get this completed together?

cli/src/main/java/com/devonfw/tools/ide/os/WindowsHelperImpl.java
jakozian: The output of `reg query <path> /s` looks something like this:
image

That means the `HKCU` of the path is written out in the output and in this case HK = HKEY and CU=Current User. For that reason i did not check for `line.startsWith(registryBasePath)`.
And the goal is to find the path to the tool and every path (doesn't matter which base path, LM or CU) begins with `HKEY_` so that seemed to be the right decision in my opinion.

hohwille: Thanks for clarification. You are completely right and I did not properly understand your code that is correct.

cli/src/test/java/com/devonfw/tools/ide/os/WindowsHelperMock.java
hohwille: OK, lets move forward: I created issue #1991 and therefore merge this PR as is since otherwise we always get stuck when someone started a PR and then is not available any more.
For the test-infrastructure we can accept some pragmatic hack for a while and will do a cleanup with #1991 later.

hohwille:
This PR is adding new functionality (API) that is planned to be used by GlobalCommandlet but not used on main.
JUnits are green and it works on Windows so I assign QA label.

PR #1878, opened 2026-04-30
#1695: Clone settings to temporary directory, analyse, and then move

### LINKED ISSUE #1695: Clone settings to temporary directory, analyse, and then move
### Feature idea

In the future, we should 
1. analyse the result: 

This has several advantages:

### Additional context

areinicke:
### Implemented changes:

* When creating a new project, the settings or code repository is first cloned to a temporary location (at `$IDE_ROOT/_ide/tmp/projects/<project name>`) where it is analyzed for validity. We check whether an ide.properties file exist either at the top level or within a settings folder at the top level. Only if this passes, do we create a new project at `$IDE_ROOT/<project name>` and move the files from the temporary location to the final location. The temporary folder is fully deleted after this process.
* Removed the --code option when creating a project. We now automatically determine if the repository is a code or settings repository and act accordingly.

---

## Checklist for this PR

  with `internal`#

Note: In some places, I used `getFileAccess()` to copy and subsequently delete files. I am aware that there is a move function. However, this was always failing for some reason


cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,69 @@ protected void doRun() {
...
 
+  /**
+   * This method is invoked when a new porject is created. It analyzes the cloned repository to check if it is a valid IDEasy repository. The repository can either be a settings repository (with ide.properties or devon.properties on the top level)
+   * or a code repository (with a settings folder on the top level containing such a file). Otherwise, the project creatio fails and an error message is logged.
+   */
+  private void analyze_project() {
+    // Settings repository: ide.properties on top levels (or devon.properties for legacy users)
+    // Code repository: settings folder on top level with ide.properties inside (or devon.properties for legacy users)
+    String projectName = this.context.getProjectName();
+    Path actualProjectPath = null;
```
satorus: the code conventions mention to never initialize non-static variables outside of constructors:
https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc#field-initialization

so probably just keep `Path actualProjectPath`

satorus:
```suggestion
    Path actualProjectPath;
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,69 @@ protected void doRun() {
...
+    FileAccess fileAccess = this.context.getFileAccess();
+
+    //Check if a file called ide.properties or devon.properties in settingsPath
+    Path SettingsPath = this.context.getSettingsPath();
+    if (Files.exists(SettingsPath.resolve(EnvironmentVariables.DEFAULT_PROPERTIES)) || Files.exists(SettingsPath.resolve(EnvironmentVariables.LEGACY_PROPERTIES))) {
```
satorus: We could think about extracting the conditions for this if-elseif to their own functions like `isSettingsRepository()`, which would help the readability and make the comments obsolete

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,69 @@ protected void doRun() {
...
+      // Repository is a settings repository: ide.properties on top levels (or devon.properties for legacy users)
+      LOG.info("The repository seems to be a settings repository based on the presence of " + EnvironmentVariables.DEFAULT_PROPERTIES + " or " + EnvironmentVariables.LEGACY_PROPERTIES + " on the top level.");
+      actualProjectPath = this.context.getIdeRoot();
+      moveProject(this.context.getIdeHome(), actualProjectPath);
+
+    } else if (Files.exists(SettingsPath.resolve("settings/" + EnvironmentVariables.DEFAULT_PROPERTIES)) || Files.exists(SettingsPath.resolve("settings/" + EnvironmentVariables.LEGACY_PROPERTIES))) {
```
satorus: Same here with something like `isCodeRepository()`

areinicke:
Thanks for the review. I have integrated the changes.

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,85 @@ protected void doRun() {
...
+   * This method is invoked when a new porject is created. It analyzes the cloned repository to check if it is a valid IDEasy repository. The repository can either be a settings repository (with ide.properties or devon.properties on the top level)
+   * or a code repository (with a settings folder on the top level containing such a file). Otherwise, the project creatio fails and an error message is logged.
+   */
+  private void analyze_project() {
+    // Settings repository: ide.properties on top levels (or devon.properties for legacy users)
+    // Code repository: settings folder on top level with ide.properties inside (or devon.properties for legacy users)
+    String projectName = this.context.getProjectName();
+    Path actualProjectPath;
+    FileAccess fileAccess = this.context.getFileAccess();
+    Path SettingsPath = this.context.getSettingsPath();
```
satorus: I know it is not from your changes, but it is triggering me anyway...maybe we can clean this up quickly here and name the variable `settingsPath` 😅

satorus (approved):

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,85 @@ protected void doRun() {
     createStartScripts();
   }
 
+  /**
+   * This method is invoked when a new porject is created. It analyzes the cloned repository to check if it is a valid IDEasy repository. The repository can either be a settings repository (with ide.properties or devon.properties on the top level)
+   * or a code repository (with a settings folder on the top level containing such a file). Otherwise, the project creatio fails and an error message is logged.
+   */
+  private void analyze_project() {
```
laim2003:
```suggestion
  private void analyzeProject() {
```

the coding conventions mention camel case, im not entirely sure but i would say that also applies to functions? 🙂 @satorus

laim2003 (changes requested):

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
areinicke: Thanks. I have updated the function name.

laim2003 (approved):

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,85 @@ protected void doRun() {
     createStartScripts();
   }
 
+  /**
+   * This method is invoked when a new porject is created. It analyzes the cloned repository to check if it is a valid IDEasy repository. The repository can either be a settings repository (with ide.properties or devon.properties on the top level)
+   * or a code repository (with a settings folder on the top level containing such a file). Otherwise, the project creatio fails and an error message is logged.
```
hohwille:
```suggestion
   * or a code repository (with a settings folder on the top level containing such a file). Otherwise, the project creation fails and an error message is logged.
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,85 @@ protected void doRun() {
...
+   */
+  private void moveProject(Path oldPath, Path newPath) {
+    FileAccess fileAccess = this.context.getFileAccess();
+    try {
+      fileAccess.copy(oldPath, newPath, FileCopyMode.COPY_TREE_OVERRIDE_FILES);
+      fileAccess.delete(oldPath);
+    } catch (Exception e) {
+      LOG.error("Failed to move project from {} to {}. Please move it manually.", oldPath, newPath, e);
+    }
+  }
```
hohwille: I am not getting why we move by creating a copy and then deleting the original source when we could use an atomic move operation. A clone code-repo can potentially be very big - this seems wasteful.

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,85 @@ protected void doRun() {
...
+  private boolean isSettingsRepository(Path repositoryPath) {
+    return Files.exists(repositoryPath.resolve(EnvironmentVariables.DEFAULT_PROPERTIES)) || Files.exists(repositoryPath.resolve(EnvironmentVariables.LEGACY_PROPERTIES));
+  }
+
+  /**
+   * Checks whether te given repository is a code repository by checking for the presence of ide.properties or devon.properties within a settings folder on the top level.
+   * @param repositoryPath - The path of the repository to be checked.
+   */
+  private boolean isCodeRepository(Path repositoryPath) {
+    return Files.exists(repositoryPath.resolve("settings").resolve(EnvironmentVariables.DEFAULT_PROPERTIES)) || Files.exists(repositoryPath.resolve("settings").resolve(EnvironmentVariables.LEGACY_PROPERTIES));
```
hohwille: Reuse and avoid redundancies:
```suggestion
    return isSettingsRepository(repositoryPath.resolve(IdeContext.FOLDER_SETTINGS));
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -99,6 +101,10 @@ protected void doRun() {
 
     if (!this.context.isSettingsRepositorySymlinkOrJunction() || this.context.isForceMode() || forcePull.isTrue()) {
       updateSettings();
+      // Check if instance of create commandlet. Only then will we analyze the project
+      if (this instanceof CreateCommandlet) {
+        analyzeProject();
+      }
```
hohwille: This design can be improved.
Option 1: you assume that you only need to analyse this when a project is initially created but not when updated.

Then simply override `updateSettings()` in `CreateCommandlet` and call `analyseProject()` after `super.updateSettings()`. Then only `CreateCommandlet` knows of `analyseProject()` (SoC).

Option 2: you need to analyse whenever you need to clone the settings.

Then this code is wrong since you skip this in case of `UpdateCommandlet`.
In such case you would need to do this after the cloning of the settings.

IMHO the option to ask for the settings in case of `ide update` is a kind of auto-repair feature of IDEasy.
Therefore `ide update` should behave also in the same way that it detects if we cloned an regular settings repo or a code repo with settings included.
But to get all this clean, we might need to revisit the design how create and update try to reuse the existing code/methods from `AbstractUpdateCommandlet`.
IMHO it would be better to do this on create:
* prepare a temp folder
* trigger the reusable functionality that clones the settings into a given project folder where here we provide the temporary folder. If it detected a code repo, it can move the settings to main workspace and proper folder name from the GitUrl and create the symlink to preserve a proper `settings` folder in the projects root dir.
* if that fails (neither regular settings repo nor code repo with settings found) we cleanup and abort.
* if that succeeded we can move the temp folder to the real project folder and continue with the regular logic where we re-initialize the IdeContext on the real project and do the full create/update process.

In case of `ide update` we can keep everything as is and reuse the same functionality described above in the 2nd point.

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,85 @@ protected void doRun() {
...
+    // Check whether the repository is a valid settings repository, code repository, or neither
+    if (isSettingsRepository(settingsPath)) {
+      LOG.info("The repository seems to be a settings repository based on the presence of " + EnvironmentVariables.DEFAULT_PROPERTIES + " or " + EnvironmentVariables.LEGACY_PROPERTIES + " on the top level.");
+      actualProjectPath = this.context.getIdeRoot();
+      moveProject(this.context.getIdeHome(), actualProjectPath);
+
+    } else if (isCodeRepository(settingsPath)) {
+      LOG.info(EnvironmentVariables.DEFAULT_PROPERTIES + " or " + EnvironmentVariables.LEGACY_PROPERTIES + " found in settings subfolder. This indicates a code repository with a settings folder on the top level.");
+      // Move settings folder contents containing code into workspace/main/<project_name>
+      actualProjectPath = this.context.getIdeRoot().resolve(projectName).resolve("workspaces/main/").resolve(projectName);
```
hohwille: The name should be derived from the git repo (see `GitUrl`)?
If I create `ide create cool-project https://github.com/foo/awesome-project.git`
I would expect `$IDE_ROOT/cool-project/workspaces/main/awesome-project/settings` and `$IDE_ROOT/cool-project/settings` being a symlink to it.

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,85 @@ protected void doRun() {
...
+      for (Path child : fileAccess.listChildren(settingsPath, f -> true)) {
+        System.out.println("Child: " + child);
```
hohwille: Please remove such debug output. Otherwise use trace-logging but some more reasonable message then.
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,85 @@ protected void doRun() {
...
+    if (isSettingsRepository(settingsPath)) {
+      LOG.info("The repository seems to be a settings repository based on the presence of " + EnvironmentVariables.DEFAULT_PROPERTIES + " or " + EnvironmentVariables.LEGACY_PROPERTIES + " on the top level.");
+      actualProjectPath = this.context.getIdeRoot();
+      moveProject(this.context.getIdeHome(), actualProjectPath);
+
+    } else if (isCodeRepository(settingsPath)) {
+      LOG.info(EnvironmentVariables.DEFAULT_PROPERTIES + " or " + EnvironmentVariables.LEGACY_PROPERTIES + " found in settings subfolder. This indicates a code repository with a settings folder on the top level.");
+      // Move settings folder contents containing code into workspace/main/<project_name>
+      actualProjectPath = this.context.getIdeRoot().resolve(projectName).resolve("workspaces/main/").resolve(projectName);
+      for (Path child : fileAccess.listChildren(settingsPath, f -> true)) {
```
hohwille: We have cloned a git repo. You should not just "copy and delete" every child.
Simply move (not copy) the entire `settings` folder to get this simple, efficient and robust.

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,85 @@ protected void doRun() {
...
+        System.out.println("Child: " + child);
+        moveProject(child, actualProjectPath);
+      }
+      // Move remaining folders into IDE_ROOT/<project_name>
+      actualProjectPath = this.context.getIdeRoot();
+      moveProject(this.context.getIdeHome(), actualProjectPath);
```
hohwille: This reveals the SoC topic: the logic assumes implementation details of `CreateCommandlet` and only works in this scenario but is implemented in `AbstractUpdateCommandlet`.
I would completely keep this out of this logic and only do it in `CreateCommandlet` after the settings have been properly cloned, analysed, and post-processed within the project folder (that may in such case be a temp folder but in case of `ide update` not).

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,85 @@ protected void doRun() {
...
+      // Delete empty settings folder in IDE_ROOT/<project_name> so we can create a symlink in the next step
+      fileAccess.delete(actualProjectPath.resolve(projectName).resolve("settings"));
```
hohwille: move it, then you do not have to delete it.
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,85 @@ protected void doRun() {
...
+      // Link settings folder in IDE_HOME to settings folder in code repository
+      fileAccess.symlink(actualProjectPath.resolve(projectName).resolve("workspaces/main").resolve(projectName).resolve("settings"), actualProjectPath.resolve(projectName).resolve("settings"));
+      // Final cleanup in temp location
+      fileAccess.delete(this.context.getIdeHome());
+
```
hohwille: IMHO again does not belong here due to SoC:

```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -108,6 +114,85 @@ protected void doRun() {
...
+    } else {
+      // Repository seems to be invalid. Clean up temporary location and return error
+      fileAccess.delete(this.context.getIdeHome());
+      throw new CliException("This repository does not include an " + EnvironmentVariables.DEFAULT_PROPERTIES + " or " + EnvironmentVariables.LEGACY_PROPERTIES + " file at the top level or a settings folder with such a file. "
+      + "The repository does not seem to be a valid IDEasy repository. Please verify the repository and try again.");
```
hohwille: dont delete the entire project.
You can either delete the settings here or IMHO even better would be to just return failure (e.g. `false` while otherwise `true` is returned for success) and let the calling logic handle the cleanup so in case of `ide update` we can just delete the `settings` and in case of `ide create` we can delete the temp project.

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -206,17 +290,10 @@ private GitUrl getOrAskSettingsUrl() {
...
-      defaultUrl = null;
-      LOG.info(MESSAGE_CODE_REPO_URL);
-    } else {
-      userPromt = "Settings URL [" + IdeContext.DEFAULT_SETTINGS_REPO_URL + "]:";
-      defaultUrl = IdeContext.DEFAULT_SETTINGS_REPO_URL;
-      LOG.info(MESSAGE_SETTINGS_REPO_URL, this.context.getSettingsPath());
-    }
+    String userPromt = "Repository URL [" + IdeContext.DEFAULT_SETTINGS_REPO_URL + "]:";
+    String defaultUrl = IdeContext.DEFAULT_SETTINGS_REPO_URL;
+    LOG.info(MESSAGE_SETTINGS_REPO_URL, this.context.getSettingsPath());
```
hohwille, line 215: fine but you then also need to include an update to the documentation that otherwise gets inconsistent:
https://github.com/devonfw/IDEasy/blob/main/documentation/settings.adoc#code-repository

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -84,15 +79,10 @@ private void initializeProject(Path newInstancePath) {
...
-  @Override
-  protected boolean isCodeRepository() {
-    return this.codeRepositoryFlag.isTrue();
-  }
-
   @Override
   protected String getStepMessage() {
 
-    return "Create (clone) " + (isCodeRepository() ? "code" : "settings") + " repository";
+    return "Creating (Cloning) repository";
```
hohwille: For steps we always use infinitive (e.g. to add prefixes in log message like starting step ..., step ... failed, or even failed to ...)

```suggestion
    return "Create (clone) repository";
```

hohwille (changes requested):
@areinicke thanks for your PR. Great that you implemented a test for it and finally started improving `GitContextMock` what I suggested long time ago when `GitContextMock` was supposed to be implemented. 👍 
This is not a very easy task. Seems in such cases it makes sense to start with a design session before doing the complete PR implementation in the future...
However, should still not be a big change since most of your PR is already perfectly correct.

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
areinicke: This only deletes the temporary location if the verification fails. Why shouldn't that one be deleted if it is invalid anyway?

hohwille: 1. `this.context.getIdeHome()` usually gives me my project home e.g. `~/projects/IDEasy`.
2. This code is used for `ide create` and for `ide update`. You are assuming that this will refer to some temporary directory since you are thinking of your changes in `ide create` but for `ide update` your assumption is wrong.
3. If the project become "invalid" this can have many reasons. Maybe by accident the user deleted some file or folder. `fileAccess.delete` is a destructive operation. If you as a user will run into the situation that your entire `~/projects/IDEasy` folder gets deleted and all your uncomitted work is gone because IDEasy due to some mistake thought that something is "invalid" you might not like the result ;)

Believe me - we already had all such situations and I do not want to go there again.
If we "delete" user content like entire IDE projects or workspaces and we have a real reason to do so, we should use `fileAccess.backup` since then the data is still there and the user could somehow recover his work in case of an accident.
However, I am still convinced that this is wrong. Keep your clenup of temporary directory only in `CreateCommandlet` where you created it and can ensure you do not cause severe harm.

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -107,6 +106,8 @@ protected void doRun() {
     createStartScripts();
   }
 
+
+
```
hohwille, line 110:
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -57,16 +58,15 @@ protected void doRun() {
 
     String newProjectName = this.newProject.getValue();
     Path newProjectPath = this.context.getIdeRoot().resolve(newProjectName);
+    Path tempProjectPath = this.context.getIdeRoot().resolve("_ide/tmp/projects").resolve(newProjectName);
```
hohwille, line 61:
```suggestion
    Path tempProjectPath = this.context.getTempPath().resolve(IdeContext.FOLDER_PROJECTS).resolve(newProjectName);
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -57,16 +58,15 @@ protected void doRun() {
...
 
     LOG.info("Creating new IDEasy project in {}", newProjectPath);
     if (!this.context.getFileAccess().isEmptyDir(newProjectPath)) {
```
hohwille, line 64: IMHO we should also check the final target project folder for existence.
If we call `ide create IDEasy -` but `$IDE_ROOT/IDEasy` already exists, do we want to create a new IDEasy project in tmp folder and then populate everything in order to then fail when we want to move the project?
If the target project already exists, we should not create at all and fail. The user can then call `ide update` on the target project or he can do `rm -rf «project»` if the project is totally broken and he wants to recreate it.

cli/src/test/java/com/devonfw/tools/ide/commandlet/CreateCommandletTest.java
```diff
@@ -217,9 +170,38 @@ void testWelcomeMessageDisplayed() {
...
+    // act - run the create command
+    assertThatThrownBy(() -> cc.run())
+        .isInstanceOf(CliException.class)
+        .hasMessageContaining("This repository does not include an " + EnvironmentVariables.DEFAULT_PROPERTIES + " or " + EnvironmentVariables.LEGACY_PROPERTIES + " file at the top level or a settings folder with such a file.")
+        .hasMessageContaining("The repository does not seem to be a valid IDEasy repository. Please verify the repository and try again.");
+
+    // assert
+    Path newProjectPath = context.getIdeRoot().resolve(NEW_PROJECT_NAME);
+    assertThat(newProjectPath).doesNotExist();
+    assertThat(context.getIdeRoot().resolve("_ide/tmp/projects").resolve(NEW_PROJECT_NAME)).doesNotExist();
```
hohwille, line 202:
```suggestion
    assertThat(context.getTempPath().resolve(IdeContext.FOLDER_PROJECTS).resolve(NEW_PROJECT_NAME)).doesNotExist();
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
hohwille: This point is still not addressed. The developer taking over this PR should keep the project renaming (move) locally inside the `CreateCommandlet` and not do this in this class that then makes assumptions about very specific implementation details of `CreateCommandlet` that do not make any sense for `UpdateCommandlet` that also extends from this class.

hohwille:
@laim2003 what is the state of this? Can you finally complete this PR or hand it over to someone else completing it.
This PR was created 1/4 year ago and it should be completed and finally merged.

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
laim2003: > Then simply override updateSettings() in CreateCommandlet and call analyseProject() after super.updateSettings(). Then only CreateCommandlet knows of analyseProject() (SoC).

From what I can tell this point has already been implemented. Regarding the other points i'm still looking into it.

laim2003: Yes, Option 1 has already been implemented by @areinicke

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
laim2003: This recommendation has also already been implemented

laim2003:
The new PR implementing these changes can be found under: #2249

PR #1888, opened 2026-05-04
#1849: Add support for VSCodium

### LINKED ISSUE #1849: Support VSCodium
### Feature idea

With original VSCode from Microsoft there are some problems/bugs esp. #797 so VSCode does not work on Mac (x86).

An alternative tool is [VSCodium](https://vscodium.com/) that is an Open-Source edition/fork of VSCode.
Ideally we can support this as an `edition` of `vscode` in IDEasy.
However, there will be some challenges:
* The structure of `VSCodium` is slightly different to `VSCode`. So they even have different binary names. But maybe we could handle that in `getBinaryName()` of `VsCode` commandlet and add an if condition checking if the edition is `vscodium`.
* We need to check if the versions between VSCode and VSCodium are kind of syntax compatible.

### Additional context

In the end users should be able to switch via `VSCODE_EDITION=vscodium` but still be able to run `ide vscode` to install and run the tool.

laert-ll:
### Implemented changes:

* Added VSCodium as alternative edition to VSCode
* Adjusted plugin installation error logging due to mismatch between VSCode & VSCodium plugins
* Added test cases for implemented changes

---

## Checklist for this PR

  with `internal`


AdemZarrouki:
I think we should add VSCodium to the table of tools in [LICENSE.asciidoc](https://github.com/devonfw/IDEasy/blob/main/documentation/LICENSE.adoc) as it has a different license (see https://github.com/VSCodium/vscodium/blob/master/LICENSE).

Also we need to add the tool to both Help.properties and Help_de.properties

AdemZarrouki (commented):
I would suggest moving vscodium to a seperate folder and not puting it together with vscode to improve readability but its not a must. Maybe @hohwille can suggest if this is needed here or not. Otherwise, everything is fine from my point of view.
Nice job @laert-ll

AdemZarrouki:
Also i think we are missing tests for the VSCodium. Maybe something like https://github.com/devonfw/IDEasy/blob/main/cli/src/test/java/com/devonfw/tools/ide/tool/vscode/VscodeTest.java

AdemZarrouki (approved):

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -55,7 +71,14 @@ protected void installPlugins(Collection<ToolPluginDescriptor> plugins, ProcessC
...
       pb.close();
-    });
+    }, suppressLevel);
+    if (isVscodium && !this.vscodiumUnavailablePlugins.isEmpty()) {
+      LOG.warn("{} plugin(s) could not be installed on VSCodium due to not being available on open-vsx or other errors:\n  - {}\n"
+              + "For full plugin support, set VSCODE_EDITION=vscode to use Microsoft's distribution.\n"
+              + "For more detailed information on why plugins failed to install, check the IDEasy logfile at $IDE_ROOT/_ide/logs/.",
+          this.vscodiumUnavailablePlugins.size(),
+          String.join("\n  - ", this.vscodiumUnavailablePlugins));
+    }
```
laert-ll, line 83: Issue: VSCodium does not support all the plugins that VSCode does.

I figured it would be better to suppress the very large red error messages that normally come when not being able to install the plugins, since it gives the impression that this is unexpected behavior. This is a temporary workaround for potentially a future story where we would have a separate list of plugins that are compatible with VSCodium.

url-updater/src/main/java/com/devonfw/tools/ide/url/tool/vscode/VsCodiumUrlUpdater.java
```diff
@@ -0,0 +1,77 @@
+package com.devonfw.tools.ide.url.tool.vscode;
+
+import com.devonfw.tools.ide.url.model.folder.UrlVersion;
+import com.devonfw.tools.ide.url.updater.GithubUrlTagUpdater;
+
+/**
+ * {@link GithubUrlTagUpdater} for the "vscodium" edition of vscode (<a href="https://vscodium.com/">VSCodium</a>).
+ */
+public class VsCodiumUrlUpdater extends GithubUrlTagUpdater {
```
hohwille: There are GitHub releases available for VSCodium:
https://github.com/VSCodium/vscodium/releases/

As we already discussed in such case `GithubUrlReleaseUpdater` should be preferred.
See also #1949

hohwille (approved):
@laert-ll thanks for your PR to support and integrate VSCodium. This looks pretty much complete with URLUpdater, Commandlet, Tests, Help, etc. 👍 
Great that you also added special support for incompatible plugins.
I hope the statefulness (adding this `List<String>` as field `vscodiumUnavailablePlugins`) will not cause problems in the future but for now it should work fine and I could not find any idea how to solve this stateless.

PR #1979, opened 2026-05-27
#1936: Improve localization of gui

### LINKED ISSUE #1936: Improve localization of the GUI
### Feature idea

Currently, all the GUI values and labels are hardcoded. As a user, i might want to use a localized version of the GUI with the correct translation.

### Additional context

when using FXML files for the gui, localization is quite simple, as you can just pass the ResourceBundle with the correct localization when loading an FXML file via
```
FXMLLoader loader = new FXMLLoader(
    getClass().getResource("view.fxml"),
    localizedResourceBundle
);
```
Example FXML file: 
```
<Label text="%username"/>
<Button text="%login"/>
```

When wanting to access localization in a views controller file, JavaFX even automatically injects the correct ResourceBundle:
```
public class LoginController {

    @FXML
    private ResourceBundle resources;

    @FXML
    public void initialize() {
        System.out.println(resources.getString("login"));
    }
}
```

KarimALotfy:
### Implemented changes:

Added localization support to the JavaFX GUI — users can select a display language, and the preference is persisted across restarts.

### How it works

**Locale resolution** — on startup, `NlsService`  resolves the active locale in order of priority: persisted user preference → system default → eg: English. The persisted value is stored in `IDE_OPTIONS` in `~/.ide/ide.properties`.

**Bundle loading** — resource bundles live under `src/main/resources/nls/` as UTF-8 `.properties` files (`messages.properties` for English, `messages_<tag>.properties` for other languages). Bundles are loaded with a custom `UTF8Control` to correctly handle non-ASCII characters.

**Locale discovery** — available locales are detected dynamically at runtime by scanning the classpath for `nls/messages_*.properties` files, covering both exploded directories and JARs — dropping a new bundle file with `CurrentLanguage` property is enough to make a language appear.

**UI integration** — the language combo box is populated from the discovered locales. Each entry is labelled using the `CurrentLanguage` property  from that locale's bundle, with a fallback to a generated display name. Selecting a language applies it immediately across the UI by calling `setLocale` and persists it to `~/.ide/ide.properties`.

### Adding a new language

Add `src/main/resources/nls/messages_<tag>.properties` with the full key set matching `messages.properties`. The language will appear in the combo box automatically on the next build.

----------------
### Testing instructions

1. Launch the app through `AppLauncher.java`  — the language combo box should list all bundled languages
2. Switch to a non-English language — all UI labels should update immediately
3. Close and reopen the app — the selected language should be restored
4. Add a new `messages_<tag>.properties` file with a `CurrentLanguage` property (ex: `CurrentLanguage=French (fr))`, rebuild, and launch — the new language should appear in the combo box without any code change.

---

### Checklist for this PR

  with `internal`


gui/src/main/java/com/devonfw/ide/gui/i18n/I18nService.java
```diff
@@ -0,0 +1,140 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Service for managing localization (i18n) in the JavaFX GUI.
+ */
+public class I18nService {
+
+  private static final Logger LOG = LoggerFactory.getLogger(I18nService.class);
+
+  private static final String BUNDLE_NAME = "i18n.messages";
```
laim2003: i would say that naming this something like "localization" would probably be a bit more readable (altough i understand where you are coming from with i18n)

At least that what I would think coming from an android development background, where this is how they name it😅

gui/src/main/java/com/devonfw/ide/gui/i18n/I18nService.java
```diff
@@ -0,0 +1,140 @@
...
+    return instance;
+  }
+
+  public static synchronized I18nService getInstance() {
+
+    if (instance == null) {
+      return getInstance(null);
+    }
+    return instance;
+  }
```
laim2003: @hohwille said that we should avoid Singleton patterns where possible and use Dependency Injection. I know i used this pattern a lot in my own state management implementation, but I will also be moving away from Singeltons there. Maybe we can discuss about whether its okay to use them in some instances in the Next daily, probably would also help me clear up some questions. I would probably leave this how it is here for now, but then open a second issue for the future where we update this to DI.

gui/src/main/java/com/devonfw/ide/gui/i18n/I18nService.java
```diff
@@ -0,0 +1,140 @@
...
+    return this.resourceBundle;
+  }
+
+  public String get(String key) {
+
+    try {
+      return this.resourceBundle.getString(key);
+    } catch (MissingResourceException e) {
+      LOG.warn("Missing translation key: {} for locale: {}", key, this.locale);
+      return MISSING_KEY_PREFIX + key + MISSING_KEY_SUFFIX;
```
laim2003: Im not sure what the use of the MISSING_KEY_PREFIX/SUFFIX constants is, since they both have the same value and seem to be only used in this method? Wouldn't it be just enough to concatenate the string here?

```suggestion
      return "!" + key + "!";
```

gui/src/main/java/com/devonfw/ide/gui/i18n/I18nService.java
```diff
@@ -0,0 +1,140 @@
...
+public class I18nService {
+
+  private static final Logger LOG = LoggerFactory.getLogger(I18nService.class);
+
+  private static final String BUNDLE_NAME = "i18n.messages";
+
+  private static final String MISSING_KEY_PREFIX = "!";
+
+  private static final String MISSING_KEY_SUFFIX = "!";
+
```
laim2003: See this [comment](https://github.com/devonfw/IDEasy/pull/1979/changes#r3421162273)
```suggestion
```

gui/src/main/java/com/devonfw/ide/gui/localization/LocalizationService.java
```diff
@@ -0,0 +1,140 @@
...
+
+    return this.resourceBundle;
+  }
+
+  public String get(String key) {
+
+    try {
+      return this.resourceBundle.getString(key);
+    } catch (MissingResourceException e) {
+      LOG.warn("Missing translation key: {} for locale: {}", key, this.locale);
```
laim2003: I think if no string is found, this probably should be an error rather than a warning. Maybe you can help me here, basically if the string is not found in ANY translation, this error will occur? Or does it already occur if the key is simply not found in the specified Locale?

laim2003 (changes requested):
looks good! just left some thoughts😊

gui/src/main/java/com/devonfw/ide/gui/i18n/I18nService.java
KarimALotfy: I agree with you, it would be clearer to be named as "Localization", done ✅

gui/src/main/java/com/devonfw/ide/gui/i18n/I18nService.java
KarimALotfy: You're totally right ✅

gui/src/main/java/com/devonfw/ide/gui/localization/LocalizationService.java
KarimALotfy: I noticed actually that if a resource key is missing, the application fails to start. This is because `fxmlLoader.load()` in the application's `start` method throws an exception before the custom exception in the `get` method is reached.

Resulting error:
```
Failed to load the application: Resource "button.open" not found
/C:/.../main-view.fxml
```

gui/src/main/resources/nls/messages.properties
```diff
@@ -0,0 +1,17 @@
+# IDEasy GUI - English (default) Translation File
```
hohwille, line 1: In `cli` we named the folder `nls` for "Native Language Support".
For homogeneity, it would be nice to have both folder named the same.
If the team better understands `localization` we can also change `nls` but it is short and maybe more speaking than `l10n` what other projects use as shortcut for `localization`.

gui/src/main/resources/localization/messages.properties
```diff
@@ -0,0 +1,17 @@
+# IDEasy GUI - English (default) Translation File
+
+# Navigation Panel
+label.project=Project
```
hohwille: does it really make sense to specify the technical aspect such as label/button/prompt here?
If we have the term say "Open" or "Save" and want to translate it to e.g. German ("Öffnen" oder "Speichern") - wouldn't we then end up with lots of duplications like `label.open` and `button.open` and `menu-item.open`, etc.?

gui/src/test/java/com/devonfw/ide/gui/localization/LocalizationServiceTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    assertThat(service.getLocale()).isEqualTo(englishLocale);
+    assertThat(service.getResourceBundle()).isNotNull();
+  }
+
+  @Test
+  public void testGetInstanceSingleton() {
+
+    LocalizationService service1 = LocalizationService.getInstance(Locale.ENGLISH);
+    LocalizationService service2 = LocalizationService.getInstance();
```
hohwille: Better do not rely on default locale in JUnits.
If the user has a different locale configured than English, the test will fail.
This will happen sooner or later and cause confusion if some developers have red tests and others have green tests.

gui/src/test/java/com/devonfw/ide/gui/nls/NlsServiceTest.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  @Test
+  public void testSetLocale() {
+
+    LocalizationService service = LocalizationService.getInstance(Locale.ENGLISH);
+    service.setLocale(Locale.GERMAN);
+
+    assertThat(service.getLocale().getLanguage()).isEqualTo("de");
+    assertThat(service.getResourceBundle()).isNotNull();
```
hohwille, line 77: I would also expect that we lookup some key for German and assert the expected translation.
If you choose a generic key that will not likely change, the test should also stay stable.

gui/src/test/java/com/devonfw/ide/gui/nls/NlsServiceTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+  @BeforeEach
+  public void setUp() {
+
+    // Reset the singleton instance before each test
+    LocalizationService.resetInstance();
+    // Preload English bundle keys used by bundle-completion tests
+    ResourceBundle englishBundle = ResourceBundle.getBundle("localization.messages", Locale.ENGLISH);
+    this.englishKeys = extractKeys(englishBundle);
+  }
```
hohwille, line 57: Only needed for testing the German keys match the English keys.
I usually expect that setup method does some generic setup used for all test methods.
Nothing wrong here but you could simply do a single test that checks that English keys and German keys are identical and then move this from setup to that single test method.
Surely nice to have...

gui/src/main/java/com/devonfw/ide/gui/MainController.java
```diff
@@ -60,10 +73,62 @@ public MainController(String directoryPath) {
...
+  @FXML
+  private void dispose() {
+    LocalizationService.getInstance().removeLocaleChangeListener(this::updateTexts);
+  }
+
+  private void initLanguageComboBox() {
+
+    // Initialize language choices
+    selectedLanguage.getItems().clear();
+    selectedLanguage.getItems().addAll("English", "Deutsch");
```
hohwille: we should not hardcode the lanugages.
Instead simply add a key like `CurrentLanguage` so e.g.
```
CurrentLanguage=English (en)
```

or 
```
CurrentLanguage=Deutsch (de)
```

Then we can dynamically lookup the available resource bundles and add new languages without changing the code.

gui/src/main/java/com/devonfw/ide/gui/MainController.java
```diff
@@ -60,10 +73,62 @@ public MainController(String directoryPath) {
...
+    // Set ComboBox prompts
+    selectedProject.setPromptText(localizationService.get("prompt.chooseProject"));
+    selectedWorkspace.setPromptText(localizationService.get("prompt.chooseWorkspace"));
+    selectedLanguage.setPromptText(localizationService.get("prompt.chooseLanguage"));
+
+    // Set Button texts
+    androidStudioOpen.setText(localizationService.get("button.open"));
+    eclipseOpen.setText(localizationService.get("button.open"));
+    intellijOpen.setText(localizationService.get("button.open"));
+    vsCodeOpen.setText(localizationService.get("button.open"));
```
hohwille: This also looks unmaintainable.
We cannot hard-code all labels, buttons, etc. of all dialogues here.

KISS solution: To change the language, the GUI has to be restarted.
Otherwise research how to tell JavaFx to reload the messages (or the entire FXML).

gui/src/main/java/com/devonfw/ide/gui/nls/NlsService.java
```diff
@@ -0,0 +1,136 @@
...
+    loadBundle();
+    LOG.info("Locale set to: {}", locale.getLanguage());
+    for (Runnable listener : this.localeListeners) {
+      try {
+        listener.run();
+      } catch (Exception e) {
+        LOG.warn("Locale change listener threw: {}", e.getMessage());
+      }
+    }
+  }
```
hohwille, line 127: usually the language is set on operating system level and then Java takes it from there via the default locale. Would be sufficient for KISS. Custom language for IDEasy can IMHO be set via
```
export IDE_OPTIONS=-Duser.lang=de
```
in `~/.ide/ide.properties` what should already work for the CLI0.
It is nice that you add the option to change in individually in the GUI.
But if you do this, we have to persist the value so that if I close the GUI and reopen it, I do not have to change it every time again.

FYI: If you want to go for KISS, you may keep this out of this first PR and we will add that later. But when we already introduce it, we should IMHO do it properly.

hohwille (changes requested):
@KarimALotfy thanks for your PR.
Very nice that you figured out how to do NLS (i18n and l10n) with JavaFx (using `Locale`, `ResourceBundle` and the `%` syntax in fxml files). Great that you also added a proper JUnit test and some advanced features. 👍 
I left quite some comments for rework.
In some cases you may decide to go for more KISS and keep some advanced features out of this first PR and introduce in a later improving PR after this one is merged or you should go for the full solution if you already start it.

gui/src/main/java/com/devonfw/ide/gui/MainController.java
hohwille: BTW: Adding the ISO-Codes in parenthesis improves UX if you see languages that you cannot read but still want to get an idea of. Example
```
แบบไทย (th)
```

gui/src/main/resources/nls/messages.properties
KarimALotfy: So should `localization`  be changed to `nls` , and accordingly the usage of the service also to `nls` ?

gui/src/main/java/com/devonfw/ide/gui/localization/LocalizationService.java
```diff
@@ -0,0 +1,436 @@
...
+   * @return immutable list of supported locales.
+   */
+  private List<Locale> detectAvailableLocales() {
+
+    Set<Locale> detectedLocales = new LinkedHashSet<>();
+    detectedLocales.add(Locale.ENGLISH);
+
+    String resourceBase = BUNDLE_NAME.replace('.', '/');
+    String packagePath = resourceBase.substring(0, resourceBase.lastIndexOf('/') + 1);
+    String filePrefix = resourceBase.substring(resourceBase.lastIndexOf('/') + 1) + "_";
```
hohwille: It is so much easier to assemble/compose values that to do the opposite.

So best practice would be to do this instead:
```
private static final String BUNDLE_PACKAGE = "localization";
private static final String BUNDLE_BASE_NAME = "messages";
private static final String BUNDLE_NAME = BUNDLE_PACKAGE + "." + BUNDLE_BASE_NAME;
```

Then you can simply do
```
String packagePath = BUNDLE_PACKAGE.replace('.', '/');
String filePrefix = BUNDLE_BASE_NAME + "_";
```

gui/src/main/java/com/devonfw/ide/gui/nls/NlsService.java
```diff
@@ -0,0 +1,436 @@
...
+
+  private void collectLocalesFromDirectory(Path dir, String filePrefix, Set<Locale> locales) {
+
+    try (var stream = Files.list(dir)) {
+      stream.map(p -> p.getFileName().toString())
+          .filter(n -> n.startsWith(filePrefix) && n.endsWith(".properties"))
+          .forEach(n -> addLocaleFromTag(n.substring(filePrefix.length(), n.length() - ".properties".length()), locales));
+    } catch (IOException e) {
```
hohwille, line 296: Always use constants instead of repeating identical String literals:
```
private static final String EXTENSION_PROPERTIES = ".properties";
```

gui/src/main/java/com/devonfw/ide/gui/localization/LocalizationService.java
```diff
@@ -0,0 +1,436 @@
...
+    String path = url.getPath();
+    String jarFilePath = path.substring(0, path.indexOf('!'));
+    String entryPrefix = packagePath + filePrefix;
+    try {
+      Path jar = Path.of(new URI(jarFilePath));
+      try (JarFile jarFile = new JarFile(jar.toFile())) {
+        jarFile.stream()
+            .map(JarEntry::getName)
+            .filter(n -> n.startsWith(entryPrefix) && n.endsWith(".properties"))
+            .forEach(n -> addLocaleFromTag(n.substring(entryPrefix.length(), n.length() - ".properties".length()), locales));
```
hohwille: same here

gui/src/main/java/com/devonfw/ide/gui/localization/LocalizationService.java
```diff
@@ -0,0 +1,436 @@
...
+      }
+    } catch (Exception e) {
+      LOG.debug("Could not scan JAR {} for locales: {}", url, e.getMessage());
```
hohwille: https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc#catching-and-handling-exceptions

cli/src/main/java/com/devonfw/tools/ide/variable/IdeVariables.java
```diff
@@ -126,13 +126,17 @@ public interface IdeVariables {
   /** {@link VariableDefinition} for support of overriding the default pycharm jvm options. */
   VariableDefinitionString PYCHARM_VM_ARGS = new VariableDefinitionString("PYCHARM_VM_ARGS", null);
 
+  /** {@link VariableDefinition} for selected GUI locale persisted in user properties. */
+  VariableDefinitionString IDE_LOCALE = new VariableDefinitionString("IDE_LOCALE");
```
hohwille: should we really introduce a new variable for that?
There already is a Java system property for it and that is the official standard: `user.lang`
So what if the user sets `IDE_OPTIONS=-Duser.lang=de` and `IDE_LOCALE=es`?
IMHO this would cause some confusion.

gui/src/main/java/com/devonfw/ide/gui/App.java
```diff
@@ -25,22 +27,28 @@ public class App extends Application {
...
+  private LocalizationService localizationService;
+
   private static final Logger LOG = LoggerFactory.getLogger(App.class);
 
   @Override
   public void start(Stage primaryStage) throws IOException {
 
+    this.primaryStage = primaryStage;
+
+    this.localizationService = LocalizationService.getInstance(null);
```
hohwille:
```suggestion
    this.localizationService = LocalizationService.getInstance();
```

gui/src/main/java/com/devonfw/ide/gui/localization/LocalizationService.java
```diff
@@ -0,0 +1,436 @@
...
+    return instance;
+  }
+
+  public static synchronized LocalizationService getInstance() {
+
+    if (instance == null) {
+      return getInstance(null);
+    }
+    return instance;
+  }
```
hohwille: This API is IMHO flawed:
If my code is calling:
```
LocalizationService serviceDe = LocalizationService.getInstance("de");
LocalizationService serviceDefault = LocalizationService.getInstance();
LocalizationService serviceEn = LocalizationService.getInstance("en");
```

Then all 3 instances will point to the same one which is technically the first one (here `de` ).
Why do we support the `Locale` argument at all?
Do we ever use this in productive code?
IMHO not so why not removing it and just offer `getInstance()` what would then be compliant with singleton-pattern.

gui/src/main/java/com/devonfw/ide/gui/localization/LocalizationService.java
```diff
@@ -0,0 +1,436 @@
...
+
+  /**
+   * Resets the singleton instance for tests or re-initialization.
+   */
+  public static synchronized void resetInstance() {
+
+    instance = null;
+  }
```
hohwille: Do we really need this method?
If you follow my suggestion to clean singleton pattern, we would not need this.

gui/src/main/java/com/devonfw/ide/gui/nls/NlsService.java
```diff
@@ -0,0 +1,436 @@
...
+  private void persistLocale(Locale localeToPersist) {
+
+    EnvironmentVariables environmentVariables = getUserEnvironmentVariables();
+    try {
+      environmentVariables.set(IdeVariables.IDE_LOCALE.getName(), localeToPersist.toLanguageTag());
+      environmentVariables.save();
+    } catch (RuntimeException e) {
+      LOG.warn("Failed to persist GUI locale: {}", e.getMessage());
+    }
+  }
```
hohwille: using `EnvironmentVariablesType.USER` is fully correct since the GUI exists globally and language will not be specific for a particular project.
As I already commented to `IDE_LOCALE` variable, we should instead use `IDE_OPTIONS` and add or update `-Duser.lang=...` in it.

gui/src/main/java/com/devonfw/ide/gui/localization/LocalizationService.java
```diff
@@ -0,0 +1,436 @@
...
+      String resourceName = toResourceName(toBundleName(baseName, locale), "properties");
+      try (InputStream inputStream = loader.getResourceAsStream(resourceName)) {
+        if (inputStream == null) {
+          return null;
+        }
+        return new PropertyResourceBundle(new InputStreamReader(inputStream, StandardCharsets.UTF_8));
+      }
+    }
+
+  }
```
hohwille: It is great and best practice to have resource bundles as UTF8.
However, this is standard in Java since Java 9 if I am not mistaken:
https://openjdk.org/jeps/226
Since we know that we will run in JDK 25+ this is IMHO pointless.

gui/src/main/java/com/devonfw/ide/gui/MainController.java
hohwille: Thanks. Very nice improvement 👍

gui/src/main/java/com/devonfw/ide/gui/nls/NlsService.java
hohwille: IMHO you did not entirely get my point when you introduced a new variable `IDE_LOCALE`.
However, I added a new review comment and can now resolve this one.

hohwille (approved):
@KarimALotfy thanks for your rework. Great that you figured the proper solution to find the locales from the classpath and reload the GUI elegantly after language change, etc. Great work 👍 
Still I left some comments for improvement. Please have a look. When resolved, we can finally merge.

KarimALotfy:
> @KarimALotfy thanks for your rework. Great that you figured the proper solution to find the locales from the classpath and reload the GUI elegantly after language change, etc. Great work 👍 Still I left some comments for improvement. Please have a look. When resolved, we can finally merge.

@hohwille  

Thanks for the review feedback,  what's recently changed:
- Replaced IDE_LOCALE with -Duser.lang=<tag> inside the existing IDE_OPTIONS 
- Removed the UTF8Control inner class
- Replaced the LocalizationService singleton with constructor injection via App → MainController
- Updated existing tests and added three new ones for the IDE_OPTIONS persistence logic
- Implemented the other suggested changes

gui/src/main/resources/nls/messages.properties
hohwille: That would be the easiest way to align with existing `cli` naming.

gui/src/main/java/com/devonfw/ide/gui/localization/LocalizationService.java
```diff
@@ -0,0 +1,401 @@
...
+   * @param explicitLocale the locale requested by the caller, or {@code null}.
+   */
+  private void initialize(Locale explicitLocale) {
+
+    Locale localeToApply = explicitLocale;
+    if (localeToApply == null) {
+      localeToApply = loadPersistedLocale();
+      if (localeToApply == null) {
+        localeToApply = Locale.getDefault();
+      }
```
hohwille: You do not need to load the default locale manually.
That is already done by Java automatically when set via system property `user.lang`.

```suggestion
      localeToApply = Locale.getDefault();
```

gui/src/main/java/com/devonfw/ide/gui/nls/NlsService.java
```diff
@@ -0,0 +1,401 @@
...
+    if (ideOptions == null || ideOptions.isBlank()) {
+      return null;
+    }
+    Matcher matcher = USER_LANG_PATTERN.matcher(ideOptions);
+    if (matcher.find()) {
+      String tag = matcher.group(1);
+      return tag.isBlank() ? null : Locale.forLanguageTag(tag);
+    }
+    return null;
+  }
```
hohwille:
```suggestion
```

gui/src/main/java/com/devonfw/ide/gui/localization/LocalizationService.java
```diff
@@ -0,0 +1,401 @@
...
+
+    EnvironmentVariables environmentVariables = getUserEnvironmentVariables();
+    try {
+      String ideOptions = environmentVariables.get(IDE_OPTIONS);
+      String userLangFlag = "-Duser.lang=" + localeToPersist.toLanguageTag();
+      String updated;
+      if (ideOptions == null || ideOptions.isBlank()) {
+        updated = userLangFlag;
+      } else if (USER_LANG_PATTERN.matcher(ideOptions).find()) {
+        updated = USER_LANG_PATTERN.matcher(ideOptions).replaceFirst(userLangFlag);
```
hohwille: Matcher of regex should be reused since it is expensive.
```
Matcher matcher = USER_LANG_PATTERN.matcher(ideOptions);
if (matcher.find()) {
  updated = matcher.replaceFirst(userLangFlag);
}
```

hohwille: BTW: If you extract a method with default visibility like this:
```
String update = updateLocale(String ideOptions, String userLangFlag);
```

Then it is much easier to test and cover all possible combinations directly via `updateLocale`.

hohwille (approved):
@KarimALotfy thanks for the update. Now only very tiny comments left. Then ready for merge. 👍

oanding-blrng:
NOT successfully tested with 2026.07.001-07_07_06-SNAPSHOT on Windows. In the GUI with `ide gui` the dropdown menu for languages doesn't appear.

Bug-Ticket #2126 was created.

```
$ ide -d -p gui
Running commandlet Gui[gui]
IDE environment variables have been set for ~/projects/project in workspace main
Running command 'C:\Program Files\Git\mingw64\bin\git.exe' with arguments 'branch' '--show-current' ...
Running command 'C:\Program Files\Git\mingw64\bin\git.exe' with arguments 'config' '--get' 'branch.main.remote' ...
Running command 'C:\Program Files\Git\mingw64\bin\git.exe' with arguments 'config' '--get' 'branch.main.merge' ...
Running command 'C:\Program Files\Git\mingw64\bin\git.exe' with arguments 'rev-parse' '@{u}' ...
The path ~/projects/project/settings/.git/*****_**** was last updated PT27.891S ago and caching duration is PT5M.
Skipping git fetch on ~/projects/project/settings because last fetch was just recently to avoid overhead.
The path ~/projects/_ide/urls/.git/**** was last updated PT7M24.481S ago and caching duration is PT30M.
Skipping git ****/***** on ~/projects/_ide/urls because last fetch was just recently to avoid overhead.
Running command 'C:\Program Files\Git\mingw64\bin\git.exe' with arguments 'diff-index' '--quiet' 'HEAD' ...
Running command 'C:\Program Files\Git\mingw64\bin\git.exe' with arguments 'ls-files' '--other' '--directory' '--exclude-standard' ...
Resolved version pattern * to version 3.9.16
No CVEs found for version 3.9.16 of tool mvn.
Tool mvn has 1 other tool(s) as dependency
Resolved version pattern 25* to version 25.0.3_9
No CVEs found for version 25.0.3_9 of tool java.
Tool java has 0 other tool(s) as dependency
Version 25.0.3_9 of tool java is already installed
Version 3.9.16 of tool mvn is already installed
Resolved version pattern 25.* to version 25.0.3_9
No CVEs found for version 25.0.3_9 of tool java.
Tool java has 0 other tool(s) as dependency
Version 25.0.3_9 of tool java is already installed
Starting GUI via commandlet
Running command 'C:\Users\oanding\projects\IDEasy\software\mvn\bin\mvn.cmd' with arguments '-f' 'C:\Users\oanding\projects\_ide\installation\gui\pom.xml' 'exec:exec' '-Dexec.executable=java' '-Dexec.classpathScope=compile' '-Dexec.args=-classpath %classpath com.devonfw.ide.gui.AppLauncher' ...
BASH_PATH environment variable was not found
A proper bash executable was found in your PATH environment variable at: C:/Program Files/Git/usr/bin/bash.***
Step 'ide' ended successfully.
```

PR #1995, opened 2026-06-01
#797: zip symlink pure java

### LINKED ISSUE #797: vscode not working on mac
### Expected behavior

As a IDEasy user, I want to install and run vscode via ide command so that I can easily use it.

### Actual behavior

```
$ ide vscode
Expected folder was not found at /Users/hohwille/projects/IDEasy/settings/vscode/workspace/setup
Start: Configuring workspace main for IDE vscode
Successfully ended step 'Configuring workspace main for IDE vscode'.
Successfully completed ide (vscode)
/Users/hohwille/projects/IDEasy/software/vscode/bin/code: line 38: /Users/hohwille/projects/_ide/software/default/vscode/vscode/1.95.3/Visual Studio Code.app/Contents/MacOS/Electron: Permission denied
```

1. `ide vscode`

### Comments/Hints:

Also plugin installations failed for this reason.
### Affected version:

* OS:Mac
* 2024.12.001-beta-SNAPSHOT

shodiBoy1:
### Implemented changes:

- On macOS, installing VSCode was failing because the ZIP extractor could not handle .framework folders. These folders use symlinks that point to other symlinks (a chain). The old code tried to resolve the full path of each symlink before creating it, which fails when the next symlink in the chain does not exist yet.                                                               
- New behavior: write each symlink with the exact text from the ZIP, the same way unzip does. No path resolution, no existence check. The chain just works once all files are on disk.                                                                          
---

### Testing instructions                                                   
                                                                                                                                   
  1. Check out the PR:
  gh pr checkout 1995                                                                                                      
          
  2. Run on a Mac:                                                                                                                 
  - Open the project in IntelliJ.
  - go to 'com.devonfw.tools.ide.cli.Ideasy' -> Modify Run Configuration.                                                       
  - Program arguments: vscode                                      
  - Working directory: any empty folder (e.g. ~/test-ideasy).                                                                      
  - Click Run.   

---

### Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -780,7 +783,8 @@ public void extractZip(Path file, Path targetDir) {
    */
   private void extractZipWithJava(Path file, Path targetDir) {
 
-    final List<PathLink> links = new ArrayList<>();
+    // link path -> raw target string from the archive; LinkedHashMap keeps entry order for deterministic creation
+    final Map<Path, String> symlinks = new LinkedHashMap<>();
```
hohwille: what is wrong with `PathLink` object?
I am not getting why you decided not to use it anymore?
IMHO it is still containing the same information in a slightly more structured way.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -794,13 +798,12 @@ private void extractZipWithJava(Path file, Path targetDir) {
...
             String linkTarget = IOUtils.toString(entryStream, StandardCharsets.UTF_8);
-            Path parent = entryPath.getParent();
-            Path source = parent.resolve(linkTarget).normalize();
-            resolveRelativePathSecure(source, root, linkTarget);
-            links.add(new PathLink(source, entryPath, PathLinkType.SYMBOLIC_LINK));
-            mkdirs(parent);
+            // security: refuse archives whose link would resolve outside the extraction root
+            resolveRelativePathSecure(entryPath.getParent().resolve(linkTarget).normalize(), root, linkTarget);
+            mkdirs(entryPath.getParent());
+            symlinks.put(entryPath, linkTarget);
```
hohwille: Still I cannot see a reason for this change. Still the same thing is happening here.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
```diff
@@ -814,9 +817,23 @@ private void extractZipWithJava(Path file, Path targetDir) {
...
+          Files.createSymbolicLink(linkPath, Path.of(rawTarget));
+        } catch (FileSystemException e) {
+          // on Windows without symlink permission fall back to a junction, same as link() does
+          if (SystemInfoImpl.INSTANCE.isWindows()) {
+            Path absoluteSource = linkPath.getParent().resolve(rawTarget).normalize();
+            mklinkOnWindows(Path.of(rawTarget), absoluteSource, linkPath, PathLinkType.SYMBOLIC_LINK, true);
+          } else {
+            throw e;
+          }
+        }
```
hohwille: you are slightly reimplementing our `link` function here.
Could the entire fix instead be to simply change `true` to `false` here, what would also fix the same bug in "untar" (that we might not yet have discovered)?
https://github.com/devonfw/IDEasy/blob/6080ee79bcab0b1292121e35b99b3157558bb23d/cli/src/main/java/com/devonfw/tools/ide/io/FileAccess.java#L140-L142

hohwille (changes requested):
@shodiBoy1 thanks for your PR and your analysis of the root cause. I like that you always take such special problem and analyse it properly trying to come up with a rather small fixing PR👍 
Surely we could merge this as is as a quickfix.
However, you revealed here that we have actually implemented our `link` function wrong (and therefore our "unpack" functions - not only for zip but also for tar).
IMHO your PR is not fixing that bug but adding a workaround with quite some redundancies.

My goal is that we aim for true quality so I would love to fix the `link` function properly.
Can you maybe extract a JUnit test that reproduces the scenario with invoking our `link` function?
BTW: We meanwhile also have `ide ln -s <source> <link>` to test such things. If you want to test with `relative` being `true` you can do `ide ln -s -r <source> <link>`.
If that is buggy, please lets fix the root cause.

cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java
hohwille: OK, seems that there is another "bug" in our `link` implementation:
https://github.com/devonfw/IDEasy/blob/6080ee79bcab0b1292121e35b99b3157558bb23d/cli/src/main/java/com/devonfw/tools/ide/io/FileAccessImpl.java#L528-L530

This was introduced in commit [083a35d](https://github.com/devonfw/IDEasy/commit/083a35d4fa9fe45b250504fb891106dd641fd518#diff-8f970ebcc8a0742f2d859c3fa7af41b8ba95b892ab30804eaff3bb3aaf9ad268).

My goal is still to get our `link` and "unpack" methods finally fixed.
IMHO that change was wrong.

hohwille: if `relative` is `false`, then why should our `link` function fiddle around with any of the given `Path` values at all?
IMHO this is all wrong. Maybe it is needed by the fallback for mklink but then it should be done only there if required.

shodiBoy1:
Thanks for the review.

My initial idea was to preserve the raw symlink target from the ZIP entry, because macOS framework bundles can contain chained symlinks where the final target is created later. That is why I first handled ZIP symlinks directly.

After your feedback I agree that this was too ZIP-specific and bypassed the existing PathLink/link(...) abstraction. I reworked it so ZIP and TAR both go through PathLink -> link(PathLink), and the actual fix is now in the shared link handling.

The shared link code now avoids resolving symlink targets too early, so archive symlinks can point through links that are created later. I also added a regression test that calls FileAccess.link(PathLink) directly for the macOS framework-style case.

Tested with FileAccessImplTest, LnCommandletTest, and a manual VS Code install from IntelliJ.

hohwille (approved):
@shodiBoy1 thanks for rework. Now approach seems reasonable. Changing to approved 👍 
Team-review should still test that link creation still works as expected and have a quick look again and the diff.

tineff96 (commented):
Overall, the PR looks good to me.

I ran:

```bash
mvn clean test
```

Results:

* Ubuntu/WSL: passes without changes
* Windows: fails on this PR branch in:

  * `testSymlinkShortcutPaths`
  * `testSymlinkAbsolutePassingRelativeSource`

The Windows failures seem related to the `mklink` fallback. If `relative == false`, passing the relative `source` to `mklink` causes Windows to resolve it against the process working directory instead of the link parent directory.

I tested the following adjustment in `FileAccessImpl.java`, line 481:

```java
Path finalSource = relative ? source : absoluteSource;
```

With this change, the focused test suite passes for me on Windows as well. This should keep the ZIP/macOS path able to preserve relative symlink targets when needed, while the Windows fallback keeps the previous absolute/adapted behavior for `relative == false`.

I still think the original macOS VSCode case should be verified manually on a Mac. I will ask a teammate with macOS to run the `ide vscode` scenario as well.

AdemZarrouki:
tested on mac(26.5.1) arm64 [Mac OS X@arm64]

```
ide vscode
Start: Configuring workspace main for IDE vscode
Successfully ended step 'Configuring workspace main for IDE vscode'.
No CVEs found for version 1.124.2 of tool vscode.
Could not determine file extension from URL https://update.code.visualstudio.com/1.124.2/darwin-arm64/stable - guess was zip but may be incorrect.
Trying to download vscode-1.124.2-mac-arm64.zip from https://update.code.visualstudio.com/1.124.2/darwin-arm64/stable
Downloading 100% │████████████████████│ 237/237MiB (0:01:26 / 0:00:00) 2,8MiB/s
SHA-256 checksum d0955f81c63181d69e59f78616b7c6deac695878a542d7bf74b896df7cb19bca is correct.
Extracting ZIP file /Users/Library/Caches/IDEasy/downloads/default/vscode-1.124.2-mac-arm64.zip to /Users/Projects/_ide/tmp/extract-vscode-1.124.2-mac-arm64.zip
Extracting 100% │████████████████████│ 237/237MiB (0:00:05 / 0:00:00) 47,6MiB/s
Resetting all installed plugins...
Installing Plugins 100% │██████│ 54/54 Plugins (0:08:35 / 0:00:00) ,1 Plugins/s
...
Successfully installed vscode in version 1.124.2 at /Users/Projects/test-ideasy/software/vscode
Successfully completed ide (vscode)
```

vscode was installed and launched successfully.

tineff96 (approved):

shodiBoy1:
@tineff96  thanks for catching this and for the fix. Applied and pushed it. Re-ran FileAccessImplTest on Windows myself: 0 failures.

PR #1999, opened 2026-06-02
#1392: smart completions

### LINKED ISSUE #1392: Smart completion

Caylipp:
### Implemented changes:

* Add `AutoCompletionRegistry` for tool-specific completion candidates
* Add `ToolArgumentsProperty` for completing tool arguments
* Delegate tool argument completion from `ToolArgumentsProperty` to `ToolCommandlet`
* Add default empty completion registry in `ToolCommandlet`
* Add initial Maven completion candidates in `Mvn`

---

### Testing instructions

1. Create an run configuration for IDEasy.
2. Set the program arguments to: -q complete mvn cl
3. Verify that Maven goals and options are listed, e.g.: clean

Additionally, you can use the end-to-end testing script to test the completion natively in your terminal (https://github.com/devonfw/IDEasy/pull/2011).

---

### Checklist for this PR

  with `internal`


AdemZarrouki (approved):
Nice Job. You extended the auto complition for mvn.

AdemZarrouki:
When testing with GraalVM using ide shell, i typed ide mvn -Dexec.m and then pressed tab but the shell was not responding anymore. @Caylipp can you check the cause of this. Aslo tried to run it in a junit test
```
  @Test
  void testCompleteMavenToolArguments() {

    // arrange
    AbstractIdeContext context = newContext(PROJECT_BASIC, null, false);
    CliArguments args = CliArguments.ofCompletion("mvn", "-Dexec.main");

    // act
    List<CompletionCandidate> candidates = context.complete(args, true);

    // assert
    assertThat(candidates.stream().map(CompletionCandidate::text))
        .contains("-Dexec.mainClass=");
  }
```
but the test was running for ever

Caylipp:
> When testing with GraalVM using ide shell, i typed ide mvn -Dexec.m and then pressed tab but the shell was not responding anymore. @Caylipp can you check the cause of this. Aslo tried to run it in a junit test
> 
> ```
>   @Test
>   void testCompleteMavenToolArguments() {
> 
>     // arrange
>     AbstractIdeContext context = newContext(PROJECT_BASIC, null, false);
>     CliArguments args = CliArguments.ofCompletion("mvn", "-Dexec.main");
> 
>     // act
>     List<CompletionCandidate> candidates = context.complete(args, true);
> 
>     // assert
>     assertThat(candidates.stream().map(CompletionCandidate::text))
>         .contains("-Dexec.mainClass=");
>   }
> ```
> 
> but the test was running for ever

Thanks for pointing this out.
I found and fixed the issue. The problem was that `-Dexec.m` was treated as an option because it starts with `-`. However, in this case it is not an IDEasy option, but a Maven argument.
If no matching option was found, the completion logic did not advance the current argument and ended up in an endless loop. I added a fallback to value completion for this case.
I tested `complete mvn -Dexec.m` locally in the shell.

AdemZarrouki:
I tested the new fix and the endless loop wasn't present anymore. The only problem is that in the issue its stated that
"Also be aware that those "keywords" above that end with = are not complete so they should not add a whitespace after its completion" but when testing the auto-completion with -Dexec.mainClass= and -Dexec.args= i get a whitespace after =. This need to be fixed.

Caylipp:
> I tested the new fix and the endless loop wasn't present anymore. The only problem is that in the issue its stated that "Also be aware that those "keywords" above that end with = are not complete so they should not add a whitespace after its completion" but when testing the auto-completion with -Dexec.mainClass= and -Dexec.args= i get a whitespace after =. This need to be fixed.


You’re right, thanks for pointing this out.
The issue was caused by Bash automatically appending a whitespace after completion. I’ve fixed it in the `_ide_completion` function by applying `compopt -o nospace` for candidates ending with `=`.

Caylipp:
Also fixed completion for `:` (e.g. `dependency:list`) by removing it from `COMP_WORDBREAKS` to prevent Bash from splitting the argument.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -382,8 +382,8 @@ public void setCwd(Path userDir, String workspace, Path ideHome) {
   }
 
   /**
-   * On macOS, {@code ~/Downloads} is protected by the OS (TCC) and the CLI may not be allowed to delete it, so we put the cache under
-   * {@code ~/Library/Caches} instead. Tests still use {@code ~/Downloads/ide} so existing fixtures keep working.
+   * On macOS, {@code ~/Downloads} is protected by the OS (TCC) and the CLI may not be allowed to delete it, so we put the cache under {@code ~/Library/Caches}
```
AdemZarrouki, line 388: can you remove the line break here.

AdemZarrouki (approved):
Good Job @Caylipp. You fixed the issue. I left a small comment regarding line break in javadoc that need to be removed

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -1495,15 +1494,30 @@ private void completeCommandlet(CliArguments arguments, Commandlet cmd, Completi
...
+              Property<?> valueProperty = valueIterator.next();
+              boolean success = valueProperty.apply(arguments, this, cmd, collector);
+              if (!success) {
+                LOG.trace("Completion cannot match option or value.");
+              }
+            } else {
+              LOG.trace("No value left for completion.");
+            }
+            return;
+          }
```
hohwille: Looks like copy&paste of code (from outer `else` block). Could be avoided by taking the original code out of the `else` and add a separate condition by the new flag.
However, is it correct to complete from `valueIterator` if we are inside if-condition `currentArgument.isOption()`? This code does not really make sense to me.
In case I am missing something, please let me know...

cli/src/main/package/functions
```diff
@@ -152,6 +157,15 @@ _ide_completion()
   else
     COMPREPLY=( $(ideasy -q complete ${COMP_WORDS[@]:1}) )
   fi
+
+# keywords ending with "=" remove whitespace
```
hohwille:
```suggestion
  # keywords ending with "=" remove whitespace
```

cli/src/main/package/functions
```diff
@@ -144,6 +144,11 @@ _ide_create_project()
   done
 }
 
+# remove ":" from word breaks for correct completion
+if [ -n "${BASH_VERSION:-}" ]; then
```
hohwille: I do not understand the condition. Should this not run in zsh?
I would rather expect something like this:
```suggestion
if [ -n "${COMP_WORDBREAKS}" ]; then
```
If I misunderstood, then maybe you could improve the comment above.

hohwille (approved):
@Caylipp thanks for your PR. Looks really nice now. Great work 👍 
I have to admit that I do not fully understand the workaround for the `=` handling.
Is that maybe because `=` is by default in `COMP_WORDBREAKS`?
I guess this is for situations where we have `--option=val[TAB]` and want to get `--option=value`. So what is by default braking here? Is a space added before or directly after the equals sign?
Then I get at least the `compopt -o nospace` workaround.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
hohwille: I now guess the problem is that for `mvn` you want to complete options like `-DskipTests` as part of the `ToolArgumentsProperty`.
If that is the correct understanding, then IMHO the proper fix would be that `ToolArgumentsProperty` should implicitly trigger `endOptions()` on `CliArguments`.

hohwille: For multi-valued properties like here we have this:
https://github.com/devonfw/IDEasy/blob/a4f2516fade1a9e362908bdcfefc4fe0b8f9f657/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L1577-L1579

However, in this code:
https://github.com/devonfw/IDEasy/blob/a4f2516fade1a9e362908bdcfefc4fe0b8f9f657/cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java#L1496

Wo do check for `!isEndOptions()` but not for `isSplitShortOpts()`.

hohwille: Maybe in this multi-valued property scenario we should call `endOptions()` instead of `stopSplitShortOptions()`?

Caylipp:
> @Caylipp thanks for your PR. Looks really nice now. Great work 👍 I have to admit that I do not fully understand the workaround for the `=` handling. Is that maybe because `=` is by default in `COMP_WORDBREAKS`? I guess this is for situations where we have `--option=val[TAB]` and want to get `--option=value`. So what is by default braking here? Is a space added before or directly after the equals sign? Then I get at least the `compopt -o nospace` workaround.

The whitespace issue is not caused by `=` being part of `COMP_WORDBREAKS`. Instead, this is Bash's default behavior: after inserting a completion candidate, Bash appends a space by default because it assumes the token is complete. Since in the issue (https://github.com/devonfw/IDEasy/issues/1392) is mentioned "those "keywords" above that end with = are not complete so they should not add a whitespace after its completion", this assumption would be incorrect. Using `compopt -o nospace` prevents Bash from appending this whitespace.

In case of `:` (e.g. `dependency:list`), the problem is different. Here `:` is part of `COMP_WORDBREAKS`, so Bash stops completion at `dependency:`. Therefore, adjusting `COMP_WORDBREAKS` is required to allow completing the full token `dependency:list`.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
Caylipp: Thanks, that makes sense.

The fallback was introduced to handle cases like `-Dexec...`, which are treated as options due to the leading `-`, but are actually tool arguments. So if no matching IDE option is found, the fallback allows completing them via the value logic. But I agree that this is not a cleanest solution, as it mixes option and value handling and duplicates logic from the value branch. Your suggestion makes more sense: treating ToolArgumentsProperty explicitly as values by triggering endOptions() avoids the need for the fallback. I'll rework the solution in that direction.

cli/src/main/package/functions
Caylipp: Good point. I used `BASH_VERSION` because I considered `COMP_WORDBREAKS` to be Bash specific and wanted to make sure the workaround only runs in Bash.
I did not consider the zsh case here. But checking `COMP_WORDBREAKS` directly makes more sense here.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
Caylipp: I refactored this now so we no longer jump from option completion into value completion directly.
Instead, the code now uses `isEndOptions()` to stop option parsing when a value property requires it. This way, tool arguments like `-Dexec.*` are handled by the normal value completion flow.

cli/src/main/package/functions
Caylipp: I updated the comment.

cli/src/main/package/functions
Caylipp: I updated the condition to check `COMP_WORDBREAKS` directly instead of `BASH_VERSION`.

Caylipp:
I also noticed one additional edge case while retesting: `ide mvn --<TAB>` currently does not show the expected `--...` Maven suggestions, while `ide mvn -<TAB>` works.
This seems to be a separate parsing edge case around `--`, since `--` can also be interpreted as the end-of-options marker.

Caylipp:
I did some debugging on this. For `ideasy -q complete mvn --`, the argument `--` is already consumed during the initial parsing (before `CompleteCommandlet.doRun()` is called) so it never reaches the completion logic in `CliArguments.ofCompletion(...)`.
That is why `mvn --<TAB>` currently does not work, while `mvn -<TAB>` and `mvn --a<TAB>` do.
So this seems to be a parsing limitation where `--` is treated as end-of-options too early for the completion use case.
Since the release is close and the main Maven smart completion cases are working now, I would suggest tracking this `--` edge case separately unless you think it should still be included in this PR.


> I also noticed one additional edge case while retesting: `ide mvn --<TAB>` currently does not show the expected `--...` Maven suggestions, while `ide mvn -<TAB>` works. This seems to be a separate parsing edge case around `--`, since `--` can also be interpreted as the end-of-options marker.

PR #2003, opened 2026-06-04
#1651: push and commit status.json files to ide-urls-status after running the URL update worklfow

### LINKED ISSUE #1651: Move status.json files out of ide-urls to own repo
### Task summary

All the `status.json` files are never used by IDEasy itself.
They are only needed for the UrlUpdater process that we run automatically every night.
These files cause a lot of diffs in the `ide-urls` repository that is continuously growing.
IMHO I made a design flaw that we included all these `status.json` files there.
My suggestion now is to create a new git repo for only the status of the IDE urls.
To get started, we could just move the `status.json` files there.
That is we need to change the UrlUpdaters and the GHA:
* GHA update-urls workflow is also cloning the new repo
* new repo will have the same folder structure but only contain the `status.json` files.
* UrlUpdaters will write them to the new repo (a parallel side-by-side directory structure)
* GHA update-urls workflow at the end commits and pushes both repos.
* We need a new access token to also include write access to the new repo.

### Additional context

_No response_

AdemZarrouki:
### Implemented changes:

* update the workflow in `update-urls.yml` to commit and push `status.json` files to repo `ide-urls-status`
* `status.json` should not be pushed to `ide-urls`
* The migration of status.json is alreday done in https://github.com/devonfw/ide-urls-status/pull/1
* Before Merging we need to update `ACTION_PUSH_TOKEN` so that it have access to the new create repo `ide-urls-status`. This need to be done by @hohwille as he has access to that functionality.
* After merging this we need to create another PR to delete all `status.json` files from `ide-urls` (will be done in https://github.com/devonfw/IDEasy/issues/2021).

---

### Testing instructions

1. This can be tested in a test branch but for that you need to adapt `update-urls.yml` to include your test branch. A workflow example was created and it successfully updates both `ide-urls` and` ide-urls-status`. https://github.com/AdemZarrouki/IDEasy/actions/runs/26902445623
2. Also you need to include a personal `ACTION_PUSH_TOKEN` and activate the URL Update workflow in your fork.

---

### Checklist for this PR

  with `internal`


tineff96 (commented):
I tested the workflow end-to-end in a fork.

---
### Test setup:

Created a test branch (test/2003-test-branch) in my fork:

tineff96/IDEasy
tineff96/ide-urls
tineff96/ide-urls-status


Adapted the workflow in tineff96/IDEasy to run against the test branch (test/2003-test-branch)

Executed the workflow based on the provided template
(reference run: https://github.com/AdemZarrouki/IDEasy/actions/runs/26902445623)

----

### Result:


The migration of `status.json` files from `ide-urls` to `ide-urls-status` worked as expected.

The workflow correctly commits:
- `status.json` files are committed to `ide-urls-status`
- all other files are committed to `ide-urls`


No unexpected files were committed

Important notes for reviewers / future testers:

Setting up the test environment is slightly non-trivial because it requires:

Forks of both repositories (ide-urls and ide-urls-status)
Adjusting the workflow to use the test branch


You must define the following secrets for commits to work:

ACTION_PUSH_TOKEN
BUILD_USER
BUILD_USER_EMAIL



Without BUILD_USER and BUILD_USER_EMAIL, the commit step will fail.
Conclusion:
From my perspective, the implementation behaves correctly and is ready for review.

hohwille (changes requested):
@AdemZarrouki thanks for your PR.
Are you aware that the existing code is not only writing the `status.json` files but also reading them and making choices depending on their content?
IMHO you are currently cloning both repos side-by-side and then running the `UpdateInitiator` that will not find any `status.json` without modifying the Java code when they get removed from the `ide-urls` repo.
I would have expected that you change the Java code to read and write the `status.json` files now from the separate directory.
If you rather want to do that in the workflow that may also be possible but then you also have to first delete all `status.json` files from `ide-urls` anfd then copy them from the new repo before you run the java code. After that you would then move them back and and ensure no `status.json` remains in `ide-urls` (what would automatically do the cleanup and delete all these files so we could avoid extra PRs for `ide-urls`).

However, things to consider:
* Testing workflows is really hard and therefore we should prefer to do as much code in Java and as little in workflows
* I would rather consider this rsync copying, etc. kind of hackish. In the end we have two git repos and a Java process to read and update the data and after that we want to push the changes. When you solve it in Java it IMHO becomes cleaner and the workflow stays shorter and easier to test and understand.

AdemZarrouki:
@hohwille thanks for the detailed review. So my idea was to avoid the java code change and do instead the changes in the workflow file as this requires less effort, and also i tried to test the workflow on a test branch many times to ensure that the changes are correct. So regarding this:
- "you are currently cloning both repos side-by-side and then running the UpdateInitiator that will not find any status.json without modifying the Java code when they get removed from the ide-urls repo.", yes it will not find any status.json files and it will creates new ones and then checks if they exist in `ide-urls-status` and if they are up-to-date (timestamp).
- "If you rather want to do that in the workflow that may also be possible but then you also have to first delete all status.json files from ide-urls anfd then copy them from the new repo before you run the java code" I did a test for this scenario by keeping the `status.json` files in both repos (https://github.com/AdemZarrouki/IDEasy/actions/runs/28010606987/job/82902658625) and as we can see that `status.json` were in `ide-urls-status` updated, and the ones in `ide-urls` were not.


But if you prefer to change the code so that UpdateInitiator accepts both ide-urls and ide-urls-status, we can do that.

AdemZarrouki:
I re-tested the workflow in a test branch after making the changes to the Java code (https://github.com/AdemZarrouki/IDEasy/actions/runs/28228272030). Everything is working as expected.
The files in https://github.com/AdemZarrouki/ide-urls/tree/test/test-workflow have been updated or created, and the corresponding status.json files in https://github.com/AdemZarrouki/ide-urls-status/tree/test/test-workflow have also been updated or created.

url-updater/src/main/java/com/devonfw/tools/ide/url/UpdateInitiator.java
```diff
@@ -19,49 +19,63 @@ public class UpdateInitiator {
...
       System.exit(1);
     }
 
-    String pathToRepo = args[0];
+    if (args.length < 2) {
+      logger.error("Error: Missing path to URL repository and/or status repository.");
+      logger.error(
+          "Usage: java UpdateInitiator <path_to_url_repository> <path_to_status_repository> <duration_string_format> <tool_to_test|updater_class_name>");
+      System.exit(1);
+    }
```
hohwille, line 41: We do not need two `if` now.
The first one (`if (args.length == 0)`) does not make sense any more and shall be removed.

cli/src/main/java/com/devonfw/tools/ide/url/model/folder/UrlRepository.java
```diff
@@ -7,6 +7,8 @@
  */
 public class UrlRepository extends AbstractUrlFolder<UrlTool> {
 
+  private UrlRepository statusRepository;
```
hohwille, line 10: An `UrlRepository` should not contain another `UrlRepository`.
BTW: Maybe it was wrong from the start to have the `UrlStatusFile` in `cli` module since we never read it from IDEasy itself. It is only used by UrlUpdater and nowhere else.
I would suggest to revert all these changes.
Simply determine the path to the status file from the path to the version folder (from `UrlVersion.getPath()`).

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/UpdateManager.java
```diff
@@ -70,6 +70,8 @@ public class UpdateManager extends AbstractProcessorWithTimeout {
 
   private final UrlRepository urlRepository;
 
+  private final UrlRepository statusRepository;
```
hohwille, line 73: Keep this as `Path` and simply derive the path to the `status.json` by resolving the relative path inside the repo to the version (e.g. `java/java/25.0.3_9`) to it plus the constant for `"status.json"`.
I would delete `UrlStatusFile`.
Also consider moving `StatusJson`, `UrlStatus`, and `UrlStatusState` from `cli` to `url-updater` module.
If we need caching of these files, we can simply use a map from path to the parsed `StatusJson`.
If you want to keep a wrapper object with `Path` reference and load + save methods also fine but no need to extend that from `AbstractUrlFile` anymore.

hohwille: Also this PR is IMHO still writing to the old locations. Code like this needs to be adopted:
https://github.com/devonfw/IDEasy/blob/c0773bf5848afcf14abdb765f9aedcc10246c22e/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java#L774-L779

When really deleting `UrlStatusFile` you will get compile errors revealing the problems to rework.

hohwille (changes requested):
I still have found some issues.
Since you are not available any more, I will search for someone else from the current team to take over this PR.

laert-ll:
Replaced by #2146

PR #2036, opened 2026-06-16
#1987: Fixed GUI not launching for snapshot versions

### LINKED ISSUE #1987: Gui commandlet broken for snapshots
### Actual behavior

```
$ ide gui
No CVEs found for version 3.9.16 of tool mvn.
No CVEs found for version 25.0.3_9 of tool java.
Version 25.0.3_9 of tool java is already installed
Version 3.9.16 of tool mvn is already installed
No CVEs found for version 25.0.3_9 of tool java.
Version 25.0.3_9 of tool java is already installed
```

No further message and no GUI showing up.

### Reproduce

1. `ide upgrade --mode=snapshot`
2. `ide gui`

### Expected behavior

The GUI should show up.
If there is an error happening, it should be displayed so we can trace it down.

### IDEasy status

```bash
$ ide -p status
Your version of IDEasy is 2026.06.001-05_30_03-SNAPSHOT.
You are online.
Your settings are up-to-date.
```

### Related/Dependent issues

#1724 #1968 #1964

### Comments/Hints

I found the following problems:
1. `repositories/repository` might need to be defined in `gui/pom.xml` so maven knows from where to download SNAPSHOTS.
```
 <repositories>
    <repository>
      <id>ossrh</id>
      <name>OSSRH Snapshots</name>
      <url>https://central.sonatype.com/repository/maven-snapshots</url>
      <snapshots>
        <enabled>true</enabled>
      </snapshots>
      <releases>
        <enabled>false</enabled>
      </releases>
    </repository>
  </repositories>
```
2. We need to add `-U` option to the maven call.
3. Since we do not have `parent` POM for GUI, we *maybe* want to explicitly specify the version of exec plugin and do not let it up to fate (mavens old super-pom) what version should be used (so instead of `exec:exec` we could use `rg.codehaus.mojo:exec-maven-plugin:3.1.0:exec`).

laim2003:
### Implemented changes:

* Added -U flag to the maven execution command
* Specified maven exec plugin version to 3.1.0
* Added `<repository>` definition to the gui-launcher pom file
* <del>Added `--enableLogging` / `-l` flag to the GUI commandlet. This will launch the GUI in `ProcessMode.DEFAULT`rather then `ProcessMode.BACKGROUND_SILENT`; this way all internal JavaFX/Maven logs etc are printed to the terminal, making debugging easier.</del>

---

### Testing instructions

##### a. Team-Review:
1. Build a native executable based on the state of this branch: [GraalVM build guide](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/graalvm-build-guide.adoc)
2. Run `ide upgrade --mode=snapshot`
3. Using the ideasy executable from step 1, run the gui (see step 4 in the GraalVM build guide)
4. GUI should now launch without any issues.

##### b. Retesting after Merge:
1. Upgrade your IDEasy version to the latest snapshot/nightly build: `ide upgrade --mode=snapshot`
2. run `ide gui`
3. GUI should now launch without any issues.

#### ~~Testing the new `--enableLogging` flag~~
---

### Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -27,13 +27,16 @@ public class Gui extends Commandlet {
 
   private static final Logger LOG = LoggerFactory.getLogger(Gui.class);
 
+  final FlagProperty enableExtendedLogging;
```
maybeec: Other commandlet property fields are declared `public final` (see e.g. `AbstractUpdateCommandlet`, `CreateCommandlet`). For consistency:

```suggestion
  /** {@link FlagProperty} to enable logging of the GUI process to the terminal. */
  public final FlagProperty enableExtendedLogging;
```

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -58,9 +61,16 @@ protected void doRun() {
...
 
-    mvn.installTool(mavenToolInstallRequest);
+    ToolInstallation mvnToolInstallation = mvn.installTool(mavenToolInstallRequest);
     ToolInstallation javaInstallation = java.installTool(javaToolInstallRequest);
 
+    /* Register the freshly installed mvn on the IDEasy managed PATH so that the IDEasy controlled maven is used to launch
+     the GUI instead of any maven that happens to be on the system PATH. We install via installTool (software repository
+     only) and therefore have to register the bin directory ourselves (install() would normally do this). I tried to achieve this alternatively via .withPathEntry(), however, this did not work as expected.
+     This was tested on a Mac; potentially, withPathVariable() works correctly on Windows.
+    */
```
maybeec: Per the *Code-Documentation* conventions, implementation comments should factually explain the *why*. This drops the first-person/speculative wording and wraps the long line to match the rest of the file:

```suggestion
    /*
     * Register the freshly installed mvn on the IDEasy-managed PATH so the IDEasy-controlled maven (from the software
     * repository) launches the GUI instead of any maven on the system PATH. installTool only installs into the software
     * repository, so we register the bin directory here (install() would normally do this).
     */
```

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -81,6 +92,13 @@ protected void doRun() {
...
      */
-    mvn.runTool(processContext.withPathEntry(javaInstallation.binDir()), ProcessMode.BACKGROUND_SILENT, args);
+    ProcessMode processMode = this.enableExtendedLogging.isTrue() ? ProcessMode.DEFAULT : ProcessMode.BACKGROUND_SILENT;
+    try {
+      mvn.runTool(processContext.withPathEntry(javaInstallation.binDir()), processMode, args);
+    } catch (Exception e) {
+      LOG.error(
+          "Failed to launch the GUI. If maven states issues with dependency resolution, check whether the maven M2 repo is enabled in your project.",
+          e);
+    }
```
maybeec, line 106: As written, any failure is logged and then swallowed, so `ide gui` exits successfully even when the GUI never launched. Wrapping-and-rethrowing keeps your hint but surfaces the failure to the caller (and narrows the catch to `RuntimeException`):

```suggestion
    try {
      mvn.runTool(processContext.withPathEntry(javaInstallation.binDir()), processMode, args);
    } catch (RuntimeException e) {
      throw new CliException(
          "Failed to launch the GUI. If maven reports issues with dependency resolution, check whether the maven M2 repo is enabled in your project.", e);
    }
```

maybeec (commented):
Thanks for tackling the SNAPSHOT GUI-launch bug. Switching to `context.newProcess()` and registering the freshly installed mvn on the IDEasy-managed PATH is a clean approach, and it's good to see a JUnit test covering both process modes. 👍

Reviewing only the diff against the [coding conventions](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc) (inline suggestions attached where there is an easy fix):

1. **(Medium) A failed launch now reports success** (`Gui.java` L96-102). Previously an exception from `runTool` propagated and the command failed; now `catch (Exception e)` logs and returns normally, so `ide gui` exits successfully even when the GUI never started. Logging the full exception is fine per the *Catching and handling Exceptions* section, so this isn't a convention breach — but for a CLI a failed launch should arguably surface a non-zero exit, and `catch (Exception e)` is very broad (it would also swallow programming errors). The inline suggestion wraps-and-rethrows as `CliException` while keeping your helpful hint.

2. **(Low) Comment wording and length** (`Gui.java` L67-71). Per the *Code-Documentation* section, implementation comments should factually explain the *why*. The current text is first-person/speculative ("I tried ... withPathEntry()", "tested on a Mac; potentially ... on Windows"), and L69 (~205 chars) is much longer than the wrapping used elsewhere in the file. A shortened, factual version is suggested inline.

3. **(Low) `-U` on every launch** (`Gui.java` L82). `--update-snapshots` forces a remote check on every `ide gui`, adding latency and requiring network even for released (non-SNAPSHOT) versions. Consider applying it only in snapshot mode.

4. **(Nit) Field visibility** (`Gui.java` L30). Other commandlet property fields are declared `public final` (e.g. `AbstractUpdateCommandlet`, `CreateCommandlet`); `enableExtendedLogging` is package-private. Making it `public final` keeps it consistent — suggestion inline.

Nice work overall — none of these are blocking.

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -27,13 +27,16 @@ public class Gui extends Commandlet {
...
+  final FlagProperty enableExtendedLogging;
+
   /**
    * @param context the {@link IdeContext}.
    */
   public Gui(IdeContext context) {
 
     super(context);
     addKeyword(getName());
+    enableExtendedLogging = add(new FlagProperty("--enableLogging", false, "-l"));
```
hohwille: For logging we already have properties like `--debug` or `--trace`, etc.
Further, don't use `--pascalCase` for CLI options but `--train-case` instead following POSIX and GNU standards.

After reading the code, I figured out that the property is actually changing the process mode and enforces that we do not run in background.
I would therefore completely rename the property and option to something like `--foreground` if you want to keep it explicit. If you think this is more related to the logging, then I would not add a new property at all but just trigger the feature when loglevel is DEBUG or below (--debug or --trace).

```suggestion
    this.enableExtendedLogging = add(new FlagProperty("--enable-logging", false, "-l"));
```

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
hohwille: Instead of adding an extra option flag or a log-level based hack - can't we always run maven in foreground but tell it to run `exec` in background so we get logs and errors if maven cannot launch the GUI, but the GUI will still not block the active terminal when started?
Would IMHO be the real and clean solution:
https://www.mojohaus.org/exec-maven-plugin/exec-mojo.html#async

BTW: All this feature about foreground vs. background is not related to the bugfix that snapshots cannot open GUI. It is easier to create two separate issues and according PRs. Maybe then we could already merge the snapshot bugfix and include it into the current release while now the PR is stuck.

hohwille (changes requested):
@laim2003 thanks for your PR. Nice that you improve and make progress to make the GUI launching perfect. 👍 
I left some comments and hints to guide to a maybe better solution. Please have a look.

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
hohwille: BTW: For such special flags avoid adding short options like `-l` this will easily conflict with other global flags that we might want to introduce later.

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
laim2003: I will remove the logging part from this PR and try the workaround with mavens' async. If we need to add this flag later, we can do that via a seperate PR.

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -59,29 +61,47 @@ protected void doRun() {
...
-    mvn.installTool(mavenToolInstallRequest);
+    ToolInstallation mvnToolInstallation = mvn.installTool(mavenToolInstallRequest);
     ToolInstallation javaInstallation = java.installTool(javaToolInstallRequest);
 
+    /*
+     * Register the freshly installed mvn on the IDEasy-managed PATH so the IDEasy-controlled maven (from the software
+     * repository) launches the GUI instead of any maven on the system PATH. installTool only installs into the software
+     * repository, so we register the bin directory here (install() would normally do this).
+     */
+    context.getPath().setPath(mvn.getName(), mvnToolInstallation.binDir());
```
hohwille, line 72:
```suggestion
    processContext.withPathEntry(mvnToolInstallation.binDir());
```

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -59,29 +61,47 @@ protected void doRun() {
...
     LOG.debug("Starting GUI via commandlet");
 
     Path pomPath = context.getIdeInstallationPath().resolve("gui/pom.xml");
     if (!Files.exists(pomPath)) {
       throw new CliException("Fatal error: The pom.xml file required for launching the IDEasy GUI could not be found in expected location: " + pomPath);
     }
 
-    List<String> args = List.of(
-        "-f",
+    ArrayList<String> args = new ArrayList<>(List.of(
```
hohwille:
```suggestion
    List<String> args = new ArrayList<>(List.of(
```

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -59,29 +61,47 @@ protected void doRun() {
...
+      args.add("-U"); //Adding this flag forces maven to download the latest SNAPSHOT version, but only if IDEasy is a snapshot version.
+    }
 
     /*
      * We manually update the PATH entry with our java version, as by default IDEasy includes the SymLink under /projectname/software/java/bin in the PATH
      * In case of projects using older Java Versions, this is important as the java version of the project could potentially older.
      */
-    mvn.runTool(processContext.withPathEntry(javaInstallation.binDir()), ProcessMode.BACKGROUND_SILENT, args);
+    try {
+      mvn.runTool(processContext.withPathEntry(javaInstallation.binDir()), ProcessMode.DEFAULT, args);
```
hohwille: It would be better to extract `processContext.withPathEntry(javaInstallation.binDir())` to its own statement out of the `try` block (together with the same thing adding `mvn` to the PATH.

cli/src/main/resources/nls/Help.properties
```diff
@@ -52,6 +52,7 @@ cmd.gradle=Tool commandlet for Gradle (Build-Tool).
 cmd.gradle.detail=Gradle is a build automation tool for Java, Kotlin, and other JVM-based languages. Detailed documentation can be found at https://docs.gradle.org/
 cmd.gui=Tool commandlet for running the IDEasy GUI
 cmd.gui.detail=This command will run the GUI version of IDEasy, opening up a more comprehensible project dashboard and allowing for easier management of your local IDEasy instance.
+cmd.gui.opt.--enableLogging=Run the GUI with enabled logging to the terminal.
```
hohwille: The option as removed, therefore also the help text shall be removed.

```suggestion
```

cli/src/main/resources/nls/Help_de.properties
```diff
@@ -52,6 +52,7 @@ cmd.gradle=Werkzeug Kommando für Gradle (Build-Tool).
 cmd.gradle.detail=Gradle ist ein Build-Automatisierungstool für Java, Kotlin und andere JVM-basierte Sprachen. Detaillierte Dokumentation ist zu finden unter https://docs.gradle.org/
 cmd.gui=Werkzeug Kommando, um die GUI von IDEasy zu öffnen.
 cmd.gui.detail=Dieser Befehl startet die GUI-Version von IDEasy, öffnet ein übersichtlicheres Projekt-Dashboard und erleichtert die Verwaltung Ihrer lokalen IDEasy-Instanz.
+cmd.gui.opt.--enableLogging=Führe die GUI mit aktiviertem Debug-Logging auf das Terminal aus.
```
hohwille:
```suggestion
```

cli/src/test/resources/ide-projects/gui/_ide/installation/gui/pom.xml
```diff
@@ -0,0 +1 @@
+<PlaceholderFile/>
```
hohwille: This is an invalid POM file.
Tests should be consistent otherwise this will later take revenge in maintaining the tests.
Can't you just copy the real gui pom during the test from its real location (package/gui) to the test dir (context.getIdePath().resolve("installation/gui")?

CHANGELOG.adoc
```diff
@@ -27,6 +28,7 @@ The full list of changes for this release can be found in https://github.com/dev
 
 Release with new features and bugfixes:
 
+
```
```suggestion
```

hohwille (approved):
@laim2003 thanks for the update. 👍 
I still have some review remarks but when resolved, we can finally merge.

CHANGELOG.adoc
hohwille: Done.

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
laim2003: Using withPathEntry() will lead to the following error: `Cannot set executable flag on file that does not exist: mvn` and will lead to the GUI launch to fail.

laim2003: I debugged this further and found that Maven is not passed as an absolute executable path, unlike the other tools.

For Maven, `mvn.runTool()` ends up using only:

```java
Path.of("mvn")
```

This means the executable is treated as a relative path. Later, `runTool()` calls `ProcessContextImpl.addExecutable()`, which eventually calls `makeExecutable()`. At that point, `"mvn"` is not a valid path, so the call fails.

The confusing part is that the full Maven binary directory is available, because `mvnToolInstallation.binDir()` returns the correct absolute path. So the path seems to get lost while configuring the process context.

The issue seems to originate here (inside Mvn.java):

```java
@Override
protected void configureToolBinary(ProcessContext pc, ProcessMode processMode) {
  Path mvn = Path.of(getBinaryName()); // resolves only to "mvn"
  Path wrapper = findWrapper(MVN_WRAPPER_FILENAME);
  pc.executable(Objects.requireNonNullElse(wrapper, mvn));
}
```

If no Maven wrapper is found, the executable is set to `"mvn"` instead of the resolved Maven binary path.

This also explains why the issue only occurs for Maven, or for other tools invoked by simple name. `SystemPath.findBinary()` does not search `extraPathEntries` when resolving such a simple executable name. Therefore, binaries added via `withPathEntry()` cannot be resolved in this case.

hohwille: Issue #2286 was already created for this. Thanks.
Then lets clean this up later.
BTW: It is fully intended that the `Mvn` commandlet is not responsible to resolve the absolute path to the `mvn` executable. That is the responsibility of `ProcessContext`.
We first need to fix #2286 and then we can clean this up here.
To prevent blocking this PR due to the nice-to-have cleanup, I will resolve this thread now.

cli/src/test/resources/ide-projects/gui/_ide/installation/gui/pom.xml
hohwille: Even better to remove it and just keep the folder 👍

laim2003:
Retested on 11/08/2026 on Windows. GUI launched succesfully with newest snapshot version. Output log:
```
ide gui
No CVEs found for version 3.9.16 of tool mvn.
No CVEs found for version 26.0.2_10 of tool java.
Version 3.9.16 of tool mvn is already installed
No CVEs found for version 25.0.4_7 of tool java.
Version 25.0.4_7 of tool java is already installed
Launching gui in snapshot mode
[INFO] Scanning for projects...
[INFO]
[INFO] -------------< com.devonfw.tools.IDEasy:ide-gui-launcher >--------------
[INFO] Building ide-gui-launcher 2026.08.001-SNAPSHOT
[INFO]   from pom.xml
[INFO] --------------------------------[ pom ]---------------------------------
Downloading from ossrh: https://central.sonatype.com/repository/maven-snapshots/com/devonfw/tools/IDEasy/ide-gui/2026.08.001-SNAPSHOT/maven-metadata.xml
Downloaded from ossrh: https://central.sonatype.com/repository/maven-snapshots/com/devonfw/tools/IDEasy/ide-gui/2026.08.001-SNAPSHOT/maven-metadata.xml (1.2 kB at 1.1 kB/s)
Downloading from ossrh: https://central.sonatype.com/repository/maven-snapshots/com/devonfw/tools/IDEasy/ide-cli/2026.08.001-SNAPSHOT/maven-metadata.xml
Downloaded from ossrh: https://central.sonatype.com/repository/maven-snapshots/com/devonfw/tools/IDEasy/ide-cli/2026.08.001-SNAPSHOT/maven-metadata.xml (3.2 kB at 6.3 kB/s)
[INFO]
[INFO] --- exec:3.1.0:exec (default-cli) @ ide-gui-launcher ---
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  2.318 s
[INFO] Finished at: 2026-08-11T13:44:32+02:00
[INFO] ------------------------------------------------------------------------
```

PR #1055, opened 2025-02-19
#1053: fix setup #1044: setup rewritten in Java

hohwille:
Fixes #1053 fixes #1044

* rewritten setup in Java (in `IdeasyCommandlet`)
* triggered via `ideasy -f install`
* reduced `setup` bash script to run that
* reduced `setup.bat` to find bash and run `setup`
* emptied migration and solved all in the setup directly
* added `IdeasyCommandletTest`

PR #2037, opened 2026-06-17
#1917: add scripts for gui shortcut launcher windows, linux, macos

### LINKED ISSUE #1917: Allow the creation of a desktop shortcut for the GUI
### Feature idea

As a user, i might want to launch the GUI directly from my desktop or the Windows Start Menu, so i dont have to open the terminal and then type in the gui command manually. 

### Additional context

This is comparable to how Git allows you to launch the GUI via a desktop shortcut:

Image

quando632:
#### Implemented changes

Instead of separate shell scripts (which end users never have access to), the shortcut creation
is now implemented directly in Java inside `IdeasyCommandlet.installIdeasy()`.

- **Linux**: writes `~/.local/share/applications/ideasy-gui.desktop` from a bundled template,
  calls `update-desktop-database` and marks it as trusted via `gio set metadata::trusted`
- **macOS**: creates `~/Applications/IDEasy GUI.command` (a `.command` file is executable by Finder)
- **Windows**: reads the real Desktop path from the `Shell Folders` registry key (handles
  OneDrive-redirected Desktops), then creates `.lnk` shortcuts on the Desktop and in the
  Start Menu via PowerShell `WScript.Shell`

Assets added to the release package (`cli/src/main/package/gui/`):
- `logo.png` — IDEasy logo used as icon on Linux/macOS
- `logo.ico` — pre-generated ICO for Windows shortcuts
- `linux/ideasy-gui.desktop` — `.desktop` template (version injected by Maven at build time)

The old `gui/scripts/` shell scripts are removed — they were never part of the release
distribution and never reached end users.

#### Testing instructions

**Windows**
1. Run `ideasy install` (or `setup.bat` from a fresh download)
2. A shortcut _IDEasy GUI_ appears on the Desktop and in the Start Menu
3. Open the Windows search bar, type _IDEasy GUI_, press Enter → GUI launches
4. Alternatively: double-click the Desktop shortcut → GUI launches

**Linux**
1. Run `ideasy install`
2. `~/.local/share/applications/ideasy-gui.desktop` is created
3. Open your application launcher, search for _IDEasy GUI_ → entry appears and launches the GUI

**macOS**
1. Run `ideasy install`
2. `~/Applications/IDEasy GUI.command` is created
3. Double-click the file in Finder → GUI launches (mark as trusted on first run if prompted)

#### Checklist for this PR

  with `internal`


laert-ll:
```
 laert@arch  ~/projects/IDEasy_1917/workspaces/main/IDEasy/gui/scripts/linux   feature/1917-create-desktop-shortcut ±  ./create-shortcuts.sh          
✓ Created: /home/laert/.local/share/applications/ideasy-gui.desktop
ℹ Launch from: Application Menu or Launcher
✓ Created: /home/laert//ideasy-gui.desktop
ℹ Launch from: Desktop

✓ IDEasy GUI shortcuts ready!

Usage:
  • Open your Application Menu and search for 'IDEasy GUI'
  • Or double-click the shortcut on your Desktop
  • First launch may take longer as Maven downloads dependencies

 laert@arch  ~/projects/IDEasy_1917/workspaces/main/IDEasy/gui/scripts/linux   feature/1917-create-desktop-shortcut ±  /./home/laert/ideasy-gui.desktop
/./home/laert/ideasy-gui.desktop: line 1: [Desktop: command not found
/./home/laert/ideasy-gui.desktop: line 4: GUI: command not found
/./home/laert/ideasy-gui.desktop: line 5: IDEasy: command not found
/./home/laert/ideasy-gui.desktop: line 9: IDE: command not found
```

The script works and the shortcut does appear in the Application Menu, but I cannot seem to open it either from the Launcher or by using the script in the home directory itself.

laert-ll:
Screenshot From 2026-06-18 13-34-26

```
 laert@arch  ~/projects/IDEasy_1917/workspaces/main/IDEasy/gui/scripts/linux   feature/1917-create-desktop-shortcut ±  cat ~/.ideasy-gui.log        
---- Enj 18 Qer 2026 01:33:59.MD CEST ----
Error: IDEasy is not installed or not in PATH
Please install IDEasy first: https://github.com/devonfw/IDEasy#setup
```

As for the `ideasy-gui.desktop` script, I still get the same error message.

.gitignore
```diff
@@ -19,6 +19,9 @@ eclipse-target/
 generated/
```
AdemZarrouki, line 23: these files need to be ignored when commiting and pushing changes

.gitattributes
```diff
@@ -1,5 +1,9 @@
 * eol=lf
 *.bat eol=crlf
```
AdemZarrouki, line 2: these files need to be ignored when commiting and pushing changes

AdemZarrouki (changes requested):
when running the create-shortcuts in windows i get this:
```
$ powershell -File create-shortcuts.ps1
IDEasy GUI Launcher Setup
Project: C:\Users\projects\IDEasy\workspaces\main\IDEasy

Icon generated from devonfw.png
Creating shortcut: C:\Users\OneDrive - Capgemini\Desktop\IDEasy GUI.lnk
Created: C:\Users\OneDrive - Capgemini\Desktop\IDEasy GUI.lnk
Creating shortcut: C:\Users\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\IDEasy GUI.lnk
Created: C:\Users\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\IDEasy GUI.lnk

Setup complete!
```
But when running the gui i get this error


```
Error: IDEasy is not installed or not in PATH
https://github.com/devonfw/IDEasy#setup

Press any key to continue . . .
```

AdemZarrouki:
Tested again in windows after the changes.
The Gui is successfully launched but the side terminal is present even after the gui is launched. Also when closing the GUI the side terminal still present with the same message `Starting IDEasy GUI...`.
grafik

I think we need to change this behavior of the shortcut.

AdemZarrouki:
In macos i get this
```
./create-shortcuts.sh                                                                                                                                                 ─╯
ℹ AppleScript unavailable, falling back to symlink...
✓ Created symlink: /Users/Applications/IDEasy GUI.command
ℹ Launch from: Applications > IDEasy GUI
ℹ AppleScript unavailable, falling back to symlink...
✓ Created symlink: /Users/Desktop/IDEasy GUI.command
ℹ Launch from: Desktop

✓ IDEasy GUI shortcuts ready!

Usage:
  • Open Finder and go to Applications
  • Double-click 'IDEasy GUI' to launch
  • Or double-click the Desktop shortcut
```
But then when opening the shortcut i get this

```
/Users//Projects/IDEasy/workspaces/main/IDEasy/gui/scripts/macos/launch-gui.command ; exit;
Error: IDEasy is not installed
Please install IDEasy first:
https://github.com/devonfw/IDEasy#setup
```

AdemZarrouki:
Tested again after the changes on Windows and macOS. The shortcut was successfully created, and when clicking on it, the GUI launches.
The only issue I see is that it takes 5–7 seconds for the GUI to become visible to the user. This can be addressed in a separate issue (if relevant).

AdemZarrouki (approved):

laert-ll:
I tested the scripts locally on Linux again from the branch checkout and it works as intended. The `create-shortcuts.sh` creates the `.desktop` entry, and in the application menu I'm able to launch the GUI with no issues. Initially I had a lot of difficulties with this, because it turned out `ide gui` itself was not working in my environment in the first place. I'm not sure if `ide gui` is a feature intended to work out of the box or if there's extra steps to be done, but that is outside of the scope of this issue.

Also it is worth noting that the scripts are committed without the executable bit, so I had to `chmod +x` them before running them. Not sure if this is intentional.

quando632:
Yeah you right i have added the correct executable bit for linux now

gui/scripts/linux/create-shortcuts.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/bash
+
+# IDEasy GUI Shortcut Creator for Linux
+# Creates .desktop files for easy GUI launching from the Application Menu and Desktop
+# Supports: GNOME, KDE, XFCE, and other freedesktop-compatible environments
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+LAUNCHER_SCRIPT="$SCRIPT_DIR/launch-gui.sh"
+ICON_PATH="$SCRIPT_DIR/../../src/main/resources/com/devonfw/ide/gui/assets/devonfw.png"
```
hohwille: This is the path during the development.
However you put this script into the release where this path is wrong and the image is not yet available.

gui/scripts/linux/create-shortcuts.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/bash
+
+# IDEasy GUI Shortcut Creator for Linux
+# Creates .desktop files for easy GUI launching from the Application Menu and Desktop
+# Supports: GNOME, KDE, XFCE, and other freedesktop-compatible environments
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
hohwille: `BASH_SOURCE` is used when a script is sourced but this script is supposed to be executed where you need to use `$0` instead.

`source create-shortcut.sh` vs. `./create-shortcut.sh`.

gui/scripts/linux/launch-gui.sh
```diff
@@ -0,0 +1,51 @@
...
+    exit 1
+fi
+
+cd "$PROJECT_ROOT"
+
+# Launch IDE GUI in background and exit immediately
+bash -ic "ide gui" </dev/null &
+disown
+
+exit 0
```
hohwille: As I already said in the daily: We don't need no script to launch the GUI with redundant logic.
You can just call `ide gui` to launch the gui.
To bypass the `ide` function you could also execute `ideasy gui` directly.

hohwille (changes requested):
@quando632 thanks for your PR. 👍 
However, it seems things have been not clear.

Wasn't the expectation of the story #1917 that the start-menu entry and/or desktop shortcut gets installed automatically during the installation of IDEasy?

How would the end-user run `create-shortcuts` script if it is not included in our release that the end-user will download (he usually does not have the source-code of IDEasy)?

Further, sorry, if nobody told you before but after lots of trouble with scripting hell in CMD, Power-Shell, Bash, etc. we have decided to avoid such scripting and do as much as possible in Java avoiding any scripting as much as possible - see for example #1055

Since the scripts are not working for me, I think we need some rework.
I would suggest the following:
* ensure to include our logo into the release, please also use our [IDEasy logo](https://github.com/devonfw/IDEasy/blob/main/documentation/images/logo.png) instead of `devonfw.png` what is not our project logo.
* as the logo will IMHO not change, simply generate the `logo.ico` for windows and also add it to git then you can avoid all the complex transformation scripting logic.
* put the creation of the desktop entry into the Java installation code with OS specific if else coding. In case it is not clear where this is get started here:
https://github.com/devonfw/IDEasy/blob/423b61432b15265e22076e9dbcb1b27cf71c0539/cli/src/main/java/com/devonfw/tools/ide/tool/IdeasyCommandlet.java#L267-L301
* You can still include a static asset for e.g. the linux desktop file `ideasy-gui.desktop` into the [OS specific folder](https://github.com/devonfw/IDEasy/tree/main/cli/src/main/package/system) under a new Linux subfolder. Then you can use maven filtering with placeholders so instead of hardcoding the version to `1.0` you can use
```
Version=$[project.version]
```
See here for how we have configured maven for this:
https://github.com/devonfw/IDEasy/blob/423b61432b15265e22076e9dbcb1b27cf71c0539/cli/pom.xml#L165

cli/src/test/resources/ide-projects/install/project/home/Downloads/ide-cli/gui/linux/ideasy-gui.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Version=test-SNAPSHOT
```
hohwille:
```suggestion
Version=$[project.version]
```

cli/src/test/resources/ide-projects/install/project/home/Downloads/ide-cli/gui/linux/ideasy-gui.desktop
```diff
@@ -0,0 +1,10 @@
...
+Type=Application
+Name=IDEasy GUI
+Comment=Launch IDEasy GUI
+Exec=@IDEASY_BIN@ gui
```
hohwille: Can't we come up with a simple command here that needs no additional replacement?
Something like
```
bash -c "$IDE_ROOT/_ide/installation/bin/ideasy gui"
```

cli/src/test/resources/ide-projects/install/project/home/Downloads/ide-cli/gui/linux/ideasy-gui.desktop
```diff
@@ -0,0 +1,10 @@
...
+Icon=@IDEASY_ICON@
```
hohwille: If Linux distros cannot resolve variables in `.desktop` files, we cannot use `$IDE_ROOT` variable.
But maybe `~/projects/_ide/installation/gui/logo.png` should work.
But if some advanced user would change the default `$IDE_ROOT` this would not work so here your replacement may still be the best option.

cli/src/test/resources/ide-projects/install/project/home/Downloads/ide-cli/gui/logo.ico
```diff
@@ -0,0 +1 @@
+logo.ico mock
```
hohwille: Can't you just remove this `Downloads` folder and instead in your test simply copy `../cli/src/main/package/` to the faked HOME directory and use that as "installationPath"?
Then you do not need to mock resources.
And in case you do a `*.png` or `*.ico` file it should be valid (it may be an empty image with 0x0 pixels but it should be a valid png/ico file then).
Still the easiest and most maintainable solution should be to delete this `Downloads` folder and copy the real files.

hohwille (commented):
@quando632 thanks for your rework. This looks great now. 👍 
Sorry, I could not fully complete the review today. However, I already wanted to provide the review comments I already made.
I guess there will not come more but I still have to complete the review next week.

BTW: For the `logo.png` that we here use as a small icon, should we scale it down and make it smaller? Otherwise we already have the `logo.png` in our git. If we keep it as is can't we avoid duplicating it in git by moving it and changing the references or by including it into the Linux assembly from its existing location?

cli/src/test/resources/ide-projects/install/project/home/Downloads/ide-cli/gui/linux/ideasy-gui.desktop
quando632: The problem with bash -c "$IDE_ROOT/..." is that $IDE_ROOT is set in .bashrc, which is not sourced by graphical desktop launchers (GNOME, KDE, XFCE). The variable is simply empty in that environment

quando632:
We do not need to scale it. The Icon= field in .desktop files can point to any size PNG the desktop environment (GNOME, KDE) automatically scales it for different use cases.

I considered sourcing logo.png directly from gui/src/main/resources/… via an additional <resource> block in cli/pom.xml to avoid the duplicate. I decided against it: the cross-module path reference (../gui/src/…) would couple the two modules' directory structures at the build level, and a future restructuring of the GUI module would silently break the CLI package with no compile error.

cli/src/main/java/com/devonfw/tools/ide/tool/IdeasyCommandlet.java
```diff
@@ -333,6 +335,113 @@ private void installIdeasyWindowsEnv(Path ideRoot, Path installationPath) {
...
+        "HKCU\\Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\Shell Folders", "Desktop");
+    Path desktopPath = (desktopStr != null && !desktopStr.isBlank()) ? Path.of(desktopStr) : this.context.getUserHome().resolve("Desktop");
+    this.context.getFileAccess().mkdirs(desktopPath);
+    createWindowsShortcut(desktopPath.resolve("IDEasy GUI.lnk"), ideasyExe, icoPath);
+    String startMenuStr = helper.getRegistryValue(
+        "HKCU\\Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\Shell Folders", "Programs");
+    Path startMenu = (startMenuStr != null && !startMenuStr.isBlank()) ? Path.of(startMenuStr)
+        : this.context.getUserHome().resolve("AppData\\Roaming\\Microsoft\\Windows\\Start Menu\\Programs");
+    if (Files.isDirectory(startMenu)) {
+      createWindowsShortcut(startMenu.resolve("IDEasy GUI.lnk"), ideasyExe, icoPath);
```
hohwille: One of my early learnings in IT was that special characters including white-spaces in file paths are a pointless pitfall that will sooner or later cause trouble that can easily be avoided (e.g use underscore instead of space).
For simplicity I would just call it `IDEasy.lnk`.

FYI: Just for the record. It seems that Windows not only supports icons for lnk but also for applications directly so we could also consider that `ideasy.exe` has this icon. This is just a note. Feel invited to file a new issue for this FYI comment if you like the idea but nothing to prevent your PR from being merged.
Example:
Image

cli/src/main/package/gui/logo.ico
hohwille, line 1: > We do not need to scale it. The Icon= field in .desktop files can point to any size PNG the desktop environment (GNOME, KDE) automatically scales it for different use cases.
> 
> I considered sourcing logo.png directly from gui/src/main/resources/… via an additional block in cli/pom.xml to avoid the duplicate. I decided against it: the cross-module path reference (../gui/src/…) would couple the two modules' directory structures at the build level, and a future restructuring of the GUI module would silently break the CLI package with no compile error.

I see your point regarding the module dependencies but I do not agree that we just copy copy the existing logo in Git to another place and use it as app icon.
This is because it should not be a copy.
Here we have an application icon that is not the same.
Every application icon I have seen so far has a quadratic size.
We need to design a new application icon here.
IMHO this will look totally silly otherwise...
image
Am I the first one testing this?

I even tested the code to create the lnk like this:
1. I downloaded the logo.ico from GitHub into my Downloads folder.
2. I ran this in PowerShell (copy & pasting from your PR code)
```
$ws = New-Object -ComObject WScript.Shell; $s = $ws.CreateShortcut('IDEasy.lnk'); $s.TargetPath = 'D:\projects\_ide\installation\bin\ideasy.exe'; $s.Arguments = 'gui'; $s.IconLocation = 'C:\Users\hohwille\Downloads\logo.ico'; $s.Save()
```
The icon is not displayed.
IMHO the `logo.ico` is broken.
Maybe it was added + committed to git as text instead of binary.
Don't we first need to add `*.ico` as `binary` to our https://github.com/devonfw/IDEasy/blob/main/.gitattributes and then add+commit the `logo.ico` to git.

hohwille (approved):
@quando632 thanks for the rework. I resolved almost all my concerns/comments. 👍 
However, there are still some last things before we can consider this as done and merge it.

hohwille:
Build fails:
```
[ERROR] Errors: 
[ERROR]   IdeasyCommandletTest.testInstallIdeasyUpdatesIdeRoot:165 » Cli IDEasy release is inconsistent at /home/runner/work/IDEasy/IDEasy/cli/target/ide-projects/install/project/home/Downloads/ide-cli [artifact=gui]
[INFO] 
[ERROR] Tests run: 800, Failures: 0, Errors: 1, Skipped: 4
```

UPDATE: Seems that something was merged on main that caused this error and now all PRs fail when updated from main.

cli/src/main/package/gui/logo.ico
quando632: Hi @hohwille, thanks for the detailed feedback and for actually testing the .lnk creation. 👍

You were right on both points:

1. The old logo.ico was a padded version of the wide logo banner, not a real, quadratic app icon.
2. It didn't render because it only contained a single non-standard size (168×156), so Windows had no proper icon size to display.

I've replaced it with a distinct, square application icon: just the >_ terminal symbol on a dark rounded square or should I better use the logo.png in the gui assets?

cli/src/test/resources/ide-projects/install/project/home/Downloads/ide-cli/gui/linux/ideasy-gui.desktop
hohwille: Thanks for explaining. Totally makes sense.

hohwille (approved):
@quando632 thanks for the update. Now everything is just perfect. Great job 👍 
Ico file also now working fine:
https://raw.githubusercontent.com/quando632/IDEasy/277b43d7e767e806e489978c4b6fb44d257161b7/cli/src/main/package/gui/logo.ico

quando632:
During the restest a bug was found. #2279 adressed the bug.

PR #2055, opened 2026-06-19
#2044: hacky fix for endoption token

### LINKED ISSUE #2044: mvn --<TAB> does not trigger completion
### Actual behavior

When completing Maven commands, `ide mvn --<TAB>` does not show any suggestions.

However:
- `ide mvn -<TAB>` works as expected
- `ide mvn --a<TAB>` also works

Only the exact case `--` fails to trigger completion.

### Reproduce

1. Run: ide mvn -
   → suggestions are shown

2. Run: ide mvn --a
   → suggestions are shown

3. Run: ide mvn --
   → no suggestions are shown

### Expected behavior

`ide mvn --<TAB>` should show all Maven long options, e.g.:

--also-make
--also-make-dependents
--fail-at-end
--fail-fast

### IDEasy status

```bash
IDE_ROOT is set to C:\Users\phhoang\projects
IDE_HOME is set to C:\Users\phhoang\projects\testprojekt
Your version of IDEasy is SNAPSHOT.
Your version of IDEasy is SNAPSHOT but version SNAPSHOT is available. Please run the following command to upgrade to the latest version:
ide upgrade
You are online.
```

### Related/Dependent issues

_No response_

### Comments/Hints

I investigated this and found that `--` is already consumed during the initial parsing phase of the `complete` command before it reaches `CompleteCommandlet.doRun()`.

Specifically:
- The `args` property of the `complete` command is multi-valued and consumes the remaining arguments using `args.next()`.
- During this process, `--` is interpreted as the end-of-options marker by the CLI parser and is consumed instead of being passed as a value.
- As a result, `--` never reaches `this.args.asArray()` and is not available to the completion logic (`CliArguments.ofCompletion(...)`).

This suggests that the parser should not treat `--` as an end-of-options marker in this context (e.g. when it is part of completion payload).

Caylipp:
### Implemented changes:

* Fix handling of a trailing `--` during CLI completion.
* In completion mode, `--` is now preserved as the current completion input instead of being consumed as an end-of-options marker.
* This allows tool commandlets such as `mvn` to provide matching completion candidates for long tool arguments.
* The regular CLI behavior of `--` as end-of-options marker remains unchanged outside of completion input.

---

### Testing instructions

1. Test with configuration args: `complete mvn --`
2. Verify that the following completion candidates are suggested:
   * `--also-make`
   * `--also-make-dependents`
   * `--fail-at-end`
   * `--fail-fast`

---

### Checklist for this PR

  with `internal`


tineff96 (approved):
Reviewed the diff with a focus on the changed CLI completion handling for trailing `--`, especially the changes around `CliArguments`, `Ideasy`, and `Property`.

I manually checked the completion behavior via Maven exec for the following inputs:

```bash
complete mvn --
complete mvn --a
complete mvn --f
complete mvn -- --something
```

For `complete mvn --`, the expected Maven long-option candidates are suggested, including:

```text
--also-make
--also-make-dependents
--fail-at-end
--fail-fast
```

I also ran the relevant automated tests:

```bash
mvn -pl cli -Dtest=CompleteTest,CliArgumentsTest test
mvn -pl cli -Dtest=CompleteTest#testCompleteMavenToolArgumentsForEmptyInput test
mvn -pl cli package -DskipTests
mvn -pl cli test
```

The full CLI test run passed:

```text
Tests run: 739, Failures: 0, Errors: 0, Skipped: 18
BUILD SUCCESS
```

The normal `--` behavior outside completion appears unaffected: it is still treated as the end-of-options marker during regular CLI parsing, while completion mode now preserves a trailing `--` as the current completion input.

Blocking issues: none found.

Non-blocking suggestion: I did not see focused regression tests for the newly covered manual cases such as `complete mvn --`, `complete mvn --a`, and `complete mvn --f`. Adding one or two focused tests for these cases would make this behavior easier to protect against future regressions.

Based on the reviewed diff and the tested cases, the changed behavior looks correct to me. Good work!

cli/src/main/java/com/devonfw/tools/ide/cli/CliArguments.java
```diff
@@ -156,7 +164,9 @@ public CliArgument next() {
    */
   public CliArguments copy() {
 
-    return new CliArguments(this.currentArg, this.endOptions, this.splitShortOpts);
+    CliArguments copy = new CliArguments(this.currentArg, this.endOptions, this.splitShortOpts);
+    copy.preserveEndOptionsToken = this.preserveEndOptionsToken;
```
hohwille: Nothing wrong but...
The multi-args constructor is only used internally as copy-constructor and from other constructor.
It would be cleaner to add the new argument to that constructor.
Then you can also keep this as a 1-liner.

cli/src/main/java/com/devonfw/tools/ide/cli/Ideasy.java
```diff
@@ -103,6 +103,9 @@ private void initContext(CliArguments arguments) {
       String key = current.getKey();
       Property<?> property = contextCommandlet.getOption(key);
       if (property == null) {
+        if ("complete".equals(current.get())) {
```
hohwille: Can you please use a constant from `CompleteCommandlet` to make this more transparent and avoiding a redundant String literal?

cli/src/main/java/com/devonfw/tools/ide/property/Property.java
```diff
@@ -429,10 +429,16 @@ protected boolean applyValue(String argValue, boolean lookahead, CliArguments ar
...
     if (success) {
       if (this.multivalued) {
+        boolean consumedExtra = false;
         while (success && args.hasNext()) {
           CliArgument arg = args.next();
+          consumedExtra = true;
           success = assignValueAsString(arg.get(), context, commandlet);
         }
+        if (!consumedExtra) {
+          args.next();
```
hohwille: I am not sure if I fully understand this change and if it really is correct.
This code is very tricky and when things like this were changed recently we typically fixed one bug but introduced another one.
Can you add an inline comment why we are doing this.
It would give me a much better feeling to merge this after understanding the change ;)

hohwille (approved):
@Caylipp thanks for your PR. I do not see your solution as hacky. I left some small comments but in general think this is perfectly correct and the way to go. Nice work 👍

cli/src/main/java/com/devonfw/tools/ide/property/Property.java
Caylipp: Thanks for the feedback, you were right with your suspicion about the copy constructor.

The previous change in applyValue() was mainly a fix I found while debugging, but after further debugging I realized that it did not address the actual root cause.
The real issue happens when auto-completion forks the argument parsing here:

completeCommandlet(arguments.copy(), cmd, collector);

The copy constructor calls setCurrent(arg). However, before this change preserveEndOptionsToken was only copied after the constructor had already called setCurrent(arg). Since preserveEndOptionsToken was still false at that point, setCurrent(arg) treated "--" as an endoptions token and skipped it again with arg.getNext().

So the initial parsing was not the only problem. The "--" token was also consumed again when the arguments were copied during completion matching.

I changed the copy constructor to pass preserveEndOptionsToken into the constructor before setCurrent(arg) is called. And I removed my changes in applyValue().

cli/src/main/java/com/devonfw/tools/ide/cli/CliArguments.java
Caylipp: I added `preserveEndOptionsToken` to the internal constructor so `copy()` keeps the correct `--` behavior.

cli/src/main/java/com/devonfw/tools/ide/cli/Ideasy.java
Caylipp: I replaced the redundant `"complete"` string literal with `CompleteCommandlet.NAME`.

Caylipp:
tested with 2026.07.001-07_07_06-SNAPSHOT on Windows ide mvn --<tab>

PR #2072, opened 2026-06-24
#2058 fix github authentication for vsccode

### LINKED ISSUE #2058: VSCode authentication not working with IDEasy
### Actual behavior

When I install and launch VSCode via IDEasy, I cannot properly login to services like GitHub for using Copilot AI.
* VSCode is not authenticated
* Ensure GitHub Copilot plugin is installed (install it if not yet installed)
* I select `Continue with GitHub`
* Then VSCode opens a login page in the browser
* I login via GitHub and confirm granting access for Copilot
* This redirects to some localhost URL that I need to confirm to open
* Now this does not connect to the existing VSCode instance but opens a new VSCode window
* The original VSCode window is still not authenticated (hint: when I select `View` > `Command Palette` > `Developer: Reload Window` then the existing VSCode windows reloads and is authenticated)
* I close all windows of VSCode
* I restart VSCode (`ide vscode`) again
* I am not authenticated.

### Reproduce

`ide vscode`
See actual behaviour

### Expected behavior

VSCode should somehow save my authentication so it persisted and after restart of VSCode, I am still logged into GitHub Copilot (please note that GitHub Copilot is just one example, I observe the same problems with Claude code).

### IDEasy status

```bash
Your version of IDEasy is 2026.06.001.
You are online.
```

### Related/Dependent issues

_No response_

### Comments/Hints

Bugs like this are not easy to trace down.
Most likely the root cause of this problem is actually a bug in VSCode itself.
However, for analysis it might be helpful to install VSCode globally on your system (without IDEasy) and follow the same procedure and see if it works correctly.
Observe where VSCode made changes (check users home directory, check windows registry, check windows credential store).
Reproduce the bug with IDEasy. See if probably some changes are made in the wrong location and therefore not picked up.
When the problem is traced further down...
* in case of a bug in IDEasy itself, create a PR that fixes it
* in case of a bug in VSCode itself, file a bug report in VSCode and provide the collected details and instructions how to reproduce the issue in an isolated way without IDEasy but with according commands and instructions (e.g. `code` command to launch VSCode with workspace arguments, etc.).

quando632:
This PR fixes #2058

## Implemented changes

`Vscode.java` can now isolate VSCode via a named `--profile` instead of `--user-data-dir`.

`--user-data-dir` moves VSCode's IPC lock to a custom path. The OS-level `vscode://` protocol handler (used by GitHub/Copilot OAuth callbacks) always targets VSCode's default data directory, so it could never find the running window and instead spawned a second, unauthenticated one.

`--profile` keeps the IPC lock at the default location while still giving each project and workspace its own isolated profile (separate auth sessions, extension state, settings). The profile name is `ideasy-«project»-«workspace»`: it contains the project name in addition to the workspace name because every IDEasy project has a `main` workspace by default and all of them would otherwise share a single profile.

As discussed in the daily, the new behaviour is behind a feature toggle so that it can be merged without risk:

| `VSCODE_PROFILE_ENABLED` | behaviour |
| --- | --- |
| unset or `false` (default) | unchanged: `--user-data-dir=«ideMetadataPath»/config` |
| `true` | new: `--profile=ideasy-«project»-«workspace»` |

Once we are confident that the new behaviour works in all scenarios (multiple projects, multiple workspaces, vscode/vscodium), a follow-up PR can activate it by default and eventually remove both the toggle and the old behaviour.

`--extensions-dir` (per-project plugin isolation) is unaffected in both cases.

**Note:** enabling the toggle starts with a fresh VSCode profile. Settings, extension state and logins from the previous `--user-data-dir` are not migrated, so you have to log in once after enabling it.

### Automated tests

```bash
mvn -pl cli test -Dtest=VscodeTest
```

Both toggle states are covered:

* `testConfigureToolArgsUsesIdeMetadataPathForUserData` - toggle disabled (default), `--user-data-dir` is used
* `testConfigureToolArgsUsesProfileIfFeatureToggleEnabled` - toggle enabled, `--profile` including the project name is used

### 1. Default behaviour is unchanged (toggle disabled)

Do **not** set `VSCODE_PROFILE_ENABLED` and run:

```bash
ide vscode
```

Expected result:

* VSCode starts as before with `--user-data-dir` pointing into the IDE metadata folder.
* Settings, extensions and logins are exactly as before this PR.
* The authentication problem of #2058 still reproduces. This is intended while the toggle is disabled.

### 2. Enable the feature toggle

Add this to `$IDE_HOME/conf/ide.properties` (user specific, not to the shared `settings`):

```properties
VSCODE_PROFILE_ENABLED=true
```

Then run:

```bash
ide vscode
```

Expected result:

* VSCode starts with `--profile=ideasy-«project»-«workspace»` and without any `--user-data-dir`.
* The profile is new and therefore empty, so you are not logged in yet.

### 3. Auth fix verification (toggle enabled)

1. Open Copilot and select **Continue with GitHub**.
2. Complete the browser login flow.

Expected result:

* Auth shows immediately.
* No **Developer: Reload Window** needed.

### 4. Persistence verification (toggle enabled)

1. Close VSCode.
2. Reopen via `ide vscode`.

Expected result:

* Still authenticated.

### 5. Per-project and per-workspace isolation (toggle enabled)

1. Open a second workspace, and additionally a workspace of a *different* IDEasy project.
2. Log in with a different GitHub account.

Expected result:

* Each project and workspace maintains its own isolated authentication state.
* Different GitHub accounts can be used independently.
* In particular two different projects that both use the `main` workspace do **not** share a profile.

### 6. Plugins and settings in the new profile (toggle enabled)

VSCode tracks extension *enablement* per profile, and user level settings live inside the user-data-dir.

Expected result / to be verified:

* Plugins installed by IDEasy (via `--extensions-dir`) are present **and enabled** in the new profile.
* Workspace settings from `.vscode/settings.json` (e.g. `java.jdt.ls.java.home`) are still applied, since VSCode reads them from the opened workspace folder independent of the profile.
* User level settings that lived inside the old user-data-dir are no longer read. Currently these are only telemetry opt-outs. If we want to keep them, they have to be migrated in the settings repo as a follow-up.

### Checklist for this PR

  with `internal`


laert-ll:
Tested it locally and the problem is fixed!

I have a question/concern however: if I understand this correctly, removing `--user-data-dir` means also that VSCodes per-workspace data is not isolated anymore, and all the projects share the same user settings, keybindings, cache, themes etc. Is this intended behavior? Or is there maybe an alternative to this?

quando632:
Yeah you are right. I have changed it to `--profile`. Now the VSCodes per-workspace data should be isolated. Can you check if it is working correctly now?

laert-ll:
Tested it again locally and everything works as intended!

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -115,9 +115,11 @@ protected void configureToolArgs(ProcessContext pc, ProcessMode processMode, Lis
     if (this.context.getSystemInfo().isWsl()) {
       pc.withEnvVar("DONT_PROMPT_WSL_INSTALL", "1");
     }
-    Path vsCodeConf = this.context.getWorkspacePath().resolve(".vscode/.userdata");
     pc.addArg("--new-window");
-    pc.addArg("--user-data-dir=" + vsCodeConf);
+    // Use a named profile (not --user-data-dir) so VS Code keeps its IPC lock at the default location.
+    // This lets the OS-level vscode:// protocol handler (OAuth callbacks e.g. GitHub/Copilot) find the
+    // already-running IDEasy window. Each workspace gets its own profile for isolated auth and settings.
+    pc.addArg("--profile=ideasy-" + this.context.getWorkspaceName());
```
hohwille, line 93: I do not see how the profile gets configured.
We are just providing a profile called `ideasy-main` (independent of the current IDEasy project).
How would VSCode now know that for this profile it should load the settings from our workspace where we are configuring them?
Can you check that the location to Java that we configure [here](https://github.com/devonfw/ide-settings/blob/main/vscode/workspace/update/.vscode/settings.json) would be picked up by VSCode?
IMHO this approach is not working or incomplete.
In case I am missing something, please let me know...

quando632: VS Code always reads these workspace settings (.vscode/settings.json)  from the opened workspace folder, completely independent of `--profile`  or `--user-data-dir`. The `java.jdt.ls.java.home` setting lives here and will continue to be picked up correctly.

The telemetry settings of the User/profile-level settings in `.vscode/.userdata/User/settings.json` were previously read via `--user-data-dir`. With `--profile`, VS Code's user-data-dir points to `%APPDATA%\Code\`, so that file is no longer read. However, these are purely telemetry opt-outs VS Code and all workspace functionality (Java, auth, extensions) works correctly without them. Migrating them to workspace settings should be done in the settings repo as a follow-up.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -86,9 +86,11 @@ protected void configureToolArgs(ProcessContext pc, ProcessMode processMode, Lis
...
```
hohwille: 1. The profile IMHO also needs to include the project name in order to be unique. I have many different projects on my machine and all have a main workspace and currently would use the same `ideasy-main` profile for IMHO different things.
2. We discussed this PR in the daily and agreed that we love to go into this direction. However, there is a risk that this will break things and it is not easy to test all scenarios (multiple projects, multiple workspaces, vscode/vscodium, etc.). Therefore the conclusion was to use a new variable (e.g. `VSCODE_PROFILE_FEATURE`) as a feature toggle in this PR. If that variable is not set, the old behaviour remains as is. But if the variable is set, then your new behaviour gets activated. This way we can merge your cool feature without any risk. Then we can all test it for several weeks. If everybody comes to the conclusion that it works better with the feature toggled active, then we can create another PR to activate this by default and even consider removing the toggle and the old behaviour entirely.

quando632: **1. Profile name.** You are right, `ideasy-main` was not unique: `main` is the default workspace name (`IdeContext.WORKSPACE_MAIN`), so every project on the machine would have shared one profile. The name is now `ideasy-«project»-«workspace»`.

**2. Feature toggle.** Added as `VSCODE_PROFILE_ENABLED` (`VariableDefinitionBoolean`, default `false`). If it is not set, the previous `--user-data-dir=«ideMetadataPath»/config` behaviour is used unchanged; if set to `true`, the new `--profile` behaviour is activated. I picked `..._ENABLED` to match `IDE_XML_MERGE_LEGACY_SUPPORT_ENABLED` and `IDE_WRITE_LOGFILE`, if you prefer `VSCODE_PROFILE_FEATURE` I will rename it.

Both toggle states are covered by tests (`testConfigureToolArgsUsesIdeMetadataPathForUserData`, `testConfigureToolArgsUsesProfileIfFeatureToggleEnabled`), the variable is documented in `variables.adoc`, and I merged current `main` (#2142 moved the user data into the metadata folder, which conflicted here, it also added an `ArgCapturingProcessContext` test helper, so I dropped my duplicate one)

Two things worth watching during the test phase:

- User-level settings that live inside the user-data-dir are no longer read when running with `--profile`. Today that is only the telemetry opt-outs, but if we ever put real user-level settings there they would need to move to workspace settings or be seeded into the profile.
- VSCode tracks extension *enablement* per profile. `--extensions-dir` is unchanged, but plugins installed by IDEasy may show up disabled in a freshly created profile this should be verified explicitly.

Note for anyone testing: enabling the toggle starts from an empty profile, existing logins and settings from the old `--user-data-dir` are not migrated. This is documented in `variables.adoc`.

documentation/variables.adoc
```diff
@@ -46,4 +46,5 @@ See also link:https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/
 |`INTELLIJ_VM_ARGS`|e.g. `-Xms128m -Xmx2048m`|Support for extension or overwrite of default IntelliJ JVM options.
 |`PYCHARM_VM_ARGS`|e.g. `-Xms128m -Xmx2048m`|Support for extension or overwrite of default Pycharm JVM options.
 |`ANDROID_STUDIO_VM_ARGS`|e.g. `-Xms128m -Xmx2048m`|Support for extension or overwrite of default Android Studio JVM options.
+|`VSCODE_PROFILE_ENABLED`|e.g. `true`|Feature toggle (disabled by default) to isolate VSCode via a named `--profile` (`ideasy-«project»-«workspace»`) instead of `--user-data-dir`. This keeps the VSCode IPC lock at its default location so that OAuth callbacks via the `vscode://` protocol handler (e.g. GitHub/Copilot login) reach the already running window instead of opening a second one - see issue https://github.com/devonfw/IDEasy/issues/2058[#2058]. Note that enabling this starts with a fresh VSCode profile: settings, extension state and logins from the previous `--user-data-dir` are not migrated.
```
hohwille: I would not document this property.
It is just a temporary feature toggle that will not exist for long.
I would not even have created a constant in `IdeVariables` for it that you must not forget to remove in the next PR when this new behaviour is well tested and becomes the new default.


```suggestion
```

hohwille (approved):
@quando632 thanks for your update with the feature toggle. Now ready for merge. 👍

CHANGELOG.adoc
```diff
@@ -6,6 +6,7 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/2058[#2058]: Fix github authentication for vscode
```
hohwille: Lets also remove this from the CHANGELOG and only add it once the feature toggle is removed and the new feature becomes the default.

```suggestion
```

quando632:
Retest on SNAPSHOT `2026.09.002-09_16_07-SNAPSHOT` - Windows 11: passed.

The only issue found is that plugins installed by IDEasy are not enabled in the freshly created profile. This is a known, minor issue and is already fixed by #2479 (pending review), so no separate bug issue is created here.

Issue #1166, opened 2025-03-25
Automatic project import for IntelliJ


hohwille:
As described in #1164 we want to have automatic project import also for IntelliJ.

Hint: via [repository configuration](https://github.com/devonfw/IDEasy/blob/master/documentation/repository.adoc) and by setting `import=intellij` this feature should be triggered.

Currently it is not implemented at all.
https://github.com/devonfw/IDEasy/blob/bbe3a82883d0889fb244fd4a4c665cf043e3e102/cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java#L111-L114

In Intellij class this method needs to be overridden and implemented properly.

I could not find any CLI feature in IntelliJ to archive this. Therefore this task is a little tricky.
My suggestion would be to do that manually:
1. Create a backup of your `workspaces/main/.intellij` and `workspaces/main/.idea` folders
2. Open your IntelliJ (`ide intellij`)
3. Manually import the project by right-clicking the `pom.xml` in your repository folder inside the workspace (`workspaces/main/«my-repo»/pom.xml`) and selecting `+ Add as Maven project` from the context menu - in case we want to support this first for maven (for gradle use `build.gradle`).
4. Check that your project is now properly recognized and build by IntelliJ (what we consider as "imported")
5. Close your IntelliJ
6. Do a recursive diff of the backup (from step 1.) with the current state of the IJ config in your workspace
7. Identify the relevant difference(s) in the config file(s)
8. Implement the `importRepository` method to apply that changes in the according intellij configuration file(s).

Issue #218, opened 2024-02-21
Add possibility to start a command in a new window with the process builder


MustaphaOuchen:
Follow up of https://github.com/devonfw/IDEasy/pull/200:

Create a new constant in `ProcessMode` called `BACKGROUND_NEW_WINDOW` with the following behavior:

"Open a new window with the output (could be done in the way it is implemented currently with CMD on windows but on linux/mac we need to figure out how to do that as we need to find the according terminal app and on MacOS there is not only the default MacOS Terminal but als iTerm2). Feel free to leave this option out for the moment but I can see this quite valuable for the future as well. "

Example cmd in windows:
`
...
   this.context.warning(
          "Currently starting the process as background process in windows will use 'more' command to redirect output to a new cmd window!");

      this.processBuilder.redirectOutput(Redirect.PIPE).redirectError(Redirect.PIPE);
      List<String> windowsArgs = List.of("cmd.exe", "/c", "start", "cmd.exe", "/k");

      String currentCommandToRunInCmd = this.arguments.stream().map(Object::toString).collect(Collectors.joining(" "));
      currentCommandToRunInCmd += " | more";

      this.arguments.clear();

      List<String> newArgs = new ArrayList<>(windowsArgs);
      newArgs.add(currentCommandToRunInCmd);

      this.arguments.addAll(newArgs);
...
`

PR #2073, opened 2026-06-25
#1676 import extra sdks automatically into ide 

### LINKED ISSUE #1676: Import extra SDKs automatically into IDE
### Feature idea

With #1298 we now have the ability to get extra installations of tools such as `java`, `dotnet`, etc.
Now if I make use of this feature in my project and run my IDE (Eclipse, IntelliJ, VSCode, etc.), I would love to have these extra versions of e.g. `java` configured automatically by IDEasy.

### Additional context

Similar to #1166 we could merge according tools using a specific add-on template.
Some things to consider:
* we should access our custom tools Java model to find what is configured
* in the IDE commandlet we should verify that the according extra tool installations actually exist before we try to add them
* the for each such extra installation we should merge the according config into the workspace config file of the IDE.
* the SDK instance should use the name configured by the IDEasy admin user in the JSON file. Special attention is needed that nobody configures an extra installation of `«tool»` (e.g. `java`), that he also names `«tool»`. Instead extra installations should have semantical names briefly describing their purpose (e.g. `frontend`, `backend`, `process-server`, etc.). Otherwise if you create an extra installation of `java` called `java` then the ID `java` is already taken for the main installation of `java` and the extra installation would then also have the same name while the name should be unique. We might consider adding some `extra-` prefix implicitly but I would rather not do that and encourage the end-users to take reasonable decisions to make thins work properly.

vivu001:
#### 1. Register the IntelliJ template
* Add the IntelliJ template for extra JDKs installations.
* Keep the approach reusable for other IDEs, such as _Eclipse_ and _VS Code_.

#### 2. Read extra tool installations
* Read extra tool installations from `$IDE_HOME/settings/ide-extra-tools.json`.
For example: 
``` 
{
  "java": {
    "client": {
      "version": "11.0.31_11",
	   "edition": "azul"
    },
    "process-engine": {
      "version": "21.*"
    }
  },
  "mvn": {
    "m4": {
      "version": "4.0.0-rc-5"
    }
  }
}
```
* Process each configured extra SDKs installation.
 
#### 3. Keep SDKs synchronized
* Synchronize extra SDKs during `ide create` or `ide update` and before the IDE starts.
* Configure the workspace only after tools are installed.
* Use merge logic to update SDKs without duplicates (Related PR https://github.com/devonfw/ide-settings/pull/91).
 
#### 4. Validate configuration and handle errors
* Reject an extra-installation name that equals its tool name; for example, an extra JDKs installation must not be named `java`.
* Skip missing or invalid SDK installations and write a warning to the log.

#### 5. Added GraalVM reflection metadata to [reflect-config.json](https://github.com/devonfw/IDEasy/pull/2073/changes/0cf86fb6c7b91551f1b95c8e5c61d93469345dae)
- The new entries enable native-image to instantiate `Intellij` via `IdeToolCommandlet`-based lookup.
---

#### 1. Build the native executable
- Run the following script to build the native image: [./build-local-dev.sh](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/run-ideasy-locally.adoc#run-ideasy-locally-as-native-binary-using-the-local-dev-script). 
- After the build completes, navigate to: `IDEasy/cli/target`.

#### 2. Create a test project with custom settings
- From `IDEasy/cli/target`, create a new test project using the modified settings from my test branch for `ide-settings`.

```
ideasy.exe create <test-import-sdks> https://github.com/vivu001/ide-settings.git#feature/1676-Test-import-extra-SDKs-automatically-into-IDE
```

#### 3. Verify automatic SDK import in IntelliJ

- Open a new terminal and initialize the workspace:
```
icd -p <test-import-sdks> -w
```
- Launch IntelliJ
```
ide intellij
```
- Expected result: 
All additional SDKs should be automatically imported into IntelliJ.
This step validates that workspace configuration and IDE settings are correctly applied (as IDEasy manages IDE configuration via workspace templates).

<div align="center">
image
</div>

#### 4. Validate `ide update` behavior

4.1 Prepare test scenario
- Manually remove all extra JDKs from IntelliJ.
- Additionally, delete jdks from the filesystem:
```
$IDE_HOME/software/java 
$IDE_ROOT/_ide/software/default/java
```

4.2 Run update
```
ide update
```
4.3 Verify result

All previously removed SDKs are re-downloaded and reinstalled.
IntelliJ configuration is updated accordingly, as shown at the end of step 3.
The file `$IDE_HOME/workspaces/main/.intellij/config/options/jdk.table.xml` has been updated with the extra SDK entries.

### Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -373,6 +376,28 @@ private void installExtraToolInstallations(String tool, List<ExtraToolInstallati
...
+    Intellij intellij = this.context.getCommandletManager().getCommandlet(Intellij.class);
+    if (intellij != null) {
+      this.context.newStep("Synchronize IntelliJ extra SDKs").run(() -> intellij.synchronizeExtraToolInstallations(workspacePath));
+    }
+
+    // later:
+    // Eclipse eclipse = ...
+    // VsCode vscode = ...
+  }
+
```
hohwille: IMHO this is the wrong place to trigger this:
The workspace will just be the current workspace what is typically `main` for `ide update` and will always be for `ide create`.
However, when I start `ide intellij` the workspace configuration is updated and IntelliJ is started in the currnet workspace that may differ from the one used by `ide update`.
Therefore you need to apply the changes when running the IDE.
This also gives us better SoC (separation of concerns):
With the current approach, the `AbstractUpdateCommandlet` needs to know about all IDE commandlets that can handle extra tools. Instead we should only trigger this inside the according IDE commandlet like Intellij, when the workspace is synchronized. So simply extend this method:
https://github.com/devonfw/IDEasy/blob/c0773bf5848afcf14abdb765f9aedcc10246c22e/cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java#L76

I will add the idea as extra review comment to `IdeToolCommandlet`.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -158,16 +198,147 @@ private void mergeConfig(Path repositoryPath, String configFilePath) {
...
+   * @param config the IntelliJ-specific import configuration for the tool.
+   */
+  private void mergeExtraToolInstallation(Path workspacePath, String tool, ExtraToolInstallation extraInstallation,
+      IntellijExtraSdkConfig config) {
+
+    String name = extraInstallation.name();
+    if ((config.reservedName() != null) && config.reservedName().equalsIgnoreCase(name)) {
+      LOG.warn("Skipping IntelliJ import for extra {} installation '{}': name conflicts with main {} installation.", tool, name, tool);
+      return;
+    }
```
hohwille: Valid point: You want to prevent that an extra installation of `<tool>` (e.g. `java`) also has the name `<tool>`:
```
{
  "java": {
    "java": { "version":"11.0.27_6", "edition":"azul" }
  }
}
```

I fully agree but IMHO we should add this as a validation to `ExtraTools` (or `ExtraToolsMapper`) so it already fails fast.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -158,16 +198,147 @@ private void mergeConfig(Path repositoryPath, String configFilePath) {
...
+
+    Path templateFile = this.context.getSettingsPath()
+        .resolve(TEMPLATE_LOCATION)
+        .resolve(config.templateFile());
+    if (!Files.exists(templateFile)) {
+      throw new CliException(
+          "Cannot import extra " + tool + " installation into IntelliJ: template file not found at " + templateFile + "\n"
+              + "Please do an upstream merge of your settings git repository.");
+    }
```
hohwille: Dont do this:
This is likely to happen for existing projects that have already forked their own ide-settings.
Since you do not run this code inside an extra `step`, any exception, will terminate the entire process.
That means users will be blocked and cannot start IntelliJ any more with `ide intellij`.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
hohwille, line 125: here add the following call above:

```suggestion
    synchronizeExtraToolInstallations();
    if (errors == 0) {
```

hohwille, line 43:
```suggestion
  private final Map<String, Set<Path>> extraSdkMap;
```

hohwille, line 56:
```suggestion
    assert (hasIde(tags));
    this.extraSdkMap = new HashMap<>();
```

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
```diff
@@ -131,4 +131,18 @@ public void importRepository(Path repositoryPath) {
...
+   * <p>
+   * The default implementation does nothing. IDEs such as IntelliJ may override this hook to import extra SDK/tool installations configured via
+   * {@code settings/ide-extra-tools.json}.
+   * </p>
+   *
+   * @param workspacePath the workspace root whose IDE configuration should be synchronized.
+   */
+  public void synchronizeExtraToolInstallations(Path workspacePath) {
+    // default no-op
+  }
```
hohwille, line 250:
```suggestion

  /**
   * TODO: add JavaDoc please
   */
  protected void registerExtraSdkTemplate(String sdk, Path relativeTemplatePath) {
    Set<Path> templatePaths = this.extraSdkMap.computeIfAbsent(sdk, HashSet::new);
    templatePaths.add(relativeTemplatePath);
  }
    
  /**
   * Synchronizes extra IDEasy tool installations into the current IDE workspace configuration if supported.
   *
   * <p>
   * By default nothing will happen. Your IDE commandlet has to register one or more according templates in its constructor.
   * </p>
   */
  protected void synchronizeExtraToolInstallations() {
    ExtraTools extraTools = ExtraToolsMapper.get().loadJsonFromFolder(this.context.getSettingsPath());
    if (extraTools == null) {
      return;
    }
    for (String sdk : extraTools.getSortedToolNames()) {
      Set<Path> templatePaths = this.extraSdkMap.get(sdk);
      if ((templatePaths == null) || templatePaths.isEmpty()) {
        LOG.debug("Skipping import of extra tool {} into {} because not configured or supported.", sdk, this.tool);
        continue;
      }
      List<ExtraToolInstallation> extraInstallations = extraTools.getExtraInstallations(sdk);
      synchronizeExtraToolInstallation(sdk, templatePaths, extraInstallations);
    }    
  }
  
  private void synchronizeExtraToolInstallation(String sdk, Set<Path> templatePaths, List<ExtraToolInstallation> extraInstallations) {
    for (Path templatePath : templatePaths) {
      Path workspaceFile = this.context.getWorkspacePath().resolve(templatePath);
      Path templateFile = this.context.getSettingsPath().resolve(this.tool).resolve(IdeContext.FOLDER_WORKSPACE)
        .resolve("repository") // TODO: create and use constant IdeContext.FOLDER_REPOSITORY - this string literal exists at least 16 times in our code base
        .resolve(templatePath);
      if (Files.exists(templateFile)) {
        for (ExtraToolInstallation extraInstallation : extraInstallations) {
          synchronizeExtraToolInstallation(sdk, templateFile, workspaceFile, extraInstallation);
        }
      } else {
        LOG.warn("You are missing a template file at {}.");
        IdeLogLevel.INTERACTION.log(LOG, "Please ask the IDEasy admin in your project to merge your settings with upstream.");
      }
    }  
  }
  
  private void synchronizeExtraToolInstallation(String sdk, Path templateFile, Path workspaceFile, ExtraToolInstallation installation) {
    String name = installation.name();
    Path extraToolHome = this.context.getSoftwareExtraPath().resolve(tool).resolve(name);
    if (!Files.isDirectory(extraToolHome)) {
      LOG.warn("Skipping extra tool installation import to {} because it is missing at {}", this.tool, extraToolHome);
      IdeLogLevel.INTERACTION.log(LOG, "Please run the following command to fix:\nide update");
      return;
    }
    ExtensibleEnvironmentVariables environmentVariables = new ExtensibleEnvironmentVariables(
      (AbstractEnvironmentVariables) this.context.getVariables().getParent(), this.context);
    String variablePrefix = "EXTRA_" + tool.toUpperCase(Locale.ROOT);
    environmentVariables.setValue(variablePrefix + "_NAME", name);
    environmentVariables.setValue(variablePrefix + "_HOME", extraToolHome.toString().replace('\\', '/'));
    environmentVariables.setValue(variablePrefix + "_VERSION", installation.version().toString());
    if (extraInstallation.edition() != null) {
      environmentVariables.setValue(variablePrefix + "_EDITION", installation.edition());
    }
    
    XmlMerger xmlMerger = new XmlMerger(this.context);
    XmlMergeDocument workspaceDocument = xmlMerger.load(workspaceFile);
    XmlMergeDocument templateDocument = xmlMerger.loadAndResolve(templateFile, environmentVariables);
    Document mergedDocument = xmlMerger.merge(templateDocument, workspaceDocument, false);
    xmlMerger.save(mergedDocument, workspaceFile);        
  }
```

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -53,6 +57,30 @@ public class Intellij extends IdeaBasedIdeToolCommandlet {
...
+   *
+   * <p>
+   * Only tools contained in this map are imported automatically into IntelliJ. This keeps the import logic generic for IntelliJ while allowing support to be
+   * added incrementally tool by tool.
+   * </p>
+   */
+  private static final Map<String, IntellijExtraSdkConfig> EXTRA_TOOL_CONFIGS = Map.of(
+      "java", new IntellijExtraSdkConfig("jdk-extra-java.xml", ".intellij/config/options/jdk.table.xml", "java")
+  );
+
```
hohwille: Nice that you solved this in a generic way.
IMHO this generic pattern is not specific for IntelliJ.
I suggested how to solve this fully generic in `IdeToolCommandlet`.

```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
hohwille, line 63: Now all that we need in IntelliJ is the following:
```suggestion
    super(context, "intellij", Set.of(Tag.INTELLIJ));
    registerExtraSdkTemplate("java", ".intellij/config/options/jdk.table.xml");
```

ATTENTION:
But you need to move the template again in `ide-settings` by another PR:
Instead of `intellij/workspace/repository/.idea/jdk-extra-java.xml` it should then be `intellij/workspace/repository/.intellij/config/options/jdk.table.xml`.

IMHO all other changes to IntelliJ below can be reverted.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
```diff
@@ -158,16 +198,147 @@ private void mergeConfig(Path repositoryPath, String configFilePath) {
...
+        buildDescriptorFound = true;
+        break;
       }
     }
-    LOG.warn("No supported build descriptor was found for project import in {}", repositoryPath);
+    if (!buildDescriptorFound) {
+      LOG.warn("No supported build descriptor was found for project import in {}", repositoryPath);
+    }
+
+    importExtraToolInstallations(repositoryPath);
```
hohwille: This is a misunderstanding.
In #1166 we want to import some repository into Intellij.
Therefore we need to figure out the workspace of that repository in order to change the workspace where the repository is located (`getWorkspacePath(repositoryPath)`).

In #1676 we want to ensure that all extra SDKs are configured in the IJ config of the current workspace before we actually launch IntelliJ (`this.context.getWorkspacePath()`).

hohwille (commented):
@vivu001 thanks for your PR and sorry to interrupt before team review.
I had a look into your diff and found some misunderstandings and before moving any further, I wanted to guide the way to the proper solution.
You did a lot perfectly correct, but there are some misunderstandings between #1166 and #1676 that seemed not clear.
Also preparing this implementation for VSCode, Eclipse and other potential IDEs or even future SDKs I suggest some more generic approach and according refactoring.
Please have a look and see if you can follow my idea and suggestions.
If there are any questions do not hesitate to reach out.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
vivu001: I also thought this method is called at the correct place. But actually, IntelliJ’s workspace sync runs before the `software/extra/java/...` folders exist, so it cannot import the new sdks. 

```
 "java": {
    "client": {
      "version": "11.0.31_11",
      "edition": "azul"
    },
    "process-engine": {
      "version": "21.*"
    }
  }
  ...
```
The warning is therefore shown some thing like that:

```
Skipping extra tool installation import to intellij because it is missing at [$IDE_HOME]\software\extra\java\client.
Please run the following command to fix: 
ide update

Skipping extra tool installation import to intellij because it is missing at [$IDE_HOME]\software\extra\java\process-engine.
Please run the following command to fix: 
ide update
```


But I’m struggling to find out how to fix it. :smiley:

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
vivu001: I created a new PR for `ide-settings` at https://github.com/devonfw/ide-settings/pull/91.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
vivu001: This has been removed in the latest commits.

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
vivu001: I've fixed it in the latest commits.

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
vivu001: I've addressed it in the latest commits. Thanks for the suggestion :blush:

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
vivu001: I think this problem is caused by configuring the workspace before the extra SDKs are installed. Instead, workspace synchronization should run after installation, so IntelliJ only tries to import SDKs that already exist on disk.

https://github.com/devonfw/IDEasy/blob/365f379a17135cbd245fa672c5a6c769329f7538/cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java#L66-L71

We should call it after the extra tools have been installed, not during workspace setup, so the SDK paths already exist when IntelliJ config is merged.

https://github.com/devonfw/IDEasy/blob/365f379a17135cbd245fa672c5a6c769329f7538/cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java#L344-L352

So, I made these changes in commit: https://github.com/devonfw/IDEasy/pull/2073/changes/a108c0787de2f42fc588bb99fb0c29ea6005ec8f

cli/src/main/java/com/devonfw/tools/ide/tool/extra/ExtraTools.java
```diff
@@ -51,6 +51,12 @@ public List<ExtraToolInstallation> getExtraInstallations(String tool) {
    */
   public void addExtraInstallations(String tool, ExtraToolInstallation extraInstallation) {
 
+    if ((tool != null) && (extraInstallation != null) && tool.equalsIgnoreCase(extraInstallation.name())) {
+      throw new IllegalArgumentException(
+          "Invalid extra installation name '" + extraInstallation.name() + "' for tool '" + tool
+              + "': the extra installation name must not be the same as the tool name.");
+    }
+
```
JoelAdbu, line 64: It might be worth considering handling the IllegalArgumentException here. Otherwise, the update command could fail due to a user misconfiguration.
One option could be to wrap the following call in a try-catch block in the caller (ExtraToolsJsonDeserializer#setProperty(...)):

```java
this.result.addExtraInstallations(tool, installation);
```
and log the validation error from the exception instead of aborting the entire update process. An error-level log message could be appropriate in this case.

JoelAdbu:
I went through the testing instructions and was able to reproduce the expected behavior successfully. The extra SDKs were imported correctly, and ide update reinstalled and synchronized the SDKs as expected.
Apart from my comment regarding the exception handling, everything looks good to me.

cli/src/main/java/com/devonfw/tools/ide/tool/extra/ExtraTools.java
vivu001: Hi @JoelAdbu, thanks for the suggestion.

I’d prefer not to catch this in `ExtraToolsJsonDeserializer#setProperty(...)`. This is configuration validation, so an `IllegalArgumentException` here indicates the user's `ide-extra-tools.json` is inconsistent with the expected model. If we swallow it and continue, we end up with a partially applied update: some extra installations are processed, while others are silently skipped, and the resulting workspace state no longer matches the declared configuration.
  
Failing fast is safer here because it keeps the system in a predictable state and makes the problem immediately obvious. The current load path already handles JSON-loading errors at the file boundary, so we still get a single top-level failure for the config file. 
https://github.com/devonfw/IDEasy/blob/a88c41ee2c6166911d55121ddc758e32518761a4/cli/src/main/java/com/devonfw/tools/ide/json/JsonObjectMapper.java#L52-L56

If we want better UX, I’d rather improve that outer exception message or logging so the user can clearly see which file and entry are invalid, instead of hiding the error at the property level.

JoelAdbu: Thanks for the explanation. Your reasoning around fail-fast behaviour and avoiding partially applied configurations makes sense to me.

hohwille: Simply replace the exception with a LOG.warning statement.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
hohwille: We should discuss this together to resolve it.
First, we should distinguish the following aspects:
* IDE installation (triggered by `ide create`, `ide update`, `ide install «my-ide»`). I was expecting that for this the workspace does not get configured/populated at all. If I am wrong, we need to understand if this is actually done on purpose or by accident. Maybe because when we install a plugin we are actually launching the IDE and already need some workspace configuration that redirects the IDE to use the project specific plugins directory.
* IDE workspace configuration (configureWorksapce). I expected this to be only triggered, when I run the IDE (e.g. `ide intellij` but not `ide install intellij`).

If we also need to configure the workspace during the installation, we should consider the following things:
* We should be forgiving so if an extra SDK is configured but not yet installed (existing in the expected directory), we should ignore that SDK and continue. This can also happen under other circumstances - e.g. if I update the `ide-extra-tools.json` and immediately run `ide intellij` without running `ide update` we have the same situation. If this is triggered from actually launching the IDE, we could log a warning in such case.
* We should properly orchestrate the order of the complex process of `ide create` and `ide update` so that extra tools get installed early enough to prevent problems. However, during the installation this will not really hurt and should also not render a warning (if the SDK is missing when the workspace was configured without actually the IDE being launched). Whenever I actually launch the IDE, the workspace will be configured again and potential problems will be fixed.

I hope now the picture gets clear.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
```diff
@@ -131,4 +157,86 @@ public void importRepository(Path repositoryPath) {
...
+      List<ExtraToolInstallation> extraInstallations = extraTools.getExtraInstallations(sdk);
+      synchronizeExtraToolInstallation(sdk, templatePaths, extraInstallations);
+    }
+  }
+
+  private void synchronizeExtraToolInstallation(String sdk, Set<Path> templatePaths, List<ExtraToolInstallation> extraInstallations) {
+    for (Path templatePath : templatePaths) {
+      Path workspaceFile = this.context.getWorkspacePath().resolve(templatePath);
+      Path templateFile = this.context.getSettingsPath().resolve(this.tool).resolve(IdeContext.FOLDER_WORKSPACE)
+          .resolve("repository") // TODO: create and use constant IdeContext.FOLDER_REPOSITORY - this string literal exists at least 16 times in our code base
```
hohwille: You should also address the TODO 😄

cli/src/main/java/com/devonfw/tools/ide/tool/intellij/Intellij.java
hohwille: Thank you 👍 This review comment can be resolved when we have merged the PR. I left a review comment also in that extra PR. Please have a look.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
```diff
@@ -66,10 +82,17 @@ public ProcessResult runTool(List<String> args) {
   @Override
   public ToolInstallation install(ToolInstallRequest request) {
 
-    configureWorkspace();
```
hohwille, line 69: Ah great finding. So the big picture gets clearer.
But as I already wrote maybe this was done on purpose.
Can you test that without this still the plugins get installed into the correct plugins folder inside the IDEasy project and are found as installed when I launch `ide intellij`?
If that is the case, then this should solve the problem and we simply do not need to configure the workspace at all during installation and you can also remove the extra loop in `AbstractUpdateCommandlet`.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
```diff
@@ -66,10 +82,17 @@ public ProcessResult runTool(List<String> args) {
...
   }
 
+  /**
+   * Synchronizes extra tool installations into the IDE workspace.
+   * This is called after extra tools are installed during ide create/update and before launching the IDE.
+   */
+  public void synchronizeExtraTools() {
+    configureWorkspace();
+  }
+
```
hohwille, line 100: This does not make sense to me: You are creating a new method `synchronizeExtraTools()` that does not only synchronizeExtraTools but triggers the workspace configuration (so more or less you have a copy of `configureWorkspace()` with a new name). Then just call `configureWorkspace()` directly and remove this method.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
vivu001: I tested it: without `configureWorkspace()` in `IdeToolCommandlet.install()` the extra SDK installation is still unaffected. The extra SDKs are still installed by the create/update flow and the workspace synchronization is only needed later when `ide intellij` imports them.

In the `AbstractUpdateCommandlet`, removing the extra loop shows that extra SDKs still work when launching `ide intellij`, but the plugins are NOT installed into `$IDE_HOME/plugins/intellij`. So this workspace sync is still needed for plugin provisioning, even though the SDK part can already be handled at IDE launch.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
vivu001: I tested the current behavior. Extra SDKs can be handled at IDE launch, but plugin provisioning still depends on installation-time workspace sync, otherwise plugins are not installed into intellij. 

So this appears to be the current implementation behavior, not something I can confirm as intentional. If I am right, we can create a new issue from here to separate plugin provisioning from workspace synchronization.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
hohwille: What is the current state and how can we finally resolve this.
IMHO this change is logically correct and also if noting is left in this method you can even remove the method override since it is pointless now.

However, as you also found out and confirmed the plugin installation requires somehow the workspace configuration to be triggered.
Reading your PR diff again, I now found that you are triggering this in `IdeaBasedToolCommandlet` when a plugin is to be installed.
However, then we are still not done:
1. [ ] How about other IDEs like VSCode or Eclipse - can we install plugins there without `configureWorkspace()` being invoked. Or is this only an IDEA/JetBrains specific thing not needed for other IDEs (because during `configureWorkspace()` we create the `idea.properties` in the workspace what makes this different from non-IDEA based IDEs)? IMHO for Eclipse and VSCode we will now never configure the workspace any more when we change your PR as is so it would be broken.
2. [ ] Performance impact: If we install 20 plugins, we will now configure the workspace 20 times. This seems to be a waste and in that case the old approach to trigger it once during the installation still seems more reasonable unless we somehow can remember that we already configured the workspace and do not rerun it for every plugin. Besides if there is an error or warning for a workspace template the user will be annoyed if he gets the same message 20 times so it is not only about performance.

I really like to merge this PR finally but IMHO this is still not 100% cleared to the end.
Maybe we have to clarify this at the end of the next daily since review ping-pong seems to take way too long...

vivu001: Thanks for the review and for suggesting that we align once more before finalizing the implementation of this issue. Based on your feedback, I made a few [changes](https://github.com/devonfw/IDEasy/pull/2073/changes/3b254b8bedef16d61338bfdb096d0a48d26b3b9e) to address your concerns. Hopefully, we'll be able to finalize this issue at the next daily.

- Generalized workspace configuration: a `runTool()` override was added to the `IdeToolCommandlet`. It calls `configureWorkspace()` only when the IDE is launched. This allows _Eclipse_ and _VS Code_ to configure their workspace at launch without any IDE‑specific code.
I also removed the `install()` override from `IdeToolCommandlet` because it does not add behavior anymore.
- Removed IDEA-specific workspace configuration: the IDEA-specific `configureWorkspace()` was called from `IdeaBasedIdeToolCommandlet.runTool()`, preventing workspace setup from being triggered on every plugin installation.

In addition, to ensure the update flow always performs synchronization: the loop that calls `configureWorkspace()` after extra‑tool installation was moved out of the `if (extraTools != null) {}` block in `AbstractUpdateCommandlet`. The method now runs on every `ide update` and `synchronizeExtraToolInstallations()` simply becomes a no‑op when there are no extra tools.

### Summary of current behavior by IDE

| Feature | IntelliJ / IDEA | Eclipse | VS Code |
|----------|----------------|----------|---------|
| Workspace config on IDE launch (`ide <ide>`) | ✅ via `IdeToolCommandlet.runTool()` | ✅ via `IdeToolCommandlet.runTool()` | ✅ via `IdeToolCommandlet.runTool()` |
| Workspace config during plugin install | ✅ skipped (plugin install uses `DEFAULT`) | ✅ skipped | ✅ skipped |
| Extra-SDKs sync on IDE update | ✅ always invoked after extra-tools install | ✅ (same) | ✅ (same) |
| Extra-SDKs template registration | ✅ (e.g. Java → `jdk.table.xml`) | ❌ none registered yet | ❌ none registered yet |

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java
```diff
@@ -41,13 +41,13 @@ public RepositoryCommandlet(IdeContext context) {
...
     addKeyword("setup");
-    this.repository = add(new RepositoryProperty("", false, "repository"));
+    this.repository = add(new RepositoryProperty("", false, IdeContext.FOLDER_REPOSITORY));
   }
 
   @Override
   public String getName() {
 
-    return "repository";
+    return IdeContext.FOLDER_REPOSITORY;
```
hohwille, line 50: Great that you did a global refactoring extracting this constant.
However the name of the repository commandlet is not exactly the same thing as the `FOLDER_REPOSITORY` so if we would change the value of that constant in the future, we do not also want to change the commandlet and property name here.

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
```diff
@@ -64,10 +79,12 @@ public ProcessResult runTool(List<String> args) {
...
   @Override
-  public ToolInstallation install(ToolInstallRequest request) {
+  public ProcessResult runTool(ProcessContext pc, ProcessMode processMode, List<String> args) {
 
-    configureWorkspace();
-    return super.install(request);
+    if (processMode == ProcessMode.BACKGROUND) {
+      configureWorkspace();
+    }
+    return super.runTool(pc, processMode, args);
```
hohwille, line 87: You found a good solution for this. However, assuming `ProcessMode.BACKGROUND` could be a little fragile.
I am not expecting that we may install plugins in background in the future, but maybe we change this in `IdeToolCommandlet`:
```
  public ProcessResult runTool(List<String> args) {

    return runTool(ProcessMode.BACKGROUND, null, args);
  }
```

We have issues like #218 and we might then want to use the new mode for all IDEs changing the above code.
So all I am suggesting here is to make this more robust:

```suggestion
    if ((processMode != null) && processMode.isBackground()) {
      configureWorkspace();
    }
    return super.runTool(pc, processMode, args);
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -350,6 +351,18 @@ private void doUpdateSoftwareStep(Step step) {
...
+    // synchronizeExtraToolInstallations() handles the no-op case internally)
+    List<String> ides = IdeVariables.CREATE_START_SCRIPTS.get(this.context);
+    if (ides != null) {
+      for (String ideName : ides) {
+        ToolCommandlet ideCommandlet = commandletManager.getToolCommandlet(ideName);
+        if (ideCommandlet instanceof IdeToolCommandlet) {
+          ((IdeToolCommandlet) ideCommandlet).configureWorkspace();
+        }
+      }
+    }
```
hohwille, line 365: I am snot getting this.
The variable `CREATE_START_SCRIPTS` should only trigger what already `AbstractUpdateCommandlet.createStartScripts()` does.
It has nothing to do with `configureWorkspace` what will happen, when the end user runs the generated start script.

```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
hohwille: I was not so sure what for IntelliJ the cell `✅ skipped (plugin install uses DEFAULT)` finally means and if maybe the plugins will be installed globally in home directory and then be missing when the IDE gets configured and started later.

I therefore cloned your feature-branch and installed a local release using our excellent `./build-local-dev.sh` script.
Then I created a new test project:
```
$ ide create extra-sdks https://github.com/devonfw/ide-settings.git#feature/test-pr-2073
```
Then I did this:
```
$ ide install intellij
No CVEs found for version 2026.2.0.1 of tool intellij.
Resetting all installed plugins...
Start: Install plugin PlantUMLIntegration (1/1)
Successfully installed plugin: PlantUMLIntegration
Successfully ended step 'Install plugin PlantUMLIntegration (1/1)'.
Successfully installed intellij in version 2026.2.0.1 at D:\projects\extra-sdks\software\intellij
Successfully completed ide (install,intellij)
```

This revealed exactly the bug I proposed before:
```
$ ls -la plugins/intellij/
total 0
drwxr-xr-x 1 hohwille 1049089 0 Aug 10 16:36 .
drwxr-xr-x 1 hohwille 1049089 0 Aug 10 16:36 ..
```

When I now run Intellij the plugin PlantUMLIntegration is not preinstalled.

I conclude that we need to fix this properly and cannot include this into the current release.

hohwille (changes requested):
Sorry, forgot to submit the review

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
JoelAdbu: This can be fixed by running configureWorkspace in the postInstall() step of PluginBasedCommandlet.
Something like:

```java
 @Override
  protected void postInstall(ToolInstallRequest request) {
    super.postInstall(request);
    Path pluginsInstallationPath = getPluginsInstallationPath();
    if (!request.isAlreadyInstalled() || this.forcePluginReinstall.isTrue()) {
      LOG.info("Resetting all installed plugins...");
      deleteAllPlugins(pluginsInstallationPath);
    }
    this.context.getFileAccess().mkdirs(pluginsInstallationPath);
    // the fix.
    if (this instanceof IdeToolCommandlet) {
      ((IdeToolCommandlet) this).configureWorkspace();
    }
    installPlugins(request.getProcessContext());
  }
```

This way, workspace configuration is executed once per IDE during installation, before plugin installation starts — instead of being triggered for each individual plugin. This ensures the IDE configuration (e.g., idea.properties) is in place so plugins get installed to the correct location.
This should solve the problem.

JoelAdbu:
This PR will be continued here: https://github.com/devonfw/IDEasy/pull/2320

PR #2085, opened 2026-06-29
#854: Fix Rancher Desktop path

### LINKED ISSUE #854: Kubectl not working on linux
Kubectl doesnt work on Linux (tested on Ubuntu 24.04.1 LTS)

1. execute `ide kubectl`

### Comments/Hints:

```
ide kubectl
We need to run the following privileged command(s):
curl -s https://download.opensuse.org/repositories/isv:/Rancher:/stable/deb/Release.key | gpg --dearmor | sudo dd status=none of=/usr/share/keyrings/isv-rancher-stable-archive-keyring.gpg
echo 'deb [signed-by=/usr/share/keyrings/isv-rancher-stable-archive-keyring.gpg] https://download.opensuse.org/repositories/isv:/Rancher:/stable/deb/ ./' | sudo dd status=none of=/etc/apt/sources.list.d/isv-rancher-stable.list
sudo apt update
sudo apt install -y --allow-downgrades rancher-desktop=1.16.0*
This will require root permissions!
Hit:1 https://download.opensuse.org/repositories/isv:/Rancher:/stable/deb ./ InRelease
Hit:2 http://security.ubuntu.com/ubuntu noble-security InRelease
Hit:3 http://archive.ubuntu.com/ubuntu noble InRelease
Hit:4 http://archive.ubuntu.com/ubuntu noble-updates InRelease
Hit:5 http://archive.ubuntu.com/ubuntu noble-backports InRelease
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
56 packages can be upgraded. Run 'apt list --upgradable' to see them.
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
rancher-desktop is already the newest version (1.16.0-2.1).
Selected version '1.16.0-2.1' (isv:Rancher:stable:download.opensuse.org [amd64]) for 'rancher-desktop'
0 upgraded, 0 newly installed, 0 to remove and 56 not upgraded.
An unexpected error occurred!
https://github.com/devonfw/IDEasy/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=IllegalStateException%3A+Running+command+%27kubectl%27+failed%3A+Cannot+run+program+%22kubectl%22%3A+error%3D2%2C+No+such+file+or+directory
java.lang.IllegalStateException: Running command 'kubectl' failed: Cannot run program "kubectl": error=2, No such file or directory
        at com.devonfw.tools.ide.process.ProcessContextImpl.run(ProcessContextImpl.java:210)
        at com.devonfw.tools.ide.tool.ToolCommandlet.runTool(ToolCommandlet.java:169)
        at com.devonfw.tools.ide.tool.ToolCommandlet.runTool(ToolCommandlet.java:152)
        at com.devonfw.tools.ide.tool.ToolCommandlet.runTool(ToolCommandlet.java:139)
        at com.devonfw.tools.ide.tool.ToolCommandlet.run(ToolCommandlet.java:130)
        at com.devonfw.tools.ide.context.AbstractIdeContext.applyAndRun(AbstractIdeContext.java:867)
        at com.devonfw.tools.ide.context.AbstractIdeContext.run(AbstractIdeContext.java:798)
Caused by: java.io.IOException: Cannot run program "kubectl": error=2, No such file or directory
        at java.base@21.0.2/java.lang.ProcessBuilder.start(ProcessBuilder.java:1170)
        at com.devonfw.tools.ide.process.ProcessContextImpl.run(ProcessContextImpl.java:174)
        ... 10 more
Caused by: java.io.IOException: error=2, No such file or directory
        at java.base@21.0.2/java.lang.ProcessImpl.forkAndExec(Native Method)
        at java.base@21.0.2/java.lang.ProcessImpl.<init>(ProcessImpl.java:295)
        at java.base@21.0.2/java.lang.ProcessImpl.start(ProcessImpl.java:225)
        at java.base@21.0.2/java.lang.ProcessBuilder.start(ProcessBuilder.java:1126)
        ... 12 more


```

laert-ll:
### Implemented changes:

* Added fallback path resolution for Rancher Desktop CLI tools in `~/.rd/bin`.
* Added warning message for first time install.

Clarification: Rancher Desktop creates `~/.rd/bin` and links its CLI tools (docker, kubectl, rdctl) there only after its first GUI launch. Therefore in order to launch docker or kubectl using IDEasy, we added a warning message for when the user tries to launch it without having launched the GUI first.

---

### Testing instructions

1. Have a working Linux environment running.
2. Run `ide kubectl` and verify that the installation is successful that the warning appears.
3. Open Rancher Desktop from the application menu.
4. Run `ide kubectl` again and verify that it runs successfully.
5. Repeat for `docker`.

---

### Checklist for this PR

  with `internal`


laert-ll:
Thank you for testing this @oanding-blrng and raising the virtualization issue with WSL! 

After some discussion, we came to the conclusion that it's currently not possible to launch `rancher-desktop` on WSL due to virtualization problems and therefore the symlink creation cannot be simulated there. However this part doesn't need testing IMHO due to the fact that its behavior is already confirmed [here](https://docs.rancherdesktop.io/faq/#:~:text=From%20versions%201.3.0%20and%20above%2C%20we%20no%20longer%20create%20symlinks%20in%20/usr/local/bin%20but%20in%20~/.rd/bin). In order to test that specific part, I would simply link the binaries manually to where `rancher-desktop` would normally link them:

```
mkdir -p ~/.rd/bin
ln -s /opt/rancher-desktop/resources/resources/linux/bin/kubectl ~/.rd/bin/kubectl
ln -s /opt/rancher-desktop/resources/resources/linux/bin/docker  ~/.rd/bin/docker
```

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -75,6 +103,17 @@ public boolean isExtract() {
     };
   }
 
+  @Override
+  protected ToolInstallation doInstall(ToolInstallRequest request) {
+
+    ToolInstallation installation = super.doInstall(request);
+    if (this.context.getSystemInfo().isLinux() && !Files.isDirectory(getRancherDesktopBinDir())) {
```
maybeec, line 110: **Blocker** - this aborts for every Linux user whose container runtime is not Rancher Desktop.

The condition only asks "is this Linux and is `~/.rd/bin` missing". It never asks whether Rancher Desktop is the runtime at all. Docker Desktop / plain docker on Linux is a supported setup in this very class - `getInstalledVersion()` below has `case LINUX -> getDockerDesktopVersionLinux()` (line 149) which greps `apt list --installed | grep docker-desktop`. For those users `~/.rd/bin` never exists, so `ide docker ps` now always exits 2 where it previously just ran.

And it is not a one-time message: `GlobalToolCommandlet.doInstall` returns from the Linux package-manager branch (`GlobalToolCommandlet.java:131-140`) *before* the already-installed short-circuit at `:145`, so `doInstall` runs on every `ide docker` / `ide kubectl` call. `ide install docker` also now exits non-zero although the install succeeded, and the `installation` computed on the line above is discarded.

The PR description calls this a warning - please make it one. `install()` already has the precedent for exactly this kind of "installed, but you must do something before it works" situation (`ToolCommandlet.java:352-358` logs a warning and returns). Suggested direction:

```java
ToolInstallation installation = super.doInstall(request);
if (this.context.getSystemInfo().isLinux() && isRancherDesktopInstalled() && !Files.isDirectory(getRancherDesktopBinDir())) {
  LOG.warn("Rancher Desktop has been installed but not launched yet. Please start Rancher Desktop once so that it sets up its "
      + "command-line tools (docker, kubectl, ...) in {}, then re-run your command.", getRancherDesktopBinDir());
}
return installation;
```

Why this matters beyond the regression: `CliException` is meant for expected user-facing aborts, and an installation that actually succeeded is not an abort - swallowing a successful install into a non-zero exit hides the real state from both the user and any script wrapping `ide install`.

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -48,23 +53,46 @@ public String getBinaryName() {
...
+   *
+   * @param command the name of the Rancher Desktop CLI to resolve.
+   * @return the {@code command} unchanged if available on PATH, otherwise its absolute path inside {@code ~/.rd/bin} or {@code null} if not found.
+   */
+  private String resolveRancherDesktopCommand(String command) {
+    if (isCommandAvailable(command)) {
+      return command;
+    }
+    Path rancherDesktopBinary = getRancherDesktopBinDir().resolve(command);
+    if (Files.exists(rancherDesktopBinary)) {
```
maybeec, line 86: **Should-fix** - `Files.exists` misses the executable extension, so this fallback is dead code on Windows.

Rancher Desktop creates `%USERPROFILE%\.rd\bin` on Windows too, but the binary there is `docker.exe`, and `dir.resolve("docker")` will not match it. `SystemPath` already encapsulates this: `EXTENSION_PRIORITY` (`SystemPath.java:53`) and `findBinaryInOrder` (`SystemPath.java:225-241`) try `.exe`/`.cmd`/`.bat`/`.msi`/`.ps1`/`""` on Windows and only `""` elsewhere. `findBinary(Path)` handles the "path with a parent" case (`SystemPath.java:299-306`) and returns the argument instance unchanged when nothing was found - the same idiom `hasBinaryOnPath` uses at `:247-251`.

Reusing it also keeps this consistent with `isCommandAvailable` right above, which goes through the very same class:

```suggestion
    Path candidate = getRancherDesktopBinDir().resolve(command);
    Path rancherDesktopBinary = this.context.getPath().findBinary(candidate);
    if (rancherDesktopBinary != candidate) {
```

(Per `coding-conventions.adoc` :: reuse of existing infrastructure - hand-rolled path probing next to a class that already does it correctly per OS is how the backslash/extension bugs keep coming back.)

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -48,23 +53,46 @@ public String getBinaryName() {
...
+      return rancherDesktopBinary.toString();
+    }
+    return null;
+  }
+
+  private Path getRancherDesktopBinDir() {
+    return this.context.getUserHome().resolve(".rd").resolve("bin");
```
maybeec, line 93: **Should-fix** - magic literals. `"bin"` already has a constant: `IdeContext.FOLDER_BIN` (`IdeContext.java:128`), used e.g. in `MacOsHelper.java:130`. And `".rd"` is repeated in `KubeCtl.java:38`, so it should become a documented `static final` constant rather than a literal in two classes.

```suggestion
  private Path getRancherDesktopBinDir() {
    return this.context.getUserHome().resolve(FOLDER_RANCHER_DESKTOP).resolve(IdeContext.FOLDER_BIN);
  }
```

with e.g.

```java
  /** The hidden folder Rancher Desktop creates in the users home directory on its first launch, containing its CLI tools. */
  private static final String FOLDER_RANCHER_DESKTOP = ".rd";
```

See `coding-conventions.adoc` :: Constants - the point is that the next person who needs this folder finds the existing name instead of typing a third literal.

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -48,23 +53,46 @@ public String getBinaryName() {
...
 
   private boolean isRancherDesktopInstalled() {
-    return isCommandAvailable("rdctl");
+    return resolveRancherDesktopCommand("rdctl") != null;
   }
 
   private String detectContainerRuntime() {
-    if (isCommandAvailable(this.tool)) {
-      return this.tool;
+    String docker = resolveRancherDesktopCommand(this.tool);
```
maybeec, line 64: **Minor** - the variable is named `docker` but it holds whatever container runtime was resolved, which is the whole point of the method it lives in (`detectContainerRuntime`). `String runtime` (or `containerRuntime`) says what it is. See `coding-conventions.adoc` :: Naming.

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -48,23 +53,46 @@ public String getBinaryName() {
   }
 
   private boolean isDockerInstalled() {
-    return isCommandAvailable("docker");
+    return resolveRancherDesktopCommand("docker") != null;
```
maybeec, line 56: **Minor** - naming: `resolveRancherDesktopCommand("docker")` reads as "resolve the Rancher Desktop docker", but `docker` here may just as well be Docker Desktop or the docker engine found on the PATH - only the *fallback* is Rancher-specific. Something like `resolveCommandWithRancherDesktopFallback` (or a neutral `resolveCommand`) would describe what the method actually does. Same for the call at line 64. Non-blocking, but the current name is what made the `doInstall` guard above look narrower than it is.

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -128,7 +167,8 @@ private VersionIdentifier getDockerDesktopVersionLinux() {
 
   private VersionIdentifier getRancherDesktopClientVersion() {
 
-    String output = this.context.newProcess().runAndGetSingleOutput("rdctl", "version");
+    String rdctl = resolveRancherDesktopCommand("rdctl");
```
maybeec, line 170: **Minor** - `resolveRancherDesktopCommand` is documented to return `null` when nothing is found, and that `null` goes straight into `runAndGetSingleOutput(rdctl, "version")`. It is safe today only because the single caller guards with `isRancherDesktopInstalled()` in `getInstalledVersion()` - an invariant that is invisible from here and easy to break later. Either assert/guard it locally, or have this method take the already-resolved command as a parameter so the guard and the use sit next to each other.

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -28,7 +33,7 @@ public class Docker extends GlobalToolCommandlet {
 
   private static final String PODMAN = "podman";
 
-  private static final Pattern RDCTL_CLIENT_VERSION_PATTERN = Pattern.compile("client version:\\s*v([\\d.]+)", Pattern.CASE_INSENSITIVE);
+  private static final Pattern RDCTL_CLIENT_VERSION_PATTERN = Pattern.compile("client version:\\s*v?([\\d.]+)", Pattern.CASE_INSENSITIVE);
```
maybeec, line 36: Good catch on the `v?` - `rdctl version` does not always prefix the client version with `v`, and without this the pattern silently yields no version. Since this is the kind of thing that regresses invisibly, please pin it with a small unit test on `resolveVersionWithPattern` covering both `Client version: v1.16.0` and `Client version: 1.16.0` (`AGENTS.md`: a bug fix starts with a failing test). **Should-fix**, as part of the missing-tests item in the summary.

cli/src/main/java/com/devonfw/tools/ide/tool/kubectl/KubeCtl.java
```diff
@@ -27,6 +29,19 @@ public KubeCtl(IdeContext context) {
     super(context, "kubectl", Set.of(Tag.KUBERNETES), Docker.class);
   }
 
+  @Override
+  protected String getBinaryName() {
```
maybeec, line 33: **Should-fix** - this duplicates `Docker.resolveRancherDesktopCommand` (`Docker.java:81-94`), and the two copies already disagree: `Docker` returns `null` when nothing is found, this one returns `this.tool`. Duplicated logic in two commandlets drifts apart on the next fix - the `v?` regex bug in this same PR is an example of what "we only fixed one of the copies" looks like.

Please move it up into one place rather than copying it: either a `protected` helper on `ToolCommandlet` next to `isCommandAvailable` (`ToolCommandlet.java:1066`), or a small package-visible `RancherDesktop` helper that both `Docker` and `KubeCtl` delegate to. Then `KubeCtl.getBinaryName()` becomes:

```java
  @Override
  protected String getBinaryName() {

    String binary = resolveCommandWithRancherDesktopFallback(this.tool);
    return (binary == null) ? this.tool : binary;
  }
```

Also note this override is `protected` while `Docker` declares the same method `public` (`Docker.java:51`) - worth aligning to the base class visibility (`ToolCommandlet.java:145`) while you are in here.

cli/src/main/java/com/devonfw/tools/ide/tool/kubectl/KubeCtl.java
```diff
@@ -27,6 +29,19 @@ public KubeCtl(IdeContext context) {
...
+
+    if (isCommandAvailable(this.tool)) {
+      return this.tool;
+    }
+    Path rancherDesktopBinary = this.context.getUserHome().resolve(".rd").resolve("bin").resolve(this.tool);
```
maybeec, line 38: **Should-fix** - same two points as in `Docker.java`: (a) `Files.exists` will not find `kubectl.exe` on Windows, where Rancher Desktop also populates `%USERPROFILE%\.rd\bin` - go through `this.context.getPath().findBinary(Path)` which applies `SystemPath.EXTENSION_PRIORITY` (`SystemPath.java:53`, `:225-241`); (b) `".rd"` and `"bin"` are literals here for the second time - `IdeContext.FOLDER_BIN` exists (`IdeContext.java:128`) and `.rd` should become a shared constant. Both disappear once this method delegates to the shared helper.

cli/src/main/java/com/devonfw/tools/ide/tool/kubectl/KubeCtl.java
```diff
@@ -27,6 +29,19 @@ public KubeCtl(IdeContext context) {
...
+    if (Files.exists(rancherDesktopBinary)) {
+      return rancherDesktopBinary.toString();
+    }
+    return this.tool;
+  }
+
   @Override
   public VersionIdentifier getInstalledVersion() {
```
maybeec, line 46: **Should-fix** - this method was not migrated, so the class is left half-fixed. It checks `isCommandAvailable(this.tool)` and then executes `this.tool` directly, both of which bypass the `~/.rd/bin` fallback you just added in `getBinaryName()`. Concretely: for a user whose `kubectl` only exists in `~/.rd/bin`, `ide kubectl` will now work but `ide get-version kubectl` still falls through to `super.getInstalledVersion()` and reports the delegate's (docker's) version.

Switching both to `getBinaryName()` keeps the class consistent:

```java
    String binary = getBinaryName();
    List<String> outputs = this.context.newProcess().runAndGetOutput(binary, "version", "--client");
```

(and drop the `isCommandAvailable(this.tool)` early-return in favour of the same resolution, so there is one notion of "where is kubectl" in this class).

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -75,6 +103,17 @@ public boolean isExtract() {
     };
   }
 
+  @Override
+  protected ToolInstallation doInstall(ToolInstallRequest request) {
+
+    ToolInstallation installation = super.doInstall(request);
+    if (this.context.getSystemInfo().isLinux() && !Files.isDirectory(getRancherDesktopBinDir())) {
+      throw new CliException("Rancher Desktop has been installed but not launched yet. Please start Rancher Desktop once so that it sets up its "
+          + "command-line tools (docker, kubectl, ...) in ~/.rd/bin, then re-run your command.", 2);
```
maybeec, line 112: **Minor** - the exit code `2` is a magic number with no meaning attached anywhere in the codebase; the only other `CliException` with an explicit code uses `1` (`AbstractVersionOrEditionGetCommandlet.java:84`). If this stays an exception at all (see the Blocker above), either drop the argument and use the `CliException(String)` constructor, or introduce a named constant explaining what `2` signals to a caller/script.

maybeec (changes requested):
Thanks for digging into this one :+1: The root-cause analysis is spot on and well documented: Rancher Desktop stopped symlinking into `/usr/local/bin` and links into `~/.rd/bin` since 1.3.0, and IDEasy never looked there. Relaxing `RDCTL_CLIENT_VERSION_PATTERN` to `v?` is a good catch, the JavaDoc on `resolveRancherDesktopCommand` actually explains the *why* and not just the *what*, the CHANGELOG entry sits under the correct milestone `2026.08.001`, the branch is up to date with `main`, CI is green and the CLA is signed. Also thanks for the very concrete testing instructions and for the manual-linking recipe in the discussion with @oanding-blrng.

However, there is one change I do not think we can merge as is, plus a set of convention items.

## Blocker

**1. The new `doInstall` abort breaks every Linux user who is not on Rancher Desktop** (`Docker.java:106-115`).

The guard is `isLinux() && !Files.isDirectory(~/.rd/bin)` - it does not check *which* container runtime is actually in play. This repo explicitly supports Docker Desktop / plain docker on Linux: `Docker.getInstalledVersion()` has `case LINUX -> getDockerDesktopVersionLinux()` (`Docker.java:149`), which greps `apt list --installed | grep docker-desktop`. For such a user `~/.rd/bin` will never exist, so `ide docker ps` - which worked before this PR - now always aborts with exit code 2. That is a functional regression on a supported platform, which is what `coding-conventions.adoc` :: Refactorings is about: never silently break something that already shipped.

Two more things fall out of the same block:

* `GlobalToolCommandlet.doInstall` returns from the package-manager branch (`GlobalToolCommandlet.java:131-140`) *before* the already-installed check at `:145`. So on Linux `doInstall` runs on **every** `ide docker` / `ide kubectl` invocation - the abort is not a one-off first-install message, it fires on every single call until the GUI has been launched.
* `ide install docker` now exits non-zero even though the installation genuinely succeeded, and the `installation` value returned by `super.doInstall(request)` is computed and then thrown away.

The PR description says "Added warning message for first time install" - but the implementation is a hard abort, not a warning. Please make the code match the intent: log a warning (`IdeLogLevel.WARNING` / `LOG.warn`, as `install()` already does for the async-installation case in `ToolCommandlet.java:352-358`) and return the `installation`, and scope the check to the case where Rancher Desktop is actually the runtime (e.g. `isRancherDesktopInstalled()` or the configured edition being `rancher`). With the `~/.rd/bin` fallback in place the subsequent run will then either work or fail with the runtime's own error, instead of IDEasy pre-emptively refusing to run.

## Should-fix

2. **Windows/macOS are not handled by the fallback.** `Files.exists(dir.resolve(command))` cannot find `docker.exe`. Rancher Desktop creates `%USERPROFILE%\.rd\bin` on Windows as well, so the fallback silently does nothing there. `SystemPath` already solves exactly this - see `EXTENSION_PRIORITY` and `findBinaryInOrder` (`SystemPath.java:53`, `:225-241`) - so reuse `this.context.getPath().findBinary(Path)` instead of a raw `Files.exists`. Details inline.

3. **The resolution logic is duplicated** between `Docker.resolveRancherDesktopCommand` and `KubeCtl.getBinaryName`, with two different contracts (`null` vs. `this.tool` as fallback). Duplicated logic across two commandlets drifts apart on the next fix - it should live in one place.

4. **Magic string literals** `".rd"` and `"bin"` in two classes. `IdeContext.FOLDER_BIN` already exists (`IdeContext.java:128`) and `.rd` deserves a documented constant next to it - see `coding-conventions.adoc` :: Constants.

5. **`KubeCtl.getInstalledVersion()` was not migrated** (`KubeCtl.java:46-54`). It still probes and executes `this.tool` directly, so `ide get-version kubectl` keeps missing the `~/.rd/bin` fallback this PR just introduced - the class is left half-migrated.

6. **No tests.** Per `AGENTS.md`, a bug fix starts with a failing test. This one is testable without touching the OS: `GlobalToolCommandletTest` already exists and even carries a Rancher-flavoured dummy commandlet, and `IdeTestContext` gives you a controlled user home, so "binary only present in `<userHome>/.rd/bin`" and "`~/.rd/bin` missing" are both reproducible and OS-independent. The `v?` regex relaxation is likewise a two-line unit test that would pin the bug you fixed. Coveralls reports "no coverable lines changed", which is the same signal.

7. **`getBinaryName()` may now return an absolute path**, which widens a contract documented as "the name of the binary executable for this tool" (`ToolCommandlet.java:143-148`). Callers assume a bare name: `Path.of(getBinaryName())` (`ToolCommandlet.java:306`, `:995`, `GlobalToolCommandlet.java:221`), `findBinaryPathByName(...)` (`:156`) and `MacOsHelper.findLinkDir(rootDir, getBinaryName())` (`:579`, which does `child.resolve(tool)` in `MacOsHelper.java:184`). It happens to hold together today, but if we widen the contract we should say so in the base-class JavaDoc so the next person does not break it.

8. **Manual QA is required before merge.** CI cannot exercise any of this: the fallback path, the first-launch message and the symlink setup all depend on a real Rancher Desktop install. Green CI is not sufficient here - please confirm a manual pass on Linux, and at minimum a smoke test on Windows and macOS that the changed `getBinaryName()`/`getInstalledVersion()` paths still behave (Rancher Desktop uses `~/.rd/bin` on macOS too). Also: issue #854 still has no milestone, which the DoD requires before it is closed.

## Minor

See the inline comments on naming (`String docker` holding a runtime, `resolveRancherDesktopCommand` being called for non-Rancher commands), the possible `null` command handed to `runAndGetSingleOutput`, and the magic exit code `2`.

## Related

@hohwille linked #2171 (RefrEnv) from this issue. That one attacks the same root cause from the other side - refreshing the shell environment after a global install so `~/.rd/bin` lands on the PATH properly. Worth a sentence in the PR description on how the two relate, so we do not end up with two overlapping mechanisms.

References: `documentation/contributing/coding-conventions.adoc`, `documentation/contributing/junit-testing.adoc`, `documentation/contributing/DoD.adoc`.

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
laert-ll: This variable only ever holds the resolved docker command (either docker or its path under `~/.rd/bin`) and podman is handled separately, thus I'd keep the name imho.

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
laert-ll: I find the first suggestion too long and the second too obscure, therefore I settled for naming it `resolveDockerCommand` since in our codebase we refer as docker to the whole container runtime installation (Docker Desktop, Rancher Desktop etc.).

Khaled-Cap:
managed to recreate the described scenario and got the following warning as output: Rancher Desktop has been installed but not launched yet. Please start Rancher Desktop once so that it sets up its command-line tools (docker, kubectl, ...) in /root/.rd/bin, then re-run your command.
The coding is clean and it implements the expected functionality. 
Issue was locally successfully tested.

PR #2087, opened 2026-06-29
#2068: Isolate Claude Code configuration per IDEasy project

### LINKED ISSUE #2067: Claude: fix startup warnings on ide claude
### Actual behavior

When running `ide claude`, two warnings appear:
1. "claude command at ~/ .local/bin/claude.exe missing or broken - run claude install to repair"
2. "Auto update failed - Run /doctor"

Claude works fine despite these warnings.

### Reproduce

1. ide install claude 
2. ide claude 

### Expected behavior

`ide claude` should start without any warnings. Since IDEasy manages the installation and version, Claude should not check for its binary at ~/.local/bin or try to auto-update itself.

### IDEasy status

```bash
IDE_ROOT is set to ~/projects
Your version of IDEasy is 2026.06.001.
You are online.
Found bash executable at: ~/AppData/****/********/Git/bin/bash.***
Found git executable at: ~/AppData/****/********/Git/***/git.***
```

### Related/Dependent issues

_No response_

### Comments/Hints

Root cause: Claude expects its binary at ~/.local/bin/ but IDEasy installs it in software/claude/ .
Additionally, Claude tries to auto-update itself, which conflicts with IDEasy's version management.

### LINKED ISSUE #2068: Claude: move config to project-local path
### Feature idea

As a developer , i want Claudes config directory to be stored in the project-local conf/claude path instead of the global ~/. claude, so that each project can have its own Claude configuration - consistent with how other tools like AWS are handled in IDEasy.

### Additional context

Currently Claude stores all settings globally in ~/. claude . In IDEasy, tool configs are project-local (e.g. AWS uses conf/aws/).

There is an undocumented env variable CLAUDE_CONFIG_DIR that could redirect this, but its not officially supported yet and has known issues:

- https://github.com/anthropics/claude-code/issues/3833 (still creates ~/.claude files)
- https://github.com/anthropics/claude-code/issues/25762 (feature request for proper support)
- https://github.com/anthropics/claude-code/issues/33430 (documentation request)

This should be implemented once Anthropic officially supports and documents CLAUDE_CONFIG_DIR. 
Implementation would follow the same pattern as Aws.java

maybeec:
### This PR fixes #2068

Running multiple Claude Code configurations side by side (different AWS Bedrock accounts, a sovereign endpoint, a company-wide instance) interferes because Claude Code defaults **all** state to the shared `~/.claude` home (settings, `.credentials.json`, `.claude.json`, MCP servers, history) and because provider/auth environment variables (`ANTHROPIC_*`, `AWS_*`, `CLAUDE_CODE_USE_BEDROCK`, …) leak between contexts — and per Claude's auth precedence a stray variable silently overrides the intended endpoint. The Claude tool added in `2026.07.001` did no isolation. This PR makes each IDEasy project a self-contained Claude configuration, fixing https://github.com/devonfw/IDEasy/issues/2068.

### Implemented changes:

* Added `EnvironmentContext.removeEnvVar(String)` (no-op default) with a real unset in `ProcessContextImpl`, so a tool can harden the *launched* process environment without ever touching the user's interactive shell (the shell-export collector keeps the no-op).
* `Claude.setEnvironment` now sets `CLAUDE_CONFIG_DIR = $IDE_HOME/conf/claude` — relocating Claude's settings, credentials, `.claude.json`, MCP servers and history into the project-local `conf/claude` — and scrubs leaking provider/auth variables (`ANTHROPIC_*`, `CLAUDE_CODE_USE_BEDROCK`/`VERTEX`/`FOUNDRY`, `CLAUDE_CODE_OAUTH_TOKEN`, `AWS_*`) from the child process. `CLAUDE_CONFIG_DIR` itself is never scrubbed. The single `setEnvironment` override covers both the launched process and `ide env` shell export.
* `Claude.postInstall` seeds a user-owned `conf/claude/settings.json` + `README.md` when absent and never overwrites them; the isolated `settings.json` `env` block is the single source of truth for provider/auth config.
* Added an `IDE_HOME`-gated `claude` shell function in `cli/src/main/package/functions` that routes bare `claude` through `ide claude` while an IDEasy project is active (transparent passthrough otherwise).
* Documented the isolation model and its accepted limitations in `documentation/ai.adoc`.

**Accepted shortcomings (by design):** on macOS, Claude.ai _subscription_ OAuth lives in the system Keychain, which `CLAUDE_CONFIG_DIR` does not relocate (Bedrock/API-token auth is unaffected); an enterprise `managed-settings.json` overrides all projects by design; a bare `claude` launched with no active IDEasy project is not scrubbed.

---

### Testing instructions

1. In an IDEasy project, run `ide install claude`, then confirm `conf/claude/settings.json` and `conf/claude/README.md` were created. Edit `settings.json`, re-run install, and confirm it is **not** overwritten.
2. Run `ide claude` and confirm Claude stores its state under `<IDE_HOME>/conf/claude` (e.g. `.claude.json`, credentials) rather than `~/.claude`.
3. `export ANTHROPIC_BASE_URL=https://leak.example` and run `ide claude`; confirm the leaked value does **not** reach Claude (it is scrubbed), while a value declared in `conf/claude/settings.json` under `env` **is** honored.
4. Set up a second IDEasy project and repeat; confirm the two projects use fully independent Claude configurations.
5. Automated: `mvn clean test`. See `ProcessContextImplTest` and `ClaudeTest`, covering config-dir injection, the env scrub, no-overwrite seeding, and an end-to-end test asserting `CLAUDE_CONFIG_DIR` reaches the launched process.

---

### Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/tool/claude/Claude.java
```diff
@@ -1,17 +1,75 @@
...
+        }
+      """;
+
+  /**
+   * Provider/auth environment variables removed from the launched Claude process so an ambient/leaked value cannot override the per-project configuration. The
+   * isolated {@code settings.json} env block is the single source of truth.
+   */
+  static final List<String> SCRUB_VARS = List.of(
+      "ANTHROPIC_API_KEY", "ANTHROPIC_AUTH_TOKEN", "ANTHROPIC_BASE_URL",
+      "ANTHROPIC_MODEL", "ANTHROPIC_SMALL_FAST_MODEL", "ANTHROPIC_CUSTOM_HEADERS",
```
hohwille, line 71: I have configured `ANTROPIC_MODEL` and some other variables in my `conf/ide.properties`.
Does it really make sense to remove them all with no way to circumvent?

hohwille: For a project it would even make sense to configure variables like `ANTROPIC_MODEL` or `BEDROCK_MODEL_ID` in `settings/ide.properties` to share such settings across the team.
I am not convinced that nuking such variables generally is a good idea.
Maybe we should remove them only if they are not comming from `ide.properties` but are inherited from System environment variables?

cli/src/main/package/functions
```diff
@@ -123,6 +123,14 @@ function icd() {
...
 }
 
+function claude() {
+  if [ -n "${IDE_HOME}" ]; then
+    ide claude "$@"
+  else
+    command claude "$@"
+  fi
+}
+
```
hohwille, line 153: nice idea to create such shim.
We once had some situation where a bug in IDEasy prevented to call a tool from IDEasy.
Such shim would take away the option to bypass IDEasy then for average users (who would not get to the idea to unload this function).
So - why would we need this or what would be the problem if we do not add this?
IMHO before running claude, we need to `ide install claude` what will already prepare the settings and set the variables. Surely if I switch projects with `cd` and not `icd` and do not also call some `ide` command before running the `claude` command directly, I might be connected to the wrong account.
Therefore I like your idea of this shim...

I am just thinking loud and challenging this so we balances pros and cons and take a clear decision.

documentation/ai.adoc
```diff
@@ -96,3 +96,23 @@ However, you can also map specific subfolders:
...
+When you run multiple Claude Code configurations (e.g. different AWS Bedrock accounts, a sovereign endpoint, a company-wide instance), they normally interfere through the shared `~/.claude` home folder and through leaked environment variables.
+
+IDEasy isolates Claude per project. Running `ide claude` (or simply `claude` while an IDEasy project is active):
+
+* sets `CLAUDE_CONFIG_DIR` to `$IDE_HOME/conf/claude`, so settings, credentials, MCP servers and history are stored per project;
+* removes leaking provider/auth variables (`ANTHROPIC_*`, `CLAUDE_CODE_USE_BEDROCK`/`VERTEX`/`FOUNDRY`, `CLAUDE_CODE_OAUTH_TOKEN`, `AWS_PROFILE`, `AWS_REGION`, `AWS_*_KEY*`, `AWS_SESSION_TOKEN`, `AWS_BEARER_TOKEN_BEDROCK`) from the launched process;
+* seeds a user-owned `conf/claude/settings.json` skeleton on first install.
+
+Put *all* provider/auth configuration in `conf/claude/settings.json` under the `env` block (see the generated `conf/claude/README.md` for Bedrock and custom-endpoint examples).
+Because scrubbed variables such as `AWS_PROFILE` are removed from the process, declare them there rather than relying on your shell.
```
hohwille: Should we add this specificity for Claude here?
IMHO we should explain the general concept and link to
https://github.com/devonfw/IDEasy/blob/main/documentation/sandbox.adoc
Then we can add such text as example.
But for Github Copilot I expect the same isolation - if not the case, we also need to file and fix a bug.
For AWS CLI we have it working and also for many other such tools.

hohwille (approved):
@maybeec thanks for your PR. Excellent fix with a holistic solution. Great job 👍 
I left some review comments for discussion and challenging (however, nothing wrong, just trying to trigger if we can improve further).

cli/src/main/package/functions
maybeec: In practice senior engineers are working in multiple projects possibly each coming with different requirements and compliance rules on AI tools. Thus, the shim should provide advanced consistent behaviour running for example claude in a specific IDE environment should not lead to a situation by accident, which falls back to a global claude installation, which might be as well installed on the machine (I have seen this scenarios too often to keep them out of mind). I would love to keep the shim as I have experienced scenarios where tools were vanishing from ide path for whatever reason making them unavailable when simply executing them like "python", but after executing "ide python" again, all worked fine. This risk should be mitigated.

documentation/ai.adoc
maybeec: Agree, I will remove it.
For GH Copilot CLI I don't have the right insights yet. Just experienced dramatic issues with claude. AWS CLI works fine and isolated as far as of my experiences.

maybeec: I changed my view on removing it. The logic to just keep variables which are defined in ide.properties + the macOS limitations make the document mandatory to allow users understand what's happening, but I shortened it to not focus on explaining the general concept of isolation again.

cli/src/main/java/com/devonfw/tools/ide/tool/claude/Claude.java
maybeec: done

oanding-blrng:
Successfully tested with 2026.07.001-07_07_06-SNAPSHOT on Windows.

Running `ide install claude` creates mentioned files `README.md` and `settings.json`:
```
$ ide install claude
No CVEs found for version 2.1.202 of tool claude.
Artifact already exists at C:\Users\oanding\Downloads\ide\default\claude-2.1.202-windows-x64.zip
To force update please delete the file and run again.
Extracting ZIP file C:\Users\oanding\Downloads\ide\default\claude-2.1.202-windows-x64.zip to C:\Users\oanding\projects\_ide\tmp\extract-claude-2.1.202-windows-x64.zip
Extracting 100% [================================================================================] 76/76MiB (0:00:03 / 0:00:00) 25.6MiB/s
Due to lack of permissions, Microsoft's mklink with junction had to be used to create a Symlink. See
https://github.com/devonfw/IDEasy/blob/main/documentation/symlink.adoc for further details. Error was: C:\Users\oanding\projects\IDEasy\software\claude: Dem Client fehlt ein erforderliches Recht
Verbindung erstellt für IDEasy\software\claude <<===>> _ide\software\default\claude\claude\2.1.202
Successfully installed claude in version 2.1.202 at C:\Users\oanding\projects\IDEasy\software\claude
```
AND
```
ls -l conf/claude/
total 5
-rw-r--r-- 1 oanding 4096 1250 Jul  7 15:08 README.md
-rw-r--r-- 1 oanding 4096   16 Jul  7 15:08 settings.json
```

Reinstalling claude within a project, there it's already installed doesn't override created files mentioned above:
```
$ ide install claude
No CVEs found for version 2.1.202 of tool claude.
Version 2.1.202 of tool claude is already installed
```

Running claude with `ide claude` within a project stores the files as described within `$IDE_HOME/conf/claude`
```
$ ls -la projects/IDEasy/conf/claude/
total 17
drwxr-xr-x 1 oanding  4096    0 Jul  7 15:18 ./
drwxr-xr-x 1 oanding  4096    0 Jul  7 15:08 ../
-rw-r--r-- 1 oanding  4096  454 Jul  7 15:18 .claude.json
drwxr-xr-x 1 oanding  4096    0 Jul  7 15:18 backups/
drwxr-xr-x 1 oanding  4096    0 Jul  7 15:15 cache/
-rw-r--r-- 1 oanding  4096 1342 Jul  7 15:14 README.md
-rw-r--r-- 1 oanding  4096  104 Jul  7 15:18 settings.json
```

Global definierte Variablen wie `ANTHROPIC_BASE_URL` werden korrekt von den Variablen in der `$IDE_HOME/conf/claude` überschrieben.

Installing/using claude in a different project creates (if needed) and uses it's own config.

PR #2091, opened 2026-06-29
CI status watch

shodiBoy1:
### Implemented changes:

* **New `ci-status-watch.yml`**  weekday-morning workflow that lists the workflows currently red on `main` and posts them as a single Teams card, so a broken CI is noticed in the Daily. Reports current state (a workflow stays listed until it's green again); creates no issues. Read-only (`actions: read`).
* **Extended `issue-pr-observer.yml`**  the daily digest now also reports new external GitHub **Discussions** (fetched via GraphQL, since they have no REST search). Issues/PRs and discussions are shown as two separate sections in the same card. Each source falls back to empty on error, so a failure in one still posts the other.

---

### Testing instructions

1. Make sure the repo secret `TEAMS_WEBHOOK_URL` is set (Teams incoming webhook).
2. **CI status watch:** Actions -> "CI status watch" → Run workflow. If any workflow on `main` is red, a card with the list appears in Teams; otherwise the run log says all green and nothing is posted.
3. **Issue/PR/Discussion observer:** Actions -> "External Issue, PR & Discussion Observer" -> Run workflow. New external issues/PRs/discussions from the look-back window appear as two sections; if there's nothing new, no message is sent.
4. Check the run logs each step prints how many items it found.

---

### Checklist for this PR

  with `internal`


.github/workflows/ci-status-watch.yml
```diff
@@ -0,0 +1,98 @@
+name: CI status watch
+
+# Posts a Teams message ONLY when one of the README badge workflows on the default
+# branch turns red (green -> red). For each it compares the two latest runs with gh; a
+# change is reported only if the newest run finished within the look-back window, so a
```
quando632: The PR description says the workflow lists the *currently red* workflows and keeps them listed until they turn green again. This header comment and the code actually implement a green-to-red *transition* that is deduplicated (a workflow that stays red is reported once, not every day). The comment here is the accurate one, so please update the PR description to match the implementation.

.github/workflows/ci-status-watch.yml
```diff
@@ -0,0 +1,98 @@
...
+  actions: read
+
+jobs:
+  watch:
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      WEBHOOK: ${{ secrets.TEAMS_WEBHOOK_URL }}
+      REPO: ${{ github.repository }}
+      BRANCH: ${{ github.event.repository.default_branch }}
```
quando632: `github.event.repository.default_branch` is not guaranteed to be populated on the `schedule` event payload. If it is empty, `gh run list --branch ""` silently widens the query to all branches. `github.ref_name` resolves to the default branch on a scheduled run and is always set.

```suggestion
      BRANCH: ${{ github.ref_name }}
```

.github/workflows/ci-status-watch.yml
```diff
@@ -0,0 +1,98 @@
...
+            .github/workflows/integration-tests.yml
+          "
+
+          # Monday catches up the weekend (Fri-Sun); other weekdays look back 24h.
+          if [ "$(date -u +%u)" -eq 1 ]; then
+            SINCE=$(date -u -d '3 days ago' +%Y-%m-%dT%H:%M:%SZ)
+          else
+            SINCE=$(date -u -d '1 day ago' +%Y-%m-%dT%H:%M:%SZ)
+          fi
+          echo "Checking badge workflows for red since $SINCE"
```
quando632: The branch drives the whole query but is never logged, which makes a failing scheduled run harder to debug. Suggest logging the resolved branch.

```suggestion
          echo "Checking badge workflows for red since $SINCE"
          echo "Resolved branch: $BRANCH"
```

.github/workflows/ci-status-watch.yml
```diff
@@ -0,0 +1,98 @@
...
+          # Compare each workflow's two latest runs; report it only if it just turned red
+          # and that run finished inside the window (so a lasting red is reported once).
+          RED=$(for WF_PATH in $WORKFLOWS; do
+              ROW=$(gh run list --repo "$REPO" --workflow "$WF_PATH" --branch "$BRANCH" \
+                --status completed --limit 2 --json conclusion,url,updatedAt,workflowName \
+                --jq '[(.[0].conclusion // ""), (.[0].updatedAt // ""), (.[0].url // ""), (.[1].conclusion // ""), (.[0].workflowName // "")] | @tsv' || true)
+              [ -z "$ROW" ] && continue
+              IFS=$'\t' read -r CUR CURTIME URL PREV NAME <<< "$ROW"
+              [ -z "$CUR" ] && continue
+              [[ "$CURTIME" > "$SINCE" ]] || continue
```
quando632: this is a lexicographic string comparison, not a date comparison. It is correct here only because both values are fixed-format ISO-8601 UTC (ending in `Z`). Worth a short inline comment so it is not later refactored into something that breaks that assumption.

.github/workflows/ci-status-watch.yml
```diff
@@ -0,0 +1,98 @@
...
+              if is_red "$CUR" && { [ "$PREV" = "success" ] || [ -z "$PREV" ]; }; then
```
quando632: A genuinely red run is swallowed when the previous run was cancelled: for runs `[failure, cancelled]`, `PREV` is `cancelled`, which is neither `success` nor empty, so nothing is reported. Since cancelled runs are pruned daily the window is small, but within the same day a real failure can be hidden. Consider treating `cancelled` as neither red nor green and comparing against the next older non-cancelled run.

.github/workflows/issue-pr-observer.yml
```diff
@@ -1,30 +1,55 @@
...
+            maybeec
+            oanding-blrng
+            Paras14
+            quando632
+            shodiBoy1
+            tineff96
+            vivu001
         run: |
+          # so a failing gh api trips the || fallback instead of being hidden by the trailing jq
+          set -o pipefail
```
quando632: This intentionally uses only `set -o pipefail` and not `set -e`, because the per-source `|| { ... ITEMS='[]'; }` fallbacks rely on the script continuing after a failed fetch. Please document that intent inline so it is not later "hardened" to `set -e`, which would break the fallbacks.

```suggestion
          set -o pipefail  # no `set -e`: the per-source `|| { ... ITEMS='[]'; }` fallbacks must keep the script running
```

quando632 (commented):
Overall solid and defensively written. One point is worth resolving before merge (PR description vs. implementation), the rest are small follow-ups.

.github/workflows/ci-status-watch.yml
hohwille:
```suggestion
              # lexicographic string comparison sufficient due to ISO-8601 format
              [[ "$CURTIME" > "$SINCE" ]] || continue
```

hohwille: Thanks 👍 I added a comment.

.github/workflows/ci-status-watch.yml
hohwille: @quando632 thanks for this finding that I agree to. 👍 
Could you suggest a fix so we can merge?

.github/workflows/ci-status-watch.yml
hohwille: Maybe it would be even smarter if we keep posting to our channel every night if something is broken so we get spammed and want to get rid of the spam by fixing the problem and nobody can excuse that he missed the notification ;)

.github/workflows/ci-status-watch.yml
hohwille: Actually I was checking this again:
1. I do not see for which workflow we actually cancel in any normal situation so this would be relevant
2. If we simply send a notification whenever something is read, we can make this so much simpler and do not actually need to dig in the history at all.

I would therefore suggest to go for KISS and simply send the notification whenever the workflow is failed.

FYI:
Initially I thought we would simply integrate this directly into the according workflow what would be even simpler.
However, the advantage of this dedicated workflow is that this extra nice-to-have feature is isolated, not redundantly copied to multiple workflows, cannot break the critical workflows like nightly-build or especially release so I now like this approach...

.github/workflows/ci-status-watch.yml
```diff
@@ -0,0 +1,100 @@
+name: CI status watch
+
+# Posts a Teams message ONLY when one of the README badge workflows on the default
+# branch turns red (green -> red). For each it compares the two latest runs with gh; a
+# change is reported only if the newest run finished within the look-back window, so a
+# workflow that stays red is not repeated every day.
```
quando632: The header describes the exact opposite of what we now want ("a workflow that stays red is not repeated every day"), so it has to be turned around. Note this also settles my earlier comment about the PR description: the description already described this behaviour, so it stays as it is and the code moves to match it.

```suggestion
# Posts a Teams message listing the README badge workflows that are currently red on the
# default branch. A workflow stays in the list until it is green again, so the reminder
# repeats every weekday morning until somebody fixes it. Creates no issues, read-only.
```

.github/workflows/ci-status-watch.yml
```diff
@@ -0,0 +1,100 @@
...
+                --jq '[(.[0].conclusion // ""), (.[0].updatedAt // ""), (.[0].url // ""), (.[1].conclusion // ""), (.[0].workflowName // "")] | @tsv' || true)
+              [ -z "$ROW" ] && continue
+              IFS=$'\t' read -r CUR CURTIME URL PREV NAME <<< "$ROW"
+              [ -z "$CUR" ] && continue
+              # lexicographic string comparison sufficient due to ISO-8601 format
+              [[ "$CURTIME" > "$SINCE" ]] || continue
+              if is_red "$CUR" && { [ "$PREV" = "success" ] || [ -z "$PREV" ]; }; then
+                jq -n --arg n "$NAME" --arg u "$URL" '{name:$n, url:$u}'
+              fi
+            done | jq -s '.')
```
quando632: The core change. Reporting current state instead of a transition removes the `SINCE` block including the Monday catch-up, the `is_red()` helper, the `PREV` comparison and the TSV parsing — and with `PREV` gone, the cancelled-run bug disappears without needing its own fix.

`cancelled`/`skipped` are now explicitly skipped rather than treated as a result, so `[failure, cancelled]` reports correctly. `--limit 5` gives headroom in case several runs in a row were cancelled.

```suggestion
          echo "Checking badge workflows on branch $BRANCH"

          # A workflow counts as red if its newest completed run that produced a real result
          # failed; cancelled/skipped runs say nothing about the code and are ignored. This is
          # exactly what the README badge shows. There is no look-back window on purpose: as
          # long as it stays red it is reported every morning, so nobody can miss it.
          RED=$(for WF_PATH in $WORKFLOWS; do
              gh run list --repo "$REPO" --workflow "$WF_PATH" --branch "$BRANCH" \
                --status completed --limit 5 --json conclusion,url,updatedAt,workflowName \
                --jq '[ .[] | select(.conclusion != "cancelled" and .conclusion != "skipped") ][0]
                      | select(. != null)
                      | select(.conclusion == "failure" or .conclusion == "timed_out" or .conclusion == "startup_failure")
                      | {name: .workflowName, url: .url, since: .updatedAt}' || true
            done | jq -s '.')
```

.github/workflows/ci-status-watch.yml
```diff
@@ -0,0 +1,100 @@
...
+
+          COUNT=$(echo "$RED" | jq 'length')
+          if [ "$COUNT" -eq 0 ]; then
+            echo "No badge workflow turned red. Skipping Teams notification."
+            exit 0
+          fi
+          echo "$COUNT badge workflow(s) turned red."
```
quando632: Follow-on: the log output still claims a transition ("turned red") which is no longer what is being detected.

```suggestion
            echo "No badge workflow is red. Skipping Teams notification."
            exit 0
          fi
          echo "$COUNT badge workflow(s) currently red."
```

.github/workflows/ci-status-watch.yml
```diff
@@ -0,0 +1,100 @@
...
+                  text: (if $count == 1 then "1 Workflow · grün → rot" else "\($count) Workflows · grün → rot" end) }
+              ]
+              + [ .[] | {
+                    type: "Container",
+                    separator: true,
+                    spacing: "Medium",
+                    items: [
+                      { type: "TextBlock", weight: "Bolder", wrap: true, text: "🔴 \(.name)" },
+                      { type: "TextBlock", spacing: "None", isSubtle: true, size: "Small", wrap: true,
+                        text: "[Run-Log öffnen](\(.url))" }
```
quando632: Same for the card: "rot geworden" and "grün → rot" would be wrong once we report standing state. The added `rot seit \(.since)` matters for the daily repeat otherwise you cannot tell at a glance what is new and what has been rotting for a week. It uses the `since` field emitted by the suggestion above.

```suggestion
                { type: "TextBlock", size: "Large", weight: "Bolder", wrap: true, color: "attention",
                  text: "🔴 CI ist rot" },
                { type: "TextBlock", spacing: "None", isSubtle: true, wrap: true,
                  text: (if $count == 1 then "1 Workflow ist rot" else "\($count) Workflows sind rot" end) }
              ]
              + [ .[] | {
                    type: "Container",
                    separator: true,
                    spacing: "Medium",
                    items: [
                      { type: "TextBlock", weight: "Bolder", wrap: true, text: "🔴 \(.name)" },
                      { type: "TextBlock", spacing: "None", isSubtle: true, size: "Small", wrap: true,
                        text: "rot seit \(.since) · [Run-Log öffnen](\(.url))" }
```

quando632 (commented):
@hohwille as requested, here is the concrete change for the KISS approach report whatever is currently red, every morning, instead of detecting a green→red transition. Four suggestions below, all in `ci-status-watch.yml`;

One thing worth deciding explicitly rather than leaving it as-is: you wrote "every night", but the cron is `0 7 * * 1-5`. Something breaking on Friday evening stays silent until Monday 09:00 CEST. `0 7 * * *` would close that gap weekday-only is defensible too, I just don't think it should stay that way by accident. Not included in the suggestions since it's your call.

hohwille:
> One thing worth deciding explicitly rather than leaving it as-is: you wrote "every night" ...

Thanks for pointing that out. Indeed I did not want to imply a workflow schedule/cron change as we do not work on weekends and need no notification on Saturday/Sunday. So I was just lax/un-precise and you are fully correct.

.github/workflows/ci-status-watch.yml
```diff
@@ -0,0 +1,87 @@
...
+
+          echo "Checking badge workflows on branch $BRANCH"
+
+          # A workflow counts as red if its newest completed run that produced a real result
+          # failed; cancelled/skipped runs say nothing about the code and are ignored. This is
+          # exactly what the README badge shows. There is no look-back window on purpose: as
+          # long as it stays red it is reported every morning, so nobody can miss it.
+          RED=$(for WF_PATH in $WORKFLOWS; do
+              gh run list --repo "$REPO" --workflow "$WF_PATH" --branch "$BRANCH" \
+                --status completed --limit 5 --json conclusion,url,updatedAt,workflowName \
```
hohwille, line 45: you increased the limit from 2 to 5. However, we only take the first result `[0]` and ignore the rest anyway.
IMHO the old limit was more reasonable (could be even 1).

```suggestion
                --status completed --limit 2 --json conclusion,url,updatedAt,workflowName \
```

hohwille (approved):
@quando632 thanks for your constructive review suggestions following my KISS idea that I all applied 👍 

I have one last suggestion, I would consider to apply before merge.
Can you quickly double-check that I am not missing something.
Then we finally merge and take this nice feature live.

.github/workflows/ci-status-watch.yml
```diff
@@ -0,0 +1,87 @@
...
+            .github/workflows/nightly-build.yml
+            .github/workflows/integration-tests.yml
+          "
+
+          echo "Checking badge workflows on branch $BRANCH"
+
+          # A workflow counts as red if its newest completed run that produced a real result
+          # failed; cancelled/skipped runs say nothing about the code and are ignored. This is
+          # exactly what the README badge shows. There is no look-back window on purpose: as
+          # long as it stays red it is reported every morning, so nobody can miss it.
```
quando632, line 42: Follow-up to the `--limit` thread: documenting why the limit is what it is, so the next reader does not read it as dead weight either.

```suggestion
          # A workflow counts as red if its newest completed run that produced a real result
          # failed; cancelled/skipped runs say nothing about the code and are ignored. This is
          # exactly what the README badge shows. There is no look-back window on purpose: as
          # long as it stays red it is reported every morning, so nobody can miss it.
          # The limit is not about how many runs we evaluate, that is always one. It defines how
          # far we can look past a streak of cancelled runs; 5 covers the longest streak seen (4).
```

.github/workflows/ci-status-watch.yml
quando632: The `[0]` does not act on the API result but on the filtered array: `[ .[] | select(...) ][0]` first drops cancelled/skipped runs, then takes the newest of what is left. So `--limit` controls how far we can look past a streak of cancelled runs, not how many results we use. With `--limit 1` a single cancelled run on top hides whatever sits underneath.

Real streaks on `main`: `nightly-build` 3 in a row (16 to 18 Dec 2025), `integration-tests` 4 in a row (9 to 17 Dec 2025). With `--limit 2` a red run below either of them would be reported as not red. I chose 5 because it covers the longest streak we have actually seen.

It is also free: `gh run list` issues the same 2 API requests for `--limit 2` and `--limit 5`.

That it reads the other way means the line needs a comment. A suggestion in this thread would only replace this line, so I put it on the comment block above (lines 39 to 42) as a separate comment.

.github/workflows/issue-pr-observer.yml
```diff
@@ -1,30 +1,55 @@
...
+            laim2003
+            marceltchanga9
+            maybeec
+            oanding-blrng
+            Paras14
+            quando632
+            shodiBoy1
+            tineff96
+            vivu001
         run: |
```
hohwille: This will be hard to maintain as our team is so volatile.
Would it be possible somehow to get this from the GitHub `IDE` Group membership or from the board somehow?
For the moment we can also merge with a hard-coded list. Maybe you can update it before I merge.
`vivu001`. `tineff96`, `areinicke`, `AdemZarrouki`, `KarimALofty` `Caylipp`, `marceltchanga9`, and `Paras14` already left.
There are several new members meanwhile.

hohwille (approved):
@shodiBoy1 thanks for your PR. Great that you could automate this and improve the observer. 👍 
I only have a small suggestion. As a quickfix we can update the hardcoded list.
Please also create a new issue that we should create a self-maintained replacement of this hard-coded list. Then we can finally merge. Sorry for the long review delay.

.github/workflows/issue-pr-observer.yml
shodiBoy1: i agree, a hardcoded list would be difficult to maintain with a frquently changing team. So i implemented the dynamic option. The workflow gets the Github teams linked to project/5 and uses their current membership. New and departed members are handled automatically. 

We only need to link the official IDE team to project/5 if this has not already been done and configure the repo secret 'TEAM_TOKEN' with projects:read and members:read. If the lookup or another API call fails, the workflow fails instead of posting incomplete results. The next run resumes from the last successful run, so no activity is missed.

.github/workflows/ci-status-watch.yml
hohwille: Thanks for explanation and improving the comment.

.github/workflows/issue-pr-observer.yml
```diff
@@ -1,87 +1,200 @@
...
 
 jobs:
   observe:
     runs-on: ubuntu-latest
     steps:
-      - name: Scan for new external issues & PRs and notify Teams
+      - name: Scan for new external issues, PRs & discussions and notify Teams
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TEAM_TOKEN: ${{ secrets.TEAM_TOKEN }}
```
hohwille, line 32: Actually reducing of permissions per purpose would be good and following zero trust.
However, GitHub Platform makes this hard for me since it forces tokens to expire and the only way to renew them is to manually regenerate them before the workflows break.
The more different tokens we create the more effort I have with the renewal.
Therefore I prefer that we stay with a single token that I have to maintain:
https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/update-access-token.adoc

Could you update the documentation to tell me what additional permission I need to add, I will regenerate it and you here reference that token instead?

hohwille (approved):
@shodiBoy1 thanks for your update. This is now great and I want to merge. 👍 
I only want to align on the token strategy. Can you have a look at my last comment please.

.github/workflows/issue-pr-observer.yml
shodiBoy1: done. I dropped TEAM_TOKEN and switched to ACTION_PUSH_TOKEN.

also updated documentation/contributing/update-access-token.adoc: extra org permissions are Members: Read and Projects: Read.

after you add those two on ACTION_PUSH_TOKEN, this should be ready to merge.

hohwille: Thanks, Excellent. I have updated the token and will now resolve and merge.

hohwille:
CI status watch still worked as expected:
https://github.com/devonfw/IDEasy/actions/runs/35717388689/job/106711989590

hohwille:
issue-pr-observer did not yet run today. The execution times do not really relate to the cron expression.
https://github.com/devonfw/IDEasy/actions/workflows/issue-pr-observer.yml

shodiBoy1:
> issue-pr-observer did not yet run today. The execution times do not really relate to the cron expression. https://github.com/devonfw/IDEasy/actions/workflows/issue-pr-observer.yml

I assumed that when we join IDEasy, we’re added to a dev team linked to the board. But the API returns no linked teams for project 5, which is why the observer failed. The existing IDE team has 86 members, including former members, so it doesnt match the current dev team. Could we create a team for the current IDEasy devs or use one if it already exists and link it to project 5 under Settings ->  Manage access? The observer already uses ACTION_PUSH_TOKEN, so no new token is needed. It will fetch the members on each run, but someone would still need to update the GitHub team when people join or leave.

PR #2093, opened 2026-06-30
#1992 improve gitcontextmock

### LINKED ISSUE #1992: Improve GitContextMock
### Task summary

As a mock for `GitContext` we have `GitContextMock`.
However, this was rather a hack than a solution.
This story should create a reasonable mock implementation.
The `clone` operation should do the following:
* Create the given `repository` directory including a `.git` subfolder.
* A file `HEAD` gets created in the `.git` folder containing: `ref: refs/head/«branch»` where `«branch»` is the `branch` property from the `GitUrl`.
* A file `FETCH_HEAD` gets created in the `.git` folder containing the `hashCode()` of the `GitUrl`.
* Inside the `.git` subfolder a `config` file is created with our IniFile support. It should contain at least the following entry:
```
[remote "origin"]
        url = «git-url-provided-as-arg»
```

Operations like `retrieveGitUrl` and `determineCurrentBranch`should return back the information stored by `clone` from the according files.
The operation ``determineRemote` can first be hardcoded to return `origin`. We can still make this more complex when needed.

A new method will be added to `GitContextMock`:
```java
void addChanges(Path repository, GitCommit... commits)`
```
Here `GitCommit` is a wrapper for a `List` of `GitChange`.
And a `GitChange` has at least the two properties:
* `Path content` (path to the file or directory to copy to target to apply the change)
* `Path target` (relative path in the repository)

So to apply a `GitChange` we simply copy `content` using `FileAccess` to the `target` path resolved in the `repository` path.
And to apply a `GitCommit` we simply apply all contained `GitChanges`.
The method `addchanges` simply puts all given `GitCommit` objects to a list or queue of pending commits.

The operations like `fetch` and `pull` will then be implemented according to the pending commits.
If there are no pending commits, they will do nothing.
A `fetch` will update `FETCH_HEAD` to the `hashCode()` of the latest pending commit (not the first one to apply but the last one in the "queue").
A `pull` will apply all changes from the pending commits in order to the given `repository` and also update the `HEAD`.

Our tests using `GitContextMock` can then be reworked and improved to make use of the new features.
Also we can properly test features like `ide upgrade` or the check if update is available, etc.

### Additional context

This is already quite complex.
Let us not over-engineer it further. So operations like `reset` can initially do nothing.
We could still later keep a `.git/backup` folder that we maintain by `GitContextMock` to also implement `reset`, etc. but IMHO this is not needed in near future.

Caylipp:
### Implemented changes:

* Improved GitContextMock behavior to better reflect real Git repository state.
* Added and updated unit tests for clone, fetch, and update detection.

---

### Testing instructions

1. Run `GitContextMockTest`.
2. Verify all tests pass.

---

### Checklist for this PR

  with `internal`


cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -31,22 +47,116 @@ public boolean hasUntrackedFiles(Path repository) {
...
+      Path refsHeads = gitFolder.resolve("refs").resolve("heads");
+      Files.createDirectories(refsHeads);
+      Path branchRefPath = refsHeads.resolve(branch);
+      Files.createDirectories(branchRefPath.getParent());
+      Files.writeString(branchRefPath, lastId);
+      // create a simple config file with remote origin url using ini-like format
+      StringBuilder cfg = new StringBuilder();
+      cfg.append("[remote \"origin\"]\n");
+      cfg.append("\turl = ").append(gitUrl.url()).append('\n');
+      Files.writeString(gitFolder.resolve("config"), cfg.toString());
```
maybeec: Per #1992 this config should be produced "with our IniFile support" rather than a hand-built string. `com.devonfw.tools.ide.io.ini.IniFileImpl` gives you `getOrCreateSection` for the `remote "origin"` section plus setting the `url` property, then write it out. That keeps the format consistent with the real `GitContextImpl` and makes the round-trip with `retrieveGitUrl` symmetric.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -59,9 +169,27 @@ public void cleanup(Path repository) {
...
+   */
   @Override
   public String retrieveGitUrl(Path repository) {
-
+    Path cfg = repository.resolve(GIT_FOLDER).resolve("config");
+    if (Files.exists(cfg)) {
+      try {
+        for (String line : Files.readAllLines(cfg)) {
+          String trimmed = line.trim();
+          if (trimmed.startsWith("url =")) {
```
maybeec: Two issues here, both solved by reading via `IniFile`/`IniFileImpl` (see the `clone` comment):
1. This returns the first line starting with `url =` regardless of section, so it isn't scoped to `[remote "origin"]` — with multiple remotes it would pick the wrong one.
2. It re-implements ini parsing by hand, which #1992 explicitly asked to avoid.

Read the `origin` section's `url` property instead, falling back to `MOCKED_URL_VALUE` when no config exists.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -31,22 +47,116 @@ public boolean hasUntrackedFiles(Path repository) {
...
+      Path branchRefPath = refsHeads.resolve(branch);
+      Files.createDirectories(branchRefPath.getParent());
+      Files.writeString(branchRefPath, lastId);
+      // create a simple config file with remote origin url using ini-like format
+      StringBuilder cfg = new StringBuilder();
+      cfg.append("[remote \"origin\"]\n");
+      cfg.append("\turl = ").append(gitUrl.url()).append('\n');
+      Files.writeString(gitFolder.resolve("config"), cfg.toString());
+    } catch (java.io.IOException e) {
+      throw new RuntimeException(e);
```
maybeec: Follow the project convention for caught exceptions. The same fix applies to every other `throw new RuntimeException(e)` in this file (lines 135, 158, 190, 257, 280, 313, 355) — each has its own suggestion below.

```suggestion
      throw new IllegalStateException("Failed to clone Git repository " + gitUrl + " to " + repository, e);
```

📖 Convention: [`coding-conventions.adoc` → "Catching and handling Exceptions"](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc) — always wrap a caught exception as `IllegalStateException("<meaningful message>", e)` carrying the original cause `e`. A bare `throw new RuntimeException(e)` violates this: it uses the wrong type and, without a message, discards diagnostic context.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -31,22 +47,116 @@ public boolean hasUntrackedFiles(Path repository) {
...
+      String branch = determineCurrentBranch(repository);
+      String lastId = String.valueOf(last.hashCode());
+      Path refsHeads = gitFolder.resolve("refs").resolve("heads");
+      Files.createDirectories(refsHeads);
+      Path branchRefPath = refsHeads.resolve(branch);
+      Files.createDirectories(branchRefPath.getParent());
+      Files.writeString(branchRefPath, lastId);
+      Files.writeString(gitFolder.resolve(FILE_FETCH_HEAD), lastId);
+    } catch (java.io.IOException e) {
+      throw new RuntimeException(e);
```
maybeec:
```suggestion
      throw new IllegalStateException("Failed to pull repository " + repository, e);
```

📖 Convention: [`coding-conventions.adoc` → "Catching and handling Exceptions"](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc) — always wrap a caught exception as `IllegalStateException("<meaningful message>", e)` carrying the original cause `e`. A bare `throw new RuntimeException(e)` violates this: it uses the wrong type and, without a message, discards diagnostic context.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -31,22 +47,116 @@ public boolean hasUntrackedFiles(Path repository) {
...
+    if (list == null || list.isEmpty()) {
+      return;
+    }
+    GitCommit last = list.get(list.size() - 1);
+    Path gitFolder = repository.resolve(GIT_FOLDER);
+    try {
+      Files.createDirectories(gitFolder);
+      Files.writeString(gitFolder.resolve(FILE_FETCH_HEAD), String.valueOf(last.hashCode()));
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
maybeec:
```suggestion
      throw new IllegalStateException("Failed to fetch repository " + repository, e);
```

📖 Convention: [`coding-conventions.adoc` → "Catching and handling Exceptions"](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc) — always wrap a caught exception as `IllegalStateException("<meaningful message>", e)` carrying the original cause `e`. A bare `throw new RuntimeException(e)` violates this: it uses the wrong type and, without a message, discards diagnostic context.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -59,9 +169,27 @@ public void cleanup(Path repository) {
...
+    if (Files.exists(cfg)) {
+      try {
+        for (String line : Files.readAllLines(cfg)) {
+          String trimmed = line.trim();
+          if (trimmed.startsWith("url =")) {
+            return trimmed.substring("url =".length()).trim();
+          }
+        }
+      } catch (IOException e) {
+        throw new RuntimeException(e);
```
maybeec:
```suggestion
        throw new IllegalStateException("Failed to read Git config for repository " + repository, e);
```

📖 Convention: [`coding-conventions.adoc` → "Catching and handling Exceptions"](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc) — always wrap a caught exception as `IllegalStateException("<meaningful message>", e)` carrying the original cause `e`. A bare `throw new RuntimeException(e)` violates this: it uses the wrong type and, without a message, discards diagnostic context.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -75,45 +203,249 @@ public Path findGit() {
...
+        Path refPath = gitFolder.resolve(ref);
+        if (!Files.exists(refPath)) {
+          // cannot resolve HEAD ref -> treat as no update
+          return false;
+        }
+        h = Files.readString(refPath).trim();
+      }
+      return !f.equals(h);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
maybeec:
```suggestion
      throw new IllegalStateException("Failed to check for repository update in " + repository, e);
```

📖 Convention: [`coding-conventions.adoc` → "Catching and handling Exceptions"](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc) — always wrap a caught exception as `IllegalStateException("<meaningful message>", e)` carrying the original cause `e`. A bare `throw new RuntimeException(e)` violates this: it uses the wrong type and, without a message, discards diagnostic context.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -75,45 +203,249 @@ public Path findGit() {
...
+    Path fetch = gitFolder.resolve(FILE_FETCH_HEAD);
+    try {
+      if (!Files.exists(fetch) || !Files.exists(trackedCommitIdPath)) {
+        return false;
+      }
+      String f = Files.readString(fetch).trim();
+      String t = Files.readString(trackedCommitIdPath).trim();
+      return !f.equals(t);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
maybeec:
```suggestion
      throw new IllegalStateException("Failed to check for repository update in " + repository, e);
```

📖 Convention: [`coding-conventions.adoc` → "Catching and handling Exceptions"](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc) — always wrap a caught exception as `IllegalStateException("<meaningful message>", e)` carrying the original cause `e`. A bare `throw new RuntimeException(e)` violates this: it uses the wrong type and, without a message, discards diagnostic context.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -75,45 +203,249 @@ public Path findGit() {
...
+      if (content.startsWith("ref:")) {
+        int idx = content.lastIndexOf('/');
+        if (idx >= 0 && idx + 1 < content.length()) {
+          return content.substring(idx + 1);
+        }
+      }
+      // fallback: return content
+      return content;
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
maybeec:
```suggestion
      throw new IllegalStateException("Failed to determine current branch for repository " + repository, e);
```

📖 Convention: [`coding-conventions.adoc` → "Catching and handling Exceptions"](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc) — always wrap a caught exception as `IllegalStateException("<meaningful message>", e)` carrying the original cause `e`. A bare `throw new RuntimeException(e)` violates this: it uses the wrong type and, without a message, discards diagnostic context.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -75,45 +203,249 @@ public Path findGit() {
...
+          Path rel = source.relativize(s);
+          Path dest = target.resolve(rel);
+          if (Files.isDirectory(s)) {
+            Files.createDirectories(dest);
+          } else {
+            Files.createDirectories(dest.getParent());
+            Files.copy(s, dest, StandardCopyOption.REPLACE_EXISTING);
+          }
+        } catch (IOException e) {
+          throw new RuntimeException(e);
```
maybeec:
```suggestion
          throw new IllegalStateException("Failed to copy directory " + source + " to " + target, e);
```

📖 Convention: [`coding-conventions.adoc` → "Catching and handling Exceptions"](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc) — always wrap a caught exception as `IllegalStateException("<meaningful message>", e)` carrying the original cause `e`. A bare `throw new RuntimeException(e)` violates this: it uses the wrong type and, without a message, discards diagnostic context.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -75,45 +203,249 @@ public Path findGit() {
...
+        }
+      });
+    }
+  }
+
+  /**
+   * Represents a file or directory change to apply when pulling a commit.
+   */
+  public static class GitChange {
```
maybeec: Minor: `GitChange` overrides neither `equals` nor `hashCode`, so `GitCommit.hashCode()` (delegating to `List<GitChange>.hashCode()`) ends up identity-based. That's fine for distinguishing "before vs. after" state within one test run, but two commits with identical content get different ids and ids aren't reproducible across runs. Consider making `GitChange`/`GitCommit` records (or adding `equals`/`hashCode`) if you want content-stable ids.

maybeec (commented):
Thanks for tackling this — the mock now models the full clone → fetch → pull → update-detection lifecycle with a pending-commits queue, and CI is green. Nicely structured and the new `GitContextMockTest` reads well.

I have two blocking items that come straight from the issue text / project conventions, plus a scope question. Inline comments give one-click suggestions for the convention fix.

**Must-fix**
1. **Use the project's `IniFile` support for the `.git/config` file.** #1992 explicitly asks that the config be created "with our IniFile support". Both the write (in `clone`) and the read (in `retrieveGitUrl`) are currently hand-rolled. `com.devonfw.tools.ide.io.ini.IniFile` / `IniFileImpl` already exist — using them also removes the fragile line parsing (see inline comments).
2. **Exception handling convention.** `coding-conventions.adoc` (§"Catching and handling Exceptions") requires wrapping as `throw new IllegalStateException("<meaningful message>", e)` — a message *and* the cause. There are 8 bare `throw new RuntimeException(e)` sites; I've left commit-able suggestions on each.

**Scope question (non-blocking, worth confirming)**
The issue closes with: *"Our tests using GitContextMock can then be reworked and improved… Also we can properly test features like `ide upgrade` or the check if update is available."* This PR delivers the mock infrastructure and self-tests, but the existing consumer (`UpdateCommandletTest`) isn't reworked and there are no feature-level tests exercising the new capabilities. Is that intended as a follow-up, or should it land here to fully close #1992?

**Minor**
- `GitChange` has no `equals`/`hashCode`, so `GitCommit.hashCode()` (which delegates to `List<GitChange>.hashCode()`) is identity-based — fine for a single run, just not content-stable. See inline note.
- `java.io.IOException` FQN in `clone`/`pull` vs. the imported `IOException` elsewhere is inconsistent (cosmetic).
- Heads-up: `IdeTestContext` defaults to `GitContextMock`, so flipping these methods from no-ops to active FS operations changes semantics for every consumer (e.g. `updateUrls()` now creates a `.git` tree under the urls path). CI is green so nothing breaks today — just noting the widened side-effect surface.

Also needs a rebase — the branch is currently behind `main`.

maybeec (changes requested):
Requesting changes — there are blocking items that should be resolved before merge (details are in the inline comments on the previous review):

**Blocking**
1. **Use the project's `IniFile` support for `.git/config`** (write in `clone`, read in `retrieveGitUrl`). #1992 explicitly requires this, and it removes the fragile hand-rolled parsing that currently isn't scoped to the `[remote "origin"]` section.
2. **Exception handling convention** — replace the 8 `throw new RuntimeException(e)` sites with `IllegalStateException("<message>", e)` (message + cause), per `coding-conventions.adoc`. One-click suggestions are provided on each line.

**Please also clarify (scope of #1992)**
- The issue asks that the existing tests using `GitContextMock` be reworked and that features like `ide upgrade` / update-detection be properly tested with the new mock. That part isn't in this PR — is it a planned follow-up, or should it land here to fully close #1992?

Once the two blocking items are addressed (and the branch rebased onto `main`), I'm happy to re-review. Thanks again — the core lifecycle modelling is solid.

Caylipp:
> Requesting changes — there are blocking items that should be resolved before merge (details are in the inline comments on the previous review):
> 
> **Blocking**
> 
> 1. **Use the project's `IniFile` support for `.git/config`** (write in `clone`, read in `retrieveGitUrl`). #1992 explicitly requires this, and it removes the fragile hand-rolled parsing that currently isn't scoped to the `[remote "origin"]` section.
> 2. **Exception handling convention** — replace the 8 `throw new RuntimeException(e)` sites with `IllegalStateException("<message>", e)` (message + cause), per `coding-conventions.adoc`. One-click suggestions are provided on each line.
> 
> **Please also clarify (scope of #1992)**
> 
> * The issue asks that the existing tests using `GitContextMock` be reworked and that features like `ide upgrade` / update-detection be properly tested with the new mock. That part isn't in this PR — is it a planned follow-up, or should it land here to fully close #1992?
> 
> Once the two blocking items are addressed (and the branch rebased onto `main`), I'm happy to re-review. Thanks again — the core lifecycle modelling is solid.

Thanks for the review.

I addressed the two blocking items:
- .git/config is now written using IniFile/IniFileImpl, and retrieveGitUrl reads the url from the remote "origin" section.
- The bare RuntimeException usages were replaced with IllegalStateException including meaningful messages and the original cause.

I also added equals/hashCode for GitChange and GitCommit to address the minor hash stability note.

Regarding the scope question: I intended this PR to focus on the GitContextMock infrastructure and its self-tests. Reworking consumer tests such as UpdateCommandletTest would be a follow-up, unless you prefer to include it here.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -9,19 +22,26 @@ public class GitContextMock implements GitContext {
...
   @Override
   public void pullOrCloneIfNeeded(GitUrl gitUrl, Path repository) {
-
+    Path gitFolder = repository.resolve(GIT_FOLDER);
+    if (Files.exists(gitFolder)) {
+      pull(repository);
+    } else {
+      clone(gitUrl, repository);
+    }
   }
```
hohwille: We should either create an abstract super-class or extend this `GitContextMock` now from `GitContextImpl`.
For many cases we can reuse the general implementation and in this example we have a special caching feature that should also be covered in the tests.
This will also apply to many other methods and you explicitly implement here from the interface but that you can simply inherit then.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -31,22 +51,116 @@ public boolean hasUntrackedFiles(Path repository) {
...
+      Path refsHeads = gitFolder.resolve("refs").resolve("heads");
+      Files.createDirectories(refsHeads);
+      Path branchRefPath = refsHeads.resolve(branch);
+      Files.createDirectories(branchRefPath.getParent());
+      Files.writeString(branchRefPath, lastId);
+      // create .git/config using the project's IniFile support
+      IniFile cfg = new IniFileImpl();
+      IniSection originSection = cfg.getOrCreateSection("remote \"origin\"");
+      originSection.setProperty("url", gitUrl.url());
+      Files.writeString(gitFolder.resolve("config"), cfg.toString());
```
hohwille: Great job 👍 
But can't we use `FileAccess` also here to avoid low-level IO fiddling and allow better reuse of our higher level infrastructure?

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -75,45 +210,295 @@ public Path findGit() {
...
+          throw new IllegalStateException("Failed to copy directory " + source + " to " + target, e);
+        }
+      });
+    }
+  }
+
+  /**
+   * Represents a file or directory change to apply when pulling a commit.
+   */
+  public static class GitChange {
```
hohwille: Since this is designed immutable (what is good 👍), it should be a record.
This saves you from boiler-plate code like `equals` and `hashCode` (or `toString` that is currently not implemented reasonable).

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -75,45 +210,295 @@ public Path findGit() {
...
+      int result = this.content.hashCode();
+      result = 31 * result + this.target.hashCode();
+      return result;
+    }
+  }
 
+  /**
+   * Represents a commit containing one or more {@link GitChange}s.
+   */
+  public static class GitCommit {
```
hohwille: Same here. Since you designed this immutable, you could better use a record.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -75,45 +210,295 @@ public Path findGit() {
...
+        return;
+      }
+      Path parent = trackedCommitIdPath.getParent();
+      if (parent != null) {
+        Files.createDirectories(parent);
+      }
+      Files.writeString(trackedCommitIdPath, commitId);
+    } catch (IOException e) {
+      throw new IllegalStateException("Failed to save current commit id to " + trackedCommitIdPath, e);
+    }
```
hohwille: This is IMHO wrong and the implementation should be the same as in GitContextImpl.

hohwille (approved):
@Caylipp thanks for your PR. Nice job with great JUnit test. 👍 
However, I left some comments for rework. Please have a look and improve before we can merge. Thanks in advance.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
Caylipp: Thanks. GitContextMock now extends GitContextImpl and reuses the caching logic. I also adapted the tests accordingly.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
Caylipp: Done. I replaced the low level file operations with FileAccess.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
Caylipp: Done. GitChange is now a record so the generated equals, hashCode, and toString are used.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
Caylipp: GitCommit is also now a record and stores an immutable copy of its changes.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
Caylipp: Fixed. GitContextMock now inherits saveCurrentCommitId from GitContextImpl and only overrides determineCurrentCommitId for the mocked repository structure.

CHANGELOG.adoc
```diff
@@ -17,6 +17,7 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/2102[#2102]: IDE_ROOT environment variable is not available to ideasy.exe during MSI installation
 * https://github.com/devonfw/IDEasy/issues/1594[#1594]: Add release commandlet
 * https://github.com/devonfw/IDEasy/issues/1982[#1982]: Modified fallback link creation on Windows
+* https://github.com/devonfw/IDEasy/issues/1992[#1992]: Improve GitContextMock
```
hohwille: Needs to be moved up to the upcoming release.

CHANGELOG.adoc
```diff
@@ -119,8 +120,6 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/1904[#1904]: Add Inso CLI to IDEasy commandlets
 * https://github.com/devonfw/IDEasy/issues/1952[#1952]: Ability for platform specific dependencies
 * https://github.com/devonfw/IDEasy/issues/1950[#1950]: Fix exit autocompletion
-* https://github.com/devonfw/IDEasy/issues/1936[#1936]: Improve localization of GUI
```
hohwille, line 122: done on purpose or by accident? Looks like an accident to me.

hohwille (approved):
@Caylipp thanks for the excellent rework. Perfect now 👍 
I will fix the CHANGELOG and then merge.

CHANGELOG.adoc
hohwille: As the story is internal and you even marked your PR as internal, no change in CHANGELOG is expected here at all.

CHANGELOG.adoc
hohwille: Was by accident trying to remove an empty line. I fixed it.

hohwille:
Re-test: Nothing for e2e testing. Just infrastructure for JUnit tests.

PR #2113, opened 2026-07-05
#1659: abort runTool with warning when global tool installer runs in background

### LINKED ISSUE #1659: Global tools run tool while installation is running in background
### Actual behavior

When I run a global tool e.g. `ide docker ps` that is not yet installed, then the following is happening:
`runTool` is invoked that calls `install` method to ensure the tool is installed.
Since the tool is not installed and it is a global tool this typically means that we run an interactive installer in the background. In our example on Windows this would download e.g. the rancher desktop Windows installer and run it in the background showing me some first screen of the installer.
Since then the `install` method returns, we continue with `runTool` to execute the actual command which here is `docker ps`.
This will end up with an error since the `docker` command is not found (installation is still running).
Even after the installation is completed, the command `docker` would still be unavailable since the `PATH` was changed in the Windows registry and that change is not visible in my existing change.

```
docker: command not found
```

### Reproduce

`ide docker ps`

### Expected behavior

After `docker` installer is launched in the background, we should get the following warning logged:
```
The installation of docker/rancher@1.21.0[*] is currently running in the background!
You need to complete the installation, potentially reboot and rerun your 'ide' command in a new terminal session after the installation has completed.
```

### IDEasy status

```bash
You are online.
Found bash executable at: C:/Program Files/Git/usr/bin/bash.exe
Found git executable at: C:/Program Files/Git/mingw64/bin/git.exe
Your settings are up-to-date.
Your version of IDEasy is 2025.11.001.
```

### Related/Dependent issues

#47

### Comments/Hints

When we create a global installation by running an installer in the background here:
https://github.com/devonfw/IDEasy/blob/de740dfbc5ef6b640c32ec1bb5ae2a71fd7fb717/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java#L153-L167

We should somehow have a new flag in `ToolInstallation` telling us that we have a background installation is running or in other words that the tool is not yet available. This flag should be set here to `true` while it is `false` by default (in all other places also in cases when we install a global tool via package manager on Linux since that will be synchronous waiting for the installation to complete).
Now we can change this code:
https://github.com/devonfw/IDEasy/blob/de740dfbc5ef6b640c32ec1bb5ae2a71fd7fb717/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java#L209

So we store the returned `ToolInstallation` in a local `installation` variable and check the new flag.
If that flag is `true`, we log a warning and abort the further process:
```
ToolInstallation installation = install(request);
if (installation.isInstalledAsynchronously()) {
  this.context.warning("The installation of {} is currently running in the background!\nYou need to complete the installation, potentially reboot and rerun your 'ide' command in a new terminal session after the installation has completed.", request.getRequested());
  return new ProcessResultImpl(this.tool, this.tool, 0, List.of());
}
```

Ali-Shariati-Najafabadi:
### Implemented changes:

* Added `installedAsynchronously` flag to `ToolInstallation` record to signal that a background installer was launched and the tool is not yet available
* In `GlobalToolCommandlet.doInstall()`, return `ToolInstallation` with `installedAsynchronously=true` instead of throwing a `CliException` when the installation path is not yet available after launching the background installer
* In `ToolCommandlet.runTool()`, check the new flag and log a warning to the user instead of trying to execute the not-yet-available binary

---

### Testing instructions

1. Run unit test: `mvn test -pl cli -Dtest=GlobalToolCommandletTest`
2. For manual testing on Windows (Docker not installed):
   - Temporarily change `GlobalToolCommandlet.java` line 175 to always return `new ToolInstallation(null, null, null, resolvedVersion, true, true)`
   - Run: `Ideasy.main()` with Program arguments : docker ps
   - Expected output: warning message about background installation instead of `docker: command not found`
   - Revert the temporary change afterwards

---

### Checklist for this PR

- [x] You have added the issue implemented by your PR in CHANGELOG.adoc unless issue is labeled with `internal`


quando632 (commented):
You did a really great job. I just have one small correction for the `ide install <globalTool>`  test because it loses the log to reboot

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -272,7 +273,13 @@ public ProcessResult runTool(ToolInstallRequest request, ProcessMode processMode
       // we render this warning so the error gets detected and can be fixed but we do not block the user by skipping the installation.
       LOG.warn("Preventing infinity loop during installation of {}", request.getRequested(), new RuntimeException());
     } else {
-      install(request);
+      ToolInstallation installation = install(request);
+      if (installation != null && installation.installedAsynchronously()) {
```
quando632, line 277: The background-install warning lives only in runTool ([ToolCommandlet.java:277]. The direct install path (ide install docker) calls install(request) here and never checks installedAsynchronously(). Previously the removed CliException in doInstall covered both paths. Now ide install <globalTool> completes silently (exit 0), so the user loses the "complete installation / reboot / rerun in new terminal" guidance. 

### Fix:
Move the "is async → warn" handling into install()/doInstall itself, or a shared helper, so every caller is covered and can't forget the check. That also removes the fragility of the all-null ToolInstallation being handed back to callers.

cli/src/test/java/com/devonfw/tools/ide/tool/GlobalToolCommandletTest.java
```diff
@@ -0,0 +1,76 @@
...
+   */
+  static class AsyncInstallerToolCommandlet extends GlobalToolCommandlet {
+
+    AsyncInstallerToolCommandlet(IdeContext context) {
+
+      super(context, TOOL_NAME, Set.of(Tag.DOCKER));
+    }
+
+    @Override
+    public ToolInstallation install(ToolInstallRequest request) {
```
quando632: The test overrides install() entirely, so the actual changed line GlobalToolCommandlet.java:177 is never executed, it only exercises ToolCommandlet.runTool. A test driving the real doInstall (mocked installer + getInstallationPath == null) would cover the core change and would surface the finding above.

CHANGELOG.adoc
```diff
@@ -14,6 +14,8 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/1517[#1517]: Fix IDEasy MSI does not configure Windows Terminal for git-bash
 * https://github.com/devonfw/IDEasy/issues/1227[#1227]: Added Setup instrcutions to the Windows installer
 * https://github.com/devonfw/IDEasy/issues/865[#865]: az not working on Mac
+* https://github.com/devonfw/IDEasy/issues/1659[#1659]: Abort runTool with warning when global tool installer runs in background
+
```
hohwille: Do not add extra newlines to changelog entries

```suggestion
* https://github.com/devonfw/IDEasy/issues/1659[#1659]: Abort runTool with warning when global tool installer runs in background
```

hohwille (approved):
@Ali-Shariati-Najafabadi thanks for your PR. You did implement the story exactly as specified in the issue. 👍 
An extra JUnit test would have been nice.
After reading the review I also came to the question if it makes more sense to throw a `CliAbortException` instead of logging a warning and returning an empty `ProcessResult`.
Maybe in some cases sub-sequent processes will continue and rely on a wrong result.
However, I do not see a current problem but only an improvement here, so I will merge this and include it in the current release.
We can still improve in yet another PR after the merge + release.

CHANGELOG.adoc
hohwille: I resolved the conflict and also this review remark just now.

quando632:
Successfully tested with 2026.07.002 on Windows

PR #2144, opened 2026-07-09
#1976: extend tool commandlet installation logic

### LINKED ISSUE #1976: Extend ToolCommandlet installation logic to allow software-repository-only installation requests.
### Task summary

Currently, when installing LocalToolCommandlets, the installation logic requires us to be inside a project when installing a tool, e.g. Java. This is, because the installation logic always tries to create symlinks (if not via install(), installToolDependencies() ends up doing this nevertheless). For some use-cases however (like launching the GUI using java and maven), we do need dependencies like java to be installed in the software repo, however, the requirement for being inside a project in order to install those dependencies creates an unnecessary obstacle (example: we want to be able to launch the GUI from everywhere, not only inside project folders, while at the same time still being able to use the java tool installed in the software repo.)

### Additional context

Before you assign yourself to this issue: Please note that the installation routine and logic in IDEasy from `ToolCommandlet` and `LocalToolCommandlet` are rather complex. So this is not an easy task and not good for beginners in IDEasy. You should already have collected some experience with the installation routine or should be good at reading and understanding code...

We are talking about this:
https://github.com/devonfw/IDEasy/blob/63240371fefba87c3eca231e7ecd563939463f84/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java#L158

That function is also installing dependencies and for each of them ends up here:
https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java#L276

This should check if we are actually inside a project and if not then do not try to get the configured version and try to install the tool in the project but just create an additional installation from the version range in the software repo.

Another idea of the story above is to add a boolean option here:
https://github.com/devonfw/IDEasy/blob/63240371fefba87c3eca231e7ecd563939463f84/cli/src/main/java/com/devonfw/tools/ide/tool/ToolInstallRequest.java#L17

This would allow to ignore the current project so the above behaviour to not do the project installation would happen even if I am inside an IDEasy project in case that boolean flag was explicitly set to true:
```
installRequest.setIgnoreProject(true);
```

Caylipp:
### This PR fixes #1976

  ### Implemented changes:

  * Added `ignoreProject` support to `ToolInstallRequest`
  * Added CLI option `--ignore-project` to `InstallCommandlet`
  * Updated dependency installation to skip project symlinks when ignoring project or outside project context
  * Updated GUI launcher dependencies to install in the software repository
  * Added tests for ignore project flag and inheritance

  ---

  ### Testing instructions

  1. Run `ide install java --ignore-project` inside an IDEasy project and verify Java is installed in the
  software repository without creating a symlink under `software/java`.
  2. Run `ide install java --ignore-project` outside any project and verify the tool is installed in the
  software repository.
  3.  Run added tests.

### Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/commandlet/InstallCommandlet.java
```diff
@@ -42,6 +46,7 @@ public InstallCommandlet(IdeContext context) {
     addKeyword(getName());
     this.tool = add(new ToolProperty("", false, "tool"));
     this.version = add(new VersionProperty("", false, "version"));
+    this.ignoreProject = add(new BooleanProperty("", false, "ignore-project"));
```
quando632: **`--ignore-project` is registered as a positional value, not a CLI flag.**

This `BooleanProperty` is created with an empty name and only an alias:

```java
this.ignoreProject = add(new BooleanProperty("", false, "ignore-project"));
```

`Property.isOption()` returns `name.startsWith("-")`, so with an empty name this property is a positional **value**, not an option. In `Commandlet.add(...)` only `isOption()` properties are put into the option map, so `--ignore-project` is never matched on the command line it becomes a third positional argument after `tool` and `version`. Every other flag in the code base uses e.g. `new FlagProperty("--skip-tools")`.

The tests do not catch this because they set the value directly (`install.ignoreProject.setValue(true)`), bypassing CLI parsing.

```suggestion
    this.ignoreProject = add(new FlagProperty("--ignore-project"));
```

Please also add a test that drives the real arg parser (`ide install <tool> --ignore-project`).

cli/src/main/resources/nls/Help_de.properties
```diff
@@ -62,6 +62,7 @@ cmd.install=Installiert das selektierte Werkzeug.
 cmd.install-plugin=Installiert die selektierte Erweiterung für das selektierte Werkzeug.
 cmd.install-plugin.detail=Erweiterung können nur für Werkzeuge installiert werden, die diese unterstützen. Erweiterungen müssen in den Projekteinstellungen definiert sein um zur Installation verfügbar zu sein. Falls eine gewünschte Erweiterung nicht verfügbar ist, bitte den ide-admin kontaktieren.
 cmd.install.detail=Die Liste an zur Installation verfügbaren Werkzeugen können mit dem Befehl "ide help" ausgegeben werden.
+cmd.install.opt.--ignore-project=Ignoriert das aktuelle Projekt und installiert das Werkzeug ausschließlich im Software-Repository.
```
quando632: **EN/DE help keys for ignore-project are inconsistent.**

`cmd.install.opt.--ignore-project` exists only in German. With the current (value) property, `NlsBundle.get` resolves `val.ignore-project`, so this `opt.--ignore-project` key is currently dead. After fixing the property to `FlagProperty("--ignore-project")` (see the InstallCommandlet comment) the lookup becomes `opt.--ignore-project`, which is present in German but **missing in English** (`Help.properties` only has `val.ignore-project`) the English help would then render `?opt.--ignore-project`.

Please align both files: use `cmd.install.opt.--ignore-project` in EN and DE, and drop the now-redundant `val.ignore-project`.

cli/src/main/java/com/devonfw/tools/ide/commandlet/InstallCommandlet.java
```diff
@@ -86,6 +92,7 @@ protected void doRun() {
     }
     ToolInstallRequest request = ToolInstallRequest.ofDirect();
     request.setRequested(new ToolEditionAndVersion(version));
+    request.setIgnoreProject(Boolean.TRUE.equals(this.ignoreProject.getValue()));
```
quando632: **Direct install ignores `ignoreProject` when resolving the version.**

Just above (lines 90-92):

```java
if (version == null) {
  version = commandlet.getConfiguredVersion();
}
```

When `--ignore-project` is set without an explicit version and we are inside a project, this still uses the project's configured version the opposite of the flag's intent. `installAsDependency` deliberately avoids `getConfiguredVersion()` when `ignoreProject` is true. Consider skipping the configured version here as well when the flag is set (falling back to the default/latest).

quando632:
Nice work on this 👍 The `ignoreProject` flag is threaded cleanly through `ToolInstallRequest` the standalone-vs-project branching in `installAsDependency` reads clearly, and I like that the skip path is covered by explicit tests for both the in-project and outside-project cases. Left a few inline comments the main one is about how the CLI flag is declared. 🙂

cli/src/main/java/com/devonfw/tools/ide/commandlet/InstallCommandlet.java
Caylipp: Fixed. The configured project version is no longer resolved when --ignore-project is set. I also added a test for this case.

cli/src/main/java/com/devonfw/tools/ide/commandlet/InstallCommandlet.java
Caylipp: Thanks, fixed. I changed it to FlagProperty("--ignore-project") and added a test using the real CLI argument parser.

cli/src/main/resources/nls/Help_de.properties
Caylipp: Fixed. The EN and DE help files now both use cmd.install.opt.--ignore-project, and the obsolete val.ignore-project entries were removed.

quando632 (changes requested):

cli/src/main/java/com/devonfw/tools/ide/commandlet/InstallCommandlet.java
```diff
@@ -53,7 +58,8 @@ public String getName() {
   @Override
   public boolean isIdeRootRequired() {
 
-    return this.tool.getValueCount() > 0;
+    return (this.tool.getValueCount() > 0)
+        && !Boolean.TRUE.equals(this.ignoreProject.getValue());
```
quando632: **Blocker: NPE when run outside of IDE_ROOT.**

`Commandlet.isIdeHomeRequired()` delegates to `isIdeRootRequired()` by default, so this override drops *both* guards. Without IDE_ROOT there is no software repository, and `installTool` NPEs (`LocalToolCommandlet:186`) instead of failing with `CliException` / `NO_IDE_ROOT` as before. The flag should only drop the IDE_HOME requirement:

```suggestion
    return this.tool.getValueCount() > 0;
  }

  @Override
  public boolean isIdeHomeRequired() {

    return (this.tool.getValueCount() > 0)
        && !Boolean.TRUE.equals(this.ignoreProject.getValue());
```

quando632 (commented):
_(superseded - see the inline comments.)_

cli/src/main/java/com/devonfw/tools/ide/commandlet/InstallCommandlet.java
```diff
@@ -81,11 +87,13 @@ protected void doRun() {
     ToolCommandlet commandlet = this.tool.getValue();
     VersionIdentifier versionIdentifier = this.version.getValue();
     VersionIdentifier version = versionIdentifier;
-    if (version == null) {
+    boolean ignoreProject = Boolean.TRUE.equals(this.ignoreProject.getValue());
+    if ((version == null) && !ignoreProject) {
```
quando632: **This guard is undone further down — the configured version is still used.**

`ToolCommandlet.completeRequestRequested` (line 432, not in this diff) fills the version straight back in and never checks `request.isIgnoreProject()`. Verified on `834d7e4`: with `JAVA_VERSION=17.0.6` configured and the flag set, 17.0.6 is installed anyway.

Fix belongs there, keeping this guard as-is:

```java
version = request.isIgnoreProject() ? VersionIdentifier.LATEST : getConfiguredVersion();
```

Same applies to the edition two lines above (`getConfiguredEdition()`), which raises the question whether `--ignore-project` should ignore the whole project config or only the version.

cli/src/test/java/com/devonfw/tools/ide/commandlet/InstallCommandletTest.java
```diff
@@ -302,5 +303,88 @@ public void testInstallCommandletOfflineUpdateWithoutCachedDownload(WireMockRunt
...
+    // assert
+    assertThat(context.getSoftwarePath().resolve("java")).doesNotExist();
+    assertThat(context.getSoftwareRepositoryPath()
+        .resolve(DefaultToolRepository.ID_DEFAULT)
+        .resolve("java")
+        .resolve("java")
+        .resolve("17.0.6")).doesNotExist();
+    assertThat(context)
+        .logAtSuccess()
+        .hasMessageContaining("Successfully installed java in version");
```
quando632: **This test passes even with the fix reverted — it does not test the flag.**

The `install` fixture has an empty `ide.properties`, so `getToolVersion` returns `LATEST` on both paths and `17.0.6 doesNotExist` is trivially true. Verified: reverting the guard in `doRun` keeps this test green. The project needs a version that differs from the latest (fixture has 17.0.6, 17.0.10, 21.0.8_9):

```suggestion
    // arrange - configure a project version that differs from the latest available version
    IdeTestContext context = newContext(PROJECT_INSTALL, wmRuntimeInfo);
    VersionSetCommandlet versionSet = context.getCommandletManager().getCommandlet(VersionSetCommandlet.class);
    versionSet.tool.setValueAsString("java", context);
    versionSet.version.setValueAsString("17.0.6", context);
    versionSet.run();
    InstallCommandlet install = context.getCommandletManager().getCommandlet(InstallCommandlet.class);
    install.tool.setValueAsString("java", context);
    install.ignoreProject.setValue(true);

    // act
    install.run();

    // assert - the configured 17.0.6 must be ignored and the latest version installed instead
    assertThat(context.getSoftwareRepositoryPath().resolve(DefaultToolRepository.ID_DEFAULT).resolve("java").resolve("java")
        .resolve("17.0.6")).doesNotExist();
    assertThat(context.getSoftwareRepositoryPath().resolve(DefaultToolRepository.ID_DEFAULT).resolve("java").resolve("java")
        .resolve("21.0.8_9")).exists();
    assertThat(context.getSoftwarePath().resolve("java")).doesNotExist();
```

Red until the `completeRequestRequested` fix lands — which is the point.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -105,13 +105,16 @@ private ToolInstallation doInstallStep(ToolInstallRequest request) {
     FileAccess fileAccess = this.context.getFileAccess();
     boolean ignoreSoftwareRepo = isIgnoreSoftwareRepo();
     Path toolPath = request.getToolPath();
-    if (!ignoreSoftwareRepo) {
+    if (!ignoreSoftwareRepo && !request.isIgnoreProject() && toolPath != null) {
```
quando632, line 115: **Silent no-op for `isIgnoreSoftwareRepo()` tools.**

For Node/Python and the `PackageManagerBasedLocalToolCommandlet` subclasses `getInstallationPath()` returns `getToolPath()` (`IDE_HOME/software/<tool>`), so the payload is in the project before we reach this branch. Verified on `834d7e4` — `ide install node --ignore-project` inside a project reports success, `IDE_HOME/software/node` exists, and the repo has no node at all.

These tools cannot use the software repository, so I would reject the combination instead of silently ignoring the flag:

```java
if (request.isIgnoreProject() && isIgnoreSoftwareRepo()) {
  throw new CliException("The tool " + this.tool + " does not support the software repository and therefore cannot be installed with --ignore-project.");
}
```

quando632 (commented):
_(superseded - see the inline comments.)_

quando632 (commented):
After my earlier comments I went through every project-dependent input in the install path instead of reporting findings one at a time. The pattern is structural, so I think it is worth looking at as a whole rather than as separate bugs

cli/src/main/java/com/devonfw/tools/ide/commandlet/InstallCommandlet.java
```diff
@@ -81,11 +87,13 @@ protected void doRun() {
...
       version = commandlet.getConfiguredVersion();
     }
     ToolInstallRequest request = ToolInstallRequest.ofDirect();
     request.setRequested(new ToolEditionAndVersion(version));
+    request.setIgnoreProject(ignoreProject);
     ToolInstallation installation = commandlet.install(request);
     if (versionIdentifier != null) {
```
quando632, line 90: **`--ignore-project` modifies the project configuration.**

This `setVersion(...)` call is unguarded, and `setVersion` writes to `settings/ide.properties`. Verified on `834d7e4`: `ide install java 17.0.6 --ignore-project` appends `JAVA_VERSION=17.0.6` to the project settings — the flag that promises to ignore the project ends up changing it.

```suggestion
    if ((versionIdentifier != null) && !ignoreProject) {
```

With this, the settings file stays untouched and `InstallCommandletTest`, `LocalToolCommandletTest` and `ToolInstallRequestTest` remain green (19 tests).

cli/src/main/java/com/devonfw/tools/ide/commandlet/InstallCommandlet.java
Caylipp: Fixed. completeRequestRequested now uses LATEST instead of the configured project version when ignoreProject is set.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
Caylipp: Fixed. Installations with --ignore-project are now rejected for tools that do not support the software repository.

quando632:
Thanks for the fast turnaround 🙌 All five points are addressed the `isIdeHomeRequired` split, the `completeRequestRequested` version fix, the guarded `setVersion`, the rewritten test, and the `CliException` for `isIgnoreSoftwareRepo` tools. I checked out the current head and the affected suite is green (18 tests, BUILD SUCCESS).

One small open question, no blocker: should `--ignore-project` also ignore the configured *edition* (`getConfiguredEdition()` is still used unconditionally), or is version-only intended? Happy either way just let me know and I'll approve.

Caylipp:
Thanks for checking. Version-only is intended for this issue, so the configured edition should remain unchanged.

quando632 (approved):

cli/src/main/java/com/devonfw/tools/ide/commandlet/InstallCommandlet.java
```diff
@@ -42,6 +46,7 @@ public InstallCommandlet(IdeContext context) {
     addKeyword(getName());
     this.tool = add(new ToolProperty("", false, "tool"));
     this.version = add(new VersionProperty("", false, "version"));
+    this.ignoreProject = add(new FlagProperty("--ignore-project"));
```
hohwille: My thought was that this is only for internal edge-cases:
As described in #1976 for launching the GUI what requires mvn what again requires java but we want to be able to launch the GUI independent from an IDEasy project so it should not fail if I am not inside an IDEasy project and also it should not modify the current IDEasy project in case I am inside one when launching the GUI.

Do you see any use-case or scenario where an end-user would explicitly want to invoke this feature?
I currently do not see that and therefore would not expose this via CLI.
But maybe I am missing something so if you have an idea in mind why end-users would want to use this, please let me know and we could keep it exposed.

CHANGELOG.adoc
```diff
@@ -15,6 +15,7 @@ The full list of changes for this release can be found in https://github.com/dev
 == 2026.07.002
 
 Release with new features and bugfixes:
+* https://github.com/devonfw/IDEasy/issues/1976[#1976]: Extend Tool Commandlet installation logic
```
hohwille: Sorry, but needs to be moved up to 2026.08.001.

hohwille (approved):
@Caylipp thanks for your PR. Nice solution and good job 👍 
I was actually looking for the place where tool dependencies get processed and wanted to check that you did not forget to set the flag from the parent request but you nicely solved that already in the `ToolInstallRequest` constructor what is more elegant so it is automatically inherited but could still be changed after construction in an edge-case. Very nice!

I did not expect that we expose this feature via the CLI to end-users. Have a look at my comments and lets see if we should revet this making this PR quite smaller or if we see value in it and should keep it.

BTW: While reviewing the Gui commandlet again via this PR, I was wondering, why we have to trigger installation of both `java` and `mvn` explicitly there even though `mvn` has a dependency to `java`. If I am not mistaken, we could now remove that (the redundant `java` installation) and before that was only a workaround because this story was not implemented.
Feel free to consult @laim2003 - in case we still need this redundant dependent java installation explicitly, it would be worth adding a comment why we need it.

CHANGELOG.adoc
Caylipp: Thanks for the hint!

cli/src/main/java/com/devonfw/tools/ide/commandlet/InstallCommandlet.java
Caylipp: Thanks for the clarification. I do not have an additional enduser use case in mind, so I agree that ignoreProject should remain an internal installation option. I will remove the CLI exposure and keep the internal request handling for GUI and dependency installations.

PR #2077, opened 2026-06-26
#1651: updated doc after generating new access token

hohwille:
### Implemented changes:

* Generated new access token and updated to our `ACTION_PUSH_TOKEN`
* Took screenshots during the process
* Updated the screenshots and documentation accordingly.

---

### Testing instructions

Please add conscise, understandable instructions on how a reviewer can test/verify the functionality of your contribution here:

1. There is nothing that can be tested explicitly.
2. Review the documentation changes and view the new result on GitHub (in Browser)
3. Observe our GitHub action workflows that all is still working (and when we merge PR #2003)

---

### Checklist for this PR

Make sure everything is checked before merging this PR. For further info please also see
our [DoD](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc).

- [x] When running `mvn clean test` locally all tests pass and build is successful
- [x] PR title is of the form `#«issue-id»: «brief summary»` (e.g. `#921: fixed setup.bat`). If no issue ID exists, title only.
- [x] PR top-level comment summarizes what has been done and contains link to addressed issue(s)
- [x] PR and issue(s) have suitable labels
- [x] Issue is set to `In Progress` and assigned to you *or* there is no issue (might happen for very small PRs)
- [x] You followed all [coding conventions](https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc)
- [x] You have added the issue implemented by your PR in [CHANGELOG.adoc](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc) unless issue is labeled
  with `internal`
- [x] You have formulated clear instructions on how to test your contribution under "Testing instructions"

PR #2146, opened 2026-07-10
#1651: Push and commit status.json files to ide-urls-status after running the URL update workflow

### LINKED ISSUE #1651: Move status.json files out of ide-urls to own repo

laert-ll:
### This PR fixes #1651 

Continuation of [PR#2003](https://github.com/devonfw/IDEasy/pull/2003)

### Implemented changes:

* Moved `status.json` handling out of the `cli` URL model into the `url-updater` module: `StatusJson`, `UrlStatus` and `UrlStatusState` are relocated, and the old `AbstractUrlFile`-based `UrlStatusFile` is replaced by a lightweight `Path`-based wrapper.
* The `status.json` path is now derived from the version folder instead of maintaining a parallel `UrlRepository` tree; the status repository is passed around as a plain `Path`.
* `UpdateManager` / `UpdateInitiator` accept the separate `ide-urls-status` repository, and the `update-urls` workflow checks it out and pushes to it.

---

### Checklist for this PR

  with `internal`


url-updater/src/main/java/com/devonfw/tools/ide/url/updater/status/UrlStatusFile.java
```diff
@@ -1,38 +1,54 @@
...
 import com.devonfw.tools.ide.json.JsonMapping;
-import com.devonfw.tools.ide.url.model.file.json.StatusJson;
-import com.devonfw.tools.ide.url.model.folder.UrlVersion;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
-/**
- * {@link UrlFile} for the "status.json" file.
- */
-public class UrlStatusFile extends AbstractUrlFile<UrlVersion> {
+public class UrlStatusFile {
```
maybeec, line 16: **Should-fix:** `UrlStatusFile` lost its class-level JavaDoc in this move (the old class had `{@link UrlFile} for the "status.json" file.`). Since this is now a plain, non-`AbstractUrlFile` wrapper with its own load/save/delete lifecycle, a short JavaDoc explaining what it represents and how that lifecycle works (lazily loaded in the constructor, `save()` only writes when `modified` was set) would help future readers understand the contract. See `coding-conventions.adoc` § Code-Documentation.

cli/src/main/java/com/devonfw/tools/ide/url/model/folder/UrlVersion.java
```diff
@@ -204,7 +185,7 @@ protected UrlFile<?> newChild(String name) {
   @Override
   protected boolean isAllowedChild(String name, boolean folder) {
 
-    return true;
+    return !"status.json".equals(name);
```
maybeec: **Should-fix:** `"status.json"` here duplicates `UrlStatusFile.STATUS_JSON`. Since `UrlStatusFile` now lives in `url-updater`, which depends on `cli` (not the other way round), it can't be imported back here to avoid the literal — but that also means nothing keeps this string in sync if the status filename in `UrlStatusFile` ever changes. Consider a local `private static final String STATUS_JSON = "status.json";` constant here with a comment noting it must stay in sync with `url-updater`'s `UrlStatusFile.STATUS_JSON`, per `coding-conventions.adoc` § Constants.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/status/UrlStatusFile.java
```diff
@@ -89,4 +80,36 @@ public void cleanup() {
       this.modified = true;
     }
   }
+
+  /**
+   * Saves this status file to disk if it has been modified.
+   */
+  public void save() {
```
maybeec, line 99: **Should-fix:** The old `UrlStatusFileTest` (which verified JSON deserialization of timestamps, error code/message) was deleted together with the `cli`-based `UrlStatusFile` and isn't replaced by an equivalent test for this new `Path`-based implementation. This class now has non-trivial lifecycle logic — lazy load in the constructor, `modified`-flag-gated `save()`, `delete()` — that's currently only exercised indirectly through `UrlUpdaterTest`/`AbstractUrlUpdaterTest`. A focused unit test under `url-updater/src/test/java/.../status/` (load existing file, load missing file defaults to empty `StatusJson`, `save()` is a no-op when unmodified, `delete()` removes the file) would keep this testable in isolation, per `documentation/contributing/junit-testing.adoc`.

url-updater/src/test/java/com/devonfw/tools/ide/url/updater/UrlUpdaterTest.java
```diff
@@ -123,15 +123,18 @@ void testStatusJsonUpdateOnFirstError(@TempDir Path tempDir, WireMockRuntimeInfo
     UrlEdition urlEdition = urlTool.getOrCreateChild(editionName);
     UrlVersion urlVersion = urlEdition.getOrCreateChild(versionName);
     // we create the structure of our tool version and URL to simulate it was valid last moth
-    UrlStatusFile statusFile = urlVersion.getOrCreateStatus();
+    Path statusPath = urlVersion.getPath().resolve(UrlStatusFile.STATUS_JSON);
+    UrlStatusFile statusFile = new UrlStatusFile(statusPath);
     UrlStatus status = statusFile.getStatusJson().getOrCreateUrlStatus(url);
     UrlStatusState successState = new UrlStatusState(lastMonth); // ensure that we trigger a recheck of the URL
     status.setSuccess(successState);
+    statusFile.setStatusJson(statusFile.getStatusJson()); // mark as modified so save() persists it
```
maybeec: **Should-fix:** This `setStatusJson(getStatusJson())` self-assignment to force `modified = true` mirrors the pre-existing hack in `AbstractUrlUpdater.java` (`// hack to set modified (better solution welcome)`), and this PR now triples it across this test file (also lines 189 and 244). Since this PR already reworks `UrlStatusFile`'s persistence API, it's a good opportunity to add a small `markModified()`/`touch()` method instead of relying on the no-op setter trick in three more places.

.github/workflows/update-urls.yml
```diff
@@ -17,6 +17,12 @@ jobs:
           repository: devonfw/ide-urls
           path: ide-urls
           token: ${{ secrets.ACTION_PUSH_TOKEN }}
+      - name: Checkout ide-urls-status
+        uses: actions/checkout@v3
+        with:
+          repository: devonfw/ide-urls-status
+          path: ide-urls-status
+          token: ${{ secrets.ACTION_PUSH_TOKEN }}
```
maybeec, line 25: **Should-fix (manual verification):** This checks out `devonfw/ide-urls-status` (and, further down at the "Commit and push status.json files" step, pushes to it) reusing `secrets.ACTION_PUSH_TOKEN`. Issue #1651 explicitly calls out "We need a new access token to also include write access to the new repo" — worth confirming `ACTION_PUSH_TOKEN` actually has write access to `ide-urls-status` before merging, since a permission problem here would only surface on the nightly cron run, not in this PR's CI.

maybeec (commented):
Thanks for picking this back up from #2003 👍 This is a clean split: `status.json` handling moves out of the `cli` URL model into `url-updater`, the `UrlStatusFile`/`StatusJson`/`UrlStatus`/`UrlStatusState` classes move with it, and the GHA workflow now checks out, updates and pushes both `ide-urls` and the new `ide-urls-status` repo. Nice that the old `UpdateManager(Path, UrlFinalReport, Instant)` constructor was kept and just delegates to the new one — `BuildSecurityJsonFiles` in the `security` module still compiles against it unchanged, which is exactly the right way to avoid a breaking signature change.

Against issue #1651's requirements: workflow checks out the new repo ✅, `UrlUpdater`s write status files into the new repo's mirrored folder structure ✅, workflow commits/pushes both repos ✅. One item from the issue doesn't look addressed here: "We need a new access token to also include write access to the new repo" — see the inline comment on the workflow file.

A few should-fix points below, nothing blocking. Also two small housekeeping notes: the PR is currently shown as **behind `main`** by GitHub, so please rebase/update before merging; and the PR title has a small typo ("worklfow" → "workflow").

.github/workflows/update-urls.yml
hohwille: @maybeec thanks for your support, review and caring that nothing will break. 👍 
However, I already updated the token and with #2077 I updated the [documentation that is now here](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/update-access-token.adoc).

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
hohwille, line 866: Before the parent was ensuring that this is saved recursively.
Do we still ensure this now?

hohwille, line 895: same here...

hohwille: Seems this is called either from `updateExistingVersion` or from `doAddVersionUrlIfNewAndValid` and both methods explicitly ensure changes are saved.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
```diff
@@ -695,6 +702,10 @@ private boolean doAddVersionUrlIfNewAndValid(String edition, String url, UrlVers
     doUpdateStatusJson(success, statusCode, edition, urlVersion, url, urlDownloadFile, update);
 
     urlVersion.save();
+    UrlStatusFile urlStatusFile = getStatusFile(urlVersion, false);
+    if (urlStatusFile != null) {
+      urlStatusFile.save();
+    }
```
hohwille: The `status.json` contains the entire status of a tool/edition/version folder so it combines the information of all urls files (`urlVersion`).
The original idea was that `save()` method is only called once per file so we avoid writing updates to the same file again multiple times.
I have the following idea for simplification:
1. You could change the `statusFiles` cache (`Map<Path, UrlStatusFile>`) to `Map<UrlVersion, UrlStatusFile>` and make the method `getStatusFile(UrlVersion urlVersion, boolean create)` easier for accessing the Map. Also then it gets more clear, what the meaning of the map key is about. A Path could be pointing to the UrlVersion folder, the status file, or something else.
2. You could move the `Map` to the `UpdateManager` to keep this state global for the process instead of creating a Map again for each single updater. We sometimes have multiple updaters for the same tool but different editions or OS that would write to the same `status.json` and could not reuse the cache then. Simply pass the `UpdateManager` to the `AbstractUrlUpdater` just like we do for `urlFinalReport` and then the updater can access status files via API of `UpdateManager` reusing the cache.
3. At the end either in `UpdateInitiator` or in both `updateAll()` and `update(String)` of `UpdateManager`, we would trigger saving all status changes. Therefore simply iterate over the values in the map and call the save method. This would guarantee that we both save every file exactly once (nice-to-have performance tuning) and that we never can forget saving a status file (that is what I currently fear could happen with your current PR).

hohwille (approved):
@laert-ll thank you for taking over this story. You perfectly catched up the requirements from story and review remarks to bring this to the finishing line. 👍 
My remarks are all about managing and saving the new status.json files and the main question I have is, whether after merging the current state, it could happen that we modify a status file in memory but forget to write it to disk what would be a bug causing potential problems. If this is not the case and we finally want to get this done, we can also create a cleanup ticket and merge this PR as is and do the cleanup later when we have time for it.
If you think you can address the review comments quickly and prefer to include them into this PR that is also fine even in case there is no bug.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/UpdateManager.java
```diff
@@ -89,15 +91,28 @@ public class UpdateManager extends AbstractProcessorWithTimeout {
...
+    this.statusRepositoryPath = pathToStatusRepository;
     this.urlFinalReport = urlFinalReport;
     setExpirationTime(expirationTime);
   }
 
+  /**
+   * The constructor.
+   *
+   * @param pathToRepository the {@link Path} to the {@code ide-urls} repository to update.
+   * @param expirationTime for GitHub actions url-update job.
```
hohwille, line 116: Parameter `urlFinalReport` is not documented in both constructors.
Was already before so not a problem of your PR.
But if we create a cleanup ticket, we should consider this for completeness.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
laert-ll: Thank you for the detailed suggestion!

Cache is now `Map<UrlVersion, UrlStatusFile>` and lives in `UpdateManager`, updaters get it via `setUpdateManager`  and saveStatusFiles at the end of `updateAll`/`update(String)` writes everything at once. However I had to change the tests given that they need an `UpdateManager` to run the updater.

hohwille (approved):
@laert-ll thanks for your patience and endurance to keep updating this until perfection. 👍 
Now I am confident that all status.json files updates are guaranteed to be saved and the design looks clean to me. Great job!

hohwille:
Since the workflow is merged, the status files are changed here as expected:
https://github.com/devonfw/ide-urls-status/commits/main/

Recent commits on ide-urls do not include status.json files anymore.
Example:
https://github.com/devonfw/ide-urls/commit/185c988b26b0f2f4edefe77e33f8ec4b36ec3c43

So IMHO all is working as planned. Setting QA approved label.

PR #2153, opened 2026-07-13
#2133: add repository navigation option to icd command

### LINKED ISSUE #2133: Add repository navigation option (-r) to icd command
### Feature idea

Extend the `icd` command by introducing a repository navigation option `-r` that allows users to directly navigate not only to a project but also into a specific repository within that project.

The new option should enhance the current navigation behavior by supporting both explicit and implicit repository selection.

### Expected behavior:

1. Basic usage (no argument)

```
icd -p IDEasy -r
```

No repository name is provided → fallback to the `«project-name»`.
If no folder `«project-name»` exists in the workspace(s) (start searching in `main` then other workspaces if no `-w` option is given) but only a single folder exists (not starting with a dot) in the `main` workspace (or the explicitly specified workspace) then take that folder.
Otherwise only cd into the workspace and log a "warning".

2. Explicit repository selection
```
icd -p IDEasy -r «repository-name»
```

The value after `-r` is interpreted as the repository (folder) name within the selected project. The repository is resolved in the context of the selected workspace or searched in all workspaces starting with `main` if no workspace is explicitly given.

3. Workspace resolution behavior
If no workspace is explicitly provided, the command should use `main` as default (workspace `main` must always exist for a valid IDEasy project).
 
4. Error handling
If no matching repository is found in any workspace, the command should fail with a clear and informative error message, e.g:
```
Error: Repository '«repository-name»' not found in any workspace.
```

### Examples:
Example A (found in `main`): 
` icd -p IDEasy -r ` 
→ navigate to `/workspaces/main/IDEasy`

Example B
`icd -p IDEasy -w dev -r backend/core`
 
→ navigate to `/workspaces/dev/backend/core`

Example C (repository not found anywhere)
`icd -p IDEasy -r non-existing-repo`

Search:
`main` ❌
`dev` ❌ 
 
```
Error: Repository 'non-existing-repo' not found in any workspace.
```
### Additional context

_No response_

Ali-Shariati-Najafabadi:
This PR fixes #2133

Implemented changes:

- Added a new `-r | --repository [<repository>]` option to the `icd` shell function in `cli/src/main/package/functions`
- If `<repository>` is omitted, the project name is used as the repository name (fallback)
- If `-w` is not given, workspace `main` is searched first, then all other workspaces; `icd` navigates into the first workspace containing the repository
- If `-w` is given explicitly, only that workspace is searched (no fallback)
- If the repository is not found in any workspace, `icd` fails with `Error: Repository '<name>' not found in any workspace.`
- Updated `icd --help` usage/options text
- Updated `documentation/cli.adoc` with a description and example of the new option

---

Testing instructions

Run the automated test suite, which covers all scenarios from #2133 in both bash and zsh:
bash cli/src/test/functions-test.sh

---

Checklist for this PR

- [x] You have added the issue implemented by your PR in CHANGELOG.adoc unless issue is labeled with `internal`


cli/src/main/package/functions
```diff
@@ -67,6 +67,8 @@ function icd() {
   fi
   local icd_project=
   local icd_workspace=
+  local icd_repository=
+  local icd_repository_given=false
```
maybeec, line 71: **Must-fix (zsh):** the two new workspace-scan loops below — `for icd_ws_dir in "${icd_workspaces_dir}"/*/` (search across workspaces, line 169) and `for icd_entry in "${icd_fallback_dir}"/*/` (single-folder fallback, line 194) — break under zsh when the glob has no match. Bash passes an unmatched glob through as a literal string (your `[ -d ... ] || continue` guards handle that), but zsh prints `no matches found` and aborts the whole `icd` call — that is its default [`NOMATCH`](https://zsh.sourceforge.io/Doc/Release/Options.html#index-NOMATCH) behavior. This matters because this file is not bash-only: `IdeasyCommandlet` wires `source "$IDE_ROOT/_ide/installation/functions"` into `.zshrc` too, and zsh is the default shell on macOS.

Reproduced with zsh 5.9 against a test `IDE_ROOT`:

```console
% icd -p proj-empty -r            # workspaces/main exists but is still empty (fresh project)
icd: no matches found: …/proj-empty/workspaces/main/*/
% icd -p no-such-project -r foo   # typo in the project name
icd: no matches found: …/no-such-project/workspaces/*/
```

In both cases the function aborts: neither the "cd into the workspace and log a warning" fallback required by #2133 nor the clean `Error: Repository ... not found` message is ever reached. In bash both cases work as specified.

The smallest fix is to give zsh bash-like glob semantics for the scope of this function. Verified with zsh 5.9: with this guard both cases above behave exactly like in bash, and the happy paths (repo in `main`, repo in another workspace) stay unchanged:

```suggestion
  local icd_repository_given=false
  # zsh aborts on unmatched globs by default (NOMATCH) - let the workspace-scan globs below behave like in bash
  if [ -n "${ZSH_VERSION:-}" ]; then
    setopt local_options null_glob
  fi
```

cli/src/main/package/functions
```diff
@@ -133,11 +147,71 @@ function icd() {
...
+            icd_found_workspace=${icd_ws_name}
+            break
+          fi
+        done
+      fi
+    fi
+    if [ "${icd_found_workspace}" != "" ]; then
+      icd_path=${icd_workspaces_dir}/${icd_found_workspace}/${icd_repo_name}
+    elif [ "${icd_repo_name_given}" = true ]; then
+      echo "Error: Repository '${icd_repo_name}' not found in any workspace." >&2
```
maybeec: **Should-address:** when `-w` is given explicitly, only that one workspace is searched (which is the intended behavior per your PR description), but the message still claims the repository was not found "in any workspace". Example: with a repository `cli` that exists in workspace `main`, `icd -w dev -r cli` reports `Error: Repository 'cli' not found in any workspace.` — which is factually wrong and may send users looking in the wrong direction instead of just dropping `-w`. Issue #2133 only specifies this message for the search-all-workspaces case, so differentiating is fine:

```suggestion
      if [ "${icd_workspace}" != "" ]; then
        echo "Error: Repository '${icd_repo_name}' not found in workspace '${icd_workspace}'." >&2
      else
        echo "Error: Repository '${icd_repo_name}' not found in any workspace." >&2
      fi
```

documentation/cli.adoc
```diff
@@ -132,3 +132,17 @@ IDE environment variables have been set for /~/projects/myproject in workspace m
 IDE environment variables have been set for /~/projects/myproject in workspace main
 [~/projects/myproject]$ echo "Wow! IDEasy is awesome!"
 ```
+
+You can also use `-r «repository»` to jump straight into a repository (folder) inside a workspace.
```
maybeec, line 136: **Minor:** issue #2133 (example B) also calls out nested repository paths — `icd -p IDEasy -w dev -r backend/core` → `workspaces/dev/backend/core` — and your implementation already supports this (verified in both the explicit `-w` and the search-all-workspaces mode). Worth a sentence here so users can discover it:

```suggestion
You can also use `-r «repository»` to jump straight into a repository (folder) inside a workspace.
The repository can also be a nested path such as `-r backend/core`.
```

maybeec (changes requested):
Thanks for this contribution — the implementation covers the issue requirements well. I verified all scenarios from #2133 hands-on: sourced the `functions` file from this PR against a synthetic `IDE_ROOT` (with a stubbed `ideasy`) and exercised every path in **bash** and in **zsh 5.9**. zsh matters here because `IdeasyCommandlet` wires `source "$IDE_ROOT/_ide/installation/functions"` into `.zshrc` as well, and zsh is the default shell on macOS.

### Verification against issue #2133

| Requirement (#2133) | Result |
|---|---|
| 1: `icd -p X -r` falls back to the project name | ✅ verified |
| 1: no project-name folder, but exactly one non-hidden folder in the workspace → enter it | ✅ verified |
| 1: otherwise cd into the workspace and log a warning | ✅ bash — ❌ zsh aborts when the workspace is empty (see inline comment) |
| 2: explicit `icd -p X -r «repo»`, incl. nested paths like `-r backend/core` (example B) | ✅ verified |
| 2/3: resolved in the given workspace, or searched `main` first, then all other workspaces | ✅ verified |
| 4: clear error message when not found | ✅ bash — ❌ zsh aborts when `workspaces` is missing/empty; wording misleading with explicit `-w` (see inline comments) |
| Help text + `documentation/cli.adoc` updated | ✅ |

### Must-fix

1. **zsh: the new workspace-scan globs abort `icd`** — see the inline comment in `functions` (includes a verified one-click fix).
2. **Missing CHANGELOG.adoc entry.** #2133 is not labeled `internal`, so per the PR checklist and the [DoD](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc) the issue must be added to [CHANGELOG.adoc](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc) under the next open release (currently the `== 2026.07.002` section):

   ```
   * https://github.com/devonfw/IDEasy/issues/2133[#2133]: Add repository navigation option (-r) to icd command
   ```

### Should-address

3. **No automated tests, although a test harness for this exact file exists and runs in CI:** [`cli/src/test/functions-test.sh`](https://github.com/devonfw/IDEasy/blob/main/cli/src/test/functions-test.sh) unit-tests the `functions` file and is executed on every PR build (see `.github/workflows/build-pr.yml`). TDD is our standard for features (see `AGENTS.md`), and your manual testing instructions map 1:1 to automatable checks using the existing pattern there (fresh `bash --noprofile --norc`, stub `ideasy` on `PATH`, temporary `IDE_ROOT` tree). Suggested minimum cases: repo found in `main`; repo found in another workspace; explicit `-w` restricting the search; not found → exit code 1 + error message; implicit `-r` resolving the project-name folder; single-folder fallback; warning + cd into the workspace for an empty/ambiguous workspace. Is this planned as a follow-up, or can it land in this PR?
4. **Misleading error message when `-w` is given explicitly** — see inline suggestion in `functions`.

### Minor

5. **Testing instructions step 5 is outdated** since the fallback behavior was added: with the tree from step 1 (`main` only contains `cli`), `icd -p IDEasy -r` now lands in `workspaces/main/cli` via the single-folder fallback — not in `workspaces/main/IDEasy` (that folder is never created in step 1). Verified in bash. Please update the PR description (e.g. create `workspaces/main/IDEasy` for step 5 and add a separate step for the single-folder fallback).
6. **Docs: nested repository paths are supported but not documented** — see inline suggestion on `cli.adoc`.

Everything else looks good: the option parsing is consistent with the existing `-w` handling, the search order matches the issue, existing `icd` invocations without `-r` are unaffected, `--help` and the docs are updated, CI is green, and the branch is up to date with `main`.

maybeec:
One design concern about the implicit `-r` fallback — this questions the requirement from #2133 rather than this implementation, which follows it faithfully (cc @vivu001):

When `-r` is given without a value and no folder matching the project name exists, the "single non-hidden folder" fallback silently drops you into a folder whose name has nothing to do with what was asked for. Example: in a project `IDEasy` whose `main` workspace only contains `ide-urls`, `icd -r` ends up in `.../main/ide-urls` without any hint. This makes the result hard to predict ("what will icd pick?") and invites misconceptions such as repositories being searched recursively (they are not — a repository is always resolved as a direct child of a workspace, or as a literal relative path like `-r backend/core`).

I would find it more intuitive to drop the auto-descend and always cd into the workspace with the existing warning when the project-name folder is not found — i.e. treat the "exactly one folder" case the same as the ambiguous case. Then `icd -r` stays fully deterministic: you either land in the folder you named (or the project-name folder), or in the workspace itself with a warning.

@vivu001 was there a specific use case behind the single-folder shortcut in #2133? If we agree to change this, the issue should be updated accordingly and this PR would actually get simpler (the fallback block shrinks to warning + cd into the workspace).

vivu001:
Hi @maybeec , thanks for the suggestion. I think the concern about implicit behavior is well-founded. To make the behavior explicit, I would summarize your suggestion as follows:

```
icd -p IDEasy -r
→ /workspaces/main/IDEasy
```

- `-r` without a value → no repository is specified
- no fallback or heuristics should be applied
- show a warning (e.g. missing repository)

About the single-folder shortcut: there does not seem to be a strong functional use case behind it. It mainly acts as a convenience shortcut, but introduces implicit behavior.

A hybrid approach (strict default behavior with optional convenience) could work. I even considered something along these lines:
E.g.: 

- By default (strict behavior):
```
icd -p IDEasy -r 
→ /workspaces/main/IDEasy 
Warning: No explicit repository was specified.
```
- With explicit user intent (convenience behavior):
```
icd -p IDEasy -r c 
→ /workspaces/main/IDEasy/cli
```


Overall, keeping repository selection explicit aligns better with the goal of making behavior clear and unambiguous.
Even so, I would still like to get @hohwille ’s opinion on this matter 👍.

maybeec (approved):

hohwille (approved):
@Ali-Shariati-Najafabadi thanks for your PR. Awesome job especially with your nice integration test. 👍 
Hope to see the integration test green tomorrow ;)

hohwille:
Retested with 2026.07.002-07_21_05-SNAPSHOT on Windows.
All working fine. I love this new feature - very handy. Thanks! 👍

PR #2162, opened 2026-07-14
#2156: Add secret placeholders for configuration templates

Caylipp:
### Implemented changes:

* added generic `$[secret:<name>]` placeholders for files under `templates/conf/`
* prompted once per distinct secret name within each template file
* supported repeated and multiple distinct placeholders
* ensured secrets are only written to generated local `conf/` files
* preserved source templates and existing local configuration files
* prevented secret values from appearing in production logs
* handled `$` and `\` in secret values safely

---

### Testing instructions

Please add conscise, understandable instructions on how a reviewer can test/verify the functionality of your contribution here:

1. Run the focused tests: mvn -pl cli -Dtest=CreateCommandletTest test
2. Run the complete CLI test suite: mvn -pl cli test
3. Optional: Run the ./build-local-dev.sh script and create a test project

---

### Checklist for this PR

Make sure everything is checked before merging this PR. For further info please also see
our [DoD](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc).

- [x] When running `mvn clean test` locally all tests pass and build is successful
- [x] PR title is of the form `#«issue-id»: «brief summary»` (e.g. `#921: fixed setup.bat`). If no issue ID exists, title only.
- [x] PR top-level comment summarizes what has been done and contains link to addressed issue(s)
- [x] PR and issue(s) have suitable labels
- [x] Issue is set to `In Progress` and assigned to you *or* there is no issue (might happen for very small PRs)
- [x] You followed all [coding conventions](https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc)
- [x] You have added the issue implemented by your PR in [CHANGELOG.adoc](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc) unless issue is labeled
  with `internal`
- [x] You have formulated clear instructions on how to test your contribution under "Testing instructions"

Issue #989, opened 2025-01-24
allow expressions in template variable definitions


hohwille:
Currently we support variable expressions like `$[IDE_HOME]` in templates.
In such case if the variable is defined, the expression is literally replaced with its value.
This is fine for most of our use-cases.

However, in some cases we might end up with resolving `$[IDE_HOME]/software/mvn` to `D:\projects\my-project/software/mvn` on Windows.
Often tools are tolerant and will accept this mixed up path syntax.
However, in some cases backslashes may need extra escaping and the result may not work.
Also we have cases like this:
https://github.com/devonfw/ide-settings/blob/3c41b5e27ad4baa90d4e3ffdf2ac16ca5e95d11a/intellij/workspace/update/.intellij/config/options/nodejs.xml#L4-L5
```
    <!-- ATTENTION: Will only work on windows so far -->
    <local-interpreter path="$[IDE_HOME]/software/node/node.exe" merge:id="name()"/>
```

We could still workaround this somehow by adding a variable `EXECUTABLE_SUFFIX` set to `.exe` in `settings/ide.properties` and then a MacOS or linux user could override the property as empty (`EXECUTABLE_SUFFIX=`) in his `config/ide.properties` to make that portable.

## Expression Syntax

The syntax should be as following:
```
@«function-name»([«arg»[,«arg»]*])
```
Or to make it more clear as regexp:
```
"@([a-z-]+)\\(([^\\)]*)\\)"
```
The first group is the `«function-name»` and the second group are the arguments.
We need to parse the individual arguments again as RegEx is not capable of expressing such list of any number of arguments properly.
For simplicity arguments are always of type String.
Multiple arguments are separated by comma - spaces around arguments are optional and will be ignored/trimmed.
Arguments will be quoted as a literal value e.g. `'My example value'` or `"My example value"`.
Such String argument itself may also contain variables `'$[IDE_HOME]/software/node/node'`.

### Functions

`@path` normalises a path 
1. argument is the path to normalise, by default backslash is replaced with slash
2. argument is optional and can be literal value `unix` (default) or `native` (on Windows use WindowsPathSyntax.WINDOWS with backslashes, etc.)

`@ask-variable` asks for a variable in plain text and `@ask-secret` asks for a secret variable (input is masked).
1. argument is the name of the requested variable. If the variable is already defined it will be returned without asking. If the empty string is given as 1st argument, then the user will always be asked.
[...]

Issue #987, opened 2025-01-24
conditional templates


hohwille:
As a IDEasy maintainer, I want to be able to configure conditional expressions for templates or template-sections so that I can configure aspects according to the project situation.

Currently we can easily have entirely static configuration blocks for IDEs (Eclipse, IntelliJ, VSCode, etc.).
Additionally, we can also use variables resolved from the project that are replaced in the template (e.g. `$[IDE_HOME]/software/mvn`).

However, we are currently preconfiguring a typical Java + Maven project in our settings.
What is some project wants to use DotNet/C# + NuGet instead?
How about C++ projects? This can be done with Eclipse but that needs totally different configuration templates.
What if many projects but not all want to use a custom AI plugin and would like to have it preconfigured but maybe if the configuration is present without the plugin, then the IDE is causing errors?

The idea of this epic is to raise ideas and finally a concept that allows us to create and maintain a generic IDE configuration in `ide-settings` that allows to be used by projects with very different demands.
Assuming we have a specific configuration template that configures a maven integration but the project does not use maven at all then this configuration does not seem to make sense. The same also applies to Java, NodeJS, etc.

# Ideas

How about the following:

* like we allow merger metadata for XML merger in templates already, we could define an attribute like `merge:condition="«conditional-expression»"` and this would mean that this element with all its children is only considered if `«conditional-expression»` evaluates to `true`. Otherwise, we consider the entire block as non-existent and in case of the root element this would mean the entire template file is ignored as if it does not exist.
* We could define a syntax for such conditional expressions allowing to check if a tool or plugin is present, a version is less than or greater than a given version, a variable has a specific value, combinations of such expressions combined with AND as well as OR, etc.
* That syntax should be powerful but still easy to read and write. There are already some conflicts coming from XML so by default allowing '|' for logical OR and `&` for logical AND, we need to consider that `&` has to be escaped in XML as `&amp;` what makes the expression harder to read. So maybe staying with `AND` + `OR` would be better. Also we should consider a simply syntax for common things like checking if a tool is present. For readability this could be `isInstalled(mvn)` but on the other hand this may already be to long for typing very common things.
[...]

Issue #412, opened 2024-06-25
consider integration of lama


hohwille:
As an IDEasy user, I want to make use of AI tools so that I can get smart coding assistance.
However, some projects have high requirements for data-security and are not allowed to use cloud based AI solutions from the hyperscalers. Therefore e.g. they host their own OpenLlama AI engine and train it with their code-base and project content in a private and secured network.
In order to use this approach with IDEasy you need to integrate that AI engine into the IDE.
This story is about the integration in IntelliJ that happens via this plugin:
https://plugins.jetbrains.com/plugin/18577-lama

All we would need to do in the first place (acceptance criteria of this story) is to have this plugin pre-configured and tagged with `ai` in our `ide-settings` repo. It should be inactive by default.

Potentially later also configuration templates (and wizards?) could be offered for projects to also automatically (pre-)configure the plugin to be connected with their private backend AI engine.

Paras14:
Hey, I picked this up and did some digging, and a fair bit has changed since this was filed ~2 years ago, so I want to write down what I found before going further.
The description leaves the plugin open ("some plugin", candidates are…) and the two links are older Llama-3 / Code-Llama how-tos, so there's no decided plugin to just wire in, picking a current, maintained one is basically the task.
The obvious candidate from the #449 side was Continue, but Continue got acquired by Cursor in mid-June 2026. The continuedev/continue repo is now read-only and not really maintained. It still runs, but I don't think we want the IDEasy standard resting on something that just went end-of-life upstream.
Two maintained, open-source alternatives that fit the self-hosted data security case, both can point at our own OpenAI-compatible backend (Ollama, LM Studio, any OpenAI-compatible endpoint), so the private OpenLlama scenario works with either:

Cline:- bigger and more mature overall, but its JetBrains plugin specifically is not open source.
Kilo Code:- the whole plugin including JetBrains is Apache-2.0 and auditable.

For an open-source, security-sensitive project like ours I'm leaning Kilo Code. Both JetBrains plugins are still fairly new though, so we'd want to validate before rollout.
The pre-config part itself is straightforward and we already do it for other AI plugin, so a new one is just id=<plugin id>, active=false, tags=ai in the same place.
The real open piece is the same one that closed ide-settings PR #67: it's not that per-project config doesn't exist, it's that a config in the repo gets pushed to everybody (incl. non-AI projects) and hard-codes a backend URL that's project-specific. 
[...]

PR #2179, opened 2026-07-21
#2165: interactive template variables

### LINKED ISSUE #2165: prompt user for template variables at apply time
### Feature idea

As an IDEasy user, I want to be asked for configuration values that cannot be preconfigured in the settings repo (like a backend URL or an API key) when the template is applied, so that plugins like AI assistants can be preconfigured without committing project-specific or secret values.
This came out of #412: to preconfigure AI coding assistants (Kilo Code, Cline) pointing to a self-hosted backend, we need config with a backend URL and an API key/token. The URL is project-specific and the key is a secret, so neither can be hardcoded in ide-settings. This is also the reason devonfw/ide-settings#67 was closed back then.
We already have variable substitution with the $[...] syntax (VariableSyntax.SQUARE) that gets resolved when templates are merged. But if a variable is not defined, AbstractEnvironmentVariables.resolveRecursive just logs a warning ("Undefined variable...") and leaves the placeholder in the file. So the resolution engine exists, but there is no way to acquire a missing value.
Interestingly we already solved this once, hardcoded for maven: Mvn.java scans settings.xml for $[...] variables, prompts via IdeContext.askForInput ("Please enter secret value for variable..."), and encrypts the value with mvn --encrypt-password. The idea of this issue is to make that generic instead of maven-only.

# Ideas

* a template can declare that a variable should be prompted from the user at apply time if undefined
distinguish two kinds of values:
    * plain values (e.g. AI_BACKEND_URL) will have normal input.
    * secrets (e.g. AI_API_KEY) will have inputs that should be masked (****). Note: askForInput currently echoes input in plain text, so a masked input mode is needed.

* storing the entered value as plain text is acceptable for now, since we have no generic encryption for external config files anyway (only maven has its own settings-security.xml mechanism). Encryption could be a follow-up.

* the syntax should be aligned with #989, which already proposes function expressions.

Related: #412, #449, #987, #989

### Additional context

_No response_

Paras14:
### Implemented changes:

* Extended `VariableSyntax.SQUARE` to additionally match `$[ask:VARIABLE]` and `$[secret:VARIABLE]` expressions.
* Added `IdeContext.askForSecret` for masked console input via `Console.readPassword`.
* During variable resolution, ask expressions for undefined variables prompt the user and persist the entered value to the user's `conf/ide.properties` so they are not asked again.
* Ask expressions for already defined variables resolve like plain variables without any interaction.
* In batch mode ask expressions are skipped with a warning.
* Updated `configurator.adoc` and added tests in `EnvironmentVariablesTest`.

As discussed with @hohwille, the prefix syntax encodes secret vs. plain in the name instead of using function arguments, which aligns it with the $[secret:name] syntax introduced by #2162. A future @function(args) syntax as proposed in #989 can be added as a further alternative in the same regex. Motivated by #412 to preconfigure AI assistant plugins with a backend URL and API key that must not be committed to the settings repository. See also #987.

---

### Testing instructions

Tested on Windows using **PowerShell**. Strongly recommended to run an actual PowerShell window, as masking needs System.console(), which is null in IDE consoles and when input is piped, and there the input stays visible
with a warning by design.

1. Build the branch and dump the classpath, from the repository root:

       mvn -pl cli -am install -DskipTests
       mvn -pl cli dependency:build-classpath "-Dmdep.outputFile=C:\temp\cp.txt"
       $CP = (Get-Content C:\temp\cp.txt -Raw).Trim()
       $CLASSES = "C:\projects\IDEasy\workspaces\main\IDEasy\cli\target\classes"
       
       _(make sure you add your path correctly for the cli target classes: "<repo>\cli\target\classes")_

2. In a test project create IDEasy\settings\workspace\update\ai-test.properties:

       ai.backend.url=$[ask:MY_URL]
       ai.api.key=$[secret:MY_TOKEN]

   Make sure MY_URL and MY_TOKEN are not yet defined in IDEasy\conf\ide.properties.

3. From the project directory, run the local build:

       cd C:\projects\IDEasy
       java -cp "$CLASSES;$CP" com.devonfw.tools.ide.cli.Ideasy update

   You are prompted before any downloads: MY_URL with visible input,
   MY_TOKEN with hidden(masked) input. Each prompt names the file requiring it.

4. 4. Check that IDEasy\workspaces\main\ai-test.properties contains the entered values
   with no $[...] left, and that both variables were written to
   conf\ide.properties exactly as typed.

5. Run the same command again. You won't get any input prompts this time, the persisted values are reused.

6. Add MY_OTHER=https://example.com to conf\ide.properties, then add the line
   ai.other=$[ask:MY_OTHER] to IDEasy\settings\workspace\update\ai-test.properties.
   Run again. There is no prompt, and IDEasy\workspaces\main\ai-test.properties
   contains ai.other=https://example.com.

7. Remove MY_URL and MY_TOKEN from IDEasy\conf\ide.properties and run with --batch.
   No input prompts, but one warning per expression, and the $[ask:...] and $[secret:...]
   text is left untouched in the output.
---

### Checklist for this PR

- [x] You have added the issue implemented by your PR in CHANGELOG.adoc


cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -273,6 +288,40 @@ private String resolveWithSyntax(final String value, final Object src, final int
...
+   *
+   * @param variableName the name of the variable to resolve.
+   * @param secret {@code true} if the input shall be masked while typing.
+   * @param resolvedVars the {@link EnvironmentVariablesResolved} used to persist the entered value.
+   * @param src the source of the expression (used in the prompt).
+   * @return the entered value or {@code null} if it could not be determined (the expression is then left untouched).
+   */
+  private String resolveAskExpression(String variableName, boolean secret, AbstractEnvironmentVariables resolvedVars, Object src) {
+
+    if ((this.context == null) || this.context.isBatchMode()) {
```
QuangAnhLe, line 303: `this.context` is set in the constructor: ( line 63)
```
if (context == null) {
this.context = parent.context; // can be null if the root parent is also null
}
```

If `parent` is null and `context` is null → throw `IllegalArgumentException`. If `parent` is not null but `parent.context` is null... the chain might be null. However, in practice, `AbstractEnvironmentVariables` is always created via `AbstractIdeContext.createVariables()` → so a context always exists. While this scenario is unlikely, it is best to include a guard. 👍

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -1061,6 +1055,36 @@ public String askForInput(String message, String defaultValue) {
     }
   }
 
+  @Override
+  public String askForSecret(String message) {
```
QuangAnhLe, line 1108: In `AbstractIdeContext.askForSecret()`, an empty input repeats the loop with no feedback. `askForInput()` has the same behaviour but at least the user can see what they typed. For masked input the user has no idea if they typed nothing or if the terminal swallowed their input. Consider logging a hint like "Empty input — please enter a value" before the next iteration. :)

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -1061,6 +1055,36 @@ public String askForInput(String message, String defaultValue) {
...
+          return null;
+        } else {
+          throw new CliAbortException();
+        }
+      }
+      String input = readSecretLine().trim();
+      if (!input.isEmpty()) {
+        return input;
+      }
+    }
```
hohwille, line 1125: I am not sure if a secret value MUST NOT be empty.
It may be odd but in some test/dev scenarios empty passwords may be used for convenience.

hohwille (commented):
@Paras14 and @QuangAnhLe this issue seems to be in team-review for quite a while.
As you already commented in the PR and we discussed in the daily your idea to implement this via #989 as according functions would make more sense and allow a generic and more future-proof solution.
I would really like to avoid that we and potential IDEasy projects start using this `ask`  and `secret` syntax and we later deprecated it replacing it with `@ask` function as already suggested and fully making sense.
Maybe it was a misunderstanding in the daily but I thought that someone had already implemented the function syntax and we should therefore use that... I could not find any PR implementing #989 so it seems I got something wrong.
Would you agree that it makes sense to update your PR to implement #989 and using the function syntax `@ask` (or `@ask-variable` / `@ask-secret` if we want to simplify usage and prevent that the secret flag needs to be passed as parameter).
This way almost no work in this PR would be in vain. Most changes are needed anyway.

UPDATE: I have refined #989 and set it as `read-to-implement`.

cli/src/main/java/com/devonfw/tools/ide/variable/VariableSyntax.java
```diff
@@ -23,13 +23,24 @@ public String create(String variableName) {
   },
 
   /**
-   * Syntax using square brackets ("$[...]").
+   * Syntax using square brackets ("$[...]"). Additionally to plain variables ("$[MY_VARIABLE]") this syntax supports the function expression
+   * "$[ask:MY_VARIABLE]" or "$[secret:MY_VARIABLE]" that will interactively ask the user for the value if the variable is undefined.
    */
-  SQUARE("\\$\\[([a-zA-Z0-9_-]+)\\]") {
+  SQUARE("\\$\\[(?:([a-zA-Z0-9_-]+)|(ask|secret):([a-zA-Z0-9_-]+))\\]") {
```
hohwille, line 29: BTW: It is better to have the prefix/function as an optional group without duplicating the variable group like this (and our convention is also to mark opening parenthesis with the according group index for readability):

```
  // ...........1.2...........3
  SQUARE("\\$\\[(@([a-z-]+):)?([a-zA-Z0-9_-]+)\\]") {
```

QuangAnhLe:
@hohwille could this ticket close and move to https://github.com/devonfw/IDEasy/issues/989 ?

Paras14:
@hohwille Agreed, reworking this around the generic function syntax is the better, more future-proof approach. Most of the existing implementation can still be reused, so the work will not be lost.

@QuangAnhLe My suggestion would be to keep #2165 open for now and revisit it once the generic function syntax in #989 is implemented.

hohwille:
>My suggestion would be to keep #2165 open for now and revisit it once the generic function syntax in #989 is implemented.

I would implement both #2164 and #989 in this PR so merging it, will close both stories.
When you create the infrastructure to allow functions, you need to implement a function anyhow as otherwise it is impossible to test the feature. Then you can already implement `ask-variable` and `ask-secret` in the same way as you did here with the now obsolete syntax. If you want to keep the PR small, you can keep the persisting of user entered values out of this PR and implement it in a separate PR. If you do so, please create an issue so it will not be forgotten since technically it would be part of implementing #989.

hohwille:
@Paras14 What is the status of this PR? Isn't this PR already replaced by PR #2282 and can be closed?

hohwille:
@Paras14 What is the status of this PR? Isn't this PR already replaced by PR #2282 and can be closed?

PR #2198, opened 2026-07-21
#2168: repository remote feature

### LINKED ISSUE #2168: repository remote feature
### Feature idea

IDEasy aims to automate the setup of development projects as much as possible.
As an IDEasy user, I also want to automate the step to add an upstream remote to my repository after cloning.
The idea of this feature is to support a new property in repository properties files:
```
git_remote=upstream:https://github.com/devonfw/ide-settings.git
```
The generic syntax would be:
```
git_remote=«remote-name»:«remote-url»[,«remote-name»:«remote-url»]*
```
So in other words we have a comma separated list of key/value pairs for origins to add to the repository where each pair consists of the name of the remote to add and its URL. The assumption is that `«remote-name»` only consists of Latin letters (e.g. "a-z") so it will neither contain colon (`:`) nor comma (`,`). Further, I hope we can assume that a `«remote-url»` will never contain a comma.

### Additional context

Add the new config here:
https://github.com/devonfw/IDEasy/blob/08a8e5bd318aad734dac3ad752647336bf3362d1/cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryConfig.java#L25
and here:
https://github.com/devonfw/IDEasy/blob/08a8e5bd318aad734dac3ad752647336bf3362d1/cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java#L22

as `List<RepositoryRemote>` (new record type consisting of `String name` and `String url`).

Then implement the new feature after cloning and before linking here (using `git remote add` command, needs to be added to `GitContext`):
https://github.com/devonfw/IDEasy/blob/08a8e5bd318aad734dac3ad752647336bf3362d1/cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java#L122-L194

QuangAnhLe:
### Implemented changes:

* Create a new `RepositoryRemote `record ( similar to `RepositotyLink`)
* Add `git_remote` property parsing to `RepositoryProperties`
* Add `remotes` field to `RepositoryConfig`
* Add a addRemote method to ``G`itContext` interface and implementation
* Call the remote addition in `RepositoryCommandlet` after cone/pull and before linking
* Add tests for a new feature.
* Changelog entry for the epic : https://github.com/devonfw/IDEasy/issues/2168

---

### Testing instructions

1. mvn test -pl cli -Dtest="RepositoryPropertiesTest"

---

### Checklist for this PR

  with `internal`


QuangAnhLe:
@hohwille If the remotes are automatically created from the `git_remote `property during repository setup, the onboarding and setup instructions https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/setup.adoc should be updated accordingly, since the manual git remote add upstream step would no longer be required.

Paras14:
Hi @QuangAnhLe, thanks for your contribution, a few things that stand out:

1. git remote add is not idempotent. The loop you added sits after the createRepository check, so it also runs when the repository already exists, meaning every re-run of ide repository setup calls git remote add upstream... again and git fails with "remote already exists". The same happens within a single run when multiple workspaces are configured, since the additional workspaces are symlinks to the first repository. Please check the existing remotes first and only add the remote if it is missing (or update the URL if it differs).

2. Please rebase onto the current main. GitContextMock was reworked in #2093 (it now extends GitContextImpl), so your branch currently conflicts in GitContextMock.java. While you are at it, the new mock also lets you record the added remotes so the RepositoryCommandlet part can actually be tested. Right now only the parsing is covered.

3. The new git_remote property needs to be documented in documentation/repository.adoc, both in the table of variables and in the example .properties block above it.

Also please replace "TodoIssueID" in the PR description with the link to #2168 and adjust the PR title to the form #2168: «brief summary». And the CLA still needs to be signed before this can be merged.

cli/src/main/java/com/devonfw/tools/ide/git/GitContext.java
```diff
@@ -254,7 +254,15 @@ default void reset(Path repository, String branch) {
...
-   * Commits the staged changes in the given repository.
+   * Adds a new git remote to the given repository using {@code git remote add}.
+   *
+   * @param repository the {@link Path} to the git repository.
+   * @param name the name of the remote to add (e.g. "upstream").
+   * @param url the URL of the remote to add.
+   */
+  void addRemote(Path repository, String name, String url);
+   /**
+   *  Commits the staged changes in the given repository.
```
Paras14:
```suggestion
  void addRemote(Path repository, String name, String url);

  /**
   * Commits the staged changes in the given repository.
```

Paras14: There should be a blank line after the opening brace('{'). Also the spaces in the comments are a bit off.

Paras14:
Hi @QuangAnhLe, thanks for the quick turnaround, the fix looks good and GitContextImpl is the right place for it. One thing that stands out:

The get-url probe uses the three argument runGitCommand, which defaults to ProcessErrorHandling.LOG_WARNING. On a first setup the remote does not exist yet, so git exits with an error and we log a warning even though everything is fine. That means the noise is moved from the second run to the first one instead of being gone. Please pass ProcessErrorHandling.NONE for the probe using the four argument overload, a failed lookup there is expected and not a problem.

QuangAnhLe:
@Paras14 thanks for your review. I have fixed all the issues you pointed out in the ticket. Could you please do one fine review before we merge it into the main branch?

Paras14:
Hello @QuangAnhLe, Looks good to me.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -256,7 +256,7 @@ public String determineRemote(Path repository) {
     return DEFAULT_REMOTE;
   }
 
-  /**
+/**
```
hohwille, line 263: Aren't you using IntelliJ with Actions on Save? You broke indentation and using the right tools this cannot happen.

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
```diff
@@ -525,6 +525,32 @@ public void saveCurrentCommitId(Path repository, Path trackedCommitIdPath) {
...
+    }
+    if (getUrlResult.isSuccessful()) {
+      // Remote exists with a different URL — update it
+      LOG.debug("Remote '{}' already exists with a different URL, updating to {}", name, url);
+      ProcessResult setResult = runGitCommand(repository, ProcessMode.DEFAULT, "remote", "set-url", name, url);
+      if (!setResult.isSuccessful()) {
+        LOG.warn("Failed to update URL for remote '{}' to {}", name, repository);
+      }
+      return;
+    }
```
hohwille: Avoid pointless duplicated check for readability.
```suggestion
    if (getUrlResult.isSuccessful()) {
      if (getUrlResult.getOut().getFirst().trim().equals(url)) {
        LOG.debug("Remote '{}' already exists with the expected URL {}", name, url);
        return;
      }
      // Remote exists with a different URL — update it
      LOG.debug("Remote '{}' already exists with a different URL, updating to {}", name, url);
      ProcessResult setResult = runGitCommand(repository, ProcessMode.DEFAULT, "remote", "set-url", name, url);
      if (!setResult.isSuccessful()) {
        LOG.warn("Failed to update URL for remote '{}' to {}", name, repository);
      }
      return;
    }
```

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
```diff
@@ -525,6 +525,32 @@ public void saveCurrentCommitId(Path repository, Path trackedCommitIdPath) {
...
+    LOG.debug("Adding remote '{}' with url '{}' to {}", name, url, repository);
+    // Check if the remote already exists and get its current URL
+    ProcessResult getUrlResult = runGitCommand(repository, ProcessMode.DEFAULT_CAPTURE, ProcessErrorHandling.NONE, "remote", "get-url", name);
+    if (getUrlResult.isSuccessful() && getUrlResult.getOut().getFirst().trim().equals(url)) {
+      LOG.debug("Remote '{}' already exists with the expected URL {}", name, url);
+      return;
+    }
+    if (getUrlResult.isSuccessful()) {
+      // Remote exists with a different URL — update it
+      LOG.debug("Remote '{}' already exists with a different URL, updating to {}", name, url);
```
hohwille: Great that you log this. However, this is a more critical situation since the user may have done this on purpose and you are overriding.
* Use `warning` instead of `debug`
* Also include the old URL since this will be lost after the operation so it can at least be found in the log.
* If you want to solve this perfectly for the future, add an optional boolean parameter `failOnOverride` allowing to explicitly call this method with `true` as arg and then throw an exception in this situation preventing the override.

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
```diff
@@ -525,6 +525,32 @@ public void saveCurrentCommitId(Path repository, Path trackedCommitIdPath) {
     }
   }
 
+  @Override
+  public void addRemote(Path repository, String name, String url) {
+
+    LOG.debug("Adding remote '{}' with url '{}' to {}", name, url, repository);
+    // Check if the remote already exists and get its current URL
+    ProcessResult getUrlResult = runGitCommand(repository, ProcessMode.DEFAULT_CAPTURE, ProcessErrorHandling.NONE, "remote", "get-url", name);
+    if (getUrlResult.isSuccessful() && getUrlResult.getOut().getFirst().trim().equals(url)) {
```
hohwille: There is `runGitCommandAndGetSingleOutput` exactly for this purpose. Why not reusing that method instead of `runGitCommand`?

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
```diff
@@ -525,6 +525,32 @@ public void saveCurrentCommitId(Path repository, Path trackedCommitIdPath) {
...
+      if (!setResult.isSuccessful()) {
+        LOG.warn("Failed to update URL for remote '{}' to {}", name, repository);
+      }
+      return;
+    }
+    // Remote does not exist — add it
+    ProcessResult result = runGitCommand(repository, ProcessMode.DEFAULT, "remote", "add", name, url);
+    if (!result.isSuccessful()) {
+      LOG.warn("Failed to add remote '{}' to {}", name, repository);
+    }
```
hohwille, line 573: The method has no return type (`void`) and specifies that it adds/sets the git remote.
Here you silently return even in case of an error.
Shouldn't we throw an exception if the `add` or `set-url` failed?
I would say yes from the PoV of this generic method. However, the feature to add the remote using this method should not fail the entire process because of such error what would be achieved by a `step` in IDEasy.
This can be archived easily by providing `ProcessErrorHandling.THROW_ERR` as argument to `runGitCommand`. Maybe we even want an explicit method `runGitCommandOrFail` (or `runRequiredGitCommand`) for that.
BTW: Currently we use `ProcessErrorHandling.LOG_WARNING` so the error is already logged automatically as warning but maybe with your explicit log message you wanted to add a more explicit log message?

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
```diff
@@ -302,6 +305,36 @@ public List<RepositoryLink> getLinks() {
...
+      if (name.isBlank() || url.isBlank()) {
+        LOG.warn("Ignoring git_remote entry {} with empty name or url from {}", trimmed, PROPERTY_GIT_REMOTE);
+        continue;
+      }
+      if (!REMOTE_NAME_PATTERN.matcher(name).matches()) {
+        LOG.warn("Ignoring git_remote entry {} with invalid remote name \"{}\" — name must consist of latin letters only from {}",
+            trimmed, name, PROPERTY_GIT_REMOTE);
+        continue;
+      }
+      remoteList.add(new RepositoryRemote(name, url));
```
hohwille: This is all implemented very nice. Just as an additional improvement:
This method is parsing a CSV and iterates over each entry in the `for` loop.
Inside the loop, it parses each entry.
Clean code suggests to extract such additional concern to another private method with a self-explaining name to make the code even more readable and keeping methods shorter and smaller to understand the big picture.
So you could simply extract a private method to parse the `remoteItem` String as `RepositoryRemote` and returns `null` for an invalid entry. Then inside the loop you only need to call that method and if non-null is returned, add it to the `List`.

Also great that you followed immutable design producing a read only list of immutable records! 👍

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -270,6 +270,15 @@ public void addChanges(Path repository, GitCommit... commits) {
...
 
+  @Override
+  public void saveCurrentCommitId(Path repository, Path trackedCommitIdPath) {
+
+    String currentCommitId = determineCurrentCommitId(repository);
+    if (currentCommitId != null) {
+      this.context.getFileAccess().writeFileContent(currentCommitId, trackedCommitIdPath);
+    }
+  }
+
```
hohwille: I do not see any reason for overriding this method with a slightly duplicated implementation?
Can we remove this?

```suggestion
```

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
```diff
@@ -350,4 +359,13 @@ public void tag(Path repository, String tagName, String message) {
   public void push(Path repository, boolean followTags) {
 
   }
+
+  private final Map<Path, Map<String, String>> remoteUrls = new HashMap<>();
```
hohwille: Java files always follow a common structure. Read Oracle and Google Java coding style and follow it (as referenced from your coding conventions).
Never mix fields between methods, etc.

Also I would suggest a better design for this:
* Do we need a transient map at all? In all other places, we simply write such metadata to files in `.git` folder so why not just using a Java properties file with `name` as key and `url` as value. Then we can also later add a trivial `getRemote` feature if needed.
* Otherwise use a central `Map<Path, GitMockProject>` and add your `remote status` as one aspect of a `GitMockProject` that can later be extended. Instead of fiddling with `Map` in `Map` structures, you can then simply call a method like `myGitMockProject.addRemote(name, url)`.

However, I would strongly suggest the first approach since that follows our current design of `GitContextImpl` and keeps the state persistent. Assuming a JUnit failed and after it terminated you manually want to check things, then you can simply open the properties file in your text editor and see if expected remotes are there or something is missing or has wrong URL.

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java
```diff
@@ -186,6 +186,9 @@ private void doImportRepository(RepositoryConfig config) {
       }
       Path linkRepositoryPath = config.isVirtualSettingsRepository() ? firstRepository : repositoryPath;
       if (Files.exists(linkRepositoryPath)) {
+        for (RepositoryRemote remote : config.remotes()) {
+          this.context.getGitContext().addRemote(linkRepositoryPath, remote.name(), remote.url());
+        }
```
hohwille, line 192: Initially I expected this to happen after cloning.
However, thinking further during this review, I came to the point that maybe the settings get changed and a new remote is added or an existing is updated. Then it is great that you are not only doing this after cloning.
After understanding the flow again, I come to the point that this is the perfect sport. Nice work 👍 
I will however submit this review comment as positive feedback and immediately set it to resolved.

hohwille (approved):
@QuangAnhLe thanks for your PR. This is great work and functionally correct. 👍 
I added some code style and design suggestions for further improvements (perfection).
Please have a look and thanks in advance.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
QuangAnhLe: I am using VS Code with Format on Save and automatic code actions. I have now also enabled insertFinalNewline and trimTrailingWhitespace to avoid formatting-only diffs in future PRs.

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
QuangAnhLe: hi Jörg, thanks for your review. :). But i have 
```
  // Before (current):
  ProcessResult getUrlResult = runGitCommand(repository, ProcessMode.DEFAULT_CAPTURE, ProcessErrorHandling.NONE, "remote", "get-url", name);
  if (getUrlResult.isSuccessful() && getUrlResult.getOut().getFirst().trim().equals(url)) { ... }
  if (getUrlResult.isSuccessful()) {
      String oldUrl = getUrlResult.getOut().getFirst().trim();
      ...
  }

```
```
  // After (refactored):
  String existingUrl = runGitCommandAndGetSingleOutput("Failed to get remote URL", repository, "remote", "get-url", name);
  if (existingUrl != null && existingUrl.trim().equals(url)) { ... }
  if (existingUrl != null) { ... }
```

 But there's a problem. runGitCommandAndGetSingleOutput logs a warning when the command fails (line 486: LOG.warn(warningOnError) in GitCOntextImpl.java ). In this method, "remote doesn't exist
  → command fails" is a normal, expected path — not an error. We'd get a misleading warning every time a new remote is added for the first time.

  Verdict: Don't refactor. The current code is correct because:

  1. No noise on the normal path — when the remote doesn't exist, we silently proceed to add it.
  2. runGitCommandAndGetSingleOutput assumes failure is unusual — it warns on failure, which doesn't fit the dual-path logic here (exists vs. doesn't exist).
  3. The only cases where that helper fits cleanly are methods like determineCurrentBranch, retrieveGitUrl, determineCurrentCommitId — where the command should succeed
  and failure is genuinely unexpected.

  If you still want to reuse it, we 'd need to change runGitCommandAndGetSingleOutput to accept a flag like suppressWarningOnError, which adds complexity for no real gain over the current code. what do you think ?

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
QuangAnhLe: hi, The override can't be removed. It is not duplicated; it serves a different purpose for testing. If we remove the override, the parent's will call determineCurrentCommitId() → which in GitContextImpl runs real git commands. The mock overrides this to read from simulated .git/HEAD files instead. Without the override, tests would fail because there's no real git repo and  use Files.writeString() → writes directly to the real filesystem, bypassing the mock FileAccess that tests rely on to control where files are written. The override is intentional. It replaces real git commands and real filesystem access with mock equivalents. Removing it breaks the mock contract. Keep it !

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
hohwille: You can simply create a cleanup ticket for this and we can then merge as is and do the cleanup later.
I would simply suggest that the `runGitCommandAndGetSingleOutput` method could get an optional additional argument telling if we want to ignore warnings and/or non-successful execution and just silently return `null` then.
But why do we have `runGitCommandAndGetSingleOutput` if we cannot use if for such use-cases?
Design for reuse is always a good practice instead of duplicating logic.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
hohwille: But the bug is that the mock is not properly implementing `determineCurrentCommitId()`.
We want to test the logic of `saveCurrentCommitId` as its real implementation.
only methods that call `git` should be mocked.
This is a common principle of mocking and testing.

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
QuangAnhLe: I agree. The current implementation works, but reusing runGitCommandAndGetSingleOutput() would be preferable from a design perspective. We could create a cleanup ticket to extend the helper method with an option to suppress warnings for expected failures and then refactor the duplicated logic afterwards.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextMock.java
QuangAnhLe: You were right, it was a near-duplicate of GitContextImpl.saveCurrentCommitId (GitContextImpl.java:511)

cli/src/main/java/com/devonfw/tools/ide/git/GitContextImpl.java
hohwille: Can you create the cleanup ticket and link it here? I will then resolve this thread and we can merge.
Then we ensure nothing gets forgotten (years ago, I always wrote such feedback and agreed on separate cleanup but the developers forgot the cleanup and left the project, that is why I would like to see the ticket before resolving).
BTW: In this case I would personally think that within the time to create such ticket, you already have implemented the same inside this PR. 😄

QuangAnhLe: Makes sense. :)) I've already addressed the cleanup in this PR by extending `runGitCommandAndGetSingleOutput()` with an option to suppress warnings for expected failures. This way we can reuse the helper and avoid creating a separate cleanup ticket.

CHANGELOG.adoc
```diff
@@ -17,19 +36,25 @@ Release with new features and bugfixes:
...
 * https://github.com/devonfw/IDEasy/issues/1917[#1917]: Allow the creation of a desktop shortcut for the GUI
 * https://github.com/devonfw/IDEasy/issues/2134[#2134]: Add auto-completion for icd command
+* https://github.com/devonfw/IDEasy/issues/2218[#2218]: Add missing 3rd party libs to LICENSE
 * https://github.com/devonfw/IDEasy/issues/2192[#2192]: Added support for auto-completion synonyms
 * https://github.com/devonfw/IDEasy/issues/1953[#1953]: Implement base functionality of cleanup commandlet
 * https://github.com/devonfw/IDEAsy/issues/2181[#2181]: Merged the VsCode and VsCodium plugin folder and added excluded-editions as an option for plugins
 * https://github.com/devonfw/IDEasy/issues/2145[#2145]: Improve documentation on symlink regarding ide ln
 * https://github.com/devonfw/IDEasy/issues/2235[#2235]: Language flag in cli dosn't take effect in some terminals
-* https://github.com/devonfw/IDEasy/issues/2168[#2168]: Add support for git_remote property to configure additional remotes after cloning
+* https://github.com/devonfw/IDEasy/issues/2264[#2264]: ide «tool» does not set the environment variables of the other installed tools
```
hohwille, line 53: This seems to be an accidental change that should be reverted.

hohwille: I will simply revert this, resolve this thread and then finally merge this PR.

hohwille: Oh, sorry for confusion - my mistake. I was looking at a partial diff where you fixed the changelog.
Nothing wrong. Will resolve without any change.

QuangAnhLe:
Successfully retested PR #2198 with IDEasy <2026.08.002-08_27_12-SNAPSHOT> on Windows <Windows 11> using Git Bash.

Test configuration:

```
git_remote=upstream:https://***.***.***.com/***/ideasy-settings.git
```

PR #2221, opened 2026-07-28
#2126 fix only german is listed in language selection dropdown

### LINKED ISSUE #2126: only german is listed in language selection dropdown
### Actual behavior

I want to change the language of the GUI, which I can start with `ide gui`. But the dropdown menu doesn't appear on the left side.

### Reproduce

- Run `ide gui` within an ide project

### Expected behavior

A dropdown menu should appear on the left side to select the language of the interface.
Further expected behaivor are mentioned in #1979.

### IDEasy status

```bash
$ ide -p status
IDE_ROOT is set to ~/projects
Your version of IDEasy is 2026.07.001-07_07_06-SNAPSHOT.
You are online.
Your settings are up-to-date.
```

### Related/Dependent issues

https://github.com/devonfw/IDEasy/issues/1936, https://github.com/devonfw/IDEasy/pull/1979

### Comments/Hints

If I run `gui/src/main/java/com/devonfw/ide/gui/AppLauncher.java` via IntelliJ, the language dropdown menu appears.

Until #2036 is merged the GUI only opens, if stable version is installed (e.g. run `ide upgrade --mode=stable`).

oanding-blrng:
### Implemented changes:

* `ResourceBundle.getBundle(...)` with `Locale.ENGLISH` would return `messages_de.properties`, because no `messages_en.properties` could be found. Therefore check for `Locale.ENGLISH` was added to use `Locale.ROOT` instead, which would result in wanted `messages.properties`
* Extended `initialize()` method in `NlsService.java` to convert `de-DE` to `de`, because they are NOT the same. The current behaviour is the following:
  1. Get system locale with `Locale.getDefault()` and generalize it (`de-DE` to `de` and `en-US` to `en`, etc.)
  2. Check if system locale is contained in supported locales of GUI
  3. If this is the case, this locale will be used. Otherwise fallback to english
* To confirm new behaviour, the test method `testGetInstanceWithLocale(...)` was extended to cover other machine setups 

---

### Testing instructions

1. Run all tests of GUI module (`mvn test -pl gui`). The the tests mentioned in the issue should run successfully again
2. Start the GUI with a run configuration pointing to `AppLauncher.java` (if #2036 is merged, `./build-local-dev.sh` could also be used)
3. Depending on the locale settings of the machine, the language should either be the system setting or English for the start
4. German and English are both listed in the dropdown menu and selecting a different language should take effect in the GUI

---

### Checklist for this PR

- [X] When running `mvn clean test` locally all tests pass and build is successful
- [X] PR title is of the form `#«issue-id»: «brief summary»` (e.g. `#921: fixed setup.bat`). If no issue ID exists, title only.
- [X] PR top-level comment summarizes what has been done and contains link to addressed issue(s)
- [X] PR and issue(s) have suitable labels
- [X] Issue is set to `In Progress` and assigned to you *or* there is no issue (might happen for very small PRs)
- [X] You followed all [coding conventions](https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc)
- [X] You have added the issue implemented by your PR in [CHANGELOG.adoc](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc) unless issue is labeled
  with `internal`
- [X] You have formulated clear instructions on how to test your contribution under "Testing instructions"


MeShehi:
I have tested on my machine, which has an English Locale, and the software starts with the right language. In addition, changing the language does indeed affect the texts in the GUI. All tests are passing too.

gui/src/main/java/com/devonfw/ide/gui/nls/NlsService.java
```diff
@@ -68,20 +68,23 @@ public class NlsService {
...
     Locale localeToApply = explicitLocale;
     if (localeToApply == null) {
-      localeToApply = Locale.getDefault();
+      // Locale.getDefault() returns "de_DE" (for systems set to German) which isn't the same as "de"
+      localeToApply = Locale.of(Locale.getDefault().getLanguage());
+      if (!availableLocales.contains(localeToApply)) {
+        localeToApply = Locale.ENGLISH;
+      }
     }
     applyLocale(localeToApply, false);
```
hohwille, line 89: IMHO there is a small misunderstanding of the `Locale` concept.
Internationalisation (l18n) defines a inheritance tree of:
* root (the "empty" locale as `Locale.ROOT`)
* the language level (e.g. `de`, `en`, `es`, etc.)
* country specific aspects within a language (e.g. British English as `en_GB` uses different spelling than American English `en_US`, you could even differentiate aspects to German for Austria in `de_AT`).
* there is even another level below but lets stop at this point.

So we can have a resource bundle with `de` for German translation.
However, we could also have `en_US` for American English specialities.

You might think that this does not matter since IDEasy does not do country specific differentiations. However, JavaFx does a lot of things when the locale changes.
For that we cannot simply remove such aspects here since the `localeToApply` is actually what is applied to JavaFx.
Therefore line 84 will sooner or later cause undesired bugs.

Great that you traced down the reason why our language dropdown was having problems because the implemented in our team was not aware of the difference between `de` and `de_DE`.
However, only consider this for the dropdown but do not change the Locale that is applied.
Some users with LTR instead of RTL writing will not like if their GUI behaves broken...
Also bear in mind that with AI translation becomes cheap.
We might introduce `en_US` with many other languages one day via automatic translation.

gui/src/main/java/com/devonfw/ide/gui/nls/NlsService.java
```diff
@@ -240,7 +243,10 @@ private void loadBundle() {
   }
 
   private ResourceBundle loadBundle(Locale locale) {
-
+    // ResourceBundle.getBundle(...) with Locale.ENGLISH would return German properties on system locale = German
+    if (locale.equals(Locale.ENGLISH)) {
+      return ResourceBundle.getBundle(BUNDLE_NAME, Locale.ROOT);
+    }
     return ResourceBundle.getBundle(BUNDLE_NAME, locale);
   }
```
hohwille, line 251: Same kind of error here:
The resource bundle will do the proper inheritance automatically.
Our code shall not hard-code crazy assumptions about the current structure of our `*_de.properties` and `*.properties` vs. `*_en.properties`, etc.

hohwille (changes requested):
@oanding-blrng thanks for analysing the root of this problem. That is always the most important part of the fix. 👍 
I do not agree with the fix itself. Please have a look. If there are questions remaining, please do not hesitate to ask me...

gui/src/main/java/com/devonfw/ide/gui/nls/NlsService.java
oanding-blrng: Thank you for the explanation. After some testing (documented with a comment under the issue https://github.com/devonfw/IDEasy/issues/2126#issuecomment-5117420584) on a clean branch I came up with a different solution. We could just introduce a new `messages_en.properties` file. This would mean, that the ResourceBundle would find the proper bundle and because of how ResourceBundle is working, he would go up the chain (in this case) to the root properties, which is containing the English localization.

`messages_en.properties` would be completely empty with just a comment to make additions/edits to the English localization within the root properties file.

Therefore no assumptions had to made about the structure and we don't remove certain information about the locale the user is using.

Would this be a better solution @hohwille ?

oanding-blrng: I created a new pr #2234 with the new approach. This pr is ready for the team-review, but I wanted to wait for your opinion before kicking of the process of the team review. 

If the new approach is suitable, we can close this pr and work with the new one.

hohwille:
This PR is replaced by PR #2234 and I therefore close it now.

Issue #1297, opened 2025-05-12
Migration from OSSRH to new Maven Central service


hohwille:
> As you may have heard, OSSRH is reaching end of life on June 30, 2025. OSSRH users need to migrate their namespaces to the Central Portal as soon as possible. Instructions for self migration are located here: 
https://central.sonatype.org/faq/what-is-different-between-central-portal-and-legacy-ossrh/

So we have to react or otherwise we cannot deploy releases anymore.

We also need to make changes to our `pom.xml` to change to the new plugin after the namespace is migrated:
https://central.sonatype.org/publish/publish-portal-maven/

In this context we might even consider moving from `com.devonfw` to `io.github.xxx` where `xxx` would ideally be `IDEasy`. Unfortunately someone already took this "namespace" and is not using it.

hohwille:
see https://github.com/m-m-m/mmm/issues/2

hohwille:
As I can see someone already migrated `com.devonfw` to maven central portal or this happened automatically due to the overdue deadline and the shutdown of OSSRH legacy.
So we immediately need to update our POM as explained in https://github.com/m-m-m/mmm/wiki/maven-central-portal
Only then we can continue to deploy SNAPSHOTs and releases.

hohwille:
This isssue cannot be closed yet.
https://github.com/devonfw/IDEasy/blob/f845f17473210c827f2c597e11b56e72d1a75714/cli/src/main/java/com/devonfw/tools/ide/tool/repository/MavenRepository.java#L50
The URL has changed and now needs to be
```
https://central.sonatype.com/repository/maven-snapshots/
```

Currently the download of the latest SNAPSHOT versions (`ide upgrade --mode=snapshot`) is "broken" since it only finds releases on legacy OSSRH that is discontinued.

hohwille:
The migration is not complete and the release build failed:
https://github.com/devonfw/IDEasy/actions/runs/16492700586/job/46631541944

```
[INFO] Waiting until Deployment 31ed3f09-9b68-47f2-a79a-4c3c702053f7 is validated
Error:  
Deployment 31ed3f09-9b68-47f2-a79a-4c3c702053f7 failed
pkg:maven/com.devonfw.tools.IDEasy/ide-doc@2025.07.001:
 - Project URL is not defined
 - SCM URL is not defined
 - Developers information is missing
```

I am quite surprised by this error. Surely we have these properties defined in our POMs.

hohwille:
[...]

PR #2223, opened 2026-07-28
#2224: publish the Homebrew formula as part of the release

### LINKED ISSUE #2224: Brew upgrade ideasy not working
### Actual behavior

When running `brew update` followed by `brew upgrade ideasy`, this is the observed behaviour:
```
lukasfaber@MacBook-Air-von-Lukas ~ % brew update
==> Updating Homebrew...
Already up-to-date.
lukasfaber@MacBook-Air-von-Lukas ~ % brew upgrade ideasy
Warning: devonfw/ideasy/ideasy 2026.04.002 already installed
Warning: devonfw/ideasy/ideasy 2026.04.002 already installed
lukasfaber@MacBook-Air-von-Lukas ~ % 
```

### Reproduce

1. Run `brew update`
2. Run `brew upgrade ideasy`

### Expected behavior

As a user, I want to be able to update ideasy to the latest version using `brew update` followed by `brew upgrade ideasy`.

### IDEasy status

```bash
N/A
```

### Related/Dependent issues

_No response_

### Comments/Hints

I just realized, that the automation for updating the brew tap is running using pull requests in this Repo, and it seems like for each version, this has to be manually merged. This is probably why the update only installs an outdated version.

Also, I noticed some irregularities with the update automation.

maybeec:
### This PR fixes #2224

Homebrew users are stuck on 2026.04.002: `brew upgrade ideasy` reports "already installed" because
the tap has not been updated since May. The tap's own polling automation opened five pull requests
that all carry empty `sha256` fields and that nobody merges. That is fixed on the tap side in
devonfw/homebrew-IDEasy#8; this PR makes the release actually publish the formula.

**Interpretation this PR is based on:** Homebrew is a *primary* distribution channel for IDEasy. A
decoupled post-processing automation running after the release can fail unnoticed — as it did for
three months. So publishing the formula becomes a synchronous part of this workflow: if it cannot be
published, the release run is red.

### Implemented changes:

* **Resolve the release version exactly once.** `.mvn/maven.config` was parsed independently in
  three places, and it is *rewritten twice during the run* (to the release version, then to the next
  snapshot). The same number was called `next_version` in `build-natives` and `current_version` in
  the release step, and anything reading the file afterwards resolves a different version depending
  on when it runs. A new `resolve-version` job publishes `release_version` and `snapshot_version` as
  job outputs, and every downstream job consumes them.
* **Wait until the deployment is published, not merely validated** *(precondition for everything
  below — see the note at the end)*. `central-publishing-maven-plugin` defaults to
  `waitUntil=validated`, so `mvn deploy` returns as soon as Central has validated the bundle. In the
  2026.07.002 run the GitHub release — whose notes link directly to `repo1.maven.org` — was created
  **seven seconds** after validation, well before those URLs resolved. The `deploy` profile now sets
  `autoPublish` + `waitUntil=published` with `waitMaxTime=3600`.
* **Drop the dead OSSRH configuration.** The `deploy` profile configured
  `nexus-staging-maven-plugin` against `s01.oss.sonatype.org`, which was retired on 2025-06-30. The
  release log confirms it is vestigial: the actual publishing is done by `central-publishing:0.8.0`
  inherited from `com.devonfw:maven-parent`. It is replaced by an explicit `central-publishing`
  configuration, keeping `publishingServerId=repository` so the existing
  `SONATYPE_USERNAME`/`SONATYPE_PASSWORD` credentials from `.mvn/settings.xml` continue to apply.
* **Publish the formula from the release.** Three new jobs after `release`:
  * `await-central` — waits until all four archives are served by `repo1` and verifies the
    checksums match the artifacts that were just deployed.
  * `verify-homebrew` — renders the formula and runs `brew audit --strict --online`,
    `brew install` and `brew test` on `macos-latest` and `macos-15-intel`.
  * `publish-homebrew` — pushes the formula to the tap, only after verification passed, so the tap
    is never left broken.

  Checksums are computed from the archives in `cli/target/` that were just deployed — the exact bytes
  Central will serve — so nothing is downloaded to build the formula.
* **Add the missing `linux-arm64` download** to the release notes. The assembly
  (`cli/src/main/assembly/release-linux-arm64.xml`) has been built and deployed for a while but was
  listed nowhere.

The tap is checked out with `ACTION_PUSH_TOKEN`, the same secret already used to push to
`devonfw/ide-urls` in `update-urls.yml` and `update-cve.yml`. If its scope does not cover
`devonfw/homebrew-IDEasy`, this is a one-line change to a different secret.

---

### Testing instructions

The release workflow only runs on `workflow_dispatch`, so the end-to-end path is first exercised by
the next release. What can be checked before that:

1. Review the job graph: `resolve-version` → `build-natives` → `release` → `await-central` →
   `verify-homebrew` → `publish-homebrew`. No job re-derives the version.
2. `mvn -N -Pdeploy validate` passes with the new publishing plugin configuration.
3. In devonfw/homebrew-IDEasy#8, `./wait-for-central.sh 2026.07.002` verifies against the real
   2026.07.002 artifacts, and `render-formula.sh` rejects the empty-version and empty-checksum
   inputs that produced the stale bot PRs.
4. At the next release, `publish-homebrew` must push the new version to the tap, and
   `brew update && brew upgrade ideasy` must upgrade.

---

### Checklist for this PR

- [x] When running `mvn clean test` locally all tests pass and build is successful *(no production code touched; `mvn -N -Pdeploy validate` verified the POM change)*
- [x] PR title is of the form `#«issue-id»: «brief summary»` *(cross-repository form, the issue lives in `devonfw/homebrew-IDEasy`)*
- [x] You have added the issue implemented by your PR in [CHANGELOG.adoc](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc)

Not applicable — no commandlet is added.

---

### Precondition to be aware of when merging

The `waitUntil=published` change is **required** for the Homebrew jobs (the formula must not be
published before the URLs it references resolve), but it also changes release timing for everyone:
`mvn deploy` now blocks until Maven Central reports the deployment as published instead of returning
at validation. Publication typically takes on the order of 20 minutes, bounded here by
`waitMaxTime=3600`; if Central does not publish within that window the release fails loudly instead
of producing release notes pointing at URLs that do not resolve yet.

**Merge order:** devonfw/homebrew-IDEasy#8 must be merged first — this workflow checks that
repository out and runs `render-formula.sh` and `wait-for-central.sh` from it.


maybeec:
### Note for the reviewer: what has and has not been executed

`verify-homebrew` runs `brew audit --strict --online` plus a real `brew install` and `brew test` — a
stricter gate than the tap's own CI, which only checks tap syntax. **It has never executed**, because
these jobs only run during a release. The same is true for `await-central` and `publish-homebrew`.

What *has* been verified:

* devonfw/homebrew-IDEasy#8 is green on `ubuntu-latest`, `macos-latest` and `macos-15-intel`,
  including the check that the formula is in sync with its template.
* `wait-for-central.sh` was run against the real 2026.07.002 artifacts. The three checksums it
  computed match exactly what the old (broken) auto-update workflow computed on 2026-07-26, and it
  produced the linux-arm64 checksum that has never been part of the formula.
* `render-formula.sh` rejects every failure mode behind the stale bot PRs: empty checksum, empty
  version, wrong argument count, uppercase hex.
* `mvn -N -Pdeploy validate` passes with the `central-publishing` configuration.

If you want the release-side path exercised before the next release, the **Publish formula**
workflow on the tap (`workflow_dispatch`) runs the same render → verify → publish sequence, sourcing
the checksums from Maven Central instead of from freshly built artifacts. Note that it pushes to the
tap's `main`, so it is only worth running once devonfw/homebrew-IDEasy#8 is merged.

### One deviation worth knowing about

The formula was intended to reference `repo1.maven.org` directly. Homebrew does not allow it:
`FormulaAudit/Urls` requires the `https://search.maven.org/remotecontent?filepath=...` form, and
neither override route is available — a tap level `.rubocop.yml` is not picked up, and inline
`# rubocop:disable` directives are rejected by `Style/DisableCopsWithinSourceCodeDirective`. The
required form was verified to return `301` to `repo1.maven.org` and to serve byte-identical content.
`wait-for-central.sh` therefore still checks availability against canonical `repo1.maven.org`, while
the formula uses the redirecting form that `brew audit` demands.

.github/workflows/release.yml
```diff
@@ -118,50 +148,147 @@ jobs:
...
+            echo "${platform}: ${sha}"
+            echo "sha_${platform//-/_}=${sha}" >> "$GITHUB_OUTPUT"
+          done
       - name: set next version & create gitHub release
         run: |
-          maven_config="$(cat .mvn/maven.config)"
-          current_version="${maven_config/#*-Drevision=}"
-          current_version="${current_version/ */}"
-          current_segment="${current_version/*./}"
+          release_version="${RELEASE_VERSION}"
```
hohwille: kind of pointless variable. Why not using `RELEASE_VERSION` directly instead of shadowing it again in lower-case?

.github/workflows/release.yml
```diff
@@ -118,50 +148,147 @@ jobs:
...
+        run: |
+          if git diff --quiet -- Formula/ideasy.rb; then
+            echo "Formula already up to date for ${RELEASE_VERSION}, nothing to publish."
+            exit 0
+          fi
+          git config user.email ${{ secrets.BUILD_USER_EMAIL }}
+          git config user.name ${{ secrets.BUILD_USER }}
+          git add Formula/ideasy.rb
+          git commit -m "ideasy ${RELEASE_VERSION}"
+          git push
```
hohwille: looks like a complete copy&past duplication of https://github.com/devonfw/homebrew-IDEasy/pull/8 to me.
Can't we just invoke the workflow from `homebrew-IDEasy` avoiding all the duplication?
Who will ensure that if this needs to be changed it will be updated consistently in both repos?

CHANGELOG.adoc
```diff
@@ -8,6 +8,7 @@ Release with new features and bugfixes:
 
 * https://github.com/devonfw/IDEasy/issues/2176[#2176]: Support 7z archive extraction
 * https://github.com/devonfw/IDEasy/issues/2100[#2100]: Fix Python not available for Mac x64
+* https://github.com/devonfw/homebrew-IDEasy/issues/4[homebrew-IDEasy#4]: Publish the Homebrew formula as part of the release so `brew upgrade ideasy` gets new versions
```
hohwille: Actually we should create issues in IDEasy. I will disabled issues in `homebrew-IDEasy`.
The same also never worked out for `ide-settings` and `ide-urls`.
Also for external users it is the only way if all IDEasy problems can be reported in one place.
New issue created:
```suggestion
* https://github.com/devonfw/IDEasy/issues/2224[#2224]: Brew upgrade ideasy not working
```
BTW: Comment should be the title of the issue (not PR) for end-users that may even have reported or watched the issue to recognise.

pom.xml
```diff
@@ -296,14 +296,20 @@
       <build>
         <plugins>
           <plugin>
-            <groupId>org.sonatype.plugins</groupId>
-            <artifactId>nexus-staging-maven-plugin</artifactId>
-            <version>1.7.0</version>
+            <groupId>org.sonatype.central</groupId>
+            <artifactId>central-publishing-maven-plugin</artifactId>
```
hohwille, line 308: I thought with #1297 we have done all this.
I am quite surprised we were still using the old nexus-staging with s01.
Doesn't this mean that I also have to do further technical preparation to make this work since we are actually migrating finally from OSSRH with this change?

hohwille (approved):
@maybeec thanks for your PR. Great stuff. I really appreciate this automation since I am too overloaded to keep track of manual processes for such extra details. 👍 

As you already outlined we have a dependency to the homebrew PR that needs to be merged first. I also left some review comments to clarify before merge.

.github/workflows/release.yml
maybeec: Agreed, that was pointless. The only reason was to keep the diff close to the original bash, which is
not a good reason. The step now uses `RELEASE_VERSION` directly and the local variable is gone.

Fixed in 10f668c8.

.github/workflows/release.yml
maybeec: You are right, and it is fixed: the three jobs are gone, replaced by a single call to a reusable
workflow in the tap.

```yaml
  publish-homebrew:
    needs: [ resolve-version, release ]
    uses: devonfw/homebrew-IDEasy/.github/workflows/publish-formula.yml@main
    with:
      version: ${{ needs.resolve-version.outputs.release_version }}
      sha_mac_arm64: ${{ needs.release.outputs.sha_mac_arm64 }}
      ...
    secrets: inherit
```

**Why not simply invoking the workflow** — that was the reason for the duplication, and it is worth
recording. Dispatching the tap workflow (`gh workflow run` / `workflow_dispatch`) returns immediately
with no status. This job would go green whatever happens over there, and an IDEasy release would
succeed with a stale Homebrew channel that nobody notices — exactly the failure mode this PR exists
to remove. Recovering the signal would mean hand-rolling a polling loop.

`uses:` does not have that problem: the called workflow's jobs run as part of this run, and **a
failure there fails this job and with it the release**. It cannot even be silenced by accident, since
`continue-on-error` is not supported on a job that calls a reusable workflow. So we get the
deduplication you are asking for *and* the loud failure. Both repositories are public and in the same
org, so the cross-repository call and `secrets: inherit` work.

To answer "who keeps them consistent": nobody has to any more. The sequence now exists once, in
devonfw/homebrew-IDEasy@c47074a, next to the formula and the scripts it renders. The release owns
*when* to publish, the tap owns *how*. Worth noting the logic was never duplicated — `render-formula.sh`
and `wait-for-central.sh` only ever existed in the tap; what was duplicated was ~70 lines of YAML
wiring around them.

The manual `workflow_dispatch` entry point stays, on the same workflow, as the recovery path for
catching up or re-publishing. Without checksums passed in it takes them from Maven Central; with them
it verifies Central serves exactly the released bytes.

Two consequences to be aware of:

* This makes the merge order **hard**, not just advisory: `@main` cannot resolve until
  devonfw/homebrew-IDEasy#8 is merged.
* `@main` pins the release to the tap's moving branch, so a bad commit on the tap could break a
  release. Pinning to a tag or commit SHA would be safer at the cost of bumping it on every change.
  Happy to switch if you prefer that trade-off.

pom.xml
maybeec: Good news: **#1297 did complete the migration, and no further preparation is needed.** What was left
behind here was dead config, not a live OSSRH dependency — so I deleted it rather than carrying it
forward.

Evidence from the 2026.07.002 release log:

```
[INFO]   ... total of 6 executions of maven-deploy-plugin replaced with nexus-staging-maven-plugin
[INFO] --- central-publishing:0.8.0:publish (injected-central-publishing) @ ide-cli ---
[INFO] Using Central baseUrl: https://central.sonatype.com
[INFO] Uploaded bundle successfully, deploymentId: e93b2aba-...  Deployment will publish automatically
```

`s01.oss.sonatype.org` appears nowhere in that log. The publishing was already being done by
`central-publishing-maven-plugin`, inherited from the `deploy` profile of `com.devonfw:maven-parent`
(`extensions=true`, `autoPublish=true`). The `nexus-staging` block in our own POM rebound the deploy
goal — hence the first log line — but the injected central-publishing extension is what actually
uploaded. So your surprise is justified: it looked alive, it was not.

Since it was dead, the profile is now reduced to the one thing this PR genuinely adds:

```xml
<plugin>
  <groupId>org.sonatype.central</groupId>
  <artifactId>central-publishing-maven-plugin</artifactId>
  <configuration>
    <waitUntil>published</waitUntil>
    <waitMaxTime>3600</waitMaxTime>
  </configuration>
</plugin>
```

`mvn -N help:effective-pom -Pdeploy` confirms the rest is inherited and merged correctly:

```xml
<execution>
  <id>injected-central-publishing</id>
  <phase>deploy</phase>
  <goals><goal>publish</goal></goals>
  <configuration>
    <waitUntil>published</waitUntil>
    <waitMaxTime>3600</waitMaxTime>
    <publishingServerId>repository</publishingServerId>
    <autoPublish>true</autoPublish>
  </configuration>
</execution>
```

`extensions=true` and version `0.8.0` come from the parent as well, and `nexus-staging` and
`s01.oss.sonatype.org` are both absent from the effective POM.

No new secrets or namespace steps: `publishingServerId=repository` still maps to the `repository`
server in `.mvn/settings.xml`, i.e. `SONATYPE_USERNAME`/`SONATYPE_PASSWORD`, which must already be
Central Portal credentials since publishing works today. The only behavioural change is that
`mvn deploy` now blocks until Central reports the deployment as *published* instead of returning at
*validated*.

PR #2227, opened 2026-07-28
#2226: show IDEasy project, workspace and repository in the git-bash window title

### LINKED ISSUE #2226: Show IDEasy project, workspace and repository in the Git Bash / Windows Terminal window title
## Expected behavior

When several Git Bash tabs/windows are open in Windows Terminal, each one should be identifiable at a glance (taskbar, Alt+Tab, tab strip) by the IDEasy project and workspace it is initialized for, e.g. `ideasy>devonfw>main`.

## Actual behavior

The title is rewritten on **every prompt** to `MINGW64:<full-path>` and can not be changed permanently — setting it manually (`printf '\033]0;...'`) works only until the next command finishes.

Cause: Git for Windows bakes an OSC escape sequence into `PS1` in `/etc/profile.d/git-prompt.sh`:

```sh
PS1='\[\033]0;$TITLEPREFIX:$PWD\007\]' # set window title
```

`$TITLEPREFIX` defaults to `$MSYSTEM` (`MINGW64`) and `$PWD` is re-expanded each time the prompt is drawn, so every prompt resets the title. Windows Terminal derives the window title from the active tab title, which follows exactly this escape sequence.

Setting only `TITLEPREFIX` is not sufficient — it replaces the `MINGW64` part but `:$PWD` remains.

## Proposal

Extend `cli/src/main/package/functions` to remove the title segment from `PS1` and emit an IDEasy-aware title from `PROMPT_COMMAND`.

### Title format

The title degrades gracefully depending on how much context is available:

| Situation | `IDE_HOME` | `$PWD` | Title |
| --- | --- | --- | --- |
| Outside any IDEasy project | unset | `/c/Users/me` | `ideasy` |
| Inside a project, outside a workspace | `/d/projects/devonfw` | `/d/projects/devonfw` | `ideasy>devonfw` |
| Inside a project, outside a workspace | `/d/projects/devonfw` | `/d/projects/devonfw/conf` | `ideasy>devonfw` |
| Inside a workspace, outside a repository | `/d/projects/devonfw` | `/d/projects/devonfw/workspaces/main` | `ideasy>devonfw>main` |
| Inside a repository | `/d/projects/devonfw` | `/d/projects/devonfw/workspaces/main/IDEasy/cli` | `ideasy>devonfw>main>IDEasy` |
| Inside another workspace | `/d/projects/devonfw` | `/d/projects/devonfw/workspaces/feature-x/foo` | `ideasy>devonfw>feature-x>foo` |

The repository is the counterpart of `icd -r` and is the first folder below the workspace. Consistent with the implicit `-r` resolution, hidden folders (e.g. `.vscode`) are not considered repositories, and for a nested repository such as `backend/core` only the top-level folder is shown to keep the title short.

Note that the `WORKSPACE` variable can **not** be used to decide whether we are inside a workspace: `ideasy env --bash` always emits `WORKSPACE="main"` as a default, even when the current directory is completely outside `IDE_ROOT`. The workspace and repository segments therefore have to be derived from `$PWD` relative to `${IDE_HOME}/workspaces/`, which has the added benefit that the title also follows a plain `cd`, not only `ide`/`icd`.

### OS / shell gating

This must be a no-op on macOS and Linux, and on Cygwin:

* The `PS1` title sequence described above is specific to Git for Windows / MSYS2. Rewriting `PS1` on other platforms would interfere with the user's own prompt configuration (oh-my-zsh, starship, distribution defaults, ...) for no benefit.
* `functions` is also sourced from `zsh` (see the `compinit`/`bashcompinit` block at the end of the file), where `PROMPT_COMMAND` does not exist — zsh uses `precmd` instead.

So the whole block should be guarded by the already existing helper `_ide_is_msys()` (`cli/src/main/package/functions:16`, matches `MINGW*`/`MSYS*` from `uname -s`) **plus** a `[ -n "${BASH_VERSION}" ]` check. `_ide_is_cygwin()` stays excluded, consistent with `documentation/cygwin.adoc`.

### Sketch

```sh
_ide_title() {
  local _t="ideasy" _rel _ws
  if [ -n "${IDE_HOME}" ]; then
    _t="${_t}>${IDE_HOME##*/}"
    case "${PWD}" in
      "${IDE_HOME}/workspaces/"*)
        _rel="${PWD#"${IDE_HOME}/workspaces/"}"
        _ws="${_rel%%/*}"
        [ -n "${_ws}" ] && _t="${_t}>${_ws}"
        ;;
    esac
  fi
  printf '\033]0;%s\007' "${_t}"
}

if _ide_is_msys && [ -n "${BASH_VERSION}" ]; then
  case "${PS1}" in
    '\[\033]0;'*'\007\]'*) PS1=${PS1#*'\007\]'} ;;   # drop Git Bash's title sequence
  esac
  case "${PROMPT_COMMAND}" in
    *_ide_title*) ;;                                  # idempotent when re-sourced
    *) PROMPT_COMMAND="_ide_title${PROMPT_COMMAND:+; ${PROMPT_COMMAND}}" ;;
  esac
fi
```

Implementation details to take care of:

* The `PS1` strip is guarded by a `case` pattern so it silently does nothing if a user brings their own prompt (e.g. via `~/.config/git/git-prompt.sh`, the override hook honored by `git-prompt.sh`) or if Git for Windows changes the sequence. The colored prompt itself stays untouched.
* Both the `PS1` strip and the `PROMPT_COMMAND` hook must be idempotent — `functions` is re-sourced whenever a user runs `source ~/.bashrc`.
* Git Bash ships bash 5.3, where `PROMPT_COMMAND` may be an **array**. The string concatenation above would clobber an array-valued `PROMPT_COMMAND`, so the array case needs to be handled.

## Alternatives considered

* `TITLEPREFIX` — insufficient, `:$PWD` is still appended (see above).
* Windows Terminal `"suppressApplicationTitle": true` + a fixed `tabTitle` in the git-bash profile (which IDEasy already patches in `IdeasyCommandlet.configureWindowsTerminalGitBash()`) — this pins *every* git-bash tab to the same static name and therefore defeats the purpose of distinguishing projects and workspaces.

## Tests

`cli/src/test/functions-test.sh` already sources `cli/src/main/package/functions` in a sandboxed subshell with stubbed `ideasy`/`uname`/`cygpath` and runs in CI (`.github/workflows/build-pr.yml`). The title computation can be covered there in the same style:

* the five title variants from the table above,
* no-op on non-MSYS `uname -s` (`Darwin`, `Linux`, `CYGWIN_NT-10.0`): `PS1` unchanged and `PROMPT_COMMAND` unset,
* re-sourcing `functions` twice does not strip `PS1` twice and does not add the hook twice.

## Documentation

`documentation/advanced-tooling-windows.adoc` (section `=== Tabs for Shells`) already describes the Windows Terminal integration and is the natural place to mention the title behavior.

maybeec:
### This PR fixes #2226

Git Bash bakes a window title escape sequence into `PS1` (see `/etc/profile.d/git-prompt.sh`), so the title is rewritten to `MINGW64:«current-directory»` on **every** prompt and cannot be set permanently. All git-bash tabs and windows therefore look alike in the taskbar and in `Alt`+`Tab`.

This PR removes that sequence and writes an IDEasy specific title instead, so you can see at a glance which project and workspace a shell is initialized for.

### Implemented changes:

* Added `_ide_title` to `cli/src/main/package/functions`, which writes the terminal title and degrades gracefully depending on the available context:

  | Situation | Title |
  | --- | --- |
  | Inside a repository | `ideasy>devonfw>main>IDEasy` |
  | Inside a workspace, outside any repository | `ideasy>devonfw>main` |
  | Inside a project, outside any workspace | `ideasy>devonfw` |
  | Outside any IDEasy project | `ideasy` |

* The repository is the counterpart of `icd -r`, i.e. the first folder below the workspace. Consistent with the implicit `-r` resolution, hidden folders (e.g. `.vscode`) are not treated as repositories, and for a nested repository such as `backend/core` only the top-level folder is shown to keep the title short.
* Workspace and repository are derived from `PWD` and **not** from the `WORKSPACE` variable, because `ideasy env --bash` always emits `WORKSPACE="main"` as a default - even when the current directory is completely outside `IDE_ROOT`. As a side effect the title also follows a plain `cd`, not only `ide`/`icd`.
* Extracted `_ide_path_segment` as the single place that determines a path segment below a base directory. It is used for the workspace and the repository of the title as well as for the project inference in `_icd_project_dir`, which previously duplicated this logic. Its result is passed via a variable rather than echoed, so that no subshell is forked - `_ide_title` runs on every prompt and process creation is expensive on Windows.
* The git-bash title sequence is removed from `PS1` and `_ide_title` is hooked into `PROMPT_COMMAND`. Both steps are idempotent, so sourcing the functions file again (e.g. via `source ~/.bashrc`) does not duplicate the hook or strip `PS1` twice.
* The whole block is guarded by the existing `_ide_is_msys` helper plus a `BASH_VERSION` check, so **macOS, Linux and Cygwin are not touched at all** - their prompt is configured by the user or the distribution. The `BASH_VERSION` check is needed because the functions file is also sourced from zsh, which has no `PROMPT_COMMAND` (it uses `precmd`).
* An existing `PROMPT_COMMAND` is preserved, including the array form supported since bash 5.1. The IDEasy hook is prepended so a user's own hook runs afterwards and can still override the title.
* The `PS1` rewrite only triggers when the title sequence is at the very beginning of the prompt, so a custom prompt (e.g. via the `~/.config/git/git-prompt.sh` hook honored by git-bash) is left untouched.
* Added 18 tests to `cli/src/test/functions-test.sh` and documented the behavior in `documentation/advanced-tooling-windows.adoc`.

---

### Testing instructions

Automated:

1. Run `bash cli/src/test/functions-test.sh` - this is the same entry point CI uses (`.github/workflows/build-pr.yml`). All tests must pass.
   Note that the zsh cases are skipped when zsh is not on the `PATH` (e.g. on Windows). They were verified on `ubuntu:24.04` with zsh 5.9, where the suite reports **43 of 43 passed**, including `zsh -n` parsing of the functions file.
2. Run `bash cli/src/test/icd-completion-test.sh` - `_icd_project_dir` was refactored, so the icd completion tests must still pass.

Manual (Windows, git-bash):

1. Open a git-bash terminal in an IDEasy project and run `source ~/.bashrc`.
2. `icd -r` into a repository - the window/tab title shows `ideasy>«your-project»>«workspace»>«repository»`.
3. `cd ..` up into the workspace - the title shows `ideasy>«your-project»>«workspace»`.
4. `cd` into a hidden folder of the workspace (e.g. `.vscode`) - the title still shows `ideasy>«your-project»>«workspace»`.
5. `icd` into the project root (outside `workspaces`) - the title shows `ideasy>«your-project»`.
6. `cd ~` and run `ide` - the title shows `ideasy`.
7. Verify the visible prompt itself is unchanged: colors, `$MSYSTEM`, the working directory and the git branch from `__git_ps1` are all still there.
8. Verify `icd -p«Tab»`, `icd -w«Tab»` and `icd -r«Tab»` still complete as before.

Manual (macOS/Linux, regression):

1. Open a terminal in an IDEasy project and run `source ~/.bashrc` (or `~/.zshrc`).
2. Verify your prompt and window title behave exactly as before - IDEasy must not change either.

---

### Checklist for this PR

  with `internal`


MeShehi:
The code correctly implements the requested changes from the linked issue. I tested it on both Windows and Linux (Bash), and everything works as expected. I'd be comfortable moving this to Reviews once the merge conflicts have been resolved.

PR #2228, opened 2026-07-29
#2042: quiet modus supression for question

### LINKED ISSUE #2042: -q parameter supresses ask for CVE choices
### Actual behavior

When a commandlet that triggers cveCheck (e.g. ide create) is invoked with the global -q / --quiet flag, and cveCheck determines that no safer alternative version is available for a tool that has open CVEs, the install hangs indefinitely on invisible input.

The CVE warning text is printed (at WARN level), e.g.:

Found 2 CVE(s) for version 3.14.6 of tool python:
CVE-2026-6019 with severity 6.1 ...
CVE-2026-7210 with severity 7.5 ...

…but the follow-up prompt Which version do you want to install? (raised by ToolCommandlet#cveCheck via IdeContext#question) is never displayed. The cursor sits waiting for input the user has no idea is being requested.

The interactive read from stdin still happens — so the install isn't broken, just silenced. Typing 1 (or yes) and Enter unblocks it, confirming a prompt was waiting all along. The prompt is logged at INFO level, which -q filters out; the surrounding warnings survive only because they sit at WARN.

### Reproduce

1. Create or use a settings repository that pins a tool to a version with open CVEs and no installable safer build (at the time of writing: python 3.14.6).
2. Run:
ide -q create my-project <settings-url>
3. Observe: the yellow CVE warnings print, then the terminal hangs on a blinking cursor with no further output. No question is visible.
4. For comparison, drop -q:
ide create my-project <settings-url>
4. The same flow now prints Which version do you want to install? with a numbered choice list, and the install proceeds normally.

### Expected behavior

-q / --quiet should suppress only purely informational log output. Interactive prompts that block on user input must always be visible, regardless of log-level filtering — otherwise the user sits in front of an apparently-frozen install with no way to know what is being asked.

Two acceptable fixes (either alone resolves the bug):

1. Move the cveCheck choice prompt to at least WARN level. The CVE warning text that immediately precedes it is already at WARN; the question that follows should not be a level lower.
2. Make IdeContext#question(...) (and any other synchronous, blocking prompt) unconditionally visible — INFO suppression should never silence a prompt the runtime is about to block on.

### IDEasy status

```bash
Your version of IDEasy is 2026.05.001.
Your are using the latest version of IDEasy and no update is available.
You are online.
```

### Related/Dependent issues

_No response_

### Comments/Hints

Discovered while building a wrapper installer that passes -q specifically to suppress an unrelated, misleading SETTINGS_URL warning that IDEasy emits before processing an explicit settings-repo argument. Dropping -q is the only workaround, which forces the wrapper to expose the misleading warning to its end users.

Severity is low (single-keystroke workaround once you know what's happening), but real-world UX impact is high: from the user's perspective the install is broken, and they file a support ticket rather than blindly pressing Enter into a black hole.

Relevant code path (as of main):
- cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java → cveCheck(...) → IdeContext#question("Which version do you want to install?", choices)

### LINKED ISSUE #2292: Questions appear twice
### Actual behavior

During my retests, I discovered that now our questions always appear twice:
First on interaction log level and then again on warning log level.
Example:
```
$ ide create test
Your version of IDEasy is currently 2026.08.001-08_07_04-SNAPSHOT
However, this is too old as your project requires at latest version 2027.01.001
Please run the following command to update to the latest version of IDEasy and fix the problem:
ide upgrade
Creating new IDEasy project in D:\projects\test
Directory D:\projects\test already exists. Do you want to continue?
Directory D:\projects\test already exists. Do you want to continue?
Option 1: yes
Option 2: no
```

This also happens with `ide cleanup` so this is not a bug in an individual commandlet but a generic bug in IdeContext and/or our logging.

### Reproduce

1. `mkdir -p $IDE_ROOT/test`
2. `ide create test -`

### Expected behavior

A question should only appear once on interaction log-level.

### IDEasy status

```bash
IDE_ROOT is set to D:\projects
IDE_HOME is set to D:\projects\IDEasy
Your version of IDEasy is 2026.08.001-08_07_04-SNAPSHOT.
You are online.
```

### Related/Dependent issues

_No response_

### Comments/Hints

I just checked PR #2228 that came to my mind as a potential cause of this problem but could not see what could be wrong there.
Needs further investigation.

marceltchanga9:
### Implemented changes:

Changed the log level in quiet modus

---

### Testing instructions

1. The ide-setting repo is working with the main branch of IDEasy. So in \ide-settings\repositories\IDEasy.properties change the git_url to "https://github.com/marceltchanga9/IDEasy.git" and the git branch to bug/2042_quiet_modus_supression_for_question
2. Delete all settings in settings.xml
3. create a new project using the command **_ide -q create test-project  https://github.com/marceltchanga9/ide-settings.git#bug/2042_q_parameter_suspression_**
4. Check the console and you'll be asked about the version of Python not being up-to-date

---

### Checklist for this PR

- [x] When running `mvn clean test` locally all tests pass and build is successful ✅
- [x] PR title is of the form `#«issue-id»: «brief summary»` (e.g. `#921: fixed setup.bat`). If no issue ID exists, title only.✅
- [x] PR top-level comment summarizes what has been done and contains link to addressed issue(s)✅
- [x] PR and issue(s) have suitable labels✅
- [x] Issue is set to `In Progress` and assigned to you *or* there is no issue (might happen for very small PRs)✅
- [x] You followed all [coding conventions](https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc)✅
  with `internal`✅
- [x] You have formulated clear instructions on how to test your contribution under "Testing instructions"✅


quando632 (commented):
Thanks for tackling this. The one line change does fix the reported symptom: with the threshold at `INTERACTION`, both the question from `cveCheck` and the option list are visible under `-q`.

Since only `ContextCommandlet.java` is part of the diff, GitHub does not let me attach applyable suggestions to the other files, so the proposed changes are inlined below.

## On the chosen approach

The fix works, but it deviates from the guidance in the issue, which says: "IMHO INTERACTION has to be raised to be after WARNING." This PR lowers the quiet threshold instead and leaves the enum order untouched.

Two points on that:

1. The underlying design flaw remains. Any code path that sets the threshold to WARNING or higher will still swallow prompts. This is not theoretical: see the existing workaround in `AbstractIdeContext` lines 1480 to 1485, where the log level is temporarily lowered to INFO just so the license question becomes visible. `runWithoutLogging(lambda, threshold)` can also set arbitrary thresholds.

2. The functional difference is not just cosmetic. With this variant SUCCESS also becomes visible under `-q`, while the enum reordering would keep it suppressed. Affected messages include `IDE_ROOT is set to ...`, `IDE_HOME is set to ...`, `You are online.`, `Found bash executable at: ...` and all `step.success()` output. This does match the documented behaviour of the flag, but it is a noticeable change for anyone using `-q` to keep output small.

Please either reorder the enum to `... SUCCESS, WARNING, INTERACTION, ERROR, PROCESSABLE` and keep `quiet = WARNING`, or briefly explain in the PR why this variant was chosen and align it with @hohwille.

## 1. Duplicate question logging

`cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java` line 1087

The question is logged twice: once at INTERACTION and once as WARN. The WARN line used to be the workaround so that at least the question text showed up under `-q`. Now that INTERACTION is visible in quiet mode, the question is printed twice, in normal mode as well as in quiet mode. This line should be removed as part of this fix.

```java
    assert (options.length > 0);
    IdeLogLevel.INTERACTION.log(LOG, question, args);
    LOG.warn(question, args);   // <- remove this line
    return displayOptionsAndGetAnswer(options);
```

## 2. English help text

`cli/src/main/resources/nls/Help.properties` line 188

The help text no longer describes the actual behaviour. Interactive questions are now visible too.

```properties
opt.--quiet=disable info logging (only log interaction, success, warning or error).
```

## 3. German help text

`cli/src/main/resources/nls/Help_de.properties` line 188

Same change as in the English file. The stray space after the opening bracket is fixed along the way.

```properties
opt.--quiet=Deaktiviert Info Logging (nur interaction, success, warning und error).
```

## 4. Missing CHANGELOG entry

`CHANGELOG.adoc`, section `== 2026.08.001`

The DoD requires an entry for the issue in the next open release. This change is visible to end users.

```asciidoc
* https://github.com/devonfw/IDEasy/issues/2100[#2100]: Fix Python not available for Mac x64
* https://github.com/devonfw/IDEasy/issues/2042[#2042]: -q parameter no longer suppresses interactive questions
```

## 5. Missing test

`cli/src/test/java/com/devonfw/tools/ide/commandlet/ContextCommandletTest.java`

The bugfix has no test. Without an assertion on the log level a revert would go unnoticed. `testRunWithOptions()` already sets `--quiet`, only the check is missing.

Add the import:

```java
import com.devonfw.tools.ide.log.IdeLogLevel;
```

And the assertion at the end of `testRunWithOptions()`:

```java
    assertThat(cxt.getStartContext().getLogLevelConsole()).isEqualTo(IdeLogLevel.INTERACTION);
```

On top of that, an end to end test via `IdeTestContext` would be useful: run a `question(...)` in quiet mode and assert that both the question and the option list are logged, and that the question is logged only once.

marceltchanga9:
@quando632 thank you for the review. That was really well detailed and I went through each point. I implemented what you have said and I have a question about the end-to-end test that you mentioned at the end of the fifth point. Should I write a new test for this or is there already a test that I can extend?

quando632:
Thanks, glad the review was helpful, and nice work turning it around so quickly. Good news on your question: you are already covered, there is no need for a new test class. Two existing tests take the end to end case nicely.

**1. ContextCommandletTest.testRunWithOptions()**

This one is already doing most of the work for you. It sets` --quiet` and just needs one assertion on the resulting level. A quick win at unit level.

**2. IdeContextTest**

Even better, there is already a nice block of three `question(...)` tests at the end of the class. `(testQuestionWithMultipleOptions, testQuestionWithSingleOptionAndEmptyAnswer, testQuestionWithSingleToolVersionChoiceAndEmptyAnswer).` Your end to end test fits right in there.

marceltchanga9:
@quando632 Thanks for your reply.
I'll then push my commits and I will appreciate if you can take a look and tell me, if everything is fine.

quando632 (commented):
Nice, switching to the enum reorder is exactly the right call. INTERACTION now sitting above WARNING fixes the root cause instead of the symptom, and it means any future code path that raises the threshold keeps prompts visible. Removing the duplicate `LOG.warn(question, args)` and adding the CHANGELOG entry rounds it off nicely.

One heads up before you go further: the build is red with three failures, and two of them are on me. My suggestions were written for your original approach (`quiet = INTERACTION`) and no longer fit now that the enum is reordered and `determineLogLevel()` is back to `WARNING`. You applied them faithfully, they just needed to move with the approach. Sorry for the detour.

## The new ordering, for reference

```
TRACE 0, DEBUG 1, INFO 2, STEP 3, SUCCESS 4, WARNING 5, INTERACTION 6, ERROR 7, PROCESSABLE 8
```

With `quiet = WARNING` (5) the visible levels are WARNING, INTERACTION, ERROR and PROCESSABLE. **SUCCESS (4) sits below the threshold and stays suppressed.** That is the detail that makes the two corrections below necessary.

## 1. `ContextCommandletTest.testRunWithOptions:77`

```
expected: INTERACTION
 but was: WARNING
```

The expected value simply needs to follow the approach:

```java
assertThat(cxt.getStartContext().getLogLevelConsole()).isEqualTo(IdeLogLevel.WARNING);
```

## 2. The help texts claim SUCCESS is logged

Since SUCCESS is now below the quiet threshold, the wording I gave you is wrong. It should be:

```properties
opt.--quiet=disable info logging (only log warning, interaction or error).
```

```properties
opt.--quiet=Deaktiviert Info Logging (nur warning, interaction und error).
```

Worth noting: the original text on main ("only log success, warning or error") was already inaccurate, because SUCCESS was below the threshold there too. So this is a long standing doc bug that your PR gets to fix properly.

## 3. `HelpCommandletTest` fails twice

`testRun:62` and `testRunWithCommandlet:99`, both through `assertOptionLogMessages:164`. The expected string is hardcoded in `HelpCommandletTest.java:171` and still carries the old wording:

```java
"-q | --quiet    disable info logging (only log success, warning or error).",
```

Just pull it to the same wording as item 2 and all three failures are gone.

## 4. `JulLogLevel` values drifted out of sync

`JulLogLevel` still has INTERACTION 1050, SUCCESS 1060, WARNING 1070, which now contradicts the enum order.

No rush on this one, it is harmless today: `JulConsoleHandler` filters by `ordinal()`, and for JUL itself the only thing that matters is that every custom level sits above `SEVERE` (1000), which `createJavaUtilLoggingProperties()` relies on via `.level=SEVERE`. The relative values play no role in filtering.

Still, we now have two scales that disagree, and the javadoc in `JulLogLevel` points straight at `IdeLogLevel`. Bumping INTERACTION to 1075 costs nothing and saves the next reader a confusing half hour.

## 5. Worth pinning the new order down

The whole filtering mechanism rests on declaration order, so a one line guard is cheap insurance against a future refactoring quietly sorting it back:

```java
assertThat(IdeLogLevel.INTERACTION.ordinal()).isGreaterThan(IdeLogLevel.WARNING.ordinal());
```

## 6. Small formatting nit

The blank line between `INTERACTION` and the javadoc of `ERROR` got lost. Every other constant in the enum has one.

## 7. On the end to end test

The sketch from my previous comment still applies as is, with one adjustment: pass `IdeLogLevel.WARNING` to `newContext(...)` instead of `IdeLogLevel.INTERACTION`, since that is the quiet threshold now. Everything else, including the assertion on the option lines and the INFO counter check, stays the same.

And now that `LOG.warn` is gone, you can go ahead and add the "logged exactly once" assertion right away:

```java
assertThat(context.getTestStartContext().getEntries()).filteredOn(e -> "Which option?".equals(e.message())).hasSize(1);
```

hohwille (approved):
@marceltchanga9 thanks for your PR. Nice work. 👍 
Now I also understand the reason of bug #2292 that you also fix with this PR.

hohwille:
Successfully retested on Windows with 2026.08.001-08_11_11-SNAPSHOT:
```
$ ide -q install
Sub-command install without any further arguments will perform the initial installation of IDEasy.
Since this is typically not to be called manually, you may have forgotten to specify the tool to install as extra argument.
The current command will install IDEasy on your computer. Are you sure?
Option 1: yes
Option 2: no
```

PR #1957, opened 2026-05-20
#1953: Implement base functionality of Cleanup Commandlet

areinicke:
### Implemented changes:

* Implemented the base functionality of the `cleanup` commandlet. Users can run `ide cleanup` to scan for unused tool installations and automatically remove them. A tool is considered as unused if no project links to the tool installation from its software folder at `$IDE_ROOT/<project>/software` or `$IDE_ROOT/<project>/software/extra`.
* Users will be given a summary report and are prompted before any deletion takes place. The confirmation prompt can be skipped by providing the flag `--fd`.
* Added `ide help` documentation for the new commandlet
* Added a basic test for the new commandlet. This will need to be expanded in th future though.

---

### Testing instructions

Please add conscise, understandable instructions on how a reviewer can test/verify the functionality of your contribution here:

1. [Optional] Install some additional software using IDEasy. You can install additional software for IDEasy projects by switching around the required editions and versions of a tool using the `set-edition` and `set-version` commandlet repeatedly.
2. Run ide cleanup
3. Verify that the unused tool installations have been removed.

AND

1. Run `ide help` and `ide help cleanup` to verify that the help documentation works as expected.

---

### Checklist for this PR

Make sure everything is checked before merging this PR. For further info please also see
our [DoD](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc).

- [x] When running `mvn clean test` locally all tests pass and build is successful
- [x] PR title is of the form `#«issue-id»: «brief summary»` (e.g. `#921: fixed setup.bat`). If no issue ID exists, title only.
- [x] PR top-level comment summarizes what has been done and contains link to addressed issue(s)
- [x] PR and issue(s) have suitable labels
- [x] Issue is set to `In Progress` and assigned to you *or* there is no issue (might happen for very small PRs)
- [x] You followed all [coding conventions](https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc)
- [x] You have added the issue implemented by your PR in [CHANGELOG.adoc](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc) unless issue is labeled
  with `internal`
- [x] You have formulated clear instructions on how to test your contribution under "Testing instructions"

Issue #2197, opened 2026-07-21
Python integration broken


hohwille:
### Actual behavior

```
$uv sync
warning: `VIRTUAL_ENV=D:/projects/xxx/software/python` does not match the project environment path `.venv` and will be ignored; use `--active` to target the active environment instead
```

Taming Python is a full day job.

### Reproduce

1. ide install python
2. uv sync (you need a `pyproject.toml` file for that in your current directory)

### Expected behavior

`uv sync` should do its job but as it seems it does not 🥲 

### IDEasy status

```bash
IDE_ROOT is set to D:/projects
IDE_HOME is set to D:/projects/project
Your version of IDEasy is 2026.07.002.
Your are using the latest stable version of IDEasy and no update is available.
Your operating system is windows(10.0)@x64 [Windows 11@amd64]
You are online.
Found bash executable at: C:/Program Files/Git/usr/bin/bash.***
Found git executable at: C:/Program Files/Git/mingw64/bin/git.***
Your settings are up-to-date.
```

### Related/Dependent issues

#1442

### Comments/Hints

If we cannot find a way to convince uv/python to use a python installation named `python` and it is forced to be named `.venv` then I only see the following solution:
* In Python.java we do not rename `.venv` to `python` but create `python` as symlink
* We ensure in `SystemPath` that we exclude software folders starting with a dot (`.`)
https://github.com/devonfw/IDEasy/blob/441072b5d54619d0927c7f5cecc3e2bab972db3d/cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java#L188

Issue #1580, opened 2025-11-07
Implement cleanup


hohwille:
### Feature idea

We planned the command `ide cleanup` that is already documented [here](https://github.com/devonfw/IDEasy/blob/main/documentation/software.adoc#repository).
This story is about adding this feature:
The idea is as following:
* The new commandlet scans the software repository to find all tools with all their versions
* The new commandlet scans all IDEasy projects and reads the link targets of the tools "installed" in the software folder of each project in each version.
* All tool versions that are not used by any project can be deleted.
* Before physically deleting the (unused) tools/directories, we should log a summary report and ask the user for confirmation. This should log one line per tool with all versions to be deleted followed by a summary line telling how many files and bytes will be deleted. The user can confirm with `yes` (what happens automatically via `ide -fb cleanup`) or abort with `no`. 
* Further options of this commandlet could configure to exclude tool versions/folders not older than X or skip all newer versions (so if java version X is still in use then only delete X-m if unused but keep X+n even if unused).
* Also we should discover all content from each project in `$IDE_HOME/updates` (backups from devonfw-ide) as well as `$IDE_ROOT/_ide/tmp` and and the summary report and deletion.
* Also also `~/Downloads/ide` should be scanned recursive and all files older than X will also be included in the deletion and the summary report.
* The default for X (delete files old than X) could be 1 year (option `--keep-newer-versions`).
* A CLI option of the commandlet allows to configure X like this `ide cleanup --retention-delay=PT6M10D`. Please note that `java.util.Duration` uses this `PT...` format by default for `toString` and `of` methods. See e.g. [here](https://www.influxdata.com/blog/java-time-duration-guide-influxdb/#heading1).

### Additional context

I would recommend to create a structure similar to our URL Repo with a simple `IdeRepo` containing instances of `IdeTool` containing instances of `IdeEdition` containing instances of `IdeVersion` containing the `Path` and potential attributes like 
* a `boolean delete` flag
* a `Set<String>` with the IDEasy projects using this tool+edition+version. If empty, then it can be deleted.
* potentially even the disc size in bytes 
* potentially the number of contained files and directories
We can scan the software repository recursively populating the above structure.

When we scan the projects for in `$IDE_HOME/software/*` or `$IDE_HOME/software/extra/*`.
For every such match we find the according `IdeVersion` in the above structure. We initialise the `Set<String>` as empty set and then add projects that have it linked from their software folder.

[...]

PR #2230, opened 2026-07-29
#1953: Complete cleanup commandlet implementation and review handover

### LINKED ISSUE #1953: Implement Base Functionality of Cleanup Commandlet
### Feature idea

Sub-Issue of #1580 

The planned Cleanup Commandlet should be able to delete unused tools, files with a variety of config options.
Initially, the Cleanup Commandlet should be able to scan and document the installed tools and delete those which are not used by any IDEasy project.

The user should be given a summary report before and asked for confirmation before any tool is deleted. Providing the `--fd` flag should skip the confirmation promopt.

The Cleanup Commandlet should document the installed tools in such a way that future extensions of the Commandlet are easy to make.

### Additional context

_No response_

Caylipp:
### This PR fixes #1953

This PR completes and supersedes #1957 after the original author left the team.

### Implemented changes:

- Reused the existing global `-f` / `--force` mode instead of introducing a cleanup-specific force option.
- Removed the obsolete `--force-delete` help entries.
- Kept cleanup discovery stateless by creating the discovered-tool list locally for each execution.
- Moved the cleanup model into reusable top-level classes in the lowercase `commandlet.cleanup` package.
- Scanned all repositories below `_ide/software`, including `default`, `maven`, and custom repository IDs.
- Matched project software references against resolved installation paths instead of relying on fixed path positions.
- Supported project links to subdirectories of an installation, such as `Contents/MacOS`.
- Supported the following software link structures:
  - `software/<tool>`
  - `software/extra/<tool>`
  - `software/extra/<tool>/<name>`
- Stored project usage only on the concrete tool edition version.
- Derived the unused state of editions and tools from their child versions.
- Improved JavaDoc for the cleanup model and its properties.

---

### Testing instructions

1. mvn -pl cli -Dtest=CleanupCommandletTest test
2. mvn clean test

---
### Related

- Implements #1953
- Supersedes #1957
- Addresses the remaining review feedback from #1957

### Checklist for this PR

  with `internal`

- [x] The new commandlet is tested on all platforms it is available for or tested on all platforms that are in scope of the linked issue


cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupIdeTool.java
```diff
@@ -0,0 +1,88 @@
...
+public class CleanupIdeTool {
+
+  /** The name of this tool. */
+  public final String toolName;
+
+  /** The installation {@link Path} of this tool. */
+  private final Path path;
+
+  /** A flag indicating whether the tool is marked for deletion. */
+  private boolean delete;
```
hohwille: Seems some of my review feedback got lost (was maybe set to resolved by original author):
Please use OOP and avoid pointless redundancies.
We have these 3 classes that all should have a common abstract super-class.
Each of them has 
* `name` (simply generalize to just `name` instead of `toolName`, `editionName`, `versionName`)
* `path`
* `delete`

Also include methods like getters with JavaDoc, setter, etc.
BTW: I also suggested a simplification to have `delete` flag only in `CleanupIdeToolEditionVersion` so the consistency of the state gets trivial while with the current design it is complex. E.g. assume we later reuse this in the GUI and the user may see this as a tree with checkboxes and wants to unselect something manually to prevent deletion and then the parent `deletion` flag would need to be updated.
Simply when you iterate over the structure to perform the actual deletion check if the parent directory (`edition` or `tool`) now is empty after deletion of children and then also delete it.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,337 @@
...
+
+  /**
+   * This method specifies the primary flow for the discovery of installed and unused software, and its deletion.
+   */
+  private void discoverAndDeleteUnusedSoftware() {
+    // Iterate over software in $IDE_ROOT/_ide/software repositories and save installed software to a list
+    List<CleanupIdeTool> installedCleanupIdeTools = discoverInstalledSoftware();
+
+    // Scan for IDEasy projects
+    List<Path> ideasyProjects = this.context.getFileAccess().listChildren(this.context.getIdeRoot(), Files::isDirectory);
```
hohwille: We have the same logic also in our GUI. It would make sense to add a method `IdeContext.findProjects()` for this.
And as a motivation why I always insist on creating reusable structures:
This implementation is incorrect. It will also include `_ide` as well as folders somehow created by the end-user that are not really IDEasy projects. When we create a method and once get it right, then we can reuse it everywhere. Even if there still is a bug, we can fix it in once place instead of multiple places because of copy&paste.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,337 @@
...
+
+    // Iterate through IDEasy projects and scan software in software folder. Save found software to list
+    for (Path ideasyProject : ideasyProjects) {
+      if (ideasyProject.getFileName().toString().equals("_ide")) {
+        continue;
+      }
```
hohwille: This would then be part of the central method.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,337 @@
...
+   * This method discovers all installed tools in all repositories below $IDE_ROOT/_ide/software and saves them to the list of installed tools.
+   *
+   * @return the list of discovered installed tools.
+   */
+  private List<CleanupIdeTool> discoverInstalledSoftware() {
+    List<CleanupIdeTool> installedCleanupIdeTools = new ArrayList<>();
+    Path softwareRepositoryPath = this.context.getSoftwareRepositoryPath();
+    List<Path> repositoryFolders = this.context.getFileAccess().listChildren(softwareRepositoryPath, Files::isDirectory);
+    for (Path repositoryFolder : repositoryFolders) {
+      discoverInstalledSoftwareRepository(installedCleanupIdeTools, repositoryFolder);
```
hohwille: This is correct for `default` and `maven` but not for custom-tools:
https://github.com/devonfw/IDEasy/blob/main/documentation/software.adoc#custom-tools
For the given example URL
https://some-file-server.company.com/projects/my-project
We would have `some-file-server.company.com` as `repositoryFolder` here, but the actual repo with tool/edition/version structure is in the sub-path `projects/my-project`.
Sorry for this extra complication. I also missed this point in the review of the old PR.
So IMHO we should instead explicitly go to subfolders `ToolRepository.ID_DEFAULT` and `MvnRepository.ID` as well as `this.context.getCustomToolRepository().getId()`. The latter ID may be `custom` and that folder will typically not exist. Anyhow code should always be robust and handle non existing directories even when expected (I could delete entire `_ide/software` folder and that should not break `ide cleanup` - actually `ide update` should automatically repair my project if I did such evil deletion manually).

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,337 @@
...
+   *
+   * @param installedCleanupIdeTools the list to populate with discovered tools.
+   * @param repositoryFolder The software repository folder to scan.
+   */
+  private void discoverInstalledSoftwareRepository(List<CleanupIdeTool> installedCleanupIdeTools, Path repositoryFolder) {
+    List<Path> toolFolders = this.context.getFileAccess().listChildren(repositoryFolder, Files::isDirectory);
+    for (Path toolFolder : toolFolders) {
+      CleanupIdeTool tool = new CleanupIdeTool(toolFolder.getFileName().toString(), this.context.getFileAccess().toRealPath(toolFolder));
+      installedCleanupIdeTools.add(tool);
+      discoverInstalledEditions(toolFolder, tool);
```
hohwille: Just as a thought, the methods could return a boolean, if some tool/edition/version was found that looks sane. Then you could skip adding if nothing sane was found like this:
```
boolean success = discoverInstalledEditions(toolFolder, tool);
if (success) {
  installedCleanupIdeTools.add(tool);
}
```
I am just still thinking if we have a reliable way of detecting "sanity".
We could obviously check for existence of `.ide.software.version` what would be my preferred way, but assuming this file was deleted by accident (see also #2197) it would IMHO still make sense to delete the tool installation if unused.
Then I was thinking that version folder should have a name that is a valid version.
However, we have seen some tools violating our own rules for what we think is a valid version.
Also checking that there is at least one digit in the version is wrong since we have version `latest`.
Maybe checking for `latest` or "at least one digit" might make sense.
So far just my thoughts - we can still ignore this for now and consider as improvement after the merge.

hohwille (commented):
We are closing the finishing line 👍 
Still I found some points...

cli/src/test/java/com/devonfw/tools/ide/commandlet/CleanupCommandletTest.java
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * Tests that unused software is deleted while software used by a project is retained.
+   *
+   * @throws IOException if the test setup cannot be created.
+   */
+  @Test
+  void testCleanupDeletesUnusedAndKeepsUsedGlobalSoftware() throws IOException {
+
+    // arrange
+    IdeTestContext context = newContext(PROJECT_BASIC);
```
quando632, line 33: These tests fail on Windows without symlink privileges (admin or Developer Mode). Verified locally with `mvn -Dtest=CleanupCommandletTest test`:

```
java.nio.file.FileSystemException: ...\software\cleanup-test-java: A required privilege is not held by the client
	at CleanupCommandletTest.createSoftwareLink(CleanupCommandletTest.java:187)
[ERROR] Tests run: 4, Failures: 0, Errors: 3, Skipped: 0
```

According to `AGENTS.md`, tests that create symbolic links must call `WindowsSymlinkTestHelper.assumeSymlinksSupported()` at the beginning of the test method, so they are skipped gracefully instead of erroring (see `FileAccessImplTest` for the pattern).

```suggestion
    WindowsSymlinkTestHelper.assumeSymlinksSupported();

    IdeTestContext context = newContext(PROJECT_BASIC);
```

The same is needed in `testCleanupKeepsVersionReferencedBySubdirectory` and `testCleanupKeepsVersionReferencedByNestedExtraTool`, plus the import of `com.devonfw.tools.ide.io.WindowsSymlinkTestHelper`.

quando632 (commented):
Thanks for picking this up the substance of the #1957 review is addressed and the new tests cover exactly the cases that were requested there. I went through the branch and found a few points before this can go in, three of them blocking from my side:

* the new tests error out on Windows without symlink privileges (verified locally, 3 of 4 tests)
* the CHANGELOG entry landed in an already released section
* the confirmation cannot actually be skipped with `-f` alone

The remaining ones are smaller correctness and consistency points. I left out anything already covered by the review above (common super class for the model classes, `delete` flag only on version level, central `findProjects()`).

CHANGELOG.adoc
```diff
@@ -117,6 +117,7 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/1884[#1884]: Fix java unzipping losing symlink information
 * https://github.com/devonfw/IDEasy/issues/1716[#1716]: Add commandlet for Claude Code CLI
 * https://github.com/devonfw/IDEasy/issues/1844[#1844]: Fix vscode installation hanging indefinitely in WSL Linux environments
+* https://github.com/devonfw/IDEasy/issues/1953[#1953]: Implement base functionality of cleanup commandlet
```
quando632: This entry was added to the `== 2026.05.001` section (lines 83-129), which is already released. The currently open section is `== 2026.08.001` at the top of the file.

```suggestion
```

Please add it under `== 2026.08.001` instead:

`* https://github.com/devonfw/IDEasy/issues/1953[#1953]: Implement base functionality of cleanup commandlet`

cli/src/main/resources/nls/Help.properties
```diff
@@ -11,6 +11,8 @@ cmd.cdk=Tool commandlet for AWS CDK.
 cmd.cdk.detail=The AWS Cloud Development Kit (AWS CDK) is an open-source software development framework for defining cloud infrastructure in code and provisioning it through AWS CloudFormation. Detailed documentation can be found at https://docs.aws.amazon.com/cdk/v2/guide/home.html
 cmd.claude=Tool commandlet for Claude Code CLI.
 cmd.claude.detail=Claude Code CLI is a command-line interface for interacting with the Claude AI assistant. Detailed documentation can be found at https://code.claude.com/docs/en/overview
+cmd.cleanup=Commandlet to clean up the IDEasy installation by uninstalling all unused tools.
+cmd.cleanup.detail=This will remove any installed tools that are currently not in use by an IDEasy project.
```
quando632: The issue asks for a flag that skips the confirmation prompt, and dropping `--force-delete` in favour of the global force mode is the right call. However `-f` alone does not skip it: `AbstractIdeContext.displayOptionsAndGetAnswer` only auto-confirms when `isBatchMode() && isForceMode()`, and `ContextCommandlet` sets both flags independently (`setBatchMode(...)` / `setForceMode(...)`). So users actually need `ide -b -f cleanup`.

Since this is the documented way to fulfil the requirement from #1953, it should be mentioned in the help text:

```suggestion
cmd.cleanup.detail=This will remove any installed tools that are currently not in use by an IDEasy project. Before anything is deleted you are asked for confirmation. Run "ide -b -f cleanup" to skip the confirmation.
```

`Help_de.properties:15` needs the same addition. Also, all four tests currently confirm via `context.setAnswers("yes")`, so the force path is untested, a test for it would be good.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,337 @@
...
+
+    super(context);
+    addKeyword(getName());
+  }
+
+  @Override
+  public String getName() {
+
+    return "cleanup";
+  }
```
quando632, line 39: `cleanup` operates on IDE_ROOT and not on a single project, but `isIdeHomeRequired()` defaults to `true`, so `ide cleanup` outside of a project aborts with `NO_IDE_HOME` (see `AbstractIdeContext:1391`). `CreateCommandlet` has the same characteristic and overrides it:

```suggestion
  @Override
  public String getName() {

    return "cleanup";
  }

  @Override
  public boolean isIdeHomeRequired() {

    return false;
  }
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,337 @@
...
+   * @param tool The respective tool for which we are discovering editions.
+   */
+  private void discoverInstalledEditions(Path editionFolder, CleanupIdeTool tool) {
+    List<Path> subfolders = this.context.getFileAccess().listChildren(editionFolder, Files::isDirectory);
+    for (Path subfolder : subfolders) {
+      CleanupIdeToolEdition edition = new CleanupIdeToolEdition(subfolder.getFileName().toString(), this.context.getFileAccess().toRealPath(subfolder));
+      tool.getEditions().add(edition);
+      discoverInstalledVersions(subfolder, edition);
+    }
+  }
```
quando632, line 141: The parameter names are one level off: this method receives the *tool* folder (the caller passes `toolFolder`), not an edition folder - and the loop variable that actually is the edition folder is called `subfolder`. In the review of #1957 the request was to name these `editionFolder` and `versionFolder`.

```suggestion
  /**
   * This method discovers all installed editions of a tool at $IDE_ROOT/_ide/software/<repository>/<tool> and saves them to the edition list of the tool.
   * Installed versions of the edition are then recursively discovered.
   *
   * @param toolFolder The folder where the editions are saved in ($IDE_ROOT/_ide/software/<repository>/<tool>).
   * @param tool The respective tool for which we are discovering editions.
   */
  private void discoverInstalledEditions(Path toolFolder, CleanupIdeTool tool) {
    List<Path> editionFolders = this.context.getFileAccess().listChildren(toolFolder, Files::isDirectory);
    for (Path editionFolder : editionFolders) {
      Path editionPath = this.context.getFileAccess().toRealPath(editionFolder);
      CleanupIdeToolEdition edition = new CleanupIdeToolEdition(editionFolder.getFileName().toString(), editionPath);
      tool.getEditions().add(edition);
      discoverInstalledVersions(editionFolder, edition);
    }
  }
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,337 @@
...
+   * @param edition The respective edition for which we are discovering versions.
+   */
+  private void discoverInstalledVersions(Path versionFolder, CleanupIdeToolEdition edition) {
+    List<Path> subfolders = this.context.getFileAccess().listChildren(versionFolder, Files::isDirectory);
+    for (Path subfolder : subfolders) {
+      CleanupIdeToolEditionVersion version = new CleanupIdeToolEditionVersion(subfolder.getFileName().toString(),
+          this.context.getFileAccess().toRealPath(subfolder));
+      edition.getVersions().add(version);
+    }
+  }
```
quando632, line 158: Same here: the parameter is the *edition* folder, not a version folder.

```suggestion
  /**
   * This method discovers all installed versions of an edition of a tool at $IDE_ROOT/_ide/software/<repository>/<tool>/<edition> and saves them to the version
   * list of the edition.
   *
   * @param editionFolder The folder where the versions are saved in ($IDE_ROOT/_ide/software/<repository>/<tool>/<edition>).
   * @param edition The respective edition for which we are discovering versions.
   */
  private void discoverInstalledVersions(Path editionFolder, CleanupIdeToolEdition edition) {
    List<Path> versionFolders = this.context.getFileAccess().listChildren(editionFolder, Files::isDirectory);
    for (Path versionFolder : versionFolders) {
      Path versionPath = this.context.getFileAccess().toRealPath(versionFolder);
      CleanupIdeToolEditionVersion version = new CleanupIdeToolEditionVersion(versionFolder.getFileName().toString(), versionPath);
      edition.getVersions().add(version);
    }
  }
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,337 @@
...
+          totalToolsDeleted);
+
+      // Ask for confirmation. Automatically confirmed if the global force option is provided.
+      try {
+        this.context.askToContinue("Do you want to continue?");
+
+      } catch (CliAbortException e) {
+        LOG.info("Installed Tools will not be deleted.");
+        return;
+      }
```
quando632: This is the only place in the main sources that catches `CliAbortException`. Swallowing it means `ide cleanup` exits with code 0 when the user answers "no", instead of `ProcessResult.ABORT`  a calling script cannot distinguish "aborted by user" from "completed". Letting it propagate gives the standard `Aborted by end-user.` handling for free.

```suggestion
      // Ask for confirmation. Automatically confirmed in batch mode with the global force option.
      this.context.askToContinue("Do you want to continue?");
```

The import of `com.devonfw.tools.ide.cli.CliAbortException` in line 11 then becomes unused and has to be removed, otherwise checkstyle fails.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,337 @@
...
+        }
+        if (!logOutputVersion.isBlank()) {
+          // If at least one version of the edition should be deleted
+          if (versionsDeleted < edition.getVersions().size()) {
+            logOutputVersion += "\t\t + " + (edition.getVersions().size() - versionsDeleted) + " more version(s) of this edition will not be deleted\n";
+          }
+          logOutputEdition += "\t - " + edition.editionName + "\n" + logOutputVersion;
+          editionsDeleted++;
+          totalEditionsDeleted++;
+        }
```
quando632, line 278: `editionsDeleted` / `totalEditionsDeleted` are incremented whenever *any* version of the edition is deleted, so they count "affected" rather than "deleted" editions. For a tool with edition A (1 of 2 versions deleted) and edition B (nothing deleted) the output says "1 more edition(s) of this tool will not be deleted", although in fact both editions are kept. The `Summary:` line in 268 has the same problem.

```suggestion
        if (!logOutputVersion.isBlank()) {
          // If at least one version of the edition should be deleted
          if (versionsDeleted < edition.getVersions().size()) {
            logOutputVersion += "\t\t + " + (edition.getVersions().size() - versionsDeleted) + " more version(s) of this edition will not be deleted\n";
          }
          logOutputEdition += "\t - " + edition.editionName + "\n" + logOutputVersion;
          if (edition.isDelete()) {
            editionsDeleted++;
            totalEditionsDeleted++;
          }
        }
```

`totalToolsDeleted` in line 260 needs the same guard with `tool.isDelete()`. Note this interacts with the suggestion to keep the `delete` flag only on `CleanupIdeToolEditionVersion` - in that design the counters would be derived from "all children deleted" instead.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,337 @@
...
+  private void discoverAndDeleteUnusedSoftware() {
+    // Iterate over software in $IDE_ROOT/_ide/software repositories and save installed software to a list
+    List<CleanupIdeTool> installedCleanupIdeTools = discoverInstalledSoftware();
+
+    // Scan for IDEasy projects
+    List<Path> ideasyProjects = this.context.getFileAccess().listChildren(this.context.getIdeRoot(), Files::isDirectory);
+
+    // Iterate through IDEasy projects and scan software in software folder. Save found software to list
+    for (Path ideasyProject : ideasyProjects) {
+      if (ideasyProject.getFileName().toString().equals("_ide")) {
```
quando632: `IdeContext.FOLDER_UNDERSCORE_IDE` exists for this - using the constant also makes the usage findable from the IDE. `IdeContext` is already imported.

```suggestion
      if (ideasyProject.getFileName().toString().equals(IdeContext.FOLDER_UNDERSCORE_IDE)) {
```

(Becomes obsolete if the central `IdeContext.findProjects()` suggested above is implemented.)

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupIdeTool.java
Caylipp: Thanks. I implemented with a common superclass, version level deletion state and cleanup of empty parent folders.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
Caylipp: Thanks, implemented and reused in both CleanupCommandlet and the GUI.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
Caylipp: Thanks, this is now handled centrally by findProjects().

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
Caylipp: I implemented for default, Maven, and the full custom repository path. Missing folders are skipped.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
Caylipp: I left this for a potential followup as suggested.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,347 @@
...
+    Path softwareRepositoryPath = this.context.getSoftwareRepositoryPath();
+
+    discoverInstalledSoftwareRepository(installedSoftwareTools,
+        softwareRepositoryPath.resolve(ToolRepository.ID_DEFAULT));
+
+    discoverInstalledSoftwareRepository(installedSoftwareTools,
+        softwareRepositoryPath.resolve(MvnRepository.ID));
+
+    discoverInstalledSoftwareRepository(installedSoftwareTools,
+        softwareRepositoryPath.resolve(this.context.getCustomToolRepository().getId()));
```
quando632: With `isIdeHomeRequired()` now returning `false`, `ide cleanup` can run without IDE_HOME. In that case `getSettingsPath()` is `null` and this line throws, because `CustomToolRepositoryImpl.of` passes it to `loadJsonFromFolder`, which calls `folder.resolve(...)` unguarded. Verified with a throwaway test (IDE_ROOT set, no IDE_HOME):

```
java.lang.NullPointerException: Cannot invoke "java.nio.file.Path.resolve(String)" because "folder" is null
  at StandardJsonObjectMapper.loadJsonFromFolder(StandardJsonObjectMapper.java:30)
  at CustomToolRepositoryImpl.of(CustomToolRepositoryImpl.java:172)
  at AbstractIdeContext.getCustomToolRepository(AbstractIdeContext.java:558)
Step 'Identify and remove unused software' ended with failure.
```

Neither the `isIdeHomeRequired()` change nor the explicit custom repository scan causes this on its own, only the combination does.

```suggestion
    // the custom tool repository is derived from ide-custom-tools.json in the settings, which require IDE_HOME
    if (this.context.getSettingsPath() != null) {
      discoverInstalledSoftwareRepository(installedSoftwareTools,
          softwareRepositoryPath.resolve(this.context.getCustomToolRepository().getId()));
    }
```

Making `CustomToolRepositoryImpl.of` null safe would fix this for every caller, but that is out of scope for this PR.

quando632 (commented):
All eight points from my previous review are addressed on `dc95fdea`. `mvn -Dtest=CleanupCommandletTest test` is green (5 tests, 3 skipped for missing symlink privileges), and checkstyle and spotless report no violations in the new files. Solving the counter problem by reporting "affected" tools and editions is nicer than what I suggested.

Two issues came out of the changes themselves, one of them blocking. Both only appear when the individual fixes are combined, so they are not visible in any single change.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,347 @@
...
+  }
+
+  @Override
+  protected void doRun() {
+
+    LOG.debug("Start cleanup commandlet");
+
+    // Identify and remove unused tools.
+    Step step = context.newStep("Identify and remove unused software");
+    step.run(this::discoverAndDeleteUnusedSoftware);
```
quando632: `Step.run(stepCode)` delegates to `run(stepCode, false)`, which catches `RuntimeException | Error` without rethrowing. Any runtime failure inside the cleanup, including the NPE above, is logged as a failed step while `doRun()` returns normally and the commandlet finishes without an error exit.

```suggestion
    // Identify and remove unused tools.
    Step step = this.context.newStep("Identify and remove unused software");
    step.run(this::discoverAndDeleteUnusedSoftware, true);
```

This also adds the `this.` prefix used everywhere else in this class.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,347 @@
...
+    if (logOutput.isBlank()) {
+      LOG.info("No installed tools will be deleted. All installed software is used by at least one project.");
+    } else {
+      LOG.info("The following installed tool versions will be deleted: \n" + logOutput);
+      LOG.info("Summary: {} installed tool versions across {} affected editions of {} affected tools will be deleted.", totalVersionsDeleted,
+          totalAffectedEditions, totalAffectedTools);
+
+      // Ask for confirmation. Automatically confirmed in batch mode with the global force option.
+      this.context.askToContinue("Do you want to continue?");
+      deleteUnusedSoftware(installedSoftwareTools);
```
quando632: Related to the previous comment: `CliException extends RuntimeException`, so the `CliAbortException` that now propagates out of `askToContinue` is caught by the same `step.run(...)`. My earlier suggestion to drop the try/catch therefore has no effect, and it makes the "user says no" case worse than before. Verified by answering `no`:

```
Aborted by end-user.
com.devonfw.tools.ide.cli.CliAbortException: Aborted by end-user.
Step 'Identify and remove unused software' ended with failure.
```

`StepImpl.end` logs this at ERROR level including the stack trace, where the user previously got a plain informational message. Sorry for pointing you in this direction. The cleanest fix is to keep the confirmation outside the step: let the step cover discovery and reporting, then ask and delete in `doRun()` after `step.run(...)`. With that, `rethrow = true` covers real failures and an abort takes the normal `ProcessResult.ABORT` path instead of being reported as a step error.

Caylipp: Thanks for catching this. I moved confirmation and deletion outside the step, so user cancellation now follows the normal abort path while actual step failures are rethrown.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,368 @@
...
+    // Iterate over software in $IDE_ROOT/_ide/software repositories and save installed software to a list
+    installedSoftwareTools.addAll(discoverInstalledSoftware());
+    // Scan for IDEasy projects
+    List<Path> ideasyProjects = this.context.findProjects();
+
+    // Iterate through IDEasy projects and scan software in software folder. Save found software to list
+    for (Path ideasyProject : ideasyProjects) {
+      String projectName = ideasyProject.getFileName().toString();
+      Path ideasyProjectSoftware = ideasyProject.resolve(IdeContext.FOLDER_SOFTWARE);
+      discoverUsedSoftware(installedSoftwareTools, ideasyProjectSoftware, projectName);
```
hohwille: Skip pointless recursion here?
```suggestion
      discoverUsedSoftware(installedSoftwareTools, ideasyProjectSoftware, projectName, 1);
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,368 @@
...
+      for (InstalledSoftwareEdition edition : tool.getEditions()) {
+        for (InstalledSoftwareVersion version : edition.getVersions()) {
+          Path versionPath = version.getPath();
+          if (referencedPath.equals(versionPath) || referencedPath.startsWith(versionPath)) {
+            version.addUsedBy(projectName);
+            return true;
+          }
+        }
+      }
+    }
```
hohwille: So for complexity we do this
* for every M projects
* for every N software folders in project
* for every O installed tools
* for every P installed editions per tool
* for every Q versions per tool-edition

Looks quite inefficient to me.
Our data structure with InstalledSoftwareTool, -Edition, and -Version could be smarter having an internal `Map` from the versionPath to the `InstalledSoftwareVersion` populated when the data gets added (what should always happen via methods to make it possible to keep states consistent rather than simply doing `get«SomeCollection»().add(...)`.
Then you have `O(1)` here instead of `O(O*P*Q)`.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/InstalledSoftwareVersion.java
```diff
@@ -0,0 +1,85 @@
...
+   *
+   * @param projectName the name of the project using this version.
+   */
+  public void addUsedBy(String projectName) {
+
+    if (!this.usedBy.contains(projectName)) {
+      this.usedBy.add(projectName);
+    }
+  }
+
```
hohwille, line 51: This is also odd design. What you want to archive here is exactly provided by a `Set` instead of a `List`.
This not only allows you to skip the `if` but also to make it more efficient since adding or contains check on `Set` is highly efficient unlike `contains` on `List`.

BTW: I just looked again into story #1580 where I wrote:
> a Set<String> with the IDEasy projects using this tool+edition+version. If empty, then it can be deleted.

Actually this is never used so we could also entirely omit it and just have `deleted` initialized with `true` in constructor and setting it to `false` whenever we find a project using it.
However, I think I suggested this design on purpose so we can later reuse this data structure in the GUI and show the end user a graphical tree where he can even see which projects are using what, etc.
So just change to `Set` and keep it.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -0,0 +1,368 @@
...
+    List<InstalledSoftwareTool> installedSoftwareTools = new ArrayList<>();
+
+    // Discover unused software and report what would be deleted.
+    Step step = this.context.newStep("Identify unused software");
+    step.run(() -> discoverAndReportUnusedSoftware(installedSoftwareTools), true);
+
+    if (hasSoftwareToDelete(installedSoftwareTools)) {
+      // Automatically confirmed in batch mode with the global force option.
+      this.context.askToContinue("Do you want to continue?");
+      deleteUnusedSoftware(installedSoftwareTools);
```
hohwille: Reading this gives me some puzzle in my head:
If there is a question `Do you want to continue?` where is the information logged so I can answer this question?
Fortunately `hasSoftwareToDelete` has no side effect and from the naming we can see it is done in `discoverAndReportUnusedSoftware`.
However, wouldn't it make more sense to move `logSoftwareToBeDeleted(installedSoftwareTools)` out of `discoverAndReportUnusedSoftware` renaming it to just `discoverUnusedSoftware` and then before the if explicitly invoke the logging/reporting method?

I know this is just nice to have but I would better understand such code then...

hohwille (approved):
@Caylipp thanks for taking up this story and PR. 👍 
I still have suggestions for improvement.
I hope you can address them tomorrow and I can merge the next day what would be awesome.
If you are running out of time, you may skip some "nice to haves" and I will create a cleanup issue and merge then. Thanks.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
Caylipp: Thanks, implemented a map for efficient version lookup and added controlled methods for hierarchy updates.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/InstalledSoftwareVersion.java
Caylipp: Thanks, changed usedBy to a Set and removed the redundant contains check.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
Caylipp: Thanks, moved the reporting out of discovery so the flow is now explicit before confirmation.

hohwille:
Build failed with:
```
[ERROR] Errors: 
[ERROR]   TaskWindowTest.listContentsAreCoherent:111 » ConcurrentModification
[INFO] 
[ERROR] Tests run: 56, Failures: 0, Errors: 1, Skipped: 0
```

Seems to be #2242 - I restarted the build and hope we get a green run.
Will also look at PR for #2242 and get this merged/fixed.
If that goes fast, I can also just sync this PR with main.

quando632:
Successfully tested with `ide-cli 2026.08.001-SNAPSHOT` on Windows

PR #2233, opened 2026-07-29
#2183: dynamically generate form fields for selected commandlet

### LINKED ISSUE #2183: Dynamically generate form fields for selected commandlet
As a user, when I select a commandlet, I want to see input fields for its parameters, options, and flags so I can configure the execution.

### Task summary
- extend the controller to read the selected commandlet metadata
- generate form fields based on the commandlet
  - text field for positional parameters
  - input fields for named options
  - checkboxes for boolean flags

### Additional context

_No response_

Hiepiscus:
### This PR fixes #2183

Introduces dynamic form generation for commandlets. Instead of using a static input field, the UI now creates appropriate form fields based on the properties of the selected commandlet for a more flexible configuration experience.
Show more lines

### Implemented changes:
   * Dynamically generate form fields based on the selected commandlet properties
   * `PropertyFormFieldFactory` (new): create appropriate UI based on commandlet:
     - `TextField` for positional parameters and named options 
     - `CheckBox` for boolean flags
   * Replaced static `TextField` with a `ScrollPane` + `VBox` in the FXML layout
   * Added `onCommandletSelected()` listener to regenerate the form on selection change
   * Adapted `runCommandlet()` to read property values from generated form fields

---

### Testing instructions

1. Run `AppLauncher`
2. Set a project and workspace
3. Click `Commandlets`
4. Select a commandlet from the combobox
5. Verify that the displayed form fields change depending on the selected commandlet
6. Enter parameters in the generated fields
7. Click `Run`
8. Verify that the commandlet executes successfully

---

### Checklist for this PR

  with `internal`


gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
```diff
@@ -0,0 +1,87 @@
...
+  private void populateCommandletList() {
+    commandletSelector.getItems().clear();
+    commandletSelector.getItems().addAll(context.getCommandletManager().getCommandlets().stream()
+        .map(Commandlet::getName)
+        .toList());
+  }
+
+  @FXML
+  private void runCommandlet() {
+    if (this.commandletSelector == null) {
```
quando632: `commandletSelector` is injected by FXML and is never null, so the guard does not catch an empty selection and `selectedCommandlet` is null in line 85. Checking the selected commandlet restores the behaviour the method had before the refactoring.

```suggestion
    if (this.selectedCommandlet == null) {
```

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
```diff
@@ -0,0 +1,87 @@
...
+      return;
+    }
+
```
quando632, line 101: The commandlets are singletons shared with the CLI context, so values of the previously selected commandlet stay set, and for multivalued properties `Property.setValue` appends instead of replacing, which adds one more value per click on Run. `AbstractIdeContext.applyAndRun:1384` calls `reset()` first for that reason.

```suggestion

    this.selectedCommandlet.reset();
```

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
```diff
@@ -0,0 +1,87 @@
...
+    for (javafx.scene.Node node : formContainer.getChildren()) {
+      if (node instanceof javafx.scene.layout.HBox hbox && hbox.getUserData() instanceof Property<?>
+          property) {
+        for (javafx.scene.Node child : hbox.getChildren()) {
+          if (child instanceof javafx.scene.control.TextField textField) {
+            property.setValueAsString(textField.getText(), context);
```
quando632: Blank fields are assigned as well, and `ToolProperty.parse("")` throws, so pressing Run on `install` without input ends in the default uncaught exception handler. Skipping blank input leaves the property unset, which is what the CLI does for a missing optional argument.

```suggestion
            String value = textField.getText();
            if (!value.isBlank()) {
              property.assignValueAsString(value, this.context, this.selectedCommandlet);
            }
```

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
```diff
@@ -0,0 +1,87 @@
...
+            break;
+          }
+        }
+      }
+    }
+
+    this.selectedCommandlet.run();
```
quando632: Without `validate()` a missing required value reaches `run()`, for example `create` would start with an empty project name. `validate()` gives us the message to show instead. Needs imports of `com.devonfw.tools.ide.validation.ValidationResult` and `com.devonfw.ide.gui.modal.IdeDialog`.

```suggestion
    ValidationResult result = this.selectedCommandlet.validate();
    if (!result.isValid()) {
      new IdeDialog(IdeDialog.AlertType.ERROR, result.getErrorMessage()).showAndWait();
      return;
    }
    this.selectedCommandlet.run();
```

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
```diff
@@ -0,0 +1,87 @@
...
+  }
```
quando632, line 154: `applyAndRun` also calls `ensureLicenseAgreement(cmd)` and checks `isIdeHomeRequired()` and `isIdeRootRequired()` before running. Skipping the license gate means a tool can be installed from the GUI without the agreement the CLI asks for.

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
```diff
@@ -0,0 +1,87 @@
...
+
+  private void populateCommandletList() {
+    commandletSelector.getItems().clear();
+    commandletSelector.getItems().addAll(context.getCommandletManager().getCommandlets().stream()
+        .map(Commandlet::getName)
+        .toList());
+  }
+
+  @FXML
+  private void runCommandlet() {
```
quando632, line 97: `run()` executes on the JavaFX application thread, so `install` freezes the window for the duration of the download and `shell`, which is offered in the combo box, never returns. A `javafx.concurrent.Task` around the call would keep the UI responsive. Happy to see this as a follow up issue if it is out of scope here.

gui/src/main/java/com/devonfw/ide/gui/PropertyFormFieldFactory.java
```diff
@@ -0,0 +1,93 @@
...
+    checkbox.setPadding(new Insets(2, 0, 2, 0));
+
+    checkbox.selectedProperty().addListener((obs, oldVal, newVal) -> {
+      booleanProperty.setValue(newVal);
+    });
+
+    return checkbox;
+  }
+
+  private static javafx.scene.Node createKeywordField(KeywordProperty keywordProperty) {
```
quando632, line 68: `Commandlet#addKeyword` creates the keyword with `required = true`, and the label never sets it, so `validate()` would reject every commandlet that has a keyword. Setting it here would be undone by the `reset()`, so `label.setUserData(keywordProperty)` plus `keyword.setValue(true)` in `runCommandlet` covers it.

gui/src/main/java/com/devonfw/ide/gui/PropertyFormFieldFactory.java
```diff
@@ -0,0 +1,93 @@
...
+    String displayName = booleanProperty.getName();
+    if (displayName.isEmpty()) {
+      displayName = booleanProperty.getAlias() != null ? booleanProperty.getAlias() : "flag";
+    }
+
+    CheckBox checkbox = new CheckBox(displayName);
+    checkbox.setSelected(booleanProperty.isTrue());
+    checkbox.setPadding(new Insets(2, 0, 2, 0));
+
+    checkbox.selectedProperty().addListener((obs, oldVal, newVal) -> {
```
quando632: The listener writes into the shared commandlet on every toggle, even when Run is never pressed, and a `reset()` in `runCommandlet` would clear it again. Reading `isSelected()` at run time via `checkbox.setUserData(booleanProperty)` puts flags and text fields on the same path.

gui/src/main/java/com/devonfw/ide/gui/MainController.java
```diff
@@ -185,6 +194,8 @@ private void setWorkspaceComboBox() {
       eclipseOpen.setDisable(false);
       intellijOpen.setDisable(false);
       vsCodeOpen.setDisable(false);
+      vsCodeOpen.setDisable(false);
+      commandletOpen.setDisable(false);
```
quando632: `vsCodeOpen.setDisable(false)` is set twice.

```suggestion
      commandletOpen.setDisable(false);
```

gui/src/main/java/com/devonfw/ide/gui/MainController.java
```diff
@@ -206,4 +217,21 @@ private void updateContext(String selectedProjectName, String selectedWorkspaceN
...
+  @FXML
+  private void openCommandlet() {
+    try {
+      FXMLLoader loader = new FXMLLoader(getClass().getResource("commandlet-view.fxml"));
+      loader.setResources(this.nlsService.getResourceBundle());
+      loader.setController(new CommandletController(IdeGuiStateManager.getInstance().getCurrentContext()));
+      Parent root = loader.load();
+
+      Stage stage = (Stage) selectedProject.getScene().getWindow();
+      stage.setScene(new Scene(root));
```
quando632: Replacing the scene of the primary stage leaves no way back to the main view, so the app has to be restarted after opening the commandlet view. A back button in `commandlet-view.fxml` or a separate `Stage` for the view would solve it.

gui/src/main/resources/com/devonfw/ide/gui/main-view.fxml
```diff
@@ -166,6 +166,7 @@
               </children>
             </HBox>
           </children>
+          <Button fx:id="commandletOpen" disable="true" text="Commandlets" onAction="#openCommandlet"/>
```
quando632: `text="Commandlets"` here and `text="Run"` in `commandlet-view.fxml:28` bypass the `%key` bundle the rest of the UI uses, so both stay English after a language switch. Adding `commandlets` and `run` to `nls/messages.properties` and `nls/messages_de.properties` keeps them translatable. Please also move the button inside `<children>`, everything else in this file lives there.

```suggestion
            <Button fx:id="commandletOpen" disable="true" mnemonicParsing="false" onAction="#openCommandlet" text="%commandlets"/>
```

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
quando632: The result of the run never reaches the view. A success writes everything to the console and leaves the window unchanged, and a failure escapes as an uncaught exception, for example `build` outside a project throws `CliException: Could not find build descriptor` straight out of the handler. Wrapping `run()` in a try/catch and reporting both outcomes in the view would close the loop.

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
```diff
@@ -0,0 +1,87 @@
...
+    for (Property<?> property : properties) {
+      children.add(PropertyFormFieldFactory.createFormField(property, context));
+    }
+  }
+
+  private void populateCommandletList() {
+    commandletSelector.getItems().clear();
+    commandletSelector.getItems().addAll(context.getCommandletManager().getCommandlets().stream()
+        .map(Commandlet::getName)
+        .toList());
```
quando632, line 93: nit: `.sorted()` was dropped in 8f8f191, so the combo box lists about 40 commandlets unsorted. Was that intentional?

gui/src/main/java/com/devonfw/ide/gui/PropertyFormFieldFactory.java
```diff
@@ -0,0 +1,93 @@
...
+    String alias = property.getAlias();
+
+    if (property.isOption()) {
+      if (alias != null && !alias.isEmpty()) {
+        return name + " (" + alias + ")";
+      }
+      return name;
+    }
+
+    if (name.isEmpty() && alias != null) {
```
quando632: nit: the non option branch reimplements `Property#getNameOrAlias()`.

quando632 (changes requested):
Nice split into `PropertyFormFieldFactory`, that keeps the controller readable. Checking `KeywordProperty` before `BooleanProperty` is easy to get wrong since one extends the other, and it is correct here. All three tasks from #2183 are covered.

One theme runs through the findings: `runCommandlet` executes a different order than `AbstractIdeContext.applyAndRun` (reset, assign, validate, run), while working on the commandlet singletons that `CommandletManagerImpl` shares with the CLI context.

Still open from the DoD checklist: the `CHANGELOG.adoc` entry, and tests. The gui module already has `AppBaseTest` and `HeadlessApplicationTest`, and `PropertyFormFieldFactory` is a static factory, so "property type maps to expected node type" should be cheap to assert.

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
Hiepiscus: The removal was intentional. I was considering a future approach where frequently used commandlets would be shown first. However, the current unsorted order is not predictable, so I've added .sorted() back for now.

gui/src/main/java/com/devonfw/ide/gui/MainController.java
Hiepiscus: Thanks for the suggestion. I've updated the navigation logic so that the commandlet view is displayed in the existing window by replacing the center content of the BorderPane. A back button was added to the commandlet view, allowing users to return to the main view without restarting the application.

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
Hiepiscus: Addressed. I added the validate() check before running the commandlet, but moved the validation logic into a dedicated method.

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+  }
+
+  private boolean validate(Commandlet cmd) {
+
+    KeywordProperty keyword = cmd.getFirstKeyword();
+    if (keyword != null) {
+      keyword.setValue(true);
+    }
```
quando632: `getFirstKeyword()` sets only the first of several keywords, and `RepositoryCommandlet` registers two required ones, `repository` and `setup`. Running `repository` from the GUI therefore always stops at "Error in property setup: Value is required and cannot be empty". Setting every keyword of the commandlet fixes it.

```suggestion
    for (Property<?> property : cmd.getProperties()) {
      if (property instanceof KeywordProperty keyword) {
        keyword.setValue(true);
      }
    }
```

gui/src/main/resources/com/devonfw/ide/gui/main-view.fxml
```diff
@@ -165,10 +165,16 @@
...
             </HBox>
+            <HBox>
+              <children>
+                <Button fx:id="commandletOpen" disable="true" text="Commandlets" onAction="#openCommandlet">
+                  <HBox.margin>
+                    <Insets bottom="10.0" left="10.0" right="10.0" top="10.0"/>
+                  </HBox.margin>
+                </Button>
+              </children>
+            </HBox>
```
quando632, line 206: `App.java:59` pins the minimum window size to half the screen, and at that size the tile list already fills the center, so this button is off screen and the feature cannot be reached without enlarging the window. The left navigation has room and is where the rest of the navigation lives.

Removing it here and adding this after the language block in the left navigation (around line 56) would do it, together with `commandlets=Commandlets` in `nls/messages.properties` and `nls/messages_de.properties`:

```xml
        <VBox maxHeight="-Infinity" prefWidth="100.0" styleClass="sideNavigationElement">
          <children>
            <Button fx:id="commandletOpen" disable="true" maxWidth="1.7976931348623157E308" mnemonicParsing="false"
              onAction="#openCommandlet" text="%commandlets"/>
          </children>
          <padding>
            <Insets left="8.0" right="8.0"/>
          </padding>
        </VBox>
```

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
```diff
@@ -0,0 +1,171 @@
...
+      new IdeDialog(IdeDialog.AlertType.ERROR, "Not inside an IDEasy project!").showAndWait();
+      return false;
+    }
+
+    if (cmd.isIdeRootRequired() && this.context.getIdeRoot() == null) {
+      new IdeDialog(IdeDialog.AlertType.ERROR, "IDEasy root not found!").showAndWait();
+      return false;
+    }
+
+    Path licenseAgreement = this.context.getUserHomeIde().resolve(".license.agreement");
```
quando632: On a machine where `~/.ide/.license.agreement` is missing this blocks every commandlet, and the GUI offers no way to accept the agreement, while `ensureLicenseAgreement` in the CLI asks the user and then writes the file. The path should at least come from the constant.

```suggestion
    Path licenseAgreement = this.context.getUserHomeIde().resolve(IdeContext.FILE_LICENSE_AGREEMENT);
```

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
```diff
@@ -0,0 +1,171 @@
...
+    runButton.setDisable(true);
+    try {
+      this.selectedCommandlet.run();
+      new IdeDialog(AlertType.INFORMATION, "Commandlet executed successfully.").showAndWait();
+    } catch (Exception e) {
+      LOG.error("Commandlet execution failed", e);
+      new IdeDialog(IdeDialog.AlertType.ERROR, e.getMessage()).showAndWait();
+    } finally {
+      runButton.setDisable(false);
+    }
```
quando632: `run()` blocks the JavaFX application thread, so the button never repaints as disabled and the clicks that arrive during the run are delivered right after the `finally`. A `Task` makes the disable effective and keeps the window responsive. Needs `import javafx.concurrent.Task;`.

```suggestion
    Commandlet commandlet = this.selectedCommandlet;
    Task<Void> execution = new Task<>() {
      @Override
      protected Void call() {
        commandlet.run();
        return null;
      }
    };
    execution.setOnSucceeded(event -> {
      this.runButton.setDisable(false);
      new IdeDialog(AlertType.INFORMATION, "Commandlet executed successfully.").showAndWait();
    });
    execution.setOnFailed(event -> {
      this.runButton.setDisable(false);
      Throwable error = execution.getException();
      LOG.error("Commandlet execution failed", error);
      new IdeDialog(IdeDialog.AlertType.ERROR, error.getMessage()).showAndWait();
    });
    this.runButton.setDisable(true);
    Thread thread = new Thread(execution, "commandlet-" + commandlet.getName());
    thread.setDaemon(true);
    thread.start();
```

gui/src/main/java/com/devonfw/ide/gui/PropertyFormFieldFactory.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  private static javafx.scene.Node createKeywordField(KeywordProperty keywordProperty) {
+    Label label = new Label(keywordProperty.getOptionName());
+    label.setStyle("-fx-font-weight: bold;");
+    label.setPadding(new Insets(2, 0, 2, 0));
+
+    label.setUserData(keywordProperty);
+    keywordProperty.setValue(true);
+
```
quando632: With the keywords set in `validate`, these two lines have no effect: `setValue(true)` is undone by the `reset()` in `runCommandlet`, and the user data of the label is never read.

```suggestion
```

gui/src/main/resources/com/devonfw/ide/gui/commandlet-view.fxml
```diff
@@ -0,0 +1,22 @@
...
+  </padding>
+  <children>
+    <HBox spacing="10.0" alignment="CENTER_LEFT">
+      <Button text="Back" mnemonicParsing="false" onAction="#goBack"/>
+    </HBox>
+    <ComboBox fx:id="commandletSelector" maxWidth="1.7976931348623157E308" prefWidth="400.0" VBox.vgrow="NEVER"/>
+    <ScrollPane fitToWidth="true" VBox.vgrow="ALWAYS">
+      <VBox fx:id="formContainer" spacing="5.0"/>
+    </ScrollPane>
+    <Button fx:id="runButton" alignment="BOTTOM_RIGHT" contentDisplay="BOTTOM" mnemonicParsing="false" onAction="#runCommandlet" text="Run"/>
```
quando632: `alignment` on a Button aligns the content inside the button, not the button within its parent, so Run currently sits at the bottom left. A right aligned container puts it where the attribute intends.

```suggestion
    <HBox alignment="CENTER_RIGHT">
      <children>
        <Button fx:id="runButton" mnemonicParsing="false" onAction="#runCommandlet" text="%run"/>
      </children>
    </HBox>
```

gui/src/main/resources/com/devonfw/ide/gui/main-view.fxml
```diff
@@ -165,10 +165,16 @@
...
           </children>
```
quando632, line 207: Moving the button into `<children>` dropped the `<padding>` that the center VBox has in `main`, so the IDE tiles now touch the edges of the window.

```suggestion
          </children>
          <padding>
            <Insets bottom="10.0" left="10.0" right="10.0" top="10.0"/>
          </padding>
```

gui/src/main/resources/com/devonfw/ide/gui/commandlet-view.fxml
```diff
@@ -0,0 +1,22 @@
...
+<?import javafx.scene.control.ComboBox?>
+<?import javafx.scene.control.ScrollPane?>
+<?import javafx.scene.layout.*?>
+<VBox xmlns="http://javafx.com/javafx/25.0.0" xmlns:fx="http://javafx.com/fxml/1" spacing="10.0">
+  <padding>
+    <Insets bottom="10.0" left="10.0" right="10.0" top="10.0"/>
+  </padding>
+  <children>
+    <HBox spacing="10.0" alignment="CENTER_LEFT">
+      <Button text="Back" mnemonicParsing="false" onAction="#goBack"/>
```
quando632: `Back` here and `Run` in line 20 are hardcoded, as are the dialog texts in `CommandletController:126` and `:150`, so they stay English after a language switch. Please add `back`, `run` and `commandlets` to `nls/messages.properties` and `nls/messages_de.properties`.

```suggestion
      <Button text="%back" mnemonicParsing="false" onAction="#goBack"/>
```

quando632 (changes requested):
Thanks for the thorough round, that covers almost everything. Pulling the checks into a separate `validate` method reads well, and routing the back navigation through `rootPane.setCenter` instead of swapping the scene is better than what I suggested, since the navigation stays in place. The tests and the CHANGELOG entry close both open DoD items.

Two of the new checks have side effects that I would fix before merging, and one point from the last round is still open.

gui/src/main/resources/com/devonfw/ide/gui/main-view.fxml
Hiepiscus: @quando632 Thanks for the suggestion. I agree that the current placement may not be ideal, especially on smaller window sizes. However, I'd prefer not to make this change as part of this PR. The overall navigation and layout should first be discussed to determine where this functionality should live.
I would rather create a follow-up issue and revisit this once we have aligned on the overall UX and navigation concept.

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
Hiepiscus: @quando632 Addressed. It makes sense to use the existing constant instead of a hard coded filename.
Regarding the missing license agreement handling, I would track this separately in a follow up issue.

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
Hiepiscus: @quando632 Thanks, I forgot to create a separate issue for this. I tested your suggested change and it works as expected.

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
```diff
@@ -0,0 +1,187 @@
...
+      @Override
+      protected Void call() {
+        commandlet.run();
+        return null;
+      }
+    };
+
+    execution.setOnSucceeded(event -> {
+      this.runButton.setDisable(false);
+      new IdeDialog(AlertType.INFORMATION, "Commandlet executed successfully.").showAndWait();
```
quando632: The buttons are localized now, but the dialog texts are not: this one and the messages in lines 166, 171 and 180 stay English after a language switch. `FXMLLoader` injects the bundle into a field named `resources`, so `@FXML private ResourceBundle resources;` plus the `java.util.ResourceBundle` import is enough to reach it here.

With these four next to the button keys in both bundles:

```properties
executionSucceeded=Commandlet executed successfully.
noIdeHome=Not inside an IDEasy project!
noIdeRoot=IDEasy root not found!
licenseNotAccepted=License agreement not accepted.
```

quando632 (commented):
The switch to a `Task` and the keyword loop both work as intended, and `repository` runs from the GUI now. Deferring the button placement and the license handling to follow up issues makes sense to me.

One thing from the last round is only half done, everything else is fine from my side.

gui/src/main/java/com/devonfw/ide/gui/CommandletController.java
Hiepiscus: @quando632 Addressed. The dialog texts are now localized as suggested.

Issue #2145, opened 2026-07-10
Expand Documentation on symlinks


MeShehi:
### Task summary

On [Symlinks](https://github.com/devonfw/IDEasy/blob/main/documentation/symlink.adoc) add notes on the possibility to create Symlinks with IDEasy using 
` ide ln -sr src-folder target-link-folder`

PR #2239, opened 2026-07-31
#218: add possibility to start a command in a new window with the process builder

### LINKED ISSUE #218: Add possibility to start a command in a new window with the process builder

samuelkos17:
### Implemented changes:

* Added a new `ProcessMode#BACKGROUND_NEW_WINDOW` mode that starts a command in a new terminal window while detaching it from the parent process.
* Implemented platform-specific new-window command builders in `ProcessContextImpl`:
  * **Linux**: Detects available terminal emulators (`gnome-terminal`, `konsole`, `xfce4-terminal`, `tilix`, `alacritty`, `xterm`, `x-terminal-emulator`) and uses the correct flag syntax for each, with a `disown` fallback when no terminal emulator is available.
  * **macOS**: Uses `osascript` to launch a new Terminal.app or iTerm2 window via AppleScript.
  * **Windows**: Uses `cmd.exe /c start` to open a new CMD window.
* Added `launchesNewWindow()` method to `ProcessMode` to check whether the mode should launch a process in a new window.
* Updated `isBackground()` to include the new `BACKGROUND_NEW_WINDOW` mode.
* Added comprehensive tests in `BackgroundNewWindowTest.java`:
  * Structural (mock-based) tests for each platform verifying the generated command shape.
  * E2E tests for each platform using a marker-file polling approach to verify the subprocess actually executes.
  * Windows/macOS E2E tests skip in CI (windows would leak in headless CI runners); Linux E2E runs in CI via the `disown` fallback.

---

### Testing instructions

1. Run `mvn clean test` — all tests should pass, including the 10 new tests in `BackgroundNewWindowTest.java`.
2. Pay special attention to the E2E tests on your platform:
   * **Windows**: `backgroundNewWindowShouldActuallyExecuteViaStartOnWindows` — observe a new CMD window opening.
   * **macOS**: `backgroundNewWindowShouldActuallyExecuteViaOsascriptOnMac` — observe a new Terminal.app (or iTerm2) window opening.
   * **Linux**: `backgroundNewWindowShouldActuallyExecuteOnLinux` — observe a new terminal window opening (if a desktop environment is available), or the test passing via the `disown` fallback in headless environments.
   * These E2E tests are opening new windows that might not close by themselves, this is intentional and made so that we can verify that a new window opens on each of the three OS. Before merging this I intend to either change or completly remove them since we don't want to have terminal windows opening on our local dev environments every time we run tests. (I would be happy for suggestions on how to change them)

---

### Checklist for this PR

- [x] PR title is of the form `#«issue-id»`: `«brief summary»` (e.g. `#921: fixed setup.bat`). If no issue ID exists, title only.
  with `internal`


CHANGELOG.adoc
```diff
@@ -14,6 +14,7 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/1917[#1917]: Allow the creation of a desktop shortcut for the GUI
 * https://github.com/devonfw/IDEasy/issues/2134[#2134]: Add auto-completion for icd command
 * https://github.com/devonfw/IDEasy/issues/2145[#2145]: Improve documentation on symlink regarding ide ln
+* https://github.com/devonfw/IDEasy/issues/2145[#218]: Added possibility to start a command in a new window using the process builder
```
Ali-Shariati-Najafabadi: **`CHANGELOG.adoc:17`**
Wrong issue link (points to #2145 instead of #218):
```suggestion
* https://github.com/devonfw/IDEasy/issues/218[#218]: Added possibility to start a command in a new window using the process builder
```

cli/src/test/java/com/devonfw/tools/ide/process/BackgroundNewWindowTest.java
```diff
@@ -0,0 +1,386 @@
...
+    // The bash command should contain osascript (macOS uses AppleScript for terminal control)
+    String bashCommand = capturedArgs.get(2);
+    assertThat(bashCommand).as("Bash command should contain osascript on macOS").contains("osascript");
+    assertThat(bashCommand).as("Bash command should contain Terminal or iTerm2 AppleScript")
+        .satisfiesAnyOf(
+            s -> assertThat(s).contains("iTerm2"),
+            s -> assertThat(s).contains("Terminal"));
+    // Should NOT fall back to plain disown anymore
+    assertThat(bashCommand).as("Bash command should NOT contain disown on macOS (has terminal support)")
+        .doesNotContain("disown");
```
Ali-Shariati-Najafabadi, line 170: This asserts the mac command shouldn't contain `disown`, but it does. `ProcessContextImpl` always appends `; disown` for new-window mode regardless of OS (see `ProcessContextImpl.java:454`), so the test's just wrong about what the code actually does. Ran it locally, fails every time.

cli/src/test/resources/process-context/write-marker.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# Writes a marker to stdout — used to verify background processes actually execute.
+# The caller should provide $1 as the marker file path.
+echo "background-process-ran" > "$1"
```
Ali-Shariati-Najafabadi, line 5: **`cli/src/test/resources/process-context/write-marker.sh`**
Root cause of the above needs `chmod +x` before committing.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -416,14 +445,230 @@ private void modifyArgumentsOnBackgroundProcess(ProcessMode processMode) {
...
+  }
+
+  private String windowsQuote(String value) {
+
+    if (value == null || value.isEmpty()) {
+      return "\"\"";
+    }
+
+    // Quote unconditionally and escape cmd.exe metacharacters
+    String escaped = value.replace("^", "^^").replace("%", "%%").replace("\"", "\\\"");
```
Ali-Shariati-Najafabadi: **`cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java:580`**
`windowsQuote()` escapes `^`, `%`, `"` but not `&`, `|`, `<`, `>`, which cmd.exe treats as operators even inside quotes sometimes. Worth a test with `&` or `|` in an argument. Also doubling `%` to `%%` only works inside `.bat` files, not on a plain `cmd /c` line.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -405,9 +406,37 @@ private void performLogging(ProcessResult result, int exitCode, String interpret
...
+
+    if (!processMode.isBackground()) {
+      throw new IllegalStateException(
+          "modifyArgumentsOnBackgroundProcess called for a non-background process.");
+    }
 
-    assert processMode.isBackground() : "Cannot handle non background process mode!";
+    String commandToRunInBackground = buildCommand(args);
+
+    if (this.context.getSystemInfo().isWindows()) {
```
Ali-Shariati-Najafabadi, line 455: This routes *all* background modes through `cmd.exe /c start` on Windows now, not just the new one. Before this PR, `BACKGROUND`/`BACKGROUND_SILENT` on Windows went through git-bash. That's a real behavior change for existing callers (Eclipse, Gui, IdeToolCommandlet, Tomcat, the IdeasyCommandlet self-update cleanup that runs `sleep 10 && rm -rf ...`) with no tests for the new Windows path. Intentional? If yes, worth a test; if not, maybe keep those on the old bash path.

cli/src/test/java/com/devonfw/tools/ide/process/BackgroundNewWindowTest.java
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    Path markerFile = Files.createTempFile("bg-marker-mac-", ".txt");
+    Files.delete(markerFile); // Remove so polling can detect when the script creates it
+    Path scriptPath = TEST_RESOURCES.resolve("process-context").resolve("write-marker.sh");
+
+    ProcessContextImpl processContext = new ProcessContextImpl(context);
+    // act
+    ProcessResult result = processContext.executable(scriptPath).addArg(markerFile.toString())
+        .run(ProcessMode.BACKGROUND_NEW_WINDOW);
```
Ali-Shariati-Najafabadi, line 399: Blows up with `End of answers reached!` at line 342. `write-marker.sh` got committed without the executable bit, so IDEasy prompts to chmod it and this test never answers that prompt,unlike the Linux test at `BackgroundNewWindowTest.java:276`, which does `context.setAnswers("1")`. Missing that here.

Ali-Shariati-Najafabadi (changes requested):
Nice approach overall, but I ran this locally and found a few things.

**`cli/src/test/resources/process-context/write-marker.sh`**
Root cause of the above, needs `chmod +x` before committing.


**`cli/src/test/java/com/devonfw/tools/ide/process/BackgroundNewWindowTest.java:218`**, **`:267`**, **`:319`**
The three "actually execute" E2E tests only skip in CI, not on a normal local run, so anyone running `mvn clean test` on Windows/Mac gets a CMD or Terminal window popping up. You called this out yourself in the PR description as a TODO, just don't want it to slip through review. Might be worth gating them behind a system property; the structural/mocked tests already cover the command shape fine on their own.

cli/src/test/java/com/devonfw/tools/ide/process/BackgroundNewWindowTest.java
samuelkos17: I've added `context.setAnswers("1") to that test, hopefully it works now.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
samuelkos17: This was in fact not intentional, thanks a lot for pointing that out, i've implemented a check so that only `BACKGROUND_NEW_WINDOW" will be executed via cmd.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
samuelkos17: Could you provide a suggestion on how I could test this?

cli/src/test/resources/process-context/write-marker.sh
samuelkos17: I did that, hopefully it works now.

cli/src/test/java/com/devonfw/tools/ide/process/BackgroundNewWindowTest.java
samuelkos17: Thanks for noticing that mistake, i fixed it and the test should hopefully run now.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
Ali-Shariati-Najafabadi: Looks right 👍 verified the dispatch now falls back to the old bash path for the existing modes. Nice fix.

cli/src/test/java/com/devonfw/tools/ide/process/BackgroundNewWindowTest.java
Ali-Shariati-Najafabadi: The original error's gone, thanks. New issue though: on my end the test now times out waiting for the marker file. Not 100% sure if that's a real bug or my test environment lacking iTerm2 automation permissions (a plain Terminal.app AppleScript worked fine for me, but the iTerm2-specific one didn't). Did you get this one to actually pass on a real interactive Mac yourself?

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
Ali-Shariati-Najafabadi: mirror your existing marker-file E2E pattern. Run BACKGROUND_NEW_WINDOW with an argument like & echo INJECTED > injected.txt, then assert injected.txt was not created. That proves nothing broke out of the argument, instead of just asserting the quoting looks right.

cli/src/test/java/com/devonfw/tools/ide/process/BackgroundNewWindowTest.java
samuelkos17: I couldn't try it out since i have no Mac. Maybe we ask someone else in the team to test it aswell?

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
samuelkos17: I still don't quite understand on how I should do that tbh. In the existing E2E test the command is written into a batch file, which is given to `processContext.executable()`. Do you mean i should add `& echo INJECTED > injected.txt` to that batch file or somewhere else?

Ali-Shariati-Najafabadi: Sorry, wasn't clear enough. Not inside the batch file's own content, as an *argument* passed to it, since that's what `windowsQuote()` actually processes. Something like:

```java
Path markerFile = Files.createTempFile("bg-marker-", ".txt");
Files.delete(markerFile);
Path injectedFile = markerFile.resolveSibling("injected.txt");
Path batchFile = Files.createTempFile("bg-test-", ".bat");
// batch file just echoes its first arg into the marker file
Files.writeString(batchFile, "@echo off\r\necho %1 > \"" + markerFile + "\"\r\nexit /b 0\r\n");

processContext.executable(batchFile)
    .addArg("& echo INJECTED > " + injectedFile)
    .run(ProcessMode.BACKGROUND_NEW_WINDOW);

// then assert:
// - injectedFile does NOT exist (the & didn't break out and run as a separate command)
// - markerFile DOES exist and contains the literal argument string (nothing got corrupted either)
```

That way you're checking both that it can't be exploited and that the argument still arrives intact.

cli/src/test/java/com/devonfw/tools/ide/process/BackgroundNewWindowTest.java
samuelkos17: @laim2003 do you have iTerm2 automation permission on your MacOs test env and could run this test?

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
samuelkos17: Alright, i've added the test and noticed that without adding `&, |, >, <` quoting the injected-args are actually executed. I've then added the quoting but now I've encountered a new issue: The quoted `& echo INJECTED > injectedFilePath` is passed as an argument to the batchFile-command, which leads to it not executing -> markerFile won't be created and the whole test fails. I'm now uncertain on how to continue with this. Do you have an idea?

Ali-Shariati-Najafabadi: Good, confirms the concern was real. My guess on the new issue: ^ only escapes in cmd.exe outside double quotes, inside windowsQuote()'s "...", it's just a literal character, so it might be corrupting the argument instead of escaping anything. Try removing the caret and see if quoting alone (already there) is enough.

Also, couldn't find your fix pushed yet. If you add a quick System.err.println(args) before this.processBuilder.command(args) in ProcessContextImpl.run(), the printed string would help too.

samuelkos17: Sorry, i forgot to push the changes. I've ran the new test using IntelliJ's debug mode and caputred the args before `this.processBuilder.command(args)` and i got the following:
arg 0: "cmd.exe"
arg 1: "/c"
arg 2 : start "" cmd.exe /k "C:\Users\sakostiu\AppData\Local\Temp\bg-test-5555936347970042732.bat" "^& echo INJECTED ^> C:\Users\sakostiu\AppData\Local\Temp\injected.txt"

When i run it, a cmd window opens saying: Der Befehl "C:\Users\sakostiu\AppData\Local\Temp\bg-test-5555936347970042732.bat" "&" ist entweder falsch geschrieben oder
konnte nicht gefunden werden.

Maybe this helps?

Ali-Shariati-Najafabadi: Thanks, that's actually really helpful. Look at the error: cmd resolved "the command" as "...\bg-test-....bat" "&" — the two separate quoted tokens (batch path and your argument) seem to be merging right where the escaped ^& sits. That points more at start "" mangling things than at windowsQuote() itself start's parsing of the trailing command+args is known to behave differently from a plain cmd /c/cmd /k.

Could you test one more isolation step: run just cmd.exe /k "<batchfile>" "^& echo INJECTED ^> ...\injected.txt" directly (no start "" in front at all) and see if that alone behaves correctly? If yes, the bug is in how modifyArgumentsOnBackgroundProcessWindows wraps things with start, not in the escaping logic and the fix would need to happen there instead.

samuelkos17: I've tested the following:
Arg 0: "cmd.exe"
Arg 1: "/c"
Arg 2: "cmd.exe /k "" "C:\Users\sakostiu\AppData\Local\Temp\bg-test-15823246681728237943.bat" "^& echo INJECTED ^> C:\Users\sakostiu\AppData\Local\Temp\injected.txt""

Is this what you meant? This does not work, there is no cmd window opening and the test fails (no marker-file created).

cli/src/test/java/com/devonfw/tools/ide/process/BackgroundNewWindowTest.java
laim2003: I dont have iTerm2 installed but since its my private mac I can install it and permissions should not be a problem.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
Ali-Shariati-Najafabadi: That's still not it you dropped `start ""` from the front and the empty `""` you added is in the wrong place (that's only needed for `start`, not for `/k`).

 Keep `start ""` exactly as before, just wrap the `cmd.exe /k` target in one extra pair of quotes:

 ```
 start "" cmd.exe /k ""C:\Users\...\bg-test-....bat" "^& echo INJECTED ^> C:\Users\...\injected.txt""
 ```

 In code that's a one-line change in `modifyArgumentsOnBackgroundProcessWindows`:

 ```java
 if (processMode.launchesNewWindow()) {
   args.add("start \"\" cmd.exe /k \"" + command + "\"");
 } else {
   args.add("start \"\" /b " + command);
 }
 ```

 I tested this through actual `ProcessBuilder` calls (not just typing into a terminal) and it works and the batch runs, the injected `&` stays inert.

 While I was at it I also found that `windowsQuote()`'s handling of `"` doesn't actually work, `^"` still gets counted as a real quote by cmd, so a payload containing a literal `"` breaks out even with the fix above. Simplest fix is to just reject it instead of trying to escape it:

 ```java
 private String windowsQuote(String value) {
   if (value == null || value.isEmpty()) {
     return "\"\"";
   }
   if (value.indexOf('"') >= 0) {
     throw new IllegalArgumentException("Argument must not contain a double quote character on Windows: " + value);
   }
   String escaped = value.replace("^", "^^")
       .replace("&", "^&")
       .replace("|", "^|")
       .replace(">", "^>")
       .replace("<", "^<")
       .replace("%", "%%");
   return "\"" + escaped + "\"";
 }
 ```

 (drop the old `.replace("\"", "^\"")` line, it wasn't doing anything useful)

 One more thing, separate from all this: the `else` branch (`start "" /b`, no new window) doesn't launch at all either. Tested that independently, same symptom, no marker file. That one's not fixed by the above and needs its own look before this is mergeable.

samuelkos17: Thanks a lot for the help! I really appreciate it. I've added your changes and it works good now. Regarding the else branch: this is actually code from a previous version that i just forgot to remove. The `modifyArgumentsOnBackgroundProcess()`-method already checks if the process is supposed to launch a new window and then handles it differently (I've added this change because of the other comment you wrote that revealed that now all background processes are executed via bash). The `modifyArgumentsOnBackgroundProcessWindows()` method is only used if the process is suppossed to open a new window. This essentially means that the else-branch could never be reached anyways, but I'll remove it since it's basically just dead code. Thanks for testing & pointing that out tho!

cli/src/test/java/com/devonfw/tools/ide/process/BackgroundNewWindowTest.java
laim2003: Dieser Test schlägt bei mir fehl. Was passiert, ist, das sich iTerm öffnet und das ausführt:
```
Last login: Mon Aug 17 13:50:20 on ttys002
'/bin/bash' 'src/test/resources/process-context/write-marker.sh' '/var/folders/xx/82_zv_pn7hx8mclhrn_8znnh0000gn/T/bg-marker-mac-12409771088903167698.txt'; exec bash
lukasfaber@MacBook-Air-von-Lukas ~ % '/bin/bash' 'src/test/resources/process-con
text/write-marker.sh' '/var/folders/xx/82_zv_pn7hx8mclhrn_8znnh0000gn/T/bg-marke
r-mac-12409771088903167698.txt'; exec bash
/bin/bash: src/test/resources/process-context/write-marker.sh: No such file or directory

The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.
```

Der Output des Testes lautet wie folgt:
```
BASH_PATH environment variable was not found
A proper bash executable was found in your PATH environment variable at: /bin/bash
Running command 'src/test/resources/process-context/write-marker.sh' using bash with arguments '/var/folders/xx/82_zv_pn7hx8mclhrn_8znnh0000gn/T/bg-marker-mac-12409771088903167698.txt' ...

java.lang.AssertionError: [Marker file should have been created by the new-window terminal process within 15 seconds] 
Expecting path:
  /var/folders/xx/82_zv_pn7hx8mclhrn_8znnh0000gn/T/bg-marker-mac-12409771088903167698.txt
to exist (symbolic links were followed).

	at com.devonfw.tools.ide.process.BackgroundNewWindowTest.backgroundNewWindowShouldActuallyExecuteViaOsascriptOnMac(BackgroundNewWindowTest.java:411)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)


Process finished with exit code 255
```

Ali-Shariati-Najafabadi:
> Nice approach overall, but I ran this locally and found a few things.
> 
> **`cli/src/test/resources/process-context/write-marker.sh`** Root cause of the above, needs `chmod +x` before committing.
> 
> **`cli/src/test/java/com/devonfw/tools/ide/process/BackgroundNewWindowTest.java:218`**, **`:267`**, **`:319`** The three "actually execute" E2E tests only skip in CI, not on a normal local run, so anyone running `mvn clean test` on Windows/Mac gets a CMD or Terminal window popping up. You called this out yourself in the PR description as a TODO, just don't want it to slip through review. Might be worth gating them behind a system property; the structural/mocked tests already cover the command shape fine on their own.

Tested this locally against your latest push,all green, including the actual injection test opening a real window. Nice work.
One thing before merge: no unit test for windowsQuote() throwing on ". That's the actual security check now, worth asserting directly instead of just relying on the e2e test.
Also noticed the no-new-window path on Windows now goes through the existing bash & disown instead of start /b. Probably right, just add a line in the PR description so it's clear that's on purpose.

cli/src/test/java/com/devonfw/tools/ide/process/BackgroundNewWindowTest.java
Ali-Shariati-Najafabadi: not a quoting issue, it's a working directory problem. write-marker.sh is passed as a relative path, and SystemPath.findBinary() never makes it absolute.
Fix: resolve the executable to an absolute path before building the new window command, or prepend an explicit cd to the current working directory in the generated command probably needed for all three OS builders, not just Mac.

samuelkos17: I've pushed a fix for this, @laim2003 could you re-run the test?

Ali-Shariati-Najafabadi: Re-ran the full suite here too, Also tested the windows-quoting fix manually on a real Windows machine works there too. Looks good from my end.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -203,7 +208,7 @@ public ProcessResult run(ProcessMode processMode) {
     try {
       applyRedirects(processMode);
       if (processMode.isBackground()) {
-        modifyArgumentsOnBackgroundProcess(processMode);
+        modifyArgumentsOnBackgroundProcess(processMode, args);
```
hohwille, line 211: interesting one - seems this was somehow buggy before since the internal arguments were modified after they are copied and tweaked to local variable `args` that is actually used for the process execution.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
```diff
@@ -416,14 +492,223 @@ private void modifyArgumentsOnBackgroundProcess(ProcessMode processMode) {
...
+  }
+
+  private String windowsQuote(String value) {
+
+    if (value == null || value.isEmpty()) {
+      return "\"\"";
+    }
+
+    if (value.indexOf('"') >= 0) {
+      throw new IllegalArgumentException("Argument must not contain a double quote character on Windows: " + value);
```
hohwille, line 617: Fail fast is good but this is also slightly dangerous since we will not fully control the arguments that may also come from variables.
Does Windows not support escaping of quotes?

macos-installer/Distribution.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This file is the installer configuration file for the macOS pkg-Installer. The next_version placeholder will be replaced during build with the correct version. -->
+<installer-gui-script minSpecVersion="1">
+  <title>IDEasy - 2026.08.002-SNAPSHOT</title>
```
hohwille: This file seems to be misplaced here and most likely be added by accident.

hohwille (approved):
@samuelkos17 thanks for your PR. Wow, this is great stuff - you have created solid handlings for every OS and carefully tested your work. 👍 
I have some small review comments but then we can merge this.

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
hohwille: Resolving as nothing to do here - just some positive feedback for you 😄

cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java
hohwille: Edge-case that we can address after the merge.
I will resolve it for now...

macos-installer/Distribution.xml
hohwille: I will simply delete the file - it has nothing to do with the story and it seems a leftover from MacOS PKG testing.

samuelkos17:
Successfully tested with 2026.08.002-08_26_03-SNAPSHOT on Windows.

PR #1896, opened 2026-05-06
#1870: Add generic get-version implementation for global tools under windows

jakozian:
### Implemented changes:

* Add logic in `getInstalledVersion` of `GlobalToolCommandlet` to look up tool in Windows registry.

---

## Checklist for this PR

Make sure everything is checked before merging this PR. For further info please also see
our [DoD](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc).

- [x] When running `mvn clean test` locally all tests pass and build is successful
- [x] PR title is of the form `#«issue-id»: «brief summary»` (e.g. `#921: fixed setup.bat`). If no issue ID exists, title only.
- [x] PR top-level comment summarizes what has been done and contains link to addressed issue(s)
- [x] PR and issue(s) have suitable labels
- [x] Issue is set to `In Progress` and assigned to you *or* there is no issue (might happen for very small PRs)
- [x] You followed all [coding conventions](https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc)
- [x] You have added the issue implemented by your PR in [CHANGELOG.adoc](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc) unless issue is labeled
  with `internal`

PR #2244, opened 2026-07-31
#1870: Add generic get-version implementation for global tools under windows

### LINKED ISSUE #1870: Implement generic get-version implementation for global tools under windows
### Feature idea

Currently [`getInstalledVersion`](https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java#L191) returns `null` as default implementation.

After implementation of #1869 the `WindowsHelper` can be used to achieve this. 

### Additional context

https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/os/WindowsHelper.java

vivu001:
### This PR fixes #1870 

This PR is a renewal of the previous PR #1896.

### Implemented changes:

* Add logic in `getInstalledVersion()` of `GlobalToolCommandlet` to look up tool in Windows registry.
* Override `getWindowsRegistryAppName()` in `Docker` and `PgAdmin` so Windows registry lookup uses the correct app names.

---

## Checklist for this PR

  with `internal`


MeShehi:
The PR correctly implements getInstalledVersion using the WindowsHelper interface. However, it looks like there are some merge conflicts, so I would move this to Reviews once those have been resolved

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -184,4 +184,9 @@ public String getToolHelpArguments() {
 
     return "help";
   }
+
+  @Override
+  public String getWindowsRegistryAppName() {
+    return "Docker Desktop";
```
cap-juan, line 180: I would not solely assume that Docker Desktop will be installed but the Rancher Edition as well.

```
if ("rancher".equals(getConfiguredEdition())) {
      return "Rancher Desktop";
    }

    return "Docker Desktop";
```

cli/src/main/java/com/devonfw/tools/ide/tool/pgadmin/PgAdmin.java
```diff
@@ -102,4 +102,9 @@ private Path getExecutableFolderFromWindowsRegistry() {
     }
     return null;
   }
+
+  @Override
+  public String getWindowsRegistryAppName() {
+    return "pgAdmin 4v9_is1";
```
cap-juan: I wouldn't be so specific with the naming, otherwise with every minimal change of the versioning the filw will have to be updated regularly.

My suggestion would be 

`return "pgAdmin 4";`

cap-juan (changes requested):
LGTM. Just a minor detail to be taken into account for the registry namings.

cli/src/main/java/com/devonfw/tools/ide/tool/pgadmin/PgAdmin.java
vivu001: @cap-juan thanks for the suggestion. You were absolutely right, and I've addressed it.
Could you please take another look?

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
vivu001: @cap-juan Thanks for the suggestion, but I’m leaning toward dropping it because tying the registry lookup to `getConfiguredEdition()` doesn’t make sense. That method reflects what the user configured, not what is actually installed. 

If someone configures "Desktop" but has "Rancher Desktop" installed, or the other way around, the lookup could miss the existing installation. Detection should be based on what is present, not what is configured.

cap-juan (approved):
Thanks for the good job.

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
hohwille: You both make valid points here:
1. @vivu001 you are fully correct that the `getConfiguredEdition()` is not the proper indicator here. Instead we would need to try both and if one of them was found take that one.
2. @cap-juan you are right that we only care about `Docker Desktop` here and ignore `rancher` what is a bug.

I would suggest that we create a follow up issue to clean this up for docker and even improve the feature to potentially support multiple windows registry app names (for different editions).
But lets not block this PR further. It is an improvement and a step in the right direction.

hohwille: Issue #2296 created.

cli/src/main/java/com/devonfw/tools/ide/tool/pgadmin/PgAdmin.java
hohwille: Sorry, this is not your bug, but something broken already before that we missed in the review, but this code is not acceptable:
We must store the result of the super implementation:
```
  Path path = super.getInstallationPath(edition, resolvedVersion);
  if (path == null) {
    path = getExecutableFolderFromWindowsRegistry();
  }
  return path;
```

Or am I missing something?

hohwille: I have no doubt that this code was nuts.
I simply changed and IMHO fixed it.

hohwille (approved):
@vivu001 thanks for your PR. Great improvement to our long neglected global tool infrastructure including JUnit. 👍 
Ready for merge.

hohwille:
For Rancher Desktop it was working with the current Snapshot, for pgadmin it does not work (reports * as version even after fresh installation of PgAdmin triggered via IDEasy and manually completed the installer).

hohwille:
Seems no other retest was done.
IMHO we have to file new bug for PgAdmin.

PR #2152, opened 2026-07-13
#2100: Resolve python versions via uv instead of ide-urls

Paras14:
### Implemented changes:

* Added `PythonRepository`, a `ToolRepository` that resolves the list of installable Python versions from `uv python list` instead of from `ide-urls`. `Python` now uses this repository via an override of `getToolRepository()`.
* `PythonUrlUpdater` and had no `darwin/x64` (Intel Mac) branch. Installable versions (e.g. 3.14.6) were therefore reported as "not found" on Intel Mac. Sourcing versions from `uv` (the tool that actually installs Python) removes this class of gap, since
  `uv` only lists versions it can install for the current OS/architecture. Still entries for Intel Macs(darwin/x64), Windows on arm(win32/arm64), and Linux on arm(linux/arm64).
* Added `Uv#parsePythonListJson(...)` (defensive JSON parsing of `uv python list`) and the `PythonUvListEntry` record, with non-CPython entries (e.g. PyPy) are filtered out.
* Wired `getPythonRepository()` into `IdeContext` and `AbstractIdeContext`
* Fixed a bootstrap ordering issue, where version resolution runs before tool dependencies are installed, so `fetchUvPythonList()` uses the `runTool` variant that ensures `uv`  is installed before running it.
* CVE-security and dependency metadata for Python continue to come from `ide-urls`, which only need the tool folder and
  are unaffected by the per-platform gap.
* Minor: added Javadoc to `UvRepository`.

---

### Testing instructions

Automated:

1. Run `PythonRepositoryTest.java` and `PythonTest.java`.

Manual:

2. On a machine with `uv` available, you can confirm that `uv python list --all-versions --only-downloads --output-format json --no-config` lists the expected CPython versions for your platform.
3. In a fresh IDEasy project where `uv` is **not** yet installed, run `ide install python 3.14*`. Expected: `uv` is installed first, then `uv python list` resolves 3.14.6, then Python 3.14.6 is installed successfully. Make sure you do it in a different new folder.
4. If possible, run step 3 on genuine Intel-Mac hardware, since that is the platform originally reported in #2100.

---

### Checklist for this PR

Make sure everything is checked before merging this PR. For further info please also see
our [DoD](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc).

- [x] When running `mvn clean test` locally all tests pass and build is successful
[...]

Issue #2100, opened 2026-07-02
Python not available for Mac x64


hohwille:
### Actual behavior

I have an MacBookPro with Intel CPU and wanted to setup and IDEasy project that works on other platforms but failed here:
```
Could not find any version matching '3.14*' - there are 183 version(s) available but none matched!
Did you mean one of: 3.11.4, 3.11.3, 3.11.2, 3.11.1, 3.11.0?
com.devonfw.tools.ide.cli.CliException: Could not find any version matching '3.14*' - there are 183 version(s) available but none matched!
Did you mean one of: 3.11.4, 3.11.3, 3.11.2, 3.11.1, 3.11.0?
	at com.devonfw.tools.ide.version.VersionIdentifier.resolveVersionPattern(VersionIdentifier.java:96)
	at com.devonfw.tools.ide.tool.repository.AbstractToolRepository.resolveVersion(AbstractToolRepository.java:299)
```

### Reproduce

Run this on intel mac:
`ide install python 3.14.6`

### Expected behavior

Python version should be installed.

### IDEasy status

```bash
IDE_ROOT is set to ~/projects
IDE_HOME is set to ~/projects/project
Your version of IDEasy is 2026.06.001.
Your are using the latest stable version of IDEasy and no update is available.
Your operating system is mac(15.7.5)@x64 [Mac OS X@x86_64]
You are online.
Found bash executable at: /bin/bash
Found git executable at: /usr/*****/bin/git
Your settings are up-to-date.
```

### Comments/Hints

I expect Python to be installed via `uv`.
So we do we actually need the URL to be present?
Maybe this could be solved with #2026

hohwille:
```
ide install python 3.14.6
Found 2 CVE(s) for version 3.14.6 of tool python:
CVE-2026-6019 with severity 6.1 and affected versions: [(,3.15.0)]
https://nvd.nist.gov/vuln/detail/CVE-2026-6019
CVE-2026-7210 with severity 7.5 and affected versions: [(,3.15.0)]
https://nvd.nist.gov/vuln/detail/CVE-2026-7210
Could not find any other version resolving your CVEs.
Please keep attention to this tool and consider updating as soon as security fixes are available.
Please note that by selecting an unsafe version to install, you accept the risk to be attacked.
Which version do you want to install?
Option 1: current (3.14.6 - unsafe)

No CVEs found for version 0.11.26 of tool uv.
[...]

PR #2246, opened 2026-08-03
#1965: improve dependent installation

### LINKED ISSUE #1965: Improve dependent installations
### Task summary

We already support tool dependencies via `dependencies.json` files from `ide-urls`.
This is properly handled here:
https://github.com/devonfw/IDEasy/blob/5cf051aece16d55a76211ee2b7b1465c5af6cdee/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java#L276-L301

As we can see we are creating a `ToolInstallationRequest` for the dependent installation as a child that is connected to the parent `ToolInstallationRequest`.
This way we can properly detect cycles in dependent installations that could lead to infinity loops:
https://github.com/devonfw/IDEasy/blob/5cf051aece16d55a76211ee2b7b1465c5af6cdee/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java#L158-L163

However, we also have `PackageManagerBasedLocalToolCommandlet` that implicitly causes a dependency to the package-manager (e.g. `npm` or `pip`).
Here we had infinity loop bugs causing stack-overflow errors in tests.
As a hackish quickfix we have this:
https://github.com/devonfw/IDEasy/blob/5cf051aece16d55a76211ee2b7b1465c5af6cdee/cli/src/main/java/com/devonfw/tools/ide/tool/PackageManagerBasedLocalToolCommandlet.java#L88-L99

and:
https://github.com/devonfw/IDEasy/blob/5cf051aece16d55a76211ee2b7b1465c5af6cdee/cli/src/main/java/com/devonfw/tools/ide/tool/node/Node.java#L47

The idea of this story is to improve this and create a clean solution replacing the hack.
The problem is actually already visible to end-users via our logs:
```
$ ide install nest
No CVEs found for version 11.15.0 of tool npm.
No CVEs found for version v24.15.0 of tool node.
...
No CVEs found for version 11.0.21 of tool nest.
No CVEs found for version 11.15.0 of tool npm.
No CVEs found for version v24.15.0 of tool node.
...
Successfully installed nest in version 11.0.21 replacing previous version v24.15.0 of nest at D:\projects\test\software\node
```
So still we see that installation of `npm` and `node` was called twice because loop was not detected.
After this issue was fixed we should only see:
```
$ ide install nest
No CVEs found for version 11.15.0 of tool npm.
No CVEs found for version v24.15.0 of tool node.
...
No CVEs found for version 11.0.21 of tool nest.
...
Successfully installed nest in version v24.15.0 of nest at D:\projects\test\software\node
```

### Additional context

In order to implement a clean fix we would somehow refactor `PackageManagerRequest` to somehow be compatbile with `ToolInstallRequest` or to have the ability to contain a `ToolInstallRequest` as a parent.
Then we could use the same chaining and loop detection fixing this issue.

MeShehi:
### Implemented changes:

* Fixed npm and pip-based installation to avoid double installation cycle
* Added ToolInstallRequest as a parent of PackageManagerRequest, added possibility to use PackageManagerRequest as a parent to ToolinstallRequest
---

### Testing instructions

1. Create or edit your configuration to run `-cp ide-cli` with `ide uninstall npm` and run, wait to end
2. Run the same configuration with `ide install nest`
3. No duplication should be present by the logs shown in terminal

---

### Checklist for this PR

  with `internal`


MeShehi:
Please, when checking the code give extra attention to ToolInstallRequest.detectInstallLoopRecursively, since that's where I suspect the new changes might have unsuspected failures. The tests do pass, but maybe edge cases are not covered

cli/src/main/java/com/devonfw/tools/ide/tool/ToolInstallRequest.java
```diff
@@ -106,10 +129,19 @@ private boolean detectInstallLoopRecursively(ToolEditionAndVersion toolEditionAn
...
+
+    boolean loopFound;
+
+    if (this.parentToolInstallRequest != null) {
+      loopFound = this.parentToolInstallRequest.detectInstallLoopRecursively(toolEditionAndVersion, sb);
+    } else if (this.parentPackageManagerRequest != null) {
+      if (this.parentPackageManagerRequest.getType().equals("uninstall")) {
+        return false;
+      }
+      loopFound = this.parentPackageManagerRequest.getToolInstallRequest().detectInstallLoopRecursively(toolEditionAndVersion, sb);
```
quando632: `getToolInstallRequest()` is null for every request that carries no tool installation, so only the hard-coded `"uninstall"` comparison keeps this from dereferencing null. The public one-arg `runPackageManager` makes other types reachable. Guarding on the parent itself covers the uninstall case too, since those requests never carry one.

```suggestion
    } else if (this.parentPackageManagerRequest != null) {
      // a package manager request without tool installation (e.g. uninstall) cannot be part of an installation loop
      ToolInstallRequest parentRequest = this.parentPackageManagerRequest.getToolInstallRequest();
      if (parentRequest == null) {
        return false;
      }
      loopFound = parentRequest.detectInstallLoopRecursively(toolEditionAndVersion, sb);
```

quando632 (commented):
### What works well

The root cause is correctly identified. `runPackageManager` built the nested install request without a parent, so `isInstallLoop()` could never see the cycle. A repeated `ide install nest` now skips the extra npm installation cycle that ran before, which is what the issue asks for.

`NpmBasedCommandlet.java:68` and `PipBasedCommandlet.java:113` also bring both callers in line with the contract already documented on `computeInstalledVersion`.

### Findings

**1. `PackageManagerBasedLocalToolCommandlet.java:94` no longer uses `pc.createChild()`.** The nested install now shares the caller's `ProcessContext`, including `errorHandling` and `exitCodeAcceptor`, which the child deliberately resets (`ProcessContextImpl.java:84`). Restoring it has to happen inside the constructor, since `setProcessContext` rejects a second write.

**2. `ToolInstallRequest.java:69` hard-codes `silent` to `false`, the replaced code passed `true`.** With the current `ng` fixture this moves the "already installed" lines for node and npm from DEBUG to INFO. Was that intended, or should the flag come from the parent request?

**3. The fixed duplication has no test.** `computeInstalledVersion` is only reached on a repeated installation, so a second `install()` in `NgTest` pins it. The mock npm in the `ng` fixture needs a `list` branch for this, it currently answers `--version` only:

```java
commandlet.install();
context.getTestStartContext().getEntries().clear();
commandlet.install();
assertThat(context).log().hasNoMessageContaining("No CVEs found for version 9.9.2 of tool npm");
```

**4. Step 1 of the testing instructions has no effect.** `ide uninstall npm` refuses the uninstall because `Npm.canBeUninstalled()` is `false`, and still prints "Successfully uninstalled npm". Running `ide install nest` twice reproduces the duplication reliably and would be the clearer instruction.

### Optional: one parent field instead of two

Both inline bugs come from two mutually exclusive parent fields with two constructors that must be kept in sync. Resolving the parent in `runPackageManager` via `request.getToolInstallRequest()` keeps a single field, and the nested request then inherits `silent` and `ignoreProject` automatically.

### Nits

- `ToolInstallRequest.java:90`: the private constructor has no javadoc, and its `silent` and `direct` parameters are always `false` at the single call site.
- `PackageManagerBasedLocalToolCommandlet.java:93` lost the `// See Node.postInstallOnNewInstallation` pointer that explained the npm/node cycle.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolInstallRequest.java
```diff
@@ -75,6 +87,17 @@ private ToolInstallRequest(ToolInstallRequest parent, boolean silent, boolean di
...
 
+  private ToolInstallRequest(PackageManagerRequest parent, boolean silent, boolean direct) {
+    super();
+    this.parentPackageManagerRequest = parent;
+    this.parentToolInstallRequest = null;
+    this.silent = silent;
+    this.direct = direct;
+    if (parent != null) {
+      this.processContext = parent.getProcessContext();
+    }
```
quando632, line 102: This constructor does not propagate `ignoreProject`, while the constructor above it does (line 86). An install request that ignores the project therefore loses that flag as soon as it goes through a package manager, silently and without any error.

```suggestion
    if (parent != null) {
      this.processContext = parent.getProcessContext();
      ToolInstallRequest parentRequest = parent.getToolInstallRequest();
      if (parentRequest != null) {
        this.ignoreProject = parentRequest.ignoreProject;
      }
    }
```

CHANGELOG.adoc
```diff
@@ -6,6 +6,8 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/1965[#1965]: Improve dependent installation, fixed npm and pip-based installation duplication
+* https://github.com/devonfw/IDEasy/issues/2100[#2100]: Fix Python not available for Mac x64
```
quando632: The `#2100` entry already exists further down in the same release block, added by PR #2152.

```suggestion
* https://github.com/devonfw/IDEasy/issues/1965[#1965]: Improve dependent installation, fixed npm and pip-based installation duplication
```

cli/src/main/java/com/devonfw/tools/ide/tool/PackageManagerRequest.java
```diff
@@ -166,4 +168,26 @@ public PackageManagerRequest setProcessMode(ProcessMode processMode) {
...
+  public ToolInstallRequest getToolInstallRequest() {
+    return this.toolInstallRequest;
+  }
+
+  /**
+   * @param toolInstallRequest new value of {@link #getToolInstallRequest()} ToolInstallRequest()}.
+   * @return this {@link PackageManagerRequest} for fluent API calls.
+   */
+  public PackageManagerRequest setToolInstallRequest(
+      ToolInstallRequest toolInstallRequest) {
```
quando632: The javadoc contains a copy/paste artifact, and the signature fits on one line within the 160 character limit.

```suggestion
  /**
   * @param toolInstallRequest new value of {@link #getToolInstallRequest()}.
   * @return this {@link PackageManagerRequest} for fluent API calls.
   */
  public PackageManagerRequest setToolInstallRequest(ToolInstallRequest toolInstallRequest) {
```

cli/src/main/java/com/devonfw/tools/ide/tool/ToolInstallRequest.java
```diff
@@ -58,6 +60,15 @@ public ToolInstallRequest(ToolInstallRequest parent) {
     this(parent, parent.silent, false);
   }
 
+  /**
+   * The constructor.
+   *
+   * @param parent the parent {@link PackageManagerRequest} (in case of a dependency).
+   */
+  public ToolInstallRequest(PackageManagerRequest parent) {
+    this(parent, false, false);
```
quando632: `silent` is hard-coded to `false` here, while the replaced code in `PackageManagerBasedLocalToolCommandlet` passed `true`. That moves "Version 9.9.2 of tool npm is already installed" and the node equivalent from DEBUG to INFO on every `ide install ng`, verified by running the same assertion against this branch and against the merge base.

```suggestion
    this(parent, true, false);
```

cli/src/main/java/com/devonfw/tools/ide/tool/npm/NpmBasedCommandlet.java
```diff
@@ -65,7 +65,7 @@ private VersionIdentifier runPackageManagerGetInstalledVersion(String npmPackage
     ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.THROW_CLI)
         .withExitCodeAcceptor(rc -> true); // if the tool is not installed npm list will end with exit code 1
     request.setProcessContext(pc);
-    ProcessResult result = runPackageManager(request);
+    ProcessResult result = runPackageManager(request, true);
```
quando632, line 68: `skipInstallation=true` is only reached on a repeated installation, so nothing currently covers the fix. A second `install()` in `NgTest` pins it, but the mock npm in `ide-projects/ng/repository/node/node/default/npm` needs a `list` branch first: its `echo "npm $*"` fallback is parsed as a version and `VersionIdentifier.of` throws. With both in place the test fails on the merge base and passes here.

```bash
if [ "$1" == "list" ]; then
  echo "/lib"
  echo "|-- $3@18.0.1"
  exit
fi
```

```java
commandlet.install();
context.getTestStartContext().getEntries().clear();
commandlet.install();
assertThat(context).log().hasNoMessageContaining("No CVEs found for version 9.9.2 of tool npm");
```

quando632:
LGTM now you only need to resolve the conflicts that it should be ready to merge

cli/src/main/java/com/devonfw/tools/ide/tool/ToolInstallRequest.java
```diff
@@ -18,7 +18,9 @@ public final class ToolInstallRequest {
 
   private static final Logger LOG = LoggerFactory.getLogger(ToolInstallRequest.class);
 
-  private final ToolInstallRequest parent;
+  private final PackageManagerRequest parentPackageManagerRequest;
+
+  private final ToolInstallRequest parentToolInstallRequest;
```
hohwille: This looks a little odd to me.
> we would somehow refactor PackageManagerRequest to somehow be compatbile with ToolInstallRequest or to have the ability to contain a ToolInstallRequest as a parent.

I was maybe not clear enough but I would have created a common interface or super-class or my other suggestion was to always create a "virtual instance" of `ToolInstallRequest` for a `PackageManagerRequest` representing an installation. Then we still have only one `parent` making the code easier to read, understand and to avoid bugs. WDYT?

hohwille (approved):
@MeShehi thanks for your PR. 👍 
I would like to ask you to improve the design.
It is fine that `PackageManagerRequest` contains a `ToolInstallRequest`.
Now if every `PackageManagerRequest` that triggers a tool installation (e.g. for `ide install ng` we trigger `npm install -gf @angular/cli@22.2.0`) already contains a `ToolInstallRequest` - that one should point to the actual installation of the new tool to install (in that example `ng`). Now if we already have a parent tool installation causing this, then this one should be the parent.
Hence, you do not need to modify `ToolInstallRequest` and it does not have to know about `PackageManagerRequest` what otherwise also causes a cyclic dependency and if that can be avoided this is always preferred for good design.

PR #2247, opened 2026-08-03
#788: Add support for IDE_OPTIONS variable per IDE commandlet

### LINKED ISSUE #788: Add more starting flags to IDE commandlets
As an IDEasy user I want to add custom arguments/flags when I start an IDE like f.e. intellij.
Therefore a variable `«IDE»_OPTIONS` should be honoured (e.g. `INTELLIJ_OPTIONS`) - analogue to `IDE_OPTIONS` that already exists for IDEasy itself.

For example the IntelliJ user wants to add the `nosplash` [command line option](https://www.jetbrains.com/help/idea/working-with-the-ide-features-from-command-line.html#arguments) e.g. `idea64.exe nosplash`, so he can skip the splash screen when loading intellij. Currently the user would have to directly edit the `idea.properties` file which does not seem like a good solution to me.
Instead it would be easier to just add another param for intellij to env variable `INTELLIJ_OPTIONS` once via export in shell or permanent via `conf/ide.properties`.

Another option which could be useful would be: `idea64.exe [--line <number>] [--column <number>] <path ...>`, described in the [open files from commandline](https://www.jetbrains.com/help/idea/opening-files-from-command-line.html) page.

See also #906

# Additional Information
Maybe it makes sense to create sub-tasks for Eclipse, IdeaBasedToolCommandlet, and Vscode.

Ali-Shariati-Najafabadi:
### Implemented changes:

* Added support for `«IDE»_OPTIONS` (e.g. `INTELLIJ_OPTIONS`, `ECLIPSE_OPTIONS`, `VSCODE_OPTIONS`, `ANDROID_STUDIO_OPTIONS`, `PYCHARM_OPTIONS`) to pass additional command-line arguments when starting an IDE, analogue to the existing global `IDE_OPTIONS`
* Implemented once in `IdeToolCommandlet.runTool(List<String>)` so it applies to all IDE commandlets (Eclipse, IntelliJ, Android Studio, PyCharm, VSCode/VSCodium) without per-tool duplication
* Scoped to only the actual IDE-launch call, not internal calls like plugin installation or repository import
* Registered the new variables in `IdeVariables.java` and documented them in `variables.adoc`

### Testing instructions

1. Set e.g. `INTELLIJ_OPTIONS=nosplash` in `conf/ide.properties`
2. Run `ide intellij` and verify the extra argument is passed to the IDE binary
3. Automated coverage: `mvn -Dtest=EclipseTest,IntellijTest,VscodeTest test` in the `cli` module

### Checklist for this PR

- [x] You have added the issue implemented by your PR in [CHANGELOG.adoc](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc) unless issue is labeled with `internal`


samuelkos17 (commented):
Very nice and clean implementation! I've followed your testing instructions and everything worked perfectly and all tests passed. I've no recommendations for any changes so i'll move this to In Review 😄

cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java
```diff
@@ -60,7 +64,26 @@ protected final void doRun() {
...
+
+  /**
+   * Appends the tokens of {@code «IDE»_OPTIONS} (e.g. {@code INTELLIJ_OPTIONS}) to the given {@code args}. This is the per-tool analogue of the global
+   * {@code IDE_OPTIONS} and only applies when actually starting the IDE (not for internal calls like plugin installation or repository import).
+   *
+   * @param args the command-line arguments to launch this IDE, extended in place.
+   */
+  private void addIdeOptions(List<String> args) {
+
+    String variableName = getName().toUpperCase(Locale.ROOT).replace("-", "_") + OPTIONS_ENV_SUFFIX;
```
hohwille: Please avoid code duplications when easily possible. Assuming the logic may need to be changed in the future and when it exists in multiple places the code quickly gets inconsistent.

```suggestion
    String variableName = getToolVariablePrefix(this.tool) + OPTIONS_ENV_SUFFIX;
```

cli/src/main/java/com/devonfw/tools/ide/variable/IdeVariables.java
```diff
@@ -126,13 +126,29 @@ public interface IdeVariables {
...
+
+  /** {@link VariableDefinition} for additional command-line arguments to start android studio. */
+  VariableDefinitionString ANDROID_STUDIO_OPTIONS = new VariableDefinitionString("ANDROID_STUDIO_OPTIONS", null);
+
+  /** {@link VariableDefinition} for additional command-line arguments to start pycharm. */
+  VariableDefinitionString PYCHARM_OPTIONS = new VariableDefinitionString("PYCHARM_OPTIONS", null);
+
+  /** {@link VariableDefinition} for additional command-line arguments to start vscode. */
+  VariableDefinitionString VSCODE_OPTIONS = new VariableDefinitionString("VSCODE_OPTIONS", null);
+
```
hohwille: We should only create variables when they are actually referenced.
There are also no variables `ECLIPSE_VERSION` or `ECLIPSE_EDITION` for example.
BTW: It seems I missed this in the review for these `«TOOL»_VM_ARGS` that also do not make any sense.
Please also note that when we add a new IDE it will work out of the box and therefore nobody will think of adding a `NEW_IDE_OPTIONS` and `NEW_IDE_VM_ARGS` variable here so only add variables that we actually need here. Also we have `spyder` as IDE that was not added as variable here (see #2293).

cli/src/test/resources/ide-projects/eclipse/eclipseproject/conf/ide.properties
```diff
@@ -1 +1,2 @@
 M2_REPO=~/.m2/repository
+ECLIPSE_OPTIONS=nosplash
```
hohwille: This is just a mocked test. However, tests should be realistic.
The correct option in eclipse is actually `-nosplash` so this test is IMHO confusing.
https://equinox.eclipseprojects.io/launcher/starting_eclipse_commandline.html#launcher-arguments

Fun fact: I tested the option with recent eclipse and with `ide eclipse -nosplash` it was showing the splash screen. However, when just running `eclipse -nosplash` it was working.
IMHO this reveals another problem with this feature: It seems the order of the arguments are relevant.
Seems IDEasy is somewhat buggy in this regard (already without this PR).

hohwille (approved):
@Ali-Shariati-Najafabadi thanks for your PR. Nice implementation of the story including test coverage. Good job 👍 
I only left some small remarks for rework.

Issue #599, opened 2024-09-09
Add release for windows-arm architecture


hohwille:
As a Windows on ARM user, I want to be able to use IDEasy so that I can automate my IDE.

Currently we do not provide releases for Windows on ARM.
However, as a new trend, Microsoft has massively improved support for ARM architecture in Windows 11 (24H2).
New laptops with Snapdragon X Elite arrived on the market.
Currently the percentage of ARM for Windows users might still be low but this is going to change and ARM will become much more relevant also for Windows, especially since the ARM CPUs are very efficient and that means longer lasting battery and less fan-noise on laptops.

With GraalVM and our deployment builds it should be no big deal to add `windows-arm` releases of IDEasy.

See also #563

hohwille:
https://learn.arm.com/learning-paths/cross-platform/github-arm-runners/actions/
https://github.blog/changelog/2024-09-03-github-actions-arm64-linux-and-windows-runners-are-now-generally-available/

Can we just add something like `windows-latest-arm` to make it work?

hohwille:
https://github.com/oracle/graal/issues/2757

hohwille:
https://github.com/oracle/graal/issues/8350

hohwille:
I was able to build my own native image on my ARM laptop.
To my surprise there are a lot of obstacles:
* there is no GraalVM package for windows on ARM at all and it is not even planned (see https://github.com/oracle/graal/issues/9215)
* I downloaded GraalVM for x64 and could make it work with emulation
* I build my own native image of IDEasy on that platform but it is not possible to to cross compilation with GraalVM (see see https://github.com/oracle/graal/issues/8350 or https://github.com/oracle/graal/issues/407) therefore even though I am running GraalVM on ARM64 my native image gets compiled for x86 what is wired.
* I got this log message: `Warning: The host machine does not support all features of 'x86-64-v3'. Falling back to '-march=compatibility' for best compatibility.`. This tells me that if we would add the same option to our win x64 build on github actions, our x86 binaries should also work on ARM fixing the `illegal instruction` error.
* jline is printing stacktraces for every progress bar - see #941

So as expected all feels like bleeding edge and early adopter.

samuelkos17:
I'm pretty sure we already support the windows-arm architecture since we have issues concerning bugs on windows arm (#1981)  so I guess this can be closed

hohwille:
[...]

PR #2254, opened 2026-08-03
#1981: fix ide update failing on windows arm64

### LINKED ISSUE #1981: ide update fails on windows@arm64
### Actual behavior

```
$ ide upgrade --mode=snapshot
No CVEs found for version 2026.05.001-20260527.032443-21 of tool ideasy.
Trying to download ide-cli-2026.05.001-20260527.032443-21-windows-arm64.tar.gz from https://central.sonatype.com/repository/maven-snapshots/com/devonfw/tools/IDEasy/ide-cli/2026.05.001-SNAPSHOT/ide-cli-2026.05.001-20260527.032443-21-windows-arm64.tar.gz
Logfile can be found at C:\Users\joerg\projects\_ide\logs\2026\05\27\_ide-ide-upgrade-10-33-14.log
An unexpected error occurred!
https://github.com/devonfw/IDEasy/issues/new?template=bug_report.yml&title=IllegalStateException%3A+Failed+to+download+file+from+URL+https%3A%2F%2Fcentral.sonatype.com%2Frepository%2Fmaven-snapshots%2Fcom%2Fdevonfw%2Ftools%2FIDEasy%2Fide-cli%2F2026.05.001-SNAPSHOT%2Fide-cli-2026.05.001-20260527.032443-21-windows-arm64.tar.gz+to+C%3A%5CUsers%5Cjoerg%5Cprojects%5C_ide%5Ctmp%5CDownloads%5Cide-cli-2026.05.001-20260527.032443-21-windows-arm64.tar.gz
java.lang.IllegalStateException: Failed to download file from URL https://central.sonatype.com/repository/maven-snapshots/com/devonfw/tools/IDEasy/ide-cli/2026.05.001-SNAPSHOT/ide-cli-2026.05.001-20260527.032443-21-windows-arm64.tar.gz to C:\Users\joerg\projects\_ide\tmp\Downloads\ide-cli-2026.05.001-20260527.032443-21-windows-arm64.tar.gz
        at com.devonfw.tools.ide.io.FileAccessImpl.downloadViaHttp(FileAccessImpl.java:137)
        at com.devonfw.tools.ide.io.FileAccessImpl.download(FileAccessImpl.java:102)
        at com.devonfw.tools.ide.tool.repository.AbstractToolRepository.download(AbstractToolRepository.java:208)
        at com.devonfw.tools.ide.tool.mvn.MvnRepository.getDownloadedArtifact(MvnRepository.java:163)
        at com.devonfw.tools.ide.tool.mvn.MvnRepository.download(MvnRepository.java:261)
        at com.devonfw.tools.ide.tool.mvn.MvnBasedLocalToolCommandlet.downloadTool(MvnBasedLocalToolCommandlet.java:54)
        at com.devonfw.tools.ide.tool.LocalToolCommandlet.performToolInstallation(LocalToolCommandlet.java:240)
        at com.devonfw.tools.ide.tool.LocalToolCommandlet.installTool(LocalToolCommandlet.java:206)
        at com.devonfw.tools.ide.tool.LocalToolCommandlet.doInstallStep(LocalToolCommandlet.java:96)
        at com.devonfw.tools.ide.tool.LocalToolCommandlet.doInstall(LocalToolCommandlet.java:86)
        at com.devonfw.tools.ide.tool.IdeasyCommandlet.doInstall(IdeasyCommandlet.java:147)
        at com.devonfw.tools.ide.tool.ToolCommandlet.install(ToolCommandlet.java:302)
        at com.devonfw.tools.ide.tool.ToolCommandlet.install(ToolCommandlet.java:287)
        at com.devonfw.tools.ide.commandlet.UpgradeCommandlet.doRun(UpgradeCommandlet.java:50)
        at com.devonfw.tools.ide.commandlet.Commandlet.run(Commandlet.java:232)
        at com.devonfw.tools.ide.context.AbstractIdeContext.applyAndRun(AbstractIdeContext.java:1315)
        at com.devonfw.tools.ide.context.AbstractIdeContext.run(AbstractIdeContext.java:1140)
        at com.devonfw.tools.ide.cli.Ideasy.runOrThrow(Ideasy.java:91)
        at com.devonfw.tools.ide.cli.Ideasy.run(Ideasy.java:55)
        at com.devonfw.tools.ide.cli.Ideasy.main(Ideasy.java:134)
        at java.base@25.0.2/java.lang.invoke.LambdaForm$DMH/sa346b79c.invokeStaticInit(LambdaForm$DMH)
Caused by: java.lang.IllegalStateException: Unexpected checked exception whilst communicating to https://central.sonatype.com/repository/maven-snapshots/com/devonfw/tools/IDEasy/ide-cli/2026.05.001-SNAPSHOT/ide-cli-2026.05.001-20260527.032443-21-windows-arm64.tar.gz
        at com.devonfw.tools.ide.network.NetworkStatusImpl.invokeNetworkTask(NetworkStatusImpl.java:141)
        at com.devonfw.tools.ide.io.FileAccessImpl.downloadWithHttpVersion(FileAccessImpl.java:148)
        at com.devonfw.tools.ide.io.FileAccessImpl.downloadViaHttp(FileAccessImpl.java:122)
        ... 20 more
Caused by: java.lang.IllegalStateException: Failed to stream response body from url: https://central.sonatype.com/repository/maven-snapshots/com/devonfw/tools/IDEasy/ide-cli/2026.05.001-SNAPSHOT/ide-cli-2026.05.001-20260527.032443-21-windows-arm64.tar.gz
        at com.devonfw.tools.ide.io.HttpDownloader.httpGet(HttpDownloader.java:89)
        at com.devonfw.tools.ide.io.FileAccessImpl.lambda$downloadWithHttpVersion$0(FileAccessImpl.java:150)
        at com.devonfw.tools.ide.network.NetworkStatusImpl.invokeNetworkTask(NetworkStatusImpl.java:133)
        ... 22 more
Caused by: java.lang.IllegalStateException: Download failed with status code 404
        at com.devonfw.tools.ide.io.HttpDownloader.httpGet(HttpDownloader.java:86)
        ... 24 more


```

### Reproduce

1. Install IDEasy on Windows on ARM.
2. ide upgrade

### Expected behavior

`ide upgrade` should upgrade your IDEasy installation to the latest version.

### IDEasy status

```bash
IDE_ROOT is set to ~/projects
Your version of IDEasy is 2026.05.001-05_22_03-SNAPSHOT.
Your version of IDEasy is 2026.05.001-05_22_03-SNAPSHOT but version 2026.05.001-20260527.032443-21 is available. Please run the following command to upgrade to the latest version:
ide upgrade
Your operating system is windows(10.0)@arm64 [Windows 11@aarch64]
You are online.
Found git executable at: C:/Program Files/Git/**********/bin/git.***
Your settings are up-to-date.
```

### Related/Dependent issues

#599

### Comments/Hints

My assumption is that when we upgraded GraalVM and Java to 25 (see #1782 and #599 / [0def37c](https://github.com/devonfw/IDEasy/commit/0def3c7343a4e41946f0da3803f3e773f0d9b625)) we actually fixed a bug that the native image was resolving `os.arch` wrong detecting architecture for Windows on ARM as x64.
This bugfix not leads to the fact that Windows on ARM actually correctly tries to download the ARM release of IDEasy for Windows.
Problem: We do not have such release yet due to issue #599
This leads to 404 error and the upgrade to fail.

Workaround:
```
$ ide -Dos.arch=x64 upgrade --mode=snapshot
```

Solution idea:
IMHO the bug could be fixed here:
https://github.com/devonfw/IDEasy/blob/ca409b96476981ecfa95e33648a409132265e14f/cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java#L111-L133

When we figure out the latest version we actually download the `maven-metadata.xml` file.
https://github.com/devonfw/IDEasy/blob/ca409b96476981ecfa95e33648a409132265e14f/cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java#L279-L291

We could use a `CachedValue` for the resulting XML `Document` and keep that in memory for a short delay (e.g. 1 Minute). Then we could get it (potentially again from the cache) from the `getMetadata` method.
This would allow us to check if the artefact we want to resolve actually exists.
If it does not exist and our architecture is ARM64, we could try a fallback to X64 instead if that is available.

cap-juan:
### Implemented changes:

- Added a fallback from windows-arm64 to windows-x64 when no native Windows ARM64 snapshot artifact is available.
- Added snapshot metadata inspection to verify which platform classifiers are actually published.
- Added short-term caching for Maven metadata to avoid downloading and parsing the same maven-metadata.xml multiple times.
- Kept the fallback limited to Windows ARM64 so that Linux and macOS artifact resolution remain unchanged.
- Added tests covering the ARM64 fallback, existing ARM64 artifacts, unsupported classifiers, and snapshot-version resolution.

### Testing instructions

1. Run the focused Maven repository tests:
`mvn -pl cli -Dtest=MvnRepositoryTest test`

2. Run the complete test suite:
`mvn clean test`

3. Build and install the modified IDEasy version locally:
`./build-local-dev.sh`

4. Open a new PowerShell session and simulate Windows ARM64:
`$env:IDE_OPTIONS = "-Dos.arch=arm64"`

5. Verify the simulated architecture:
`ide status`

6. Run the previously failing snapshot upgrade:
`ide upgrade --mode=snapshot`

7. Verify that IDEasy falls back to the windows-x64 artifact instead of failing with a 404 for windows-arm64.

8. Remove the architecture override after testing:
`Remove-Item Env:IDE_OPTIONS -ErrorAction SilentlyContinue`

---

### Checklist for this PR

  with `internal`

- [ ] The tool can be installed automatically (during setup via settings) or via the commandlet call
- [ ] The tool is isolated in its IDEasy project, see [Sandbox Principle](https://github.com/devonfw/IDEasy/blob/main/documentation/sandbox.adoc)
- [ ] The new tool is added to the table of tools in [LICENSE.asciidoc](https://github.com/devonfw/IDEasy/blob/main/documentation/LICENSE.adoc)
- [ ] The new commandlet is a [command-wrapper](https://github.com/devonfw/IDEasy/blob/main/documentation/cli.adoc#command-wrapper) for `«tool»`
- [ ] Proper help texts for all supported languages are added [here](https://github.com/devonfw/IDEasy/tree/main/cli/src/main/resources/nls)
- [ ] The new commandlet installs potential dependencies automatically
- [ ] The variables `«TOOL»_VERSION` and `«TOOL»_EDITION` are honored by your commandlet

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnArtifact.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnArtifact.java
@@ -18,7 +18,7 @@ public class MvnArtifact extends SoftwareArtifact {
   public static final String ARTIFACT_ID_IDEASY_CLI = "ide-cli";
 
   /** {@link #getClassifier() Classifier} of source code. */
-  public static final String CLASSIFER_SOURCES = "sources";
+  public static final String CLASSIFIER_SOURCES = "sources";
 
   /** {@link #getType() Type} of JAR file. */
   public static final String TYPE_JAR = "jar";
@@ -28,6 +28,16 @@ public class MvnArtifact extends SoftwareArtifact {
 
   /** {@link #getFilename() Filename} for artifact metadata with version information. */
   public static final String MAVEN_METADATA_XML = "maven-metadata.xml";
+
+  public static final String XML_TAG_VERSIONING = "versioning";
+
+  public static final String XML_TAG_SNAPSHOT_VERSIONS = "snapshotVersions";
+
+  public static final String XML_TAG_SNAPSHOT_VERSION = "snapshotVersion";
+
+  public static final String XML_TAG_CLASSIFIER = "classifier";
+
+  public static final String XML_TAG_EXTENSION = "extension";
 
   private static final Pattern SNAPSHOT_VERSION_PATTERN = Pattern.compile("-\\d{8}\\.\\d{6}-\\d+");
 
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
@@ -1,6 +1,7 @@
 package com.devonfw.tools.ide.tool.mvn;
 
 import java.io.InputStream;
+import java.net.URI;
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -11,8 +12,6 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
@@ -63,24 +62,16 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
 
   private static final Duration METADATA_CACHE_DURATION_SNAPSHOT = Duration.ofMinutes(5);
 
-  private static final long XML_METADATA_CACHE_RETENTION =
-      Duration.ofMinutes(1).toMillis();
+  /**
+   * Short-lived in-memory cache used to avoid duplicate XML metadata fetches within one CLI run.
+   */
+  private static final Duration XML_METADATA_CACHE_DURATION = Duration.ofMinutes(1);
 
   private static final String CLASSIFIER_WINDOWS_ARM64 = "windows-arm64";
 
   private static final String CLASSIFIER_WINDOWS_X64 = "windows-x64";
 
-  /**
-   * Matches resolved timestamped Maven snapshot versions, for example:
-   * 2026.05.001-20260527.032443-21
-   * 2025.02.001-beta-20250204.023111-1
-   */
-  private static final Pattern PATTERN_TIMESTAMPED_SNAPSHOT =
-      Pattern.compile("^(.+)-\\d{8}\\.\\d{6}-\\d+$");
-
   private final Path localMavenRepository;
-
-  private final DocumentBuilder documentBuilder;
 
   private final Map<String, CachedValue<Document>> xmlMetadataCache;
 
@@ -99,12 +90,6 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
     super(context);
     this.localMavenRepository = IdeVariables.M2_REPO.get(this.context);
     this.xmlMetadataCache = new ConcurrentHashMap<>();
-    try {
-      DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
-      this.documentBuilder = factory.newDocumentBuilder();
-    } catch (Exception e) {
-      throw new IllegalStateException("Failed to create XML document builder", e);
-    }
   }
 
   @Override
@@ -162,10 +147,14 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
       }
 
       if (CLASSIFIER_WINDOWS_ARM64.equals(classifier)) {
-        classifier = resolveSnapshotClassifier(
+        classifier = resolveWindowsArm64Classifier(
             artifact,
             classifier,
             artifact.getType());
+
+        if (CLASSIFIER_WINDOWS_X64.equals(classifier)) {
+          arch = SystemArchitecture.X64;
+        }
       }
 
       artifact = artifact.withClassifier(classifier);
@@ -185,30 +174,25 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
    * @param artifact the artifact being resolved.
    * @param classifier the originally resolved classifier.
    * @param extension the artifact extension, such as {@code tar.gz}.
-   * @return the original classifier if it exists or no suitable fallback is
-   *         available; otherwise {@code windows-x64}.
+   * @return the original classifier if it exists or no suitable fallback is available; otherwise {@code windows-x64}.
    */
-  private String resolveSnapshotClassifier(
-      MvnArtifact artifact,
-      String classifier,
-      String extension) {
+  private String resolveWindowsArm64Classifier(MvnArtifact artifact, String classifier, String extension) {
 
-    String snapshotVersion = getSnapshotBaseVersion(artifact.getVersion());
-
-    // The classifier availability is listed in version-specific metadata only
-    // for Maven snapshots. Leave normal release handling unchanged.
-    if (snapshotVersion == null) {
-      return classifier;
+    if (!artifact.isSnapshot()) {
+      logWindowsArm64Fallback(artifact.getVersion());
+      return CLASSIFIER_WINDOWS_X64;
     }
 
-    MvnArtifact metadataArtifact = artifact
-        .withVersion(snapshotVersion)
-        .withMavenMetadata();
+    String metadataUrl = getMavenUrl(artifact.withMavenMetadata());
 
-    String metadataUrl = getMavenUrl(metadataArtifact);
-    Document metadata = fetchXmlMetadata(metadataUrl);
-
-    return resolveSnapshotClassifier(metadata, classifier, extension);
+    try {
+      Document metadata = fetchXmlMetadata(metadataUrl);
+      return resolveSnapshotClassifier(metadata, classifier, extension, artifact.getVersion());
+    } catch (RuntimeException e) {
+      LOG.debug("Failed to inspect snapshot metadata from {} - falling back to {}.",
+          metadataUrl, CLASSIFIER_WINDOWS_X64, e);
+      return CLASSIFIER_WINDOWS_X64;
+    }
   }
 
   /**
@@ -217,12 +201,10 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
    * @param metadata the version-specific Maven snapshot metadata.
    * @param classifier the requested artifact classifier.
    * @param extension the requested artifact extension.
+   * @param version the artifact version.
    * @return the resolved classifier.
    */
-  String resolveSnapshotClassifier(
-      Document metadata,
-      String classifier,
-      String extension) {
+  String resolveSnapshotClassifier(Document metadata, String classifier, String extension, String version) {
 
     if (hasSnapshotArtifact(metadata, classifier, extension)) {
       return classifier;
@@ -231,35 +213,13 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
     if (CLASSIFIER_WINDOWS_ARM64.equals(classifier)
         && hasSnapshotArtifact(metadata, CLASSIFIER_WINDOWS_X64, extension)) {
 
-      LOG.warn(
-          "No Windows ARM64 artifact is available. "
-              + "Falling back to the Windows x64 artifact.");
-
+      logWindowsArm64Fallback(version);
       return CLASSIFIER_WINDOWS_X64;
     }
 
     // Preserve the original classifier so the existing download error is
     // reported when neither the requested artifact nor a fallback exists.
     return classifier;
-  }
-
-  /**
-   * Converts a resolved timestamped snapshot version back to its Maven snapshot
-   * base version.
-   *
-   * @param version the resolved artifact version.
-   * @return the corresponding {@code -SNAPSHOT} version, or {@code null} if the
-   *         version is not a timestamped snapshot.
-   */
-  String getSnapshotBaseVersion(String version) {
-
-    Matcher matcher = PATTERN_TIMESTAMPED_SNAPSHOT.matcher(version);
-
-    if (!matcher.matches()) {
-      return null;
-    }
-
-    return matcher.group(1) + "-SNAPSHOT";
   }
 
   /**
@@ -270,27 +230,42 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
    * @param extension the artifact extension.
    * @return {@code true} if the artifact is present.
    */
-  private boolean hasSnapshotArtifact(
-      Document metadata,
-      String classifier,
-      String extension) {
+  private boolean hasSnapshotArtifact(Document metadata, String classifier, String extension) {
 
-    NodeList snapshotVersions =
-        metadata.getElementsByTagName("snapshotVersion");
+    Element root = metadata.getDocumentElement();
 
-    for (int i = 0; i < snapshotVersions.getLength(); i++) {
-      Node snapshotVersion = snapshotVersions.item(i);
+    Element versioning =
+        findFirstChildElement(root, MvnArtifact.XML_TAG_VERSIONING);
 
-      String currentClassifier =
-          getChildText(snapshotVersion, "classifier");
+    if (versioning == null) {
+      return false;
+    }
 
-      String currentExtension =
-          getChildText(snapshotVersion, "extension");
+    Element snapshotVersions =
+        findFirstChildElement(versioning, MvnArtifact.XML_TAG_SNAPSHOT_VERSIONS);
 
-      if (classifier.equals(currentClassifier)
-          && extension.equals(currentExtension)) {
+    if (snapshotVersions == null) {
+      return false;
+    }
 
-        return true;
+    NodeList children = snapshotVersions.getChildNodes();
+
+    for (int i = 0; i < children.getLength(); i++) {
+      Node node = children.item(i);
+
+      if (node instanceof Element snapshotVersion
+          && MvnArtifact.XML_TAG_SNAPSHOT_VERSION.equals(snapshotVersion.getTagName())) {
+
+        String currentClassifier =
+            getChildText(snapshotVersion, MvnArtifact.XML_TAG_CLASSIFIER);
+
+        String currentExtension =
+            getChildText(snapshotVersion, MvnArtifact.XML_TAG_EXTENSION);
+
+        if (classifier.equals(currentClassifier)
+            && extension.equals(currentExtension)) {
+          return true;
+        }
       }
     }
 
@@ -300,23 +275,17 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
   /**
    * Gets the text content of a direct child node.
    *
-   * @param parent the parent node.
-   * @param childName the child-node name.
+   * @param element the parent node.
+   * @param tag the child-node name.
    * @return the child text, or {@code null} if the child does not exist.
    */
-  private String getChildText(Node parent, String childName) {
+  private String getChildText(Element element, String tag) {
 
-    NodeList children = parent.getChildNodes();
-
-    for (int i = 0; i < children.getLength(); i++) {
-      Node child = children.item(i);
-
-      if (childName.equals(child.getNodeName())) {
-        return child.getTextContent().trim();
-      }
+    Element child = findFirstChildElement(element, tag);
+    if (child == null) {
+      return null;
     }
-
-    return null;
+    return child.getTextContent().trim();
   }
 
   /**
@@ -448,19 +417,26 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
     return getDownloadedArtifact(metadata.getMvnArtifact(), metadata.getChecksums());
   }
 
-  private Element getFirstChildElement(Element element, String tag, Object source) {
+  private Element findFirstChildElement(Element element, String tag) {
 
     NodeList children = element.getChildNodes();
-    int length = children.getLength();
-    for (int i = 0; i < length; i++) {
+    for (int i = 0; i < children.getLength(); i++) {
       Node node = children.item(i);
-      if (node instanceof Element child) {
-        if (child.getTagName().equals(tag)) {
-          return child;
-        }
+      if (node instanceof Element child && child.getTagName().equals(tag)) {
+        return child;
       }
     }
-    throw new IllegalStateException("Failed to resolve snapshot version - element " + tag + " not found in " + source);
+    return null;
+  }
+
+  private Element getFirstChildElement(Element element, String tag, Object source) {
+
+    Element child = findFirstChildElement(element, tag);
+    if (child != null) {
+      return child;
+    }
+    throw new IllegalStateException(
+        "Failed to resolve snapshot version - element " + tag + " not found in " + source);
   }
 
   private Document fetchXmlMetadata(String url) {
@@ -470,7 +446,7 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
             url,
             key -> new CachedValue<>(
                 () -> downloadXmlMetadata(key),
-                XML_METADATA_CACHE_RETENTION));
+                XML_METADATA_CACHE_DURATION.toMillis()));
 
     return cachedMetadata.get();
   }
@@ -479,18 +455,18 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
 
     try {
       return this.context.getNetworkStatus().invokeNetworkTask(() -> {
-        URL xmlUrl = new URL(url);
+        URL xmlUrl = URI.create(url).toURL();
 
         try (InputStream is = xmlUrl.openStream()) {
-          synchronized (this.documentBuilder) {
-            return this.documentBuilder.parse(is);
-          }
+          DocumentBuilder documentBuilder =
+              DocumentBuilderFactory.newInstance().newDocumentBuilder();
+          return documentBuilder.parse(is);
         }
       }, url);
 
     } catch (Exception e) {
       throw new CliException(
-          "Failed to determine the latest version from " + url, e);
+          "Failed to fetch XML metadata from " + url, e);
     }
   }
 
@@ -536,5 +512,11 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
       }
       return this.checksums.iterator();
     }
+  }
+
+  private void logWindowsArm64Fallback(String version) {
+    LOG.warn("No {} artifact is published for version {} - falling back to {} "
+            + "(see https://github.com/devonfw/IDEasy/issues/599).",
+        CLASSIFIER_WINDOWS_ARM64, version, CLASSIFIER_WINDOWS_X64);
   }
 }
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
@@ -67,9 +67,9 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
    */
   private static final Duration XML_METADATA_CACHE_DURATION = Duration.ofMinutes(1);
 
-  private static final String CLASSIFIER_WINDOWS_ARM64 = "windows-arm64";
+  private static final String ARCH_ARM64 = "arm64";
 
-  private static final String CLASSIFIER_WINDOWS_X64 = "windows-x64";
+  private static final String ARCH_X64 = "x64";
 
   private final Path localMavenRepository;
 
@@ -146,13 +146,15 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
         classifier = resolvedClassifier;
       }
 
-      if (CLASSIFIER_WINDOWS_ARM64.equals(classifier)) {
-        classifier = resolveWindowsArm64Classifier(
+      if (classifier.contains(ARCH_ARM64)) {
+        String originalClassifier = classifier;
+
+        classifier = resolveArm64Classifier(
             artifact,
             classifier,
             artifact.getType());
 
-        if (CLASSIFIER_WINDOWS_X64.equals(classifier)) {
+        if (!classifier.equals(originalClassifier)) {
           arch = SystemArchitecture.X64;
         }
       }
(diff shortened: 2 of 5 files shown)


cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -112,26 +139,186 @@ public MvnArtifactMetadata getMetadata(MvnArtifact artifact, String tool, String
...
       resolvedClassifier = classifier.replace("${arch}", arch.toString());
+
       if (resolvedClassifier.equals(classifier)) {
         arch = null;
+      } else {
+        classifier = resolvedClassifier;
       }
-      artifact = artifact.withClassifier(resolvedClassifier);
+
+      if (CLASSIFIER_WINDOWS_ARM64.equals(classifier)) {
```
maybeec, line 149: **Should-fix** — this only fixes `ide upgrade --mode=snapshot`.

`resolveSnapshotClassifier` bails out via `getSnapshotBaseVersion(...) == null` for every non-timestamped version, so release resolution is untouched (which the JavaDoc states as intentional). But `ide upgrade` without `--mode` resolves `UpgradeMode.STABLE` or `UNSTABLE` for a non-SNAPSHOT installation (`IdeasyCommandlet#getConfiguredVersion`), i.e. a plain release version — and IDEasy publishes no `windows-arm64` release either (that is what #599 is about). So the exact command in the issue report still ends in the same 404, and so does a fresh install/setup on Windows on ARM.

The mechanism you used cannot be extended to releases as-is: the release-level `maven-metadata.xml` lists `<version>` elements only, never classifiers, so there is nothing to inspect. Two options:

- Simplest and covers both tracks: while #599 is open, map Windows + ARM64 to x64 unconditionally, and keep the snapshot-metadata check purely as the mechanism that *stops* the fallback once an ARM64 artifact does appear.
- Or do a cheap `HEAD` on the resolved release URL and fall back on 404.

Either way, please make `ide upgrade` (no `--mode`) work, or state explicitly in the PR body that this PR only addresses the snapshot track and that #1981 stays open for releases.

Separately: this hardcodes the fully-resolved classifier string although the `os` and `arch` enums are right there in scope two lines up. Checking `(os == OperatingSystem.WINDOWS) && (arch == SystemArchitecture.ARM64)` expresses the intent directly, survives any future change to the classifier template, and removes the need for the two `CLASSIFIER_WINDOWS_*` string constants.

maybeec (commented):
Thanks for picking this one up 👍 The diagnosis matches the analysis in #1981 exactly — read the snapshot `maven-metadata.xml`, check which classifiers actually exist, and fall back — and reusing `CachedValue` for the parsed `Document` is the right building block. Commits follow `#«issue-id»: «summary»`, the CHANGELOG entry sits under the correct upcoming milestone, and the fallback is deliberately kept narrow so Linux/macOS resolution is untouched. 👍

No blockers, so this is a **comment** review rather than request-changes. There is however one real scope gap and a chunk of logic that duplicates helpers `MvnArtifact` already provides.

### Should-fix

1. **The reported command is still broken.** `getSnapshotBaseVersion(...)` returns `null` for release versions, so the fallback only ever applies to `--mode=snapshot`. Plain `ide upgrade` on a non-SNAPSHOT installation resolves `UpgradeMode.STABLE`/`UNSTABLE` (`IdeasyCommandlet#getConfiguredVersion`), which yields a release version — and that path still produces the 404 from the issue. The issue's reproduction is "1. Install IDEasy on Windows on ARM. 2. `ide upgrade`", so as it stands `fixes #1981` is only half true. See the inline comment on `getMetadata`.
2. **Duplicated snapshot-version logic.** `PATTERN_TIMESTAMPED_SNAPSHOT` and `getSnapshotBaseVersion(String)` re-implement `MvnArtifact.SNAPSHOT_VERSION_PATTERN` / `MvnArtifact#isSnapshot()` / `MvnArtifact#getBaseVersion()`, which already exist and are already exercised by the download-URL path. Both new members can be deleted — inline suggestion below.
3. **The actual bug path has no test.** All three new tests call the package-private `resolveSnapshotClassifier(Document, String, String)` directly. The wiring in `getMetadata` and the metadata-URL derivation are untested, which is why coverage moved only +0.02%. Per the TDD rule in `AGENTS.md` the first test should be the one that reproduces the bug — and `SystemInfoMock` + `IdeTestContext#setSystemInfo` + `MvnRepositoryMock` (WireMock-backed, already serves `maven-metadata.xml` from test resources) make that straightforward.
4. **Probing metadata is not best-effort.** If the extra `maven-metadata.xml` fetch fails, `fetchXmlMetadata` throws `CliException("Failed to determine the latest version from ...")` and aborts the whole upgrade with a message that no longer describes what was happening. It also means Windows ARM64 now *always* needs network in `getMetadata`, even when the artifact is already in the local `.m2` repo.
5. **Manual QA.** CI has no windows-arm64 runner, so green checks prove nothing here. The `-Dos.arch=arm64` simulation in the testing instructions exercises resolution but not actually running the x64 binary under emulation — this needs one real pass on Windows on ARM before merge.

### PR hygiene

- The branch is **BEHIND** `main` (`mergeStateStatus: BEHIND`) — please update it before merge.
- Neither the PR nor #1981 has a milestone set; per `documentation/contributing/DoD.adoc` that needs to be set before the issue is closed.
- Minor: the CHANGELOG line restates the bug title ("ide update fails on windows@arm64") rather than the fix. Most neighbouring entries read as what changed, e.g. `Fall back to windows-x64 when no windows-arm64 artifact is published`.

Details and concrete suggestions inline.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -112,26 +139,186 @@ public MvnArtifactMetadata getMetadata(MvnArtifact artifact, String tool, String
...
+    }
+
+    MvnArtifact metadataArtifact = artifact
+        .withVersion(snapshotVersion)
+        .withMavenMetadata();
+
+    String metadataUrl = getMavenUrl(metadataArtifact);
+    Document metadata = fetchXmlMetadata(metadataUrl);
+
+    return resolveSnapshotClassifier(metadata, classifier, extension);
```
maybeec: **Should-fix** — `MvnArtifact` already provides both of these.

- `MvnArtifact#isSnapshot()` (`MvnArtifact.java:239`) is exactly the `getSnapshotBaseVersion(...) != null` test — it matches the same timestamped pattern *and* a literal `-SNAPSHOT` suffix.
- `MvnArtifact#getPath()` already routes through `getBaseVersion()` (`MvnArtifact.java:249`), which turns `2026.05.001-20260527.032443-21` into `2026.05.001-SNAPSHOT`, and `getMvnBaseUrl()` already picks `MAVEN_SNAPSHOTS` for snapshots. So `artifact.withMavenMetadata()` on its own yields exactly the URL you are building here — the explicit `withVersion(snapshotVersion)` is a no-op.

The convention (`coding-conventions.adoc` § Refactorings / Clean Code) is to move logic up or reuse it, never to duplicate it — two copies of the same snapshot-version regex will drift.

```suggestion
    // classifier availability is only listed in the version-specific metadata of maven snapshots
    if (!artifact.isSnapshot()) {
      return classifier;
    }

    String metadataUrl = getMavenUrl(artifact.withMavenMetadata());
    Document metadata = fetchXmlMetadata(metadataUrl);

    return resolveSnapshotClassifier(metadata, classifier, extension);
```

With this, `PATTERN_TIMESTAMPED_SNAPSHOT` (line 78) and `getSnapshotBaseVersion` (line 254) can both be deleted, along with `testGetSnapshotBaseVersion`.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -112,26 +139,186 @@ public MvnArtifactMetadata getMetadata(MvnArtifact artifact, String tool, String
...
+
+  /**
+   * Converts a resolved timestamped snapshot version back to its Maven snapshot
+   * base version.
+   *
+   * @param version the resolved artifact version.
+   * @return the corresponding {@code -SNAPSHOT} version, or {@code null} if the
+   *         version is not a timestamped snapshot.
+   */
+  String getSnapshotBaseVersion(String version) {
```
maybeec: **Should-fix** — duplicate of `MvnArtifact#getBaseVersion()` (`MvnArtifact.java:249`), which does the identical transformation with the identical regex (`MvnArtifact.SNAPSHOT_VERSION_PATTERN`, line 32). The only difference is the `null` return for non-snapshots, which `MvnArtifact#isSnapshot()` covers. Please drop this method and `PATTERN_TIMESTAMPED_SNAPSHOT` — see the suggestion on `resolveSnapshotClassifier` above.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -112,26 +139,186 @@ public MvnArtifactMetadata getMetadata(MvnArtifact artifact, String tool, String
...
+    if (snapshotVersion == null) {
+      return classifier;
+    }
+
+    MvnArtifact metadataArtifact = artifact
+        .withVersion(snapshotVersion)
+        .withMavenMetadata();
+
+    String metadataUrl = getMavenUrl(metadataArtifact);
+    Document metadata = fetchXmlMetadata(metadataUrl);
```
maybeec: **Should-fix** — this probe should be best-effort.

`fetchXmlMetadata` propagates failures as `CliException("Failed to determine the latest version from " + url, e)`. On this code path we are not determining a version at all, we are checking classifier availability — so any hiccup (404 on the metadata, proxy/TLS problem, offline mode) now aborts the whole upgrade with a message that misdescribes what happened, where before the user at least got the honest "Failed to download ... 404".

The consequence in offline mode is concrete: `NetworkStatusImpl#invokeNetworkTask` throws `CliOfflineException` up front (`NetworkStatusImpl.java:129`), so on Windows ARM64 `getMetadata` now requires network even when the artifact is already sitting in the local `.m2` repository — a case that used to work.

Suggest wrapping the probe and degrading gracefully:

```java
    try {
      Document metadata = fetchXmlMetadata(getMavenUrl(artifact.withMavenMetadata()));
      return resolveSnapshotClassifier(metadata, classifier, extension);
      LOG.debug("Failed to inspect snapshot metadata from {} - keeping classifier {}", metadataUrl, classifier, e);
      return classifier;
    }
```

Note the exception object is passed as the last argument so the cause survives, per `coding-conventions.adoc` § Catching and handling Exceptions.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -112,26 +139,186 @@ public MvnArtifactMetadata getMetadata(MvnArtifact artifact, String tool, String
...
-      artifact = artifact.withClassifier(resolvedClassifier);
+
+      if (CLASSIFIER_WINDOWS_ARM64.equals(classifier)) {
+        classifier = resolveSnapshotClassifier(
+            artifact,
+            classifier,
+            artifact.getType());
+      }
+
+      artifact = artifact.withClassifier(classifier);
```
maybeec, line 160: **Minor** — once the classifier falls back to `windows-x64`, the `arch` local is still `ARM64`, so the returned `MvnArtifactMetadata#getArch()` disagrees with the artifact it describes.

I checked and this is inert *today*: the only production consumer is `AbstractToolRepository#doDownload` (`AbstractToolRepository.java:93`, building the download-cache filename), and `MvnRepository` overrides `download(UrlDownloadFileMetadata)` to route to `getDownloadedArtifact` instead. But it is a trap for whoever wires the next consumer — they would get an `arm64`-named cache entry holding x64 bytes, which `Files.exists(target)` would then happily serve forever even after a real ARM64 artifact ships.

Cheapest fix: set `arch = SystemArchitecture.X64;` in the same branch that performs the fallback, so the metadata stays self-consistent.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -112,26 +139,186 @@ public MvnArtifactMetadata getMetadata(MvnArtifact artifact, String tool, String
...
+      String extension) {
+
+    if (hasSnapshotArtifact(metadata, classifier, extension)) {
+      return classifier;
+    }
+
+    if (CLASSIFIER_WINDOWS_ARM64.equals(classifier)
+        && hasSnapshotArtifact(metadata, CLASSIFIER_WINDOWS_X64, extension)) {
+
+      LOG.warn(
```
maybeec: **Minor** — `coding-conventions.adoc` § Log-Messages asks for contextual information rather than a bare statement. This message tells the user something unexpected happened but not what they got or why. Something like:

```suggestion
      LOG.warn("No {} artifact is published for version {} - falling back to {} (see https://github.com/devonfw/IDEasy/issues/599).",
          CLASSIFIER_WINDOWS_ARM64, artifact.getVersion(), CLASSIFIER_WINDOWS_X64);
```

(the `{}` placeholder style is the documented convention — § Log-message parameters). That requires threading the version into this overload, or emitting the warning from the caller instead.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -112,26 +139,186 @@ public MvnArtifactMetadata getMetadata(MvnArtifact artifact, String tool, String
...
+  }
+
+  /**
+   * Gets the text content of a direct child node.
+   *
+   * @param parent the parent node.
+   * @param childName the child-node name.
+   * @return the child text, or {@code null} if the child does not exist.
+   */
+  private String getChildText(Node parent, String childName) {
```
maybeec: **Minor** — this class already has a child-lookup helper: `getFirstChildElement(Element, String, Object)` (line 451). Two differences worth noting:

- `getFirstChildElement` filters on `node instanceof Element`; `getChildText` walks *all* child nodes including text/comment nodes and compares `getNodeName()`, which happens to work here only because no text node can be named `classifier`.
- Having two near-identical DOM walkers in one class is exactly the duplication `coding-conventions.adoc` § Clean Code warns about.

Suggest extracting a nullable `findFirstChildElement(Element, String)` and letting the existing throwing variant delegate to it, then reading `.getTextContent().trim()` at the call site.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -112,26 +139,186 @@ public MvnArtifactMetadata getMetadata(MvnArtifact artifact, String tool, String
...
+   * @param classifier the artifact classifier.
+   * @param extension the artifact extension.
+   * @return {@code true} if the artifact is present.
+   */
+  private boolean hasSnapshotArtifact(
+      Document metadata,
+      String classifier,
+      String extension) {
+
+    NodeList snapshotVersions =
```
maybeec: **Minor** — `"snapshotVersion"`, `"classifier"` and `"extension"` are the Maven metadata schema, repeated across two methods. Per `coding-conventions.adoc` § Constants these belong in `private static final String` constants with a short JavaDoc, alongside the `"versioning"`/`"snapshot"` literals already used in `resolveSnapshotVersion` — which arguably deserve the same treatment while you are here.

Also: `metadata.getElementsByTagName(...)` searches the entire document, whereas the rest of this class deliberately navigates `versioning` → `snapshot` step by step. Navigating `versioning` → `snapshotVersions` would keep the two readers consistent and would not silently match a `snapshotVersion` element that appeared somewhere unexpected.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -278,15 +465,32 @@ private Element getFirstChildElement(Element element, String tag, Object source)
...
+
+  private Document downloadXmlMetadata(String url) {
+
     try {
       return this.context.getNetworkStatus().invokeNetworkTask(() -> {
         URL xmlUrl = new URL(url);
+
         try (InputStream is = xmlUrl.openStream()) {
-          return documentBuilder.parse(is);
+          synchronized (this.documentBuilder) {
```
maybeec: **Minor** — the concurrency story here is half-applied, which is more confusing than either extreme.

`ConcurrentHashMap` + `synchronized (this.documentBuilder)` implies concurrent access is expected, but `CachedValue` itself is not thread-safe: `value` and `timestamp` are plain non-volatile fields and `get()` is unsynchronized (`CachedValue.java:55`). So under real concurrency the supplier can still run twice and a partially-published `Document` can still be observed — the guard on the builder does not close that.

Either concurrency is real, in which case the caching needs to be guarded too (or use `ThreadLocal<DocumentBuilder>` / a fresh builder per parse, which is cheap and sidesteps the fact that `DocumentBuilder` also wants `reset()` between parses), or it is not, in which case the `synchronized` block is unexplained noise in an otherwise mechanical refactoring. A one-line comment stating which it is would be enough for me.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -54,10 +63,27 @@ public class MvnRepository extends ArtifactToolRepository<MvnArtifact, MvnArtifa
 
   private static final Duration METADATA_CACHE_DURATION_SNAPSHOT = Duration.ofMinutes(5);
 
+  private static final long XML_METADATA_CACHE_RETENTION =
```
maybeec: **Minor** — the two sibling constants immediately above are `Duration`; this one converts to `long` millis at the declaration, so it reads differently from its neighbours for no reason. `CachedValue` wants millis only at the constructor call:

```suggestion
  private static final Duration XML_METADATA_CACHE_DURATION = Duration.ofMinutes(1);
```

and then `XML_METADATA_CACHE_DURATION.toMillis()` at the `new CachedValue<>(...)` site.

Worth a sanity check on the value too: the on-disk metadata cache for snapshots is 5 minutes (`METADATA_CACHE_DURATION_SNAPSHOT`), so a 1-minute in-memory retention is stricter than the layer behind it. Within one CLI invocation that difference never materialises, but it invites the question — a short comment explaining that this cache exists to deduplicate fetches *within a single run* would settle it.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -112,26 +139,186 @@ public MvnArtifactMetadata getMetadata(MvnArtifact artifact, String tool, String
...
+   * can execute Windows x64 applications, so the x64 artifact is used as a
+   * fallback when the ARM64 artifact is unavailable.
+   *
+   * @param artifact the artifact being resolved.
+   * @param classifier the originally resolved classifier.
+   * @param extension the artifact extension, such as {@code tar.gz}.
+   * @return the original classifier if it exists or no suitable fallback is
+   *         available; otherwise {@code windows-x64}.
+   */
+  private String resolveSnapshotClassifier(
```
maybeec: **Minor** — formatting, and note that `mvn spotless:apply` will *not* fix this: the Spotless config in `pom.xml:246-261` only does import order, unused imports, trailing whitespace and final newline — line wrapping is left to the IntelliJ formatter driven by `.editorconfig`.

`.editorconfig` sets `max_line_length = 160` (checkstyle `LineLength` max 160 as well), but these signatures and several statements are wrapped at roughly 80, one parameter per line, which does not match the surrounding file (compare `resolveSnapshotVersion` / `getFirstChildElement`). The same applies to `hasSnapshotArtifact`, `getChildText`, `fetchXmlMetadata` and the `throw new CliException(...)` that got re-wrapped although it was untouched otherwise.

Running the IDE formatter with the project `.editorconfig` will collapse these and also drop the unrelated blank lines inserted into `getMetadata`, which currently make the diff larger than the change.

cli/src/test/java/com/devonfw/tools/ide/tool/mvn/MvnRepositoryTest.java
```diff
@@ -231,6 +233,74 @@ void testResolveSnapshotVersion() {
     assertThat(version).hasToString("2025.02.001-beta-20250204.023111-1");
   }
 
+  @Test
+  void testResolveSnapshotClassifierFallsBackFromWindowsArm64ToX64() {
```
maybeec, line 241: **Should-fix** — these three tests all call `resolveSnapshotClassifier(Document, ...)` directly, so the part that actually failed in #1981 is never executed: the `CLASSIFIER_WINDOWS_ARM64.equals(classifier)` guard in `getMetadata`, the metadata-URL derivation, and the resulting download URL. That matches the +0.02% coverage delta.

`AGENTS.md` asks for the bug-reproducing test first, and everything needed already exists:

- `SystemInfoMock` (`cli/src/test/java/com/devonfw/tools/ide/os/SystemInfoMock.java`) — add a `WINDOWS_ARM64 = new SystemInfoImpl("Windows 11", "10.0", "aarch64")` constant and register it in `MOCKS`;
- `IdeTestContext#setSystemInfo(...)`, used the same way as `EnvironmentCommandletTest.java:102`;
- `MvnRepositoryMock`, which already stubs `maven-metadata.xml` responses through WireMock from `repository/mvn/**` test resources and overrides `getMavenUrl`.

So the missing test is roughly: set `SystemInfoMock.WINDOWS_ARM64`, place a snapshot `maven-metadata.xml` without a `windows-arm64` entry in the mock repo, call `getMetadata("ideasy", "ideasy", <timestamped snapshot>, null)` and assert `getUrls()` ends in `-windows-x64.tar.gz`. That is OS-independent and network-free, per `documentation/contributing/junit-testing.adoc`.

There is also a side effect worth guarding against: because the fallback now runs inside `getMetadata`, a contributor on a Windows ARM64 machine running `mvn clean test` will make `testGetMetadataWithRelease`/`testGetMetadataWithSnapshot` perform a live fetch against `central.sonatype.com` (and fail offline). Pinning the `SystemInfo` in those two tests instead of reading the host's would make them deterministic everywhere.

cli/src/test/java/com/devonfw/tools/ide/tool/mvn/MvnRepositoryTest.java
```diff
@@ -260,4 +330,24 @@ private static Document parseXml(String xml) {
     }
   }
 
+  @Test
```
maybeec, line 360: **Minor** — two small things:

- This test sits *after* the `parseXml` helper at the very bottom of the class; the other tests are grouped above it. Please move it up with the rest.
- The four new test methods have no JavaDoc, while the pre-existing ones in this file carry `/** Test of {@link MvnRepository#...}. */` (lines 218 and 304). Keeping that consistent is what makes the class readable as a spec (`coding-conventions.adoc` § Code-Documentation).

If you adopt the `MvnArtifact#getBaseVersion()` reuse suggested on `MvnRepository.java`, this test goes away entirely — `MvnArtifact` currently has no test for `getBaseVersion()`, so moving these three assertions into an `MvnArtifactTest` would be a genuine net win.

cap-juan:
Automated tests cover Windows ARM64 artifact resolution and x64 fallback. I could not perform a runtime validation on physical Windows ARM64 hardware, so execution under Windows-on-ARM emulation still needs manual verification before merge.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnArtifact.java
```diff
@@ -18,7 +18,7 @@ public class MvnArtifact extends SoftwareArtifact {
   public static final String ARTIFACT_ID_IDEASY_CLI = "ide-cli";
 
   /** {@link #getClassifier() Classifier} of source code. */
-  public static final String CLASSIFER_SOURCES = "sources";
+  public static final String CLASSIFIER_SOURCES = "sources";
```
hohwille, line 21: Thanks for fixing typo 👍

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -261,32 +417,56 @@ public Path download(MvnArtifactMetadata metadata) {
     return getDownloadedArtifact(metadata.getMvnArtifact(), metadata.getChecksums());
   }
 
-  private Element getFirstChildElement(Element element, String tag, Object source) {
+  private Element findFirstChildElement(Element element, String tag) {
 
     NodeList children = element.getChildNodes();
-    int length = children.getLength();
-    for (int i = 0; i < length; i++) {
+    for (int i = 0; i < children.getLength(); i++) {
```
hohwille: FYI this coding style was done on purpose:
https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc#consider-extracting-local-variable-for-multiple-method-calls

So with your change `chidlren.getLenght()` is called N times instead of just once.
It will not be a big performance issue but when you remove good patterns from the code, I want to explain that we do such things for reason. And you will not always deep dive into the implementation of `getLength()` to see how expensive the method may actually be.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -112,26 +124,170 @@ public MvnArtifactMetadata getMetadata(MvnArtifact artifact, String tool, String
...
+   * fallback when the ARM64 artifact is unavailable.
+   *
+   * @param artifact the artifact being resolved.
+   * @param classifier the originally resolved classifier.
+   * @param extension the artifact extension, such as {@code tar.gz}.
+   * @return the original classifier if it exists or no suitable fallback is available; otherwise {@code windows-x64}.
+   */
+  private String resolveWindowsArm64Classifier(MvnArtifact artifact, String classifier, String extension) {
+
+    if (!artifact.isSnapshot()) {
```
hohwille, line 183: why don't you want to fix this also for release upgrades but only for SNAPSHOTs?
Or if I got this wrong and you always fallback to windows_x64 for releases what if IDEasy fixes #599?
What if this code is used for something else than IDEasy?
This does not seem correct for me.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
```diff
@@ -112,26 +124,170 @@ public MvnArtifactMetadata getMetadata(MvnArtifact artifact, String tool, String
...
+
+      if (CLASSIFIER_WINDOWS_ARM64.equals(classifier)) {
+        classifier = resolveWindowsArm64Classifier(
+            artifact,
+            classifier,
+            artifact.getType());
+
+        if (CLASSIFIER_WINDOWS_X64.equals(classifier)) {
+          arch = SystemArchitecture.X64;
+        }
```
hohwille: You have created a very special fix for IDEasy on Windows only.
My thought would have been to fix this entirely for MvnRepository:
1. Try to retrieve artifact as planned
2. If that fails and architecture was arm64 and actually part of classifier, simply retry the same with x64 instead of arm64. This will then also fix problems on MacOS and will also work for tools other than IDEasy if they are retrieved via MvnRepository.

WDYT?

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
hohwille: Update: I saw you are using the XML metadata for SNAPSHOT versions to see if the artifact actually exists.
My idea was if the download fails, then consider the retry if criteria meet.

hohwille (approved):
@cap-juan thanks for your PR. You have done good changes to solve the IDEasy upgrading of SNAPSHOTS 👍 
I should have been more explicit in the story description giving a rough idea how to fix this.
With all the team review rounds now my feedback is quite late to question the general approach. Please have a look and see if all my suggestions make sense for you. My general approach is not to fight symptoms but to always fix issues at their very root.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MvnRepository.java
cap-juan: Thanks, I adjusted this accordingly. SNAPSHOTs still use Maven metadata to check whether the ARM64 artifact exists. For releases, the ARM64 artifact is now tried first, and only if the download fails with a 404 and the Windows ARM64 fallback criteria match, the x64 artifact is retried. This also means a future native ARM64 release (e.g. after #599) will be used directly instead of being forced to x64.

hohwille (approved):
@cap-juan thanks for your excellent rework. I also very much like your code-style how you apply clean code and extract small private methods with self-explaining names making the code and flow very readable. Great job 👍

cap-juan:
Successfully tested with 2026.09.002 on Windows 11 Pro.

If no arm is found then it will default to x64.

hohwille:
Works like a charm on my Windows@ARM laptop:
```
$ ide upgrade
No CVEs found for version 2026.09.002-20260922.073732-13 of tool ideasy.
Artifact with classifier windows-arm64 is unavailable for version 2026.09.002-20260922.073732-13 - falling back to windows-x64.
Trying to download ide-cli-2026.09.002-20260922.073732-13-windows-x64.tar.gz from https://central.sonatype.com/repository/maven-snapshots/com/devonfw/tools/IDEasy/ide-cli/2026.09.002-SNAPSHOT/ide-cli-2026.09.002-20260922.073732-13-windows-x64.tar.gz
...
```

PR #2262, opened 2026-08-04
#2255: Extract MSI build logic into shell script

### LINKED ISSUE #2255: Refactor build logic from GitHub Workflows into dedicated shell scripts
### Feature idea

During recent discussions, we identified that longer, functionally related Bash command sequences should not be implemented directly inside GitHub workflows whenever possible.

Complex build steps are difficult to develop, test, and maintain when embedded in workflow YAML files. Moving the logic into dedicated *.sh scripts improves:

- Maintainability
- Local testing capabilities
- Readability of workflows
- Reusability of build logic
- Easier troubleshooting and future modifications

### Scope

Extract build-specific logic from workflows into dedicated shell scripts and invoke them from the workflow instead.

Initial candidates:

- macOS PKG build
- Windows MSI build

The GitHub workflow should only orchestrate the process and call the corresponding script.

### Additional context

_No response_

Caylipp:
### This PR fixes #2255

No macOS PKG build exists in the current workflows, so this PR only extracts the Windows MSI build.

### Implemented changes:

- Add `windows-installer/build-msi.sh`
- Replace the inline MSI build logic in `nightly-build.yml`
- Replace the inline MSI build logic in `release.yml`
- Keep the workflows focused on orchestration

---

### Testing instructions

1. Validated the shell script syntax with:
bash -n windows-installer/build-msi.sh

---

### Checklist for this PR

  with `internal`


QuangAnhLe:
Hi @Caylipp, thanks for contribution, a few things that stand: 

what 's good: :)
 1. Correct refactor for DRY principle. The exact same 15-line block was duplicated in both workflows. Calling ./windows-installer/build-msi.sh from both is the right move.
  2. Proper set -euo pipefail. Fail-fast, no undefined variables, pipe failures caught.
  3. Directory resolution via BASH_SOURCE[0]. The script correctly resolves the repo root relative to its own location, so it works regardless of the caller's working
  directory.
  4. Workflow YAML is significantly cleaner. From 15 lines of inline script to a single run: ./windows-installer/build-msi.sh.
  5. -f documentation/pom.xml is more robust than cd documentation && mvn ... && cd .. since it doesn't depend on the working directory.
The two substantive items are:
  1. The * → . copy behavior change (likely harmless but worth a quick verification that no dotfiles in cli/target/package/ should be excluded).
  2. Non-idempotent dotnet tool install — not a blocker for CI (fresh runners each run), but makes the script harder to run locally for iteration. If the goal is developer-friendliness, adding idempotency guards would help.

Caylipp:
> Hi @Caylipp, thanks for contribution, a few things that stand:
> 
> what 's good: :)
> 
> 1. Correct refactor for DRY principle. The exact same 15-line block was duplicated in both workflows. Calling ./windows-installer/build-msi.sh from both is the right move.
> 2. Proper set -euo pipefail. Fail-fast, no undefined variables, pipe failures caught.
> 3. Directory resolution via BASH_SOURCE[0]. The script correctly resolves the repo root relative to its own location, so it works regardless of the caller's working
>    directory.
> 4. Workflow YAML is significantly cleaner. From 15 lines of inline script to a single run: ./windows-installer/build-msi.sh.
> 5. -f documentation/pom.xml is more robust than cd documentation && mvn ... && cd .. since it doesn't depend on the working directory.
>    The two substantive items are:
> 6. The * → . copy behavior change (likely harmless but worth a quick verification that no dotfiles in cli/target/package/ should be excluded).
> 7. Non-idempotent dotnet tool install — not a blocker for CI (fresh runners each run), but makes the script harder to run locally for iteration. If the goal is developer-friendliness, adding idempotency guards would help.

Thanks! I restored the original copy behavior and added a guard to avoid reinstalling WiX when it is already available.

QuangAnhLe:
add CHANGELOG.adoc

windows-installer/build-msi.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+repository_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
```
hohwille: Where does this `$(dirname "${BASH_SOURCE[0]}")` come from (I already saw this in multiple different PR reviews)? Did you copy that from our `functions`?
When coding bash you need to distinguish what is a script and what is a file to be sourced:
* script: `./script.sh`
* sourcing: `. functions` or `source functions`

This is a script and it is called as a script from the workflow.
Therefore to get the location of this script itself we use `$0` in bash.
so to get the `windows-installer` directory, you could do `$(dirname "$0")`.
BTW: `dirname` is an external program so when using this you actually launch a new process what causes overhead. You can also use bash build-in features instead of `dirname` or `basename`:
```
$ echo $0
/usr/bin/bash

$ echo ${0/*\//}
bash

$ echo ${0%/*}
/usr/bin
```

If you have something that is to be sourced, then unfortunately `$0` remains untouched as the value from the parent script or shell sourcing the file.
In bash this can somehow be resolved via `${BASH_SOURCE[0]}`.

windows-installer/build-msi.sh
```diff
@@ -0,0 +1,32 @@
...
+cd "${repository_dir}"
```
hohwille, line 6: This is kind of wired.
The only purpose of the variable `repository_dir` is that you navigate to it via `cd`.
So instead of doing
```
variable="$(cd "...something..." && pwd)"
cd "${variable}"
```
It would make more sense to just do
```
cd "...something..."
```
I always get confused by such code and start thinking why is it written so complicated? There must be some reason...

Next this "...something..." is actually `"$(dirname "${BASH_SOURCE[0]}")/.."`.

Where does this `$(dirname "${BASH_SOURCE[0]}")` come from (I already saw this in multiple different PR reviews)? Did you copy that from our `functions`?
It will IMHO always evaluate to `.` on github so what is it good for?

When coding bash you need to distinguish what is a script and what is a file to be sourced:
* script: `./script.sh`
* sourcing: `. functions` or `source functions`

This is a script and it is called as a script from the workflow.
Therefore to get the location of this script itself we use `$0` in bash.
so to get the `windows-installer` directory, you could do `$(dirname "$0")`.
BTW: `dirname` is an external program so when using this you actually launch a new process what causes overhead. You can also use bash build-in features instead of `dirname` or `basename`:
```
$ echo $0
/usr/bin/bash

$ echo ${0/*\//}
bash

$ echo ${0%/*}
/usr/bin
```

If you have something that is to be sourced, then unfortunately `$0` remains untouched as the value from the parent script or shell sourcing the file.
In bash this can somehow be resolved via `${BASH_SOURCE[0]}`.

hohwille: BTW: It seems your script must run in the root of our IDEasy git clone.
Therefore I was wrong assuming that you want to navigate into the folder where this script is located (`windows-installer`).
So IMHO if I conclude correctly this `cd` command actually has no effect at all (`cd $PWD`).

hohwille (approved):
@Caylipp / @QuangAnhLe thanks for this PR. This should work fine and cleanup redundancies. 👍 
I left some comment for some code I found very strange.

windows-installer/build-msi.sh
hohwille: @QuangAnhLe how do we proceed here? We could have already merged this meanwhile...
Can you just remove these two lines that only cause confusion but no effect so I can merge this PR?

QuangAnhLe: @hohwille Thanks for following up. Since this PR belongs to Philipp, I cannot push the requested changes to the PR branch myself. If the two lines have no functional effect, please go ahead and merge this PR. I will create a separate refactoring PR afterward to remove them and clean up the script.

hohwille:
```suggestion
```
Since you are not reacting I am taking over.

PR #2266, opened 2026-08-04
#2264: Set environment variables of all installed tools when running a tool

### LINKED ISSUE #2264: ide «tool» does not set the environment variables of the other installed tools (unlike ide standalone)
### Actual behavior

Running a tool through the `ide` prefix does **not** set up the environment variables of the other installed tools of the project. Only the invoked tool and its declared dependencies get their `setEnvironment` applied.

Concrete symptom: `ide npm run tauri:dev` fails because the rust/cargo environment variables (`CARGO_HOME`, `RUSTUP_HOME`) are missing in the child process, although rust is installed via IDEasy and `cargo` *is* found on the `PATH` (all `software/*/bin` folders are collected into `SystemPath`). The rustup proxies therefore fall back to `~/.cargo` / `~/.rustup`, which do not exist.

Two aggravating effects:

1. **Re-running the command does not help.** The `ide` shell function returns early on a non-zero exit code (`cli/src/main/package/functions:26-41`), i.e. the `eval "$(ideasy env --bash)"` at line 41 is unreachable once the command failed. So a failing `ide «cmd»` never refreshes the shell environment and the next attempt starts from the same incomplete environment.
2. **Running `ide` standalone once fixes it.** `ide` without arguments goes straight to `ideasy env --bash`, which exports the full environment (including `CARGO_HOME`/`RUSTUP_HOME`) into the interactive shell. Afterwards `ideasy npm …` inherits those variables from the parent process, which masks the defect.

So the environment is computed by **two different code paths** that do not agree:

| | `ide` (no args) → `ideasy env` | `ide «tool» «args»` |
|---|---|---|
| Env source | iterates over **all installed local tools** and calls `setEnvironment()` on each | exported `ide.properties`-layer variables + `SystemPath` + `setEnvironment()` of **only the invoked tool and its dependencies** |
| Result | `CARGO_HOME`, `RUSTUP_HOME`, `JAVA_HOME`, `UV_TOOL_DIR`, … | only the invoked tool's own variables |

Relevant code:

* `EnvironmentCommandlet.setEnvironmentVariablesInLocalTools` (`cli/src/main/java/com/devonfw/tools/ide/commandlet/EnvironmentCommandlet.java:142-159`) loops over every `LocalToolCommandlet` and calls `tool.setEnvironment(...)` for each installed tool.
* `ProcessContextImpl` (`cli/src/main/java/com/devonfw/tools/ide/process/ProcessContextImpl.java:73-78`) seeds the process environment **only** from `getVariables().collectExportedVariables()`, and line 193 sets `PATH` from `SystemPath`. Nothing applies `setEnvironment()` of the other installed tools.
* The invoked tool's own variables are added later via `ToolCommandlet.createToolInstallation` → `setEnvironment` (`ToolCommandlet.java:599`), its dependencies via `LocalToolCommandlet.installToolDependencies` (`LocalToolCommandlet.java:322-335`). Since npm declares node (not rust) as dependency, `Rust.setEnvironment` (`cli/src/main/java/com/devonfw/tools/ide/tool/rust/Rust.java:89-95`) is never called.

This affects every tool that needs another tool's environment variables without declaring it as dependency (rust/cargo from npm/tauri, `CATALINA_HOME`, `UV_TOOL_DIR`, `JAVA_HOME` for arbitrary consumers, custom scripts run via `ide`). It currently works only when the user's shell happens to carry a fresh `ideasy env` snapshot — so fresh installs, newly installed tools in an already-open terminal, and non-interactive/CI invocations are hit hardest.

### Reproduce

Original scenario:

1. Project with rust and node installed via IDEasy, rust installed *after* the current terminal was opened (so the shell env has no `CARGO_HOME`).
2. `ide npm run tauri:dev` → fails, cargo/rustup cannot find their home directories.
3. Re-run `ide npm run tauri:dev` → still fails.
4. Run `ide` (standalone), then `ide npm run tauri:dev` → works.

Minimal reproducer without rust (any project with node and java installed), showing the two code paths disagree:

```bash
$ ideasy env --bash | grep _HOME
export JAVA_HOME="/d/projects/my-project/software/java"
export MVN_HOME="/d/projects/my-project/software/mvn"
export NODE_HOME="/d/projects/my-project/software/node"

# scrub the inherited variables to see what IDEasy itself provides to the child process
$ env -u JAVA_HOME -u MVN_HOME ideasy node -e "console.log('JAVA_HOME=' + process.env.JAVA_HOME); console.log('NODE_HOME=' + process.env.NODE_HOME); console.log('MVN_HOME=' + process.env.MVN_HOME)"
JAVA_HOME=undefined
NODE_HOME=D:\projects\my-project\software\node
MVN_HOME=undefined
```

### Expected behavior

`ide «tool» «args»` should initialize the environment exactly like `ide` standalone does: the child process should receive the environment variables of **all** installed tools of the project (with the invoked tool and its dependencies taking precedence), so that a tool such as npm/tauri can call `cargo` with a correct `CARGO_HOME`/`RUSTUP_HOME`.

Additionally, a failing `ide «cmd»` should still refresh the shell environment so that a retry is not stuck with the previous (incomplete) environment.

### IDEasy status

```bash
IDE_ROOT is set to D:\projects
IDE_HOME is set to D:\projects\my-project
Your version of IDEasy is 2026.07.002.
You are online.
```

### Comments/Hints

Suggested fix, in two parts:

1. **Single source of truth for the environment.** Extract `EnvironmentCommandlet.setEnvironmentVariablesInLocalTools` into shared code and apply it when the `ProcessContext` for a tool run is created (`AbstractIdeContext.newProcess()` / `createProcessContext()`, `AbstractIdeContext.java:951-976`). Both `EnvironmentVariableCollectorContext` and `ProcessContextImpl` implement `EnvironmentContext`, so the same logic can serve both paths. The invoked tool and its dependencies already run their `setEnvironment` afterwards and would therefore correctly override. Things to watch: the cost of the `isInstalled()` scan over all commandlets (the `env` commandlet already pays it) and keeping the "installed version is never read by `setEnvironment`" assumption documented (see `EnvironmentCommandlet.java:148`).
2. **`cli/src/main/package/functions`:** capture the return code, always run the `eval "$(ideasy env --bash)"`, and return the code afterwards, so a failed command still leaves a correct shell environment.

Following the project's TDD rule, step one would be a failing test that runs one tool commandlet in a project where another tool (e.g. rust) is installed and asserts that the other tool's variables are present on the process context.

maybeec:
### This PR fixes #2264

Running a tool via `ide «tool» «args»` did not provide the environment variables of the other installed tools of the project - only the invoked tool and its declared dependencies got their `setEnvironment` applied. The environment was therefore computed by two disagreeing code paths (`ideasy env` iterated over all installed tools, the tool run did not). Symptom from the issue: `ide npm run tauri:dev` failed because `CARGO_HOME`/`RUSTUP_HOME` were missing although rust is installed via IDEasy.

### Implemented changes:

* Extracted `EnvironmentCommandlet.setEnvironmentVariablesInLocalTools` into `IdeContext.setEnvironmentOfInstalledTools(EnvironmentContext)` (implemented in `AbstractIdeContext`) as single source of truth for the tool environment - used by `ideasy env` as well as by tool runs.
* `ToolCommandlet` now seeds every newly created tool `ProcessContext` with the environment variables of all installed tools (new private `newToolProcess(...)`, used by `runTool(...)` and `completeRequestProcessContext(...)`). The invoked tool and its dependencies apply their own `setEnvironment` afterwards and therefore still take precedence.
* The extracted method gates on the existence of the tool installation directory instead of `isInstalled()`. `isInstalled()` spawns processes for some tools (e.g. `Npm.computeInstalledVersion` runs `npm --version`, npm-based commandlets run `npm list -g «package»`), which was tolerable for the rarely used `env` commandlet but would add seconds of latency to *every* `ide «tool»` call. Side effect: a leftover `software/«tool»` folder without version file now also contributes its variables - consistent with how `SystemPath` already treats those folders.
* `cli/src/main/package/functions`: a failing `ide «cmd»` no longer returns early. The shell environment is always refreshed via `eval "$(ideasy env --bash)"` and the original exit code is returned at the end, so a retry no longer starts from the same incomplete environment.
* Test: `TomcatTest.testTomcatHasEnvironmentOfOtherInstalledTools` - the tomcat test project now has rust pre-installed and `catalina.sh`/`catalina.bat` echo `CARGO_HOME`. Written first and verified to fail before the fix (`CARGO_HOME=`). The pre-existing `testTomcat` guards that precedence is unchanged (tomcat still gets the tomcat-specific `JAVA_HOME` of java `21_35`, not the project version `8u402b06`).

---

### Testing instructions

Automated:

1. `mvn -Dtest=TomcatTest test` in the `cli` folder - `testTomcatHasEnvironmentOfOtherInstalledTools` asserts that `CARGO_HOME` of the installed rust points into `software/rust/.cargo` although tomcat only depends on java.
2. `mvn -Dtest=EnvironmentCommandletTest,EnvironmentCommandletGlobalTest test` in the `cli` folder - the output of `ideasy env` is unchanged.

Manual (the original scenario):

1. Use a project with rust and node installed via IDEasy.
2. In a fresh terminal, verify that the child process receives the rust variables:
   `env -u CARGO_HOME -u RUSTUP_HOME ideasy node -e "console.log(process.env.CARGO_HOME, process.env.RUSTUP_HOME)"`
   Both must point into `«IDE_HOME»/software/rust` (before this change they were `undefined`).
3. Run a failing command such as `ide npm run does-not-exist` and check that the environment of the shell is still refreshed afterwards (e.g. `echo $CARGO_HOME` is set) while `echo $?` right after the `ide` call still reports the non-zero exit code.

---

### Checklist for this PR

  with `internal`

_Note: this PR does not add a new tool commandlet, so the commandlet checklist does not apply._


hohwille (approved):
@maybeec thanks for filing the issue and providing this PR as bugfix. Great job👍 
I was not fully sure if the change in `functions` is really a good idea. If something is fully broken in IDEasy and therefore every command fails it might even be harmful to continue executing and evaluating the env command.
However, reading the changes carefully I still think this is acceptable and I could not create a scenario where this causes a real problem.
Everything else is just perfect so I would included it in today's release.

PR #2282, opened 2026-08-06
#989: allow expressions in template variable definitions

### LINKED ISSUE #989: allow expressions in template variable definitions

### LINKED ISSUE #2165: prompt user for template variables at apply time

Paras14:
### This PR fixes #989

Implements #989 and supersedes the `$[ask:...]` / `$[secret:...]` syntax of #2179.

Closes #2165.

### Implemented changes

* Added new package `com.devonfw.tools.ide.expression` with the expression syntax `@«function-name»([«arg»[,«arg»]*])` resolved during variable resolution.
* `ExpressionFunction` - interface implemented by every function.
* `ExpressionFunctionManager` - registry to look up functions by name so further functions can be registered with new IDEasy releases.
* `ExpressionParser` - locates function calls and parses their arguments. A RegEx is only used to locate the start of a call, the argument list is scanned manually since a RegEx cannot express a balanced list of arguments containing quoted commas, quoted parenthesis or nested calls.
* `ExpressionContext` - gives a function access to the `IdeContext`, variable lookup and persistence.
* Added function `@path(«path»[, unix|native])` that normalises a path, by default replacing backslashes with slashes.
* Added functions `@ask-variable(«name»[, «question»[, «default»]])` and `@ask-secret(...)` with masked input. An already defined variable is returned without asking, an empty 1st argument always asks, and an empty string as 3rd argument permits empty input.
* Added functions `@if-windows`, `@if-mac`, `@if-linux` and `@if-unix` that insert their argument if the OS matches.
* Added `IdeContext.askForSecret(String, String)` analogous to `askForInput`. `AbstractIdeContext` implements the prompt loop and the batch mode contract and delegates reading to the new protected `readSecretLine()`, which `IdeContextConsole` overrides with `Console.readPassword()`.
* Values entered for a workspace template are persisted to `conf/ide.properties` so the question is only asked the first time. A value that could not be asked for in batch mode is not persisted.
* Arguments are resolved through the regular variable resolution, so a String argument may itself contain variables (e.g. `@path('$[IDE_HOME]/software/node')`).
* Text that does not call a registered function is left untouched.

---

### Testing instructions

Masked input needs a real console, so run this from a normal terminal and not from the IDE.

1. Build the branch and dump the classpath, from the repository root:

```
mvn -pl cli -am install -DskipTests
mvn -pl cli dependency:build-classpath "-Dmdep.outputFile=cp.txt"
$CP = (Get-Content cli\cp.txt -Raw).Trim()
$CLASSES = "C:\projects\IDEasy\workspaces\main\IDEasy\cli\target\classes"
```

_(make sure you add your path correctly for the cli target classes: "<repo>\cli\target\classes")_

2. In a test project create `IDEasy\settings\workspace\update\ai-test.properties`:

```
ai.backend.url=@ask-variable('MY_URL')
ai.api.key=@ask-secret('MY_TOKEN', 'Enter your API key (from the portal):')
ai.node.path=@path('$[IDE_HOME]/software/node/node', native)
ai.other=@ask-secret('MY_OPTIONAL', 'Password (may be empty):', '')
```

Make sure `MY_URL`, `MY_TOKEN` and `MY_OPTIONAL` are not yet defined in `IDEasy\conf\ide.properties`.

3. From the project directory, run the local build:

```
java -cp "$CLASSES;$CP" com.devonfw.tools.ide.cli.Ideasy update
```

4. You are asked three times. `MY_URL` is echoed while typing, `MY_TOKEN` and `MY_OPTIONAL` are not. The question for `MY_TOKEN` is shown as given, including the parenthesis. Press enter without typing anything for `MY_OPTIONAL`.

5. `IDEasy\workspaces\main\ai-test.properties` contains the three entered values and `ai.node.path` with backslashes. `IDEasy\conf\ide.properties` contains `MY_URL`, `MY_TOKEN` and `MY_OPTIONAL`.

6. Run again. There is no prompt and the file keeps the same values.

7. Remove `MY_URL`, `MY_TOKEN` and `MY_OPTIONAL` from `IDEasy\conf\ide.properties` and run again with `--batch`. There is no prompt, the workspace merge fails with `CliAbortException: Aborted by end-user.` and nothing is added to `IDEasy\conf\ide.properties`, as an undefined variable cannot be asked for in batch mode.

---

### Checklist for this PR

  with `internal`


maybeec (commented):
Thanks for this PR, and thanks in particular for the parser itself. Locating a call with a RegEx and then scanning the argument list by hand is exactly the right call: `findClosingParenthesis` / `parseArguments` handle quoted commas, quoted parenthesis and nesting correctly, and the parametrized `testForeignExpressionIsUntouched` covering `@media`, `@include`, `@Override` and `"@angular/core"` is precisely the test I would have asked for. Passing unknown functions through untouched is the single most important property of this feature and you got it right *and* covered it. The JavaDoc on the new types is also good.

No blockers, and nothing below needs a redesign. The two I care most about are the plaintext handling of `@ask-secret` values and `@path(..., native)` not using `WindowsPathSyntax`.

### Should-fix

1. `@ask-secret` values are persisted and logged in clear text - `AbstractEnvironmentVariables.java:414`
2. `@path(..., native)` hand-rolls separator replacement instead of `WindowsPathSyntax` - `PathFunction.java:51`
3. `isPersistent()` is hard-coded to `true`, so the "settings templates must not persist" requirement of #989 is not actually implemented - `AbstractEnvironmentVariables.java:422`, and its test only exercises a test double - `ExpressionParserTest.java:198`
4. `askForSecret` is a verbatim copy of `askForInput` - `AbstractIdeContext.java:1108`
5. `DirectoryMergerExpressionTest` depends on `Hashtable` iteration order - `DirectoryMergerExpressionTest.java:34`
6. The masking itself has no automated coverage, and the Windows git-bash path needs a manual test pass - `IdeContextConsole.java:65`
7. The new expression syntax is not documented anywhere - `CHANGELOG.adoc:9`
8. A template authoring error aborts `ide update` with a raw `IllegalArgumentException` - `ExpressionParser.java:89`

### Minor

9. Secrets are `trim()`ed - `AbstractIdeContext.java:1121`
10. Function results are re-scanned by the variable resolver - `AbstractEnvironmentVariables.java:214`
11. `AbstractIdeContextTest.TEST_RESOURCES` already exists - `DirectoryMergerExpressionTest.java:36`
12. Redundant initial `matcher.find()` - `ExpressionParser.java:57`

### Scope against #989

| Requirement | Status |
| --- | --- |
| `@<function-name>([<arg>[,<arg>]*])` syntax, args always String, comma separated, trimmed, quoted with `'` or `"` | met |
| Args may themselves contain variables | met |
| Manual argument scanning instead of a pure RegEx | met, and better than the RegEx sketched in the issue |
| `@path` with 1st arg path, optional 2nd arg `unix` (default) / `native` | partial - `native` does not use `WindowsPathSyntax` (see 2) |
| `@ask-variable` / `@ask-secret`, defined variable returned without asking, empty 1st arg always asks, default question text, 3rd arg default value | met |
| Values from a workspace template persisted to `conf/ide.properties` | met |
| "For settings templates this should not happen since these templates are only instantiated once" | missing (see 3) |
| `@if-windows` / `@if-mac` / `@if-linux` / `@if-unix` | met |
| `ExpressionFunction` interface + `ExpressionFunctionManager` registry so new releases can register more functions | met |

Also: the issue invites a follow-up story for the maven `settings.xml` password prompting/encryption (`@ask-maven-secret` or a `resolve` flag telling the function it is resolving `settings.xml`). Worth creating it and linking it here so the plaintext-storage topic from finding 1 has an owner.

### CI / DoD

All checks green, CLA signed, branch is up-to-date with `main`, CHANGELOG entry present under the correct milestone, PR title follows `#989: ...`. One nit: the commit `Update CHANGELOG for version 2026.08.002` does not follow the `#<issue-id>: <summary>` commit format (see `documentation/contributing/commit.adoc`).

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -357,6 +364,65 @@ public String toString() {
...
+      return this.resolvedVars.getValue(name, false);
+    }
+
+    @Override
+    public void setVariable(String name, String value) {
+
+      EnvironmentVariables conf = getByType(EnvironmentVariablesType.CONF);
+      if (conf instanceof EnvironmentVariablesPropertiesFile propertiesFile) {
+        propertiesFile.set(name, value);
+        propertiesFile.save();
```
maybeec, line 414: **Should-fix** - the value the user just typed behind a masked prompt is written and logged in clear text.

The mechanism, all in existing code this now feeds:

* `EnvironmentVariablesPropertiesFile.set(String, String, boolean)` logs `LOG.debug("Set variable '{}={}' in {}", name, value, this.propertiesFilePath)` (`EnvironmentVariablesPropertiesFile.java:346`), so `ide -d update` prints the API token on the console.
* Once persisted, `EnvironmentVariablesMap.getFlat` logs `LOG.trace("{}: Variable {}={}", getSource(), name, value)` (`EnvironmentVariablesMap.java:41`) on every later read.
* `EnvironmentCommandlet.doRun` calls `collectVariables()` (all variables, not only the exported ones) and prints them, so plain `ide env` dumps `AI_API_KEY=sk-...`.

So `@ask-secret` currently differs from `@ask-variable` only in how the value is *entered*, not in how it is stored or shown afterwards. That is the part that will surprise users: masked input implies the value stays secret.

Minimum I would like to see here: register the entered secret in the privacy map so `PrivacyUtil` masks it in log output (`AbstractIdeContext.initializePrivacyMap` at `AbstractIdeContext.java:1060` is the existing hook), and document explicitly in the docs that `@ask-secret` stores the value unencrypted in `conf/ide.properties`. Encryption itself is fine as the follow-up story the issue asks for.

cli/src/main/java/com/devonfw/tools/ide/expression/function/PathFunction.java
```diff
@@ -0,0 +1,59 @@
...
+  @Override
+  public String apply(List<String> args, ExpressionContext context) {
+
+    String path = args.get(0);
+    String mode = (args.size() > 1) ? args.get(1) : MODE_UNIX;
+    if (MODE_UNIX.equals(mode)) {
+      return path.replace('\\', '/');
+    } else if (MODE_NATIVE.equals(mode)) {
+      if (context.getIdeContext().getSystemInfo().isWindows()) {
+        return path.replace('/', '\\');
```
maybeec, line 51: **Should-fix** - please use `WindowsPathSyntax` here rather than a character replace. The issue asks for it by name ("on Windows use WindowsPathSyntax.WINDOWS with backslashes, etc.") and the class exists exactly for this: `WindowsPathSyntax.normalize(String)` detects the drive letter in either syntax and rewrites root *and* separators.

Why it matters mechanically: IDEasy explicitly tolerates MSYS-style paths on Windows. `WindowsPathSyntax.MSYS.getDrive` accepts `/d/...`, `AbstractIdeContext.initializePrivacyMap` (`AbstractIdeContext.java:1066-1067`) registers both the `D:\...` and the `/d/...` form of the same path, and `EnvironmentVariablesMap.getFlat` (`EnvironmentVariablesMap.java:42-48`) normalizes variable values through `pathSyntax.normalize(value)`. So a value in MSYS form can reach `@path`, and `replace('/', '\\')` turns `/d/projects/foo` into `\d\projects\foo` - the drive letter is silently lost and the resulting path is broken, which is the exact class of bug #989 was filed to remove.

```java
} else if (MODE_NATIVE.equals(mode)) {
  if (context.getIdeContext().getSystemInfo().isWindows()) {
    return WindowsPathSyntax.WINDOWS.normalize(path);
  }
  return path.replace('\\', '/');
}
```

(needs `import com.devonfw.tools.ide.os.WindowsPathSyntax;`). A test with `IDE_HOME` set to `/d/projects/my-project` alongside the existing `testPathNativeOnWindows` would lock this in.

The `MODE_UNIX` branch above is fine as-is - the issue defines that one as plain backslash-to-slash.

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -357,6 +364,65 @@ public String toString() {
...
+      } else {
+        LOG.warn("Cannot persist variable {} since no configuration file is available.", name);
+      }
+    }
+
+    @Override
+    public boolean isPersistent() {
+      return true;
```
maybeec, line 422: **Should-fix** - this is the only production implementation of `ExpressionContext`, and it returns a constant `true`. That means the requirement from #989 - "For settings templates this should not happen since these templates are only instantiated once" - is not implemented: every `@ask-*` call persists, no matter where it came from.

The abstraction cannot decide it at this point either, because `resolveRecursive` has no idea whether its caller is `PropertiesMerger` on a workspace template or something else. It needs to be threaded in from the caller, e.g. as a field on the `ResolveContext` record set by `EnvironmentVariables.resolve(...)`.

Two honest options:

1. Wire it up: add the flag to `ResolveContext` and let the merger pass it, then `isPersistent()` returns it.
2. Drop it: remove `isPersistent()` from `ExpressionContext` and always persist, plus a comment saying why. You invoked KISS in the issue yourself, and an interface method that no production caller can ever make `false` is dead weight that reads as if the feature exists.

Either is fine with me, but the current middle ground is the worst of the three because the API and the test both suggest the behaviour is there.

cli/src/test/java/com/devonfw/tools/ide/expression/ExpressionParserTest.java
```diff
@@ -0,0 +1,380 @@
...
+   * Test that a settings template does not persist the entered value since it is only instantiated once.
+   */
+  @Test
+  void testSettingsTemplateDoesNotPersist() {
+
+    // arrange
+    IdeTestContext context = newContext(PROJECT_BASIC);
+    context.setAnswers("value");
+    TestExpressionContext expressionContext = new TestExpressionContext(context);
+    expressionContext.persistent = false;
```
maybeec, line 198: **Should-fix** - paired with `AbstractEnvironmentVariables.java:422`: this test sets a field on `TestExpressionContext`, then asserts that `TestExpressionContext.setVariable` was not called. It validates the test double, not any IDEasy production code - no production code path can ever produce `isPersistent() == false` today.

Per `documentation/contributing/junit-testing.adoc` a test has to exercise the project's own logic. Once the flag is threaded through `ResolveContext` this test should assert on the real `EnvironmentExpressionContext` (i.e. that nothing landed in `conf/ide.properties`), the way `DirectoryMergerExpressionTest` already does for the positive case. If you take the "drop `isPersistent()`" route instead, this test should go with it.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -1104,6 +1104,31 @@ public String askForInput(String message, String defaultValue) {
     }
   }
 
+  @Override
+  public String askForSecret(String message, String defaultValue) {
```
maybeec, line 1108: **Should-fix** - this is `askForInput(String, String)` (lines 1083-1105) copied verbatim; the only difference in 24 lines is `readSecretLine()` instead of `readLine()` on line 1121. Duplicated logic like this drifts: the next fix to the batch-mode / force-mode / default-value contract will land in one copy only, and the two prompts will start behaving differently in ways nobody notices.

Please extract the shared loop and delegate, e.g.:

```java
@Override
public String askForInput(String message, String defaultValue) {

  return ask(message, defaultValue, false);
}

@Override
public String askForSecret(String message, String defaultValue) {

  return ask(message, defaultValue, true);
}

private String ask(String message, String defaultValue, boolean secret) {

  while (true) {
    // ... existing body, with:
    String input = secret ? readSecretLine() : readLine().trim();
  }
}
```

See `documentation/contributing/coding-conventions.adoc` - duplicated code either moves up or stays where it was.

cli/src/test/java/com/devonfw/tools/ide/merge/DirectoryMergerExpressionTest.java
```diff
@@ -0,0 +1,57 @@
...
+   * @throws Exception on error.
+   */
+  @Test
+  void testExpressionsInWorkspaceTemplate(@TempDir Path workspaceDir) throws Exception {
+
+    // arrange
+    IdeTestContext context = newContext(PROJECT_BASIC, null, true);
+    // NOTE: the answers are consumed in the order the questions are asked. PropertiesMerger iterates the Properties
+    // and therefore does not preserve the order of the lines in the template file.
+    context.setAnswers("sk-TOPSECRET", "http://llama.local");
```
maybeec, line 34: **Should-fix** - the comment on the two lines above is an accurate description of a fragile test rather than a fix for it. `PropertiesMerger.resolve` iterates `properties.keySet()` of a plain `java.util.Properties` (`PropertiesMerger.java:60`), i.e. `Hashtable` order, which is unspecified. The answers here are positional, so if that order ever changes (different JDK, different key set after someone edits the template) `sk-TOPSECRET` silently becomes the value of `backend.url` and the test fails, or worse asserts the wrong pairing.

The fix also improves production behaviour: iterate deterministically so the user is prompted in the order they see the lines in their template, and so merged files come out stable. `SortedProperties` already exists in the codebase and `PropertiesMerger.inverseMerge` uses it (`PropertiesMerger.java:82`):

```java
private void resolve(Properties properties, EnvironmentVariables variables, Object src) {

  for (String key : new TreeSet<>(properties.stringPropertyNames())) {
    properties.setProperty(key, variables.resolve(properties.getProperty(key), src, this.legacySupport));
  }
}
```

Then the comment can go and the answers become meaningful.

Good test otherwise - this is the one that actually proves the end-to-end wiring, including that `@media(...)` survives the merge untouched.

cli/src/main/java/com/devonfw/tools/ide/context/IdeContextConsole.java
```diff
@@ -52,6 +52,21 @@ protected String readLine() {
...
+  protected String readSecretLine() {
+
+    if (this.scanner == null) {
+      char[] password = System.console().readPassword();
+      if (password == null) {
+        return "";
+      }
+      return new String(password);
+    } else {
+      LOG.warn("System console not available - secret input will be visible while typing.");
```
maybeec, line 65: **Should-fix** - two things about the masking, both about the part CI cannot see.

1. No automated coverage. `AbstractIdeTestContext` overrides `readLine()` only (`AbstractIdeTestContext.java:155`) and does not override `readSecretLine()`, so every `@ask-secret` test in this PR flows through the *echoing* `readLine()` - and `AbstractIdeTestContext.readLine` even logs the answer at INTERACTION level. If someone changed `AskFunction.ask` to call `askForInput` for secrets too, every test here would stay green. Please override `readSecretLine()` in `AbstractIdeTestContext` (recording that it was used, or serving from a separate answer list) and assert it in at least one `@ask-secret` test.

2. This warn branch is the *common* case on Windows, not an exotic one: under mintty (git-bash) `System.console()` returns `null`, so the constructor takes the scanner fallback (`IdeContextConsole.java:36-42`) and the secret is echoed. That is the right behaviour given the constraints and the warning is good, but it means the feature needs a manual pass on Windows cmd/PowerShell, Windows git-bash, Linux and macOS before merge - green CI proves nothing about masking. Your testing instructions already say "run this from a normal terminal"; please state the result per OS in the PR when you have it.

CHANGELOG.adoc
```diff
@@ -6,6 +6,7 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/989[#989]: Allow expressions in template variable definitions
```
maybeec, line 9: **Should-fix** (not about this line itself) - the feature is undocumented outside the code.

`documentation/configurator.adoc:39-42` is where the workspace template syntax lives today ("Variables in the form `$[<variable-name>]` get resolved..."). The audience for `@path` / `@ask-variable` / `@ask-secret` / `@if-windows` is exactly the settings maintainers reading that page, and right now the only description of the syntax is the issue and this PR body. Please add a section there covering:

* the `@<function-name>(<args>)` syntax and the quoting rules,
* one example per function (the `nodejs.xml` case from #989 is the perfect motivating example),
* that unknown `@name(...)` is passed through untouched, so `@media` etc. are safe,
* where `@ask-*` values are stored, and what that means for secrets (see my comment on `AbstractEnvironmentVariables.java:414`),
* that expressions belong in workspace templates - putting an `@ask-*` call into a value in `ide.properties` means every variable resolution, including `ide env`, will try to prompt.

The CHANGELOG line itself is correct and under the right milestone.

cli/src/main/java/com/devonfw/tools/ide/expression/ExpressionParser.java
```diff
@@ -0,0 +1,178 @@
...
+    return sb.toString();
+  }
+
+  private String apply(ExpressionFunction function, List<String> args, String value, ExpressionContext context) {
+
+    int size = args.size();
+    int min = function.getMinArgs();
+    int max = function.getMaxArgs();
+    if ((size < min) || ((max >= 0) && (size > max))) {
+      throw new IllegalArgumentException(
```
maybeec, line 89: **Should-fix** - a wrong argument count is a configuration mistake by the settings maintainer, but this raw `IllegalArgumentException` propagates out of variable resolution and aborts `ide update` with a stack trace.

That is inconsistent with how the very same resolution step handles the sibling case: an undefined variable is reported with `LOG.atLevel(logLevel).log("Undefined variable {} in '{}'", var, src)` and resolution continues (`AbstractEnvironmentVariables.java:239-250`). The rationale there applies here too - a broken template line should not block a user from getting their IDE started.

Please either log a warning naming the file (`src`) and leave the expression untouched, or throw a `CliException` with a message pointing at the template - `CliException` is what IDEasy uses for expected, user-facing failures, and it gets rendered without a stack trace. The same applies to the `IllegalArgumentException` in `AskFunction.apply` (empty variable name without a question) and the one in `PathFunction.apply` (invalid mode): in all three the end user gets a technical stacktrace for someone else's typo.

One detail worth keeping either way: `value` is interpolated into the message, so for `@ask-secret` an already-resolved secret could end up in the exception text. Prefer the source/template reference over the raw value.

cli/src/main/java/com/devonfw/tools/ide/context/AbstractIdeContext.java
```diff
@@ -1104,6 +1104,31 @@ public String askForInput(String message, String defaultValue) {
...
+        IdeLogLevel.INTERACTION.log(LOG, message);
+      }
+      if (isBatchMode()) {
+        if (isForceMode()) {
+          return defaultValue;
+        } else {
+          throw new CliAbortException();
+        }
+      }
+      String input = readSecretLine().trim();
```
maybeec, line 1121: **Minor** - `trim()` is inherited from the `askForInput` copy, but it is wrong for a secret: leading or trailing whitespace can be part of a password, and more practically, a token pasted with a stray space is silently altered so the user gets an authentication failure with no hint why.

```suggestion
      String input = readSecretLine();
```

(the `isEmpty()` check below still does the right thing for a plain Enter).

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -206,14 +211,16 @@ private String resolveRecursive(String value, Object source, int recursion, Abst
     }
     recursion++;
 
+    String value2 = EXPRESSION_PARSER.resolve(value, new EnvironmentExpressionContext(source, recursion, resolvedVars, context));
```
maybeec, line 214: **Minor** - running the expression parser *before* `resolveWithSyntax` means the function results are then fed back through variable resolution. Since arguments are already resolved explicitly by `ExpressionParser.parseArgument` via `context.resolve(...)`, that second pass buys nothing and can only do harm: a value the user typed that happens to contain `$[` (or `${` with `legacySupport` on) is reinterpreted as a variable reference and logs an "Undefined variable" warning containing the value.

Not blocking - just noting that resolving expressions *after* the variable pass would be equivalent for every case in your tests and would keep function output opaque. If you keep the current order, a short comment here explaining why would help the next reader.

Also: `value2` reads as a scratch name. `withExpressions` or `expressionsResolved` would say what it is (`coding-conventions.adoc`, Naming).

cli/src/test/java/com/devonfw/tools/ide/merge/DirectoryMergerExpressionTest.java
```diff
@@ -0,0 +1,57 @@
...
+    DirectoryMerger merger = context.getWorkspaceMerger();
+    Path templates = Path.of("src/test/resources/templates-expression");
```
maybeec, line 36: **Minor** - `AbstractIdeContextTest` (which this test extends) already defines `protected static final Path TEST_RESOURCES = Path.of("src/test/resources")` at `AbstractIdeContextTest.java:36`.

```suggestion
    Path templates = TEST_RESOURCES.resolve("templates-expression");
```

cli/src/main/java/com/devonfw/tools/ide/expression/ExpressionParser.java
```diff
@@ -0,0 +1,178 @@
...
+   * @param context the {@link ExpressionContext}.
+   * @return the given value with all expressions of registered functions replaced by their result.
+   */
+  public String resolve(String value, ExpressionContext context) {
+
+    if (value == null) {
+      return null;
+    }
+    Matcher matcher = FUNCTION_START.matcher(value);
+    if (!matcher.find()) {
```
maybeec, line 57: **Minor** - this `find()` and the `find(pos)` on line 62 with `pos == 0` search the same region twice. The early return is a nice fast path for the overwhelmingly common "no `@` call in this value" case (and this runs on every single variable resolution, so it is worth keeping), but it can reuse the result:

```java
if (!matcher.find()) {
  return value;
}
StringBuilder sb = new StringBuilder(value.length() + EXTRA_CAPACITY);
int pos = 0;
do {
  ...
} while (matcher.find(pos));
```

which mirrors the `do { ... } while (matcher.find())` shape already used in `AbstractEnvironmentVariables.resolveWithSyntax`.

QuangAnhLe:
Thanks for the PR. Most of my review points align with those of @maybeec  mentioned earlier, so you can simply refer to my summary below; I won't be reviewing each section individually. Please feel free to contact me if you have any questions.

**Review: #989 — expression functions for template variables**

 Overall: solid, well-tested implementation. The architecture matches the ticket's design (interface + manager + parser + per-OS/path/ask functions), the parser correctly handles what a regex
  can't (quoted commas, quoted parens, nesting), foreign @... syntax is passed through untouched, and the test suite is thorough (33 tests: parser unit tests, integration through
  EnvironmentVariables, and a real DirectoryMerger run). All pass. Commit messages are clean. The core logic is correct.

  Below are findings, most-severe first. None block merge; the first two are worth a decision.

 **1. Secret values flow into logs in plaintext (masking only applies to keystrokes)**

  @ask-secret masks input while typing, but the returned value is then treated as an ordinary variable value. The integration test's own output shows it:
  Set variable 'AI_API_KEY=sk-TOPSECRET' in ...conf\ide.properties
  That log line comes from PropertiesMerger (pre-existing), but it's newly reachable now that a secret can be the value being set. The value also lands in conf/ide.properties in cleartext — which
  is the intended persistence, so that's fine. But the console/log echo of the secret is a real exposure for the "secret" case the ticket is trying to protect. Worth confirming intent; if
  secrets are meant to stay off the log, this path (and any other value-logging) should be masked for @ask-secret results.

**2. isPersistent() is hardcoded true — settings templates persist too (spec deviation)**

  AbstractEnvironmentVariables.EnvironmentExpressionContext.isPersistent() (line 421–423) always returns true, so @ask-variable/@ask-secret persist to conf/ide.properties even when resolved from
  a settings template. The ticket is explicit:

  "For settings templates this should not happen since these templates are only instantiated once."

  The Javadoc on ExpressionContext.isPersistent() (line 41–44) even documents the intended distinction, but the implementation doesn't make it — it has no way to tell whether the current
  resolution is a settings vs. workspace template. Impact is low in practice (settings are instantiated once, so the "asked only once" benefit rarely matters), but the behavior contradicts the
  spec and the doc. Either implement the settings/workspace distinction or soften the Javadoc/commit claim. Note the unit test testSettingsTemplateDoesNotPersist only exercises a hand-rolled
  TestExpressionContext with persistent=false — it doesn't prove the real AbstractEnvironmentVariables path honors the setting, so this gap is masked in tests.

**3. A defined variable's value returned by @ask-variable is not re-resolved for nested expressions**

  In AskFunction.apply, the defined-variable fast path returns context.getVariable(...) directly (line 71–73). That value is appended into value2 and then $[...] variables are resolved once more
  by resolveWithSyntax, but embedded @... expressions inside an already-defined variable's value would not be, since apply's return is never fed back through the expression parser. This is an
  asymmetry (a plain $[FOO] whose value contains @path(...) behaves the same — also not re-resolved — so it's consistent with existing variable semantics, but worth knowing). Edge case, low
  priority; flagging for completeness.

 **4. Minor**
 
  - readSecretLine fallback is unmasked: when System.console() is null (piped input, running from the IDE), IdeContextConsole.readSecretLine() logs a warning and reads via the plaintext Scanner.
  Acceptable given the constraints, and it correctly warns, but it's worth ensuring the manual-testing doc emphasizes "run from a normal terminal" (it does — good).
  - No way to escape a quote inside a quoted argument. Backslash is deliberately not an escape char (correct, tested), so a literal ' inside a single-quoted arg is unrepresentable. Fine for KISS;
  just a documented limitation.
  - EXTRA_CAPACITY = 8 is duplicated in both ExpressionParser (line 31) and AbstractEnvironmentVariables (line 34). Harmless, but could be a shared constant.
  - resolve() double-find: the guard if (!matcher.find()) (line 57) is followed by while (matcher.find(pos)) from pos=0, so the first match is found twice. Correct but slightly wasteful.

**Suggested before merge**

Items 1 and 2 are the only ones needing a judgment call — decide whether secrets should be masked in logs and whether the settings-vs-workspace persistence distinction is in scope now or a follow-up. Everything else is optional polish. The implementation and tests otherwise meet the Definition of Done for the core story.

cli/src/main/java/com/devonfw/tools/ide/expression/ExpressionParser.java
```diff
@@ -0,0 +1,179 @@
...
+ * Text that does not form a call of a {@link ExpressionFunctionManager#getFunction(String) registered function} is passed through entirely untouched. This is
+ * essential since foreign configuration formats may use an {@code @} for their own purposes (e.g. CSS {@code @media(...)}) and IDEasy must never try to resolve
+ * placeholders that are not ours.
+ */
+public class ExpressionParser {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ExpressionParser.class);
+
+  /** Locates the start of a potential function call. The group is the function name. */
+  //                                                       .1
```
hohwille: Typically auto-formatters will eat up spaces trimming comments. Further the idea of this is to put the numbers exactly at the positions of the opening parenthesis to document the group numbers.
```suggestion
  // .............................................................1
```

cli/src/main/java/com/devonfw/tools/ide/expression/ExpressionParser.java
```diff
@@ -0,0 +1,179 @@
...
+  public String resolve(String value, ExpressionContext context) {
+
+    if (value == null) {
+      return null;
+    }
+    Matcher matcher = FUNCTION_START.matcher(value);
+    if (!matcher.find()) {
+      return value;
+    }
+    StringBuilder sb = new StringBuilder(value.length() + EXTRA_CAPACITY);
```
hohwille: I do not think that the users input will somehow relate with `value.length()`.
I would go for KISS and not try to predict any capacity since we have no clue what the function will produce or the user might enter.
```suggestion
    StringBuilder sb = new StringBuilder();
```

cli/src/main/java/com/devonfw/tools/ide/environment/AbstractEnvironmentVariables.java
```diff
@@ -357,6 +367,60 @@ public String toString() {
...
+    @Override
+    public IdeContext getIdeContext() {
+
+      return AbstractEnvironmentVariables.this.context;
+    }
+
+    @Override
+    public String resolve(String value) {
+
+      return this.resolvedVars.resolveRecursive(value, this.src, this.recursion, this.resolvedVars, this.context);
```
hohwille, line 402: very nice. You support both nice reuse as well as great nesting this way.
If I understand correctly, you  already support nesting functions in function arguments with any combination of variables. Great job 👍

cli/src/main/java/com/devonfw/tools/ide/merge/PropertiesMerger.java
```diff
@@ -57,10 +57,10 @@ protected void doMerge(Path setup, Path update, EnvironmentVariables resolver, P
 
   private void resolve(Properties properties, EnvironmentVariables variables, Object src) {
 
-    Set<Object> keys = properties.keySet();
-    for (Object key : keys) {
-      String value = properties.getProperty(key.toString());
-      properties.setProperty(key.toString(), variables.resolve(value, src, this.legacySupport));
+    // resolve the keys in a deterministic order (rather than the unspecified Properties/Hashtable order) so that
+    // interactive questions are asked in a stable order and the merged file comes out consistently
+    for (String key : new TreeSet<>(properties.stringPropertyNames())) {
```
hohwille, line 62: Also very nice 👍 
BTW: Are you aware of `com.devonfw.tools.ide.environment.SortedProperties`?

cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
```diff
@@ -280,6 +280,42 @@ default String askForInput(String message) {
...
+   */
+  void addSecretVariable(String name);
+
+  /**
+   * Registers the value of a variable as secret if the variable was marked via {@link #addSecretVariable(String)}. Has to be called before the value is logged.
+   *
+   * @param name the name of the variable.
+   * @param value the value of the variable.
+   */
+  void addSecretValue(String name, String value);
```
hohwille, line 317: Great that you care about security of credentials. I appreciate that very much.
However, I see some problems here:
1. If the function was not invoked this method gets never called and hence we have no clue what variables are secrets.
2. If we really want to protected secrets, we IMHO need to store them encrypted (see how maven is doing that or [spring-boot with jasypt spring-boot starter](https://github.com/devonfw/devon4j/blob/master/documentation/spring/guide-spring-configuration.asciidoc#password-encryption) is solving it). This is a complex thing to implement and should deserve its own story if we ever need to support it. The other alternative would be not to persist secrets what would be quite simple. This works fine for templates that are instantiated only once (e.g. in `settings/templates` or in `workspace/setup`). For the moment, I would go for KISS and think about this way.
3. Sometimes it is easier and more pragmatic to decide for a convention than trying to implement a bullet proof solution: We can use the convention that secret variables have to follow specific naming rules e.g. start or end with one of `SECRET`, `PASSWORD`, or `API_KEY`. Then we can avoid this complex runtime state of secret variable names registered in `IdeContext`. Also you could even enforce that the function `ask-secret` only accepts variables following this naming pattern. Then we could also detect secret values already when we create our environment and reading the variables from the convention.

cli/src/main/java/com/devonfw/tools/ide/expression/function/AskFunction.java
```diff
@@ -0,0 +1,177 @@
...
+    return 4;
+  }
+
+  @Override
+  public String apply(List<String> args, ExpressionContext context) {
+
+    String variableName = args.get(0);
+    String question = (args.size() > 1) ? args.get(1) : null;
+    EnvironmentVariablesType location = toLocation((args.size() > 2) ? args.get(2) : null);
+    String defaultValue = (args.size() > 3) ? args.get(3) : null;
```
hohwille, line 82: All correct, but maybe we can approve (in a cleanup story after merge by someone else):
1. Simple idea: Create subclasses of `ExpressionFunction` like `UnaryExpressionFunction`, `BinaryExpressionFunction`, `TenerayExpressionFunction` that already implement `apply` verify the maximum number of arguments and then call a dedicated `apply` method variant with the proper number of arguments.
2. Advanced idea: In `ExpressionFunctionManager` we only register factories by function name and the factory then creates the `ExpressionFunction` that then becomes stateful and has a list of `Property` arguments just like a `Commandlet`. Then we can even define defaults and keep the parsing from String to dedicated enums and types out of the concrete function implementation.

Dont get me wrong. Your PR is great and already solves more than I expected. I am just always thinking further and generating a vision for the perfect future to aim for. So my suggestion would be that you create a follow up story to improve your solution and copy all the review feedback that is nice to have or not required for merge into that and then you just link the issue in a comment and we resolve all the threads and can merge this so you have a safe handover in case you need to move on.

cli/src/main/java/com/devonfw/tools/ide/expression/function/AskFunction.java
```diff
@@ -0,0 +1,177 @@
...
+    if (NULL_VALUE.equals(defaultValue)) {
+      // an explicit "null" means no default value, so empty input is not allowed
```
hohwille, line 84: I was thinking about the `null` literal so if not enclosed with quotes (argument `null` instead of `"null"` or `'null'`) this would result in `null` but also fine. Nobody will likely need `"null"` as a reasonable default value...

cli/src/main/java/com/devonfw/tools/ide/expression/function/AskFunction.java
```diff
@@ -0,0 +1,177 @@
...
+      question = "Please enter the value for the " + (this.secret ? "secret " : "") + "variable " + variableName + ":";
+    }
+    value = ask(question, defaultValue, context);
+    if (value == null) {
+      // In batch mode with force enabled the user cannot be asked and no default value was given. The expression
+      // resolves to the empty string but is NOT persisted: otherwise the variable would be defined as empty forever
+      // and the user would never be asked again on the next interactive run.
+      return "";
+    }
+    context.setVariable(variableName, value, location);
```
hohwille: so for the moment we could simply skip in case of secret.

```suggestion
    if (!this.secret) {
      context.setVariable(variableName, value, location);
    }
```

cli/src/main/java/com/devonfw/tools/ide/expression/function/AskFunction.java
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * @return the value entered by the user, the default value, or {@code null} if the user could not be asked (batch mode with force) and no default value was
+   *     given.
+   */
+  private String ask(String question, String defaultValue, ExpressionContext context) {
+
+    String prompt = question;
+    if (defaultValue != null) {
+      // show the default so the user can just hit return
+      prompt = question + " <" + defaultValue + ">";
```
hohwille: In other places we use this

```suggestion
      prompt = question + " [" + defaultValue + "]";
```

cli/src/test/java/com/devonfw/tools/ide/expression/ExpressionParserTest.java
```diff
@@ -0,0 +1,509 @@
...
+    // arrange
+    IdeTestContext context = newContext(PROJECT_BASIC);
+    TestExpressionContext expressionContext = new TestExpressionContext(context);
+    expressionContext.variables.put("IDE_HOME", "D:\\projects\\my-project");
+
+    // act
+    String result = expressionContext.resolve("@path('$[IDE_HOME]/software/mvn')");
+
+    // assert
+    assertThat(result).isEqualTo("D:/projects/my-project/software/mvn");
```
hohwille: Looks very windows specific with the drive letters.
Test should still work on Linux and MacOS but maybe to avoid confusion skip the drive letter?

cli/src/test/java/com/devonfw/tools/ide/expression/ExpressionParserTest.java
```diff
@@ -0,0 +1,509 @@
...
+
+    // arrange
+    IdeTestContext context = newContext(PROJECT_BASIC);
+    TestExpressionContext expressionContext = new TestExpressionContext(context);
+
+    // act
+    String result = expressionContext.resolve(value);
+
+    // assert
+    assertThat(result).isEqualTo(value);
```
hohwille, line 172: Great for robustness. 👍

documentation/configurator.adoc
```diff
@@ -54,6 +54,77 @@ If the user modifies such settings and reopens his IDE his changes are reverted.
...
+Besides plain variables in the form `$[«variable-name»]`, a template value may contain an expression of the syntax `@«function-name»([«arg»[,«arg»]*])`.
+This allows to apply a little transformation to a value or to ask the user for a value while the workspace is being created.
+
+=== Syntax
+
+* The function name consists of lowercase letters, digits and dashes and is written after a single `@` followed by a pair of parentheses, e.g. `@path(...)`.
+* Arguments are always strings and are separated by a comma.
+Spaces around an argument are ignored/trimmed.
+* To pass a literal value that may contain the argument separator (a comma) or a parenthesis, quote it with single or double quotes, e.g. `@ask-secret('AI_API_KEY', 'Enter your key (from the portal):')`.
+A backslash inside a quoted argument is a literal backslash and is not an escape character, since arguments regularly contain native Windows paths.
```
hohwille: IDEasy aims to be platform independent and hence also templates and configurations should be.
Therefore I do not see a valid argument to support easy handling of native windows paths since this is IMHO rather discouraged than something we should specially support.
Maybe we later even need to add escaping to support quotes in literals.

```suggestion
```

documentation/configurator.adoc
```diff
@@ -54,6 +54,77 @@ If the user modifies such settings and reopens his IDE his changes are reverted.
...
+Encryption is a separate follow-up story (it also covers the case of the Maven `settings.xml`).
+* Expressions belong in the workspace templates under `setup` and `update`.
+Do not put an `@ask-*` call into a value of `ide.properties` itself: a value that contains an expression is resolved on _every_ variable resolution, so a `@ask-*` in `ide.properties` would prompt on every `ide` invocation (including `ide env`).
+Expressions are typically used in the `update` templates to build up OS-specific or user-specific settings that would otherwise be hard to express with a plain variable, as in the following example that only resolves to a working value on the matching operating system:
++
+[source]
+----
+<local-interpreter path="@path('$[IDE_HOME]/software/node/node', native)" merge:id="name()"/>
+----
+
```
hohwille, line 129: This is great but I fear that this way it will be hard to keep the documentation in sync with the code.
Maybe we can use docjar and just link to the latest JavaDoc of the functions package.
If we later follow my suggestions to express function arguments more structural with generic parsing support rather than manual handling in apply method implementation the JavaDoc itself would be fully sufficient to lookup what is needed. Otherwise we can also just put a link to the package in the GitHub source-code.

CHANGELOG.adoc
```diff
@@ -29,6 +29,7 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/2251[#2251]: Provide generic uninstall support for globally installed tools (windows)
 * https://github.com/devonfw/IDEasy/issues/1135[#1135]: Fix PowerShell env variable initialization on Windows by sourcing functions from the PowerShell profile
 * https://github.com/devonfw/IDEasy/issues/741[#741]: Add a warning message for legacy devonfw-ide settings users
+* https://github.com/devonfw/IDEasy/issues/989[#989]: Allow expressions in template variable definitions
```
hohwille: Sorry, that it took too long but we need to move to the next release.

hohwille (approved):
@Paras14 thank you for your PR. This is really great work and you already included more than I was hoping for. Your implementation seems solid and functional. Awesome job 🥇 
I left comments for improvement. My suggestion would be that you have a look what can be done very easy and quick and create a follow up issue for the rest so we can resolve these comments here and merge this PR which is already great as is. Thanks 🙏

cli/src/main/java/com/devonfw/tools/ide/expression/function/AskFunction.java
hohwille: BTW: My mistake - in the story, I indeed wrote `angled brackets` but meant `square brackets`.

cli/src/test/java/com/devonfw/tools/ide/expression/ExpressionParserTest.java
```diff
@@ -0,0 +1,509 @@
...
+    // assert
+    assertThat(secretContext.getSecretLineCount()).isEqualTo(1);
+    assertThat(plainContext.getSecretLineCount()).isZero();
+    assertThat(secretContext).log().hasNoMessageContaining("dummy-secret-value");
+  }
+
+  /**
+   * Simple {@link ExpressionContext} for testing that also simulates the surrounding variable resolution of {@code AbstractEnvironmentVariables}.
+   */
+  private static class TestExpressionContext implements ExpressionContext {
```
hohwille: I understand that you want to create a plain unit-test here, but maybe it would make more sense to test this without mocking by testing it via EnvironmentVariablesResolved so you do not need this mock?

documentation/configurator.adoc
Paras14: This is valid point, let's move this to a separate issue, since linking the JavaDoc (or the GitHub source) and reworking the docs is best done on its own to keep them in sync.
@hohwille

cli/src/main/java/com/devonfw/tools/ide/expression/function/AskFunction.java
Paras14: These can be in a separate follow-up issue and link it here so we can resolve the thread, leaving the cleanup for after merge.
@hohwille

hohwille:
FYI: This stupid GitGuardian is still complaining since it pretends to finds a "secret" in the git history of this PR.
I will consider this problem as solved since there never was a real secret here and we will squash this PR so only the final state will go to main. Will have to overrule GitGuardian as admin then...

cli/src/test/java/com/devonfw/tools/ide/expression/ExpressionParserTest.java
hohwille: Created #2409

documentation/configurator.adoc
hohwille: Created #2409

cli/src/main/java/com/devonfw/tools/ide/expression/function/AskFunction.java
hohwille: Created #2409

cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
hohwille: Created #2409

hohwille:
Tested in combination with PR #2398 using IDEasy 2026.09.002-09_22_07-SNAPSHOT on Windows:
```
$ ide create test-ideasy ...
...
Start: Prepare repository
Please enter your GitHub login: [hohwille]

Successfully ended step 'Prepare repository'.
Start: Setup of repository IDEasy
Start: Clone or pull repository
Cloning into '.'...
remote: Enumerating objects: 41326, done.
remote: Counting objects: 100% (664/664), done.
remote: Compressing objects: 100% (420/420), done.
remote: Total 41326 (delta 467), reused 244 (delta 244), pack-reused 40662 (from 2)
Receiving objects: 100% (41326/41326), 13.53 MiB | 9.51 MiB/s, done.
Resolving deltas: 100% (16362/16362), done.
Updating files: 100% (1960/1960), done.
Already on 'main'
Your branch is up to date with 'origin/main'.
Successfully ended step 'Clone or pull repository'
...
$ cd workspaces/main/IDEasy/
$ git remote -v
origin  https://github.com/hohwille/IDEasy.git (fetch)
origin  https://github.com/hohwille/IDEasy.git (push)
upstream        https://github.com/devonfw/IDEasy.git (fetch)
upstream        https://github.com/devonfw/IDEasy.git (push)
```

So retest is successful and shows the combination of our features are working.

The output during the setup also contains confusing error log messages like:
```
Unknown command "«end»".
```
Analysis revealed that this is actually from our maven build and from JUnit tests that seem to test strange things but finally succeed.
Finally the build failed on my machine but that has other reasons not related to this story.

PR #2285, opened 2026-08-06
#2279: allow launching the GUI without an IDEasy project

### LINKED ISSUE #2279: Desktop shortcut for the GUI does not work: gui commandlet requires IDE_HOME
### Actual behavior

The desktop/start-menu shortcut created by #1917 is generated correctly, but clicking it never starts the GUI.

The shortcut runs `ideasy.exe gui`, and `createWindowsShortcut` does not set a `WorkingDirectory`. Windows therefore starts the process in the directory of the target executable (`«IDE_ROOT»\_ide\installation\bin`), which is not inside an IDEasy project. Since `Gui` does not override `isIdeHomeRequired()`, it inherits the default `true` from `Commandlet`, so the commandlet aborts:

```
The gui commandlet requires an IDEasy project to work.
Please run "icd <project-name>" before calling "ide gui".
```

A desktop shortcut is by definition launched from outside any project, so in its current form it can never work.

### Reproduce

1. Install IDEasy from a release containing #1917 (`ide install`), so that the shortcut is created.
2. Double-click the `IDEasy` shortcut on the desktop or in the start menu.
3. The GUI does not appear.

Equivalent without a shortcut — run the same command from any directory outside a project:

```
cd C:\Users\«user»\Downloads
"C:\projects\_ide\installation\bin\ideasy.exe" gui
```

Verified with the working directory set both to a neutral folder and to `«IDE_ROOT»\_ide\installation\bin` (the directory Explorer uses when the shortcut has no `WorkingDirectory`). Both produce the error above. Running the identical command from within a project directory starts the GUI correctly.

### Expected behavior

Double-clicking the shortcut starts the IDEasy GUI without requiring a preceding `icd «project»`.

The GUI already supports this: `ProjectManager` reads the project list from `IDE_ROOT` and its class comment states this is intentional — *"you might want to read the list of projects without being already in the project context"*. Project selection happens inside the GUI, so `gui` should require `IDE_ROOT` but not `IDE_HOME`.

### IDEasy status

```bash
IDE_ROOT is set to C:\projects
Your version of IDEasy is 2026.07.002
```

Note: the shortcut code is not part of release 2026.07.002 (tagged 2026-07-21); #1917 was merged on 2026-07-30. The behavior above was reproduced by invoking the `gui` commandlet directly, which is unchanged by #1917.

### Related/Dependent issues

Sub-issue of #1917

### Comments/Hints

Affects all platforms, not just Windows. The Linux `.desktop` entry uses `Exec=«installation»/bin/ideasy gui` with no `Path=` key, and the macOS `IDEasy.command` calls the binary without changing directory, so both hit the same guard.

Suggested fix: override `isIdeHomeRequired()` in `Gui` to return `false` while leaving `isIdeRootRequired()` at `true`.

The existing test coverage does not catch this — `IdeasyCommandletTest.verifyDesktopShortcut` only asserts on Windows that the two `.lnk` files exist, without checking target, arguments, or whether launching them succeeds.

quando632:
### This PR fixes #2279

The desktop shortcut added by #1917 is created correctly but never starts the GUI, because it launches `ideasy gui` from outside any IDEasy project.

### Implemented changes:

* `Gui` no longer requires `IDE_HOME`. `IDE_ROOT` is still required, since the launcher pom and the project list resolve relative to it. Project selection happens inside the GUI, which reads the projects from `IDE_ROOT` via `ProjectManager`.
* The bin directory of the resolved Maven installation is registered in the `SystemPath`. Without `IDE_HOME` there is no project software folder, so `mvn` was not resolvable and the launch failed with `CreateProcess error=2`.
* New `GuiTest` covering both the relaxed `IDE_HOME` requirement and that `IDE_ROOT` stays required.

---

### Testing instructions

1. Build the branch: `mvn -o package -DskipTests` in `cli`.
2. Open a shell that has no `IDE_HOME` set and change into a directory outside any IDEasy project, for example your home directory.
3. Run the `gui` commandlet from that directory.
4. Expected: the GUI window opens. Before this change the run aborted with `The gui commandlet requires an IDEasy project to work`.

Note that `ide-gui` must be resolvable from `USER_HOME/.m2` or a remote repository, since without `IDE_HOME` the Maven configuration folder falls back to `USER_HOME/.m2` (`Mvn.java:282-283`). On a SNAPSHOT developer installation you may need to run `mvn -pl gui -am install` first.

---

### Checklist for this PR

  with `internal`

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
@@ -50,6 +50,25 @@ public class Gui extends Commandlet {
     return false;
   }
 
+  /**
+   * Registers the {@link ToolInstallation#binDir() bin directory} of the given Maven installation in the {@link IdeContext#getPath() SystemPath} so that the
+   * Maven binary can be resolved.
+   * <p>
+   * This only applies without {@link IdeContext#getIdeHome() IDE_HOME}, where there is no project software folder to resolve Maven from. Inside a project the
+   * tool map is already filled from that folder, and since this commandlet installs the latest Maven rather than the configured one, registering it would
+   * override the version configured for the project.
+   *
+   * @param mvn the {@link Mvn} commandlet providing the tool name to register.
+   * @param mvnInstallation the {@link ToolInstallation} of Maven.
+   */
+  void registerMvnBinDir(Mvn mvn, ToolInstallation mvnInstallation) {
+
+    // The null check mirrors LocalToolCommandlet, since a null bin directory would end up in the tool map and break the next binary lookup.
+    if ((this.context.getIdeHome() == null) && (mvnInstallation.binDir() != null)) {
+      this.context.getPath().setPath(mvn.getName(), mvnInstallation.binDir());
+    }
+  }
+
   @Override
   protected void doRun() {
 
@@ -70,12 +89,7 @@ public class Gui extends Commandlet {
     ToolInstallation mvnInstallation = mvn.installTool(mavenToolInstallRequest);
     ToolInstallation javaInstallation = java.installTool(javaToolInstallRequest);
 
-    // Without IDE_HOME there is no project software folder, so SystemPath cannot resolve the mvn binary on its own.
-    // Registering the bin directory of the installation makes it findable, analogous to what LocalToolCommandlet does after a fresh installation.
-    // The null check mirrors LocalToolCommandlet, since a null bin directory would end up in the tool map and break the next binary lookup.
-    if (mvnInstallation.binDir() != null) {
-      this.context.getPath().setPath(mvn.getName(), mvnInstallation.binDir());
-    }
+    registerMvnBinDir(mvn, mvnInstallation);
 
     LOG.debug("Starting GUI via commandlet");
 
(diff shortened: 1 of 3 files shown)


cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -42,6 +42,33 @@ public String getName() {
...
+   * @param mvn the {@link Mvn} commandlet providing the tool name to register.
+   * @param mvnInstallation the {@link ToolInstallation} of Maven.
+   */
+  void registerMvnBinDir(Mvn mvn, ToolInstallation mvnInstallation) {
+
+    // The null check mirrors LocalToolCommandlet, since a null bin directory would end up in the tool map and break the next binary lookup.
+    if ((this.context.getIdeHome() == null) && (mvnInstallation.binDir() != null)) {
+      this.context.getPath().setPath(mvn.getName(), mvnInstallation.binDir());
+    }
+  }
```
hohwille, line 70: IMHO we do not need any of such things if we properly pass the same `ProcessContext` to both tool installations and then use that to run `mvn`.
Also the tool `mvn` has a dependency to `java`.
That we explicitly called an installation of `java` here was a temporary workaround until #1965 was implemented.
I am sorry if we have such dependencies and I do not have enough time to manage all our issues properly so that every developer has enough context to implement a story easily.
Can you get in touch with @laim2003 wo started all this and with whom I once discussed how the end-goal of this should look like.

quando632: Thanks for the feedback @hohwille, and no worries about the missing context.

While this PR was open, it seems like the core fix already landed on main:

#2310 made Gui.isIdeHomeRequired() return false, which is exactly what #2279 needed.
#2352 registers the Maven bin directory in the SystemPath, so mvn can be found without a project.
So this PR is obsolete and I would close it. #2279 can be closed as resolved by #2310.

PR #2023, opened 2026-06-15
#1594: Implement release commandlet

laert-ll:
### Implemented changes:

* Added ide release commandlet that automates building and deploying a release: derives the release version from .mvn/maven.config (-Drevision=), commits and tags it, runs the Maven deploy build, bumps to the next -SNAPSHOT version, and pushes after confirmation.
* Guards: aborts on uncommitted changes, warns when run on a fork, and requires the top-level project (or --force).
* Build goals are configurable via MVN_RELEASE_OPTS (default clean deploy -Dchangelist= -Pdeploy); failed builds can be retried.
* Added commit, tag, push, and retrieveGitRemotes to GitContext.

---

### Testing instructions

1. In a project using -Drevision= versioning (e.g. IDEasy itself), run ide release from the project root.
2. Confirm it proposes the correct release/next versions, then creates the set release version commit + release/<version> tag, builds, and creates the set next version commit.
3. Answer no at the final "Do you want to continue?" prompt to abort before pushing — verify nothing is pushed. To clean up: git reset --hard HEAD~2 && git tag -d release/<version>.

---

### Checklist for this PR

Make sure everything is checked before merging this PR. For further info please also see
our [DoD](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc).

- [x] When running `mvn clean test` locally all tests pass and build is successful
- [x] PR title is of the form `#«issue-id»: «brief summary»` (e.g. `#921: fixed setup.bat`). If no issue ID exists, title only.
- [x] PR top-level comment summarizes what has been done and contains link to addressed issue(s)
- [x] PR and issue(s) have suitable labels
- [x] Issue is set to `In Progress` and assigned to you *or* there is no issue (might happen for very small PRs)
- [x] You followed all [coding conventions](https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc)
- [x] You have added the issue implemented by your PR in [CHANGELOG.adoc](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc) unless issue is labeled
  with `internal`
- [x] You have formulated clear instructions on how to test your contribution under "Testing instructions"

PR #2302, opened 2026-08-11
#2178: Make Release commandlet build-tool independent

### LINKED ISSUE #2178: Make ReleaseCommandlet independent of specific build commandlet
### Task summary

As a cleanup of PR #2023 this story comes as a cleanup improvement.
Just like the `BuildCommandlet` also the `ReleaseCommandlet` should be independent of the actual build tool.
It should therefore ask the available build commandlets if they apply (just like `BuildCommandlet` does).
Then we can follow SoC, stop hardwiring `mvn` and become more open (less biased because current ReleaseCommandlet only supports maven and if we do not have a maven project, it will fail).

### Additional context

Ideally we should ask the commandlet manager for all commandlets that are build commandlets and then ask them for the build descriptor.
However, there is a trade-off and that is why currently `BuildCommandlet` still has a hardcoded list of build commandlets:

Both `npm` and `yarn` use `package.json` as build descriptor. The only additional indicator is the presence of a `yarn.lock` file.
If that is present, we should prefer `yarn`.
Now if we find commandlets in random order and first try `npm` before `yarn` then, it would find the descriptor and take the show instead of `yarn`.
On the other side, `npm` in general can work if only a `package.json` is present - independent from the presence of a `yarn.lock` file.
We could maybe consider adding special knowledge about `yarn` into `npm` that it would not apply if a `yarn.lock` was found.
Maybe thinking further there might even be a more elegant solution (some way to give `yarn` priority so it would be checked before `npm` without hardcoding such knowledge).

laert-ll:
### Implemented changes:

* Release Commandlet is now build-tool independent: it detecs the build tool from its build descriptor (like BuildCommandlet) and works througth the BuildTool interface instead of hardcoding Maven.
* Build Tool detection now in BuildCommandlet.findBuildCommandlet(...) with priority-ordered first-match, so yarn wins over npm when a yarn.lock is present.
* Top-level-project check is now based on the build descriptor rather than a hardcoded pom.xml.

---

### Testing instructions

1. Run the tests in the cli module: mvn -Dtest=BuildCommandletTest,ReleaseCommandletTest test.
2. In a Maven project (.mvn/maven.config with -Drevision=…-SNAPSHOT), run ide release and check the workflow is as described in #2023.
3. In a project with no supported build descriptor (or a Gradle-only project), run ide release and check that it fails with "Could not find a supported build tool…".

---

### Checklist for this PR

  with `internal`

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -7,6 +7,7 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 Release with new features and bugfixes:
 
 * https://github.com/devonfw/IDEasy/issues/2178[#2178]: Make ReleaseCommandlet independent of specific build commandlet and fix `ide build` using npm instead of yarn
+* https://github.com/devonfw/IDEasy/issues/2142[#2142]: Move IDE-specific metadata (.idea, .vscode) out of workspace
 * https://github.com/devonfw/IDEasy/issues/989[#989]: Allow expressions in template variable definitions
 
 The full list of changes for this release can be found in https://github.com/devonfw/IDEasy/milestone/50?closed=1[milestone 2026.09.002].
@@ -39,6 +40,7 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/1135[#1135]: Fix PowerShell env variable initialization on Windows by sourcing functions from the PowerShell profile
 * https://github.com/devonfw/IDEasy/issues/741[#741]: Add a warning message for legacy devonfw-ide settings users
 * https://github.com/devonfw/IDEasy/issues/1933[#1933]: Added a console panel to the GUI
+* https://github.com/devonfw/IDEasy/issues/1981[#1981]: Fall back to windows-x64 when no windows-arm64 artifact is published
 
 The full list of changes for this release can be found in https://github.com/devonfw/IDEasy/milestone/49?closed=1[milestone 2026.09.001].
 
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -45,6 +45,7 @@ The full list of changes for this release can be found in https://github.com/dev
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/2178[#2178]: Make ReleaseCommandlet independent of specific build commandlet
 * https://github.com/devonfw/IDEasy/issues/2197[#2197]: Fix broken python integration
 * https://github.com/devonfw/IDEasy/issues/2131[#2131]: Improve `ide upgrade --mode=` auto-completion
 * https://github.com/devonfw/IDEasy/issues/1870[#1870]: Add generic get-version implementation for global tools under windows
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -6,7 +6,9 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/2178[#2178]: Make ReleaseCommandlet independent of specific build commandlet and fix `ide build` using npm instead of yarn
 * https://github.com/devonfw/IDEasy/issues/989[#989]: Allow expressions in template variable definitions
+
 
 The full list of changes for this release can be found in https://github.com/devonfw/IDEasy/milestone/50?closed=1[milestone 2026.09.002].
 
@@ -45,7 +47,6 @@ The full list of changes for this release can be found in https://github.com/dev
 
 Release with new features and bugfixes:
 
-* https://github.com/devonfw/IDEasy/issues/2178[#2178]: Make ReleaseCommandlet independent of specific build commandlet
 * https://github.com/devonfw/IDEasy/issues/2197[#2197]: Fix broken python integration
 * https://github.com/devonfw/IDEasy/issues/2131[#2131]: Improve `ide upgrade --mode=` auto-completion
 * https://github.com/devonfw/IDEasy/issues/1870[#1870]: Add generic get-version implementation for global tools under windows
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -9,7 +9,6 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/2178[#2178]: Make ReleaseCommandlet independent of specific build commandlet and fix `ide build` using npm instead of yarn
 * https://github.com/devonfw/IDEasy/issues/989[#989]: Allow expressions in template variable definitions
 
-
 The full list of changes for this release can be found in https://github.com/devonfw/IDEasy/milestone/50?closed=1[milestone 2026.09.002].
 
 == 2026.09.001
--- a/cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
@@ -46,7 +46,7 @@ public class BuildCommandlet extends Commandlet {
 
     LocalToolCommandlet commandlet = this.context.getCommandletManager().findBuildTool(buildPath);
     if (commandlet == null) {
-      throw new CliException("Could not find build descriptor - no pom.xml, build.gradle, or package.json found!");
+      throw new CliException("Could not find a build descriptor in " + buildPath + " - no supported build tool detected.");
     }
     List<String> args = this.arguments.asList();
     if (args.isEmpty()) {
--- a/cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
@@ -8,17 +8,11 @@ import com.devonfw.tools.ide.cli.CliException;
 import com.devonfw.tools.ide.context.IdeContext;
 import com.devonfw.tools.ide.property.StringProperty;
 import com.devonfw.tools.ide.tool.LocalToolCommandlet;
-import com.devonfw.tools.ide.tool.gradle.Gradle;
-import com.devonfw.tools.ide.tool.mvn.Mvn;
-import com.devonfw.tools.ide.tool.npm.Npm;
-import com.devonfw.tools.ide.tool.yarn.Yarn;
 
 /**
  * Build tool {@link Commandlet} for automatically detecting build configuration files and running the respective tool.
  */
 public class BuildCommandlet extends Commandlet {
-
-  private static final List<Class<? extends LocalToolCommandlet>> BUILD_TOOLS = List.of(Mvn.class, Gradle.class, Yarn.class, Npm.class);
 
   /** The explicit build options to use (if empty use defaults). */
   public final StringProperty arguments;
@@ -50,7 +44,7 @@ public class BuildCommandlet extends Commandlet {
       throw new CliException("Missing current working directory!");
     }
 
-    LocalToolCommandlet commandlet = findBuildCommandlet(this.context, buildPath);
+    LocalToolCommandlet commandlet = this.context.getCommandletManager().findBuildTool(buildPath);
     if (commandlet == null) {
       throw new CliException("Could not find build descriptor - no pom.xml, build.gradle, or package.json found!");
     }
@@ -60,28 +54,6 @@ public class BuildCommandlet extends Commandlet {
       args = getDefaultToolOptions(variableName);
     }
     commandlet.runTool(args);
-  }
-
-  /**
-   * Detects the applicable build tool for the given {@code buildPath} by {@link LocalToolCommandlet#findBuildDescriptor(Path) querying} the available build
-   * commandlets (in order of priority) for a matching build descriptor (e.g. {@code pom.xml}, {@code build.gradle} or {@code package.json}).
-   *
-   * @param context the {@link IdeContext}.
-   * @param buildPath the {@link Path} to the directory to build.
-   * @return the applicable build {@link LocalToolCommandlet} or {@code null} if no build descriptor was found.
-   */
-  static LocalToolCommandlet findBuildCommandlet(IdeContext context, Path buildPath) {
-
-    if (buildPath == null) {
-      return null;
-    }
-    for (Class<? extends LocalToolCommandlet> toolClass : BUILD_TOOLS) {
-      LocalToolCommandlet toolCommandlet = context.getCommandletManager().getCommandlet(toolClass);
-      if (toolCommandlet.findBuildDescriptor(buildPath) != null) {
-        return toolCommandlet;
-      }
-    }
-    return null;
   }
 
   private List<String> getDefaultToolOptions(String buildOptionName) {
--- a/cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
@@ -72,6 +72,9 @@ public class BuildCommandlet extends Commandlet {
    */
   static LocalToolCommandlet findBuildCommandlet(IdeContext context, Path buildPath) {
 
+    if (buildPath == null) {
+      return null;
+    }
     for (Class<? extends LocalToolCommandlet> toolClass : BUILD_TOOLS) {
       LocalToolCommandlet toolCommandlet = context.getCommandletManager().getCommandlet(toolClass);
       if (toolCommandlet.findBuildDescriptor(buildPath) != null) {
--- a/cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
@@ -50,23 +50,35 @@ public class BuildCommandlet extends Commandlet {
       throw new CliException("Missing current working directory!");
     }
 
-    List<String> args = this.arguments.asList();
-    LocalToolCommandlet commandlet = null;
-    for (Class<? extends LocalToolCommandlet> toolClass : BUILD_TOOLS) {
-      LocalToolCommandlet toolCommandlet = this.context.getCommandletManager().getCommandlet(toolClass);
-      Path buildDescriptor = toolCommandlet.findBuildDescriptor(buildPath);
-      if (buildDescriptor != null) {
-        commandlet = toolCommandlet;
-        if (args.isEmpty()) {
-          String variableName = commandlet.getName().toUpperCase(Locale.ROOT) + "_BUILD_OPTS";
-          args = getDefaultToolOptions(variableName);
-        }
-      }
-    }
+    LocalToolCommandlet commandlet = findBuildCommandlet(this.context, buildPath);
     if (commandlet == null) {
       throw new CliException("Could not find build descriptor - no pom.xml, build.gradle, or package.json found!");
     }
+    List<String> args = this.arguments.asList();
+    if (args.isEmpty()) {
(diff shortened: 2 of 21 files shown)


cli/src/main/java/com/devonfw/tools/ide/commandlet/ReleaseCommandlet.java
```diff
@@ -41,15 +41,20 @@ protected void doRun() {
 
     Path projectPath = this.context.getCwd();
     GitContext git = this.context.getGitContext();
-    Mvn buildTool = this.context.getCommandletManager().getCommandlet(Mvn.class);
+
+    LocalToolCommandlet commandlet = BuildCommandlet.findBuildCommandlet(this.context, projectPath);
+    if (!(commandlet instanceof BuildTool buildTool)) {
+      throw new CliException("Could not find a supported build tool to release the project in " + projectPath + ".");
+    }
```
maybeec, line 51: **Should-fix** — this branch folds two genuinely different failures into one message:

* `commandlet == null` → there is no build descriptor here at all, so there is nothing to release;
* `commandlet` is non-`null` but not a `BuildTool` → we *did* detect the project's build tool (e.g. `Gradle` via `build.gradle`), it simply has no release support yet.

A user standing in a Gradle project is told "Could not find a supported build tool", which reads as "your project is broken" when the truth is "gradle release is not implemented". Naming the detected tool turns a dead end into an actionable message — and it is the difference between the user filing a bug and the user opening a feature request.

```suggestion
    LocalToolCommandlet commandlet = BuildCommandlet.findBuildCommandlet(this.context, projectPath);
    if (commandlet == null) {
      throw new CliException("Could not find a build descriptor in " + projectPath + " - there is nothing to release here.");
    }
    if (!(commandlet instanceof BuildTool buildTool)) {
      throw new CliException("The build tool " + commandlet.getName() + " detected in " + projectPath + " does not support releasing.");
    }
```

Note this also makes `ReleaseCommandletTest#testReleaseWithoutBuildDescriptor` (which currently only asserts `CliException.class`) distinguishable from your new `testReleaseWithUnsupportedBuildToolThrowsException` — worth tightening that older test with a `hasMessageContaining` too, the way you already did for the new one.

maybeec (commented):
Thanks for picking this up 👍 — and thanks especially for not stopping at the literal ask. Turning the detection loop from *last match wins* into *first match wins* (`BuildCommandlet#findBuildCommandlet`) quietly fixes two real bugs that were sitting in `BuildCommandlet`:

* In a yarn project (`package.json` + `yarn.lock`) the old loop had no `break`, so after `Yarn` matched it kept going and `Npm` — last in `BUILD_TOOLS` — always took over. `Yarn.findBuildDescriptor` was effectively unreachable.
* Worse, in a polyglot repo (`pom.xml` **and** `package.json`) the old code ran `npm` but resolved the defaults from `MVN_BUILD_OPTS`, i.e. it executed `npm clean install`. Hoisting the `args.isEmpty()` block out of the loop fixes that too.

Also good: the `parent == null` guard in `isTopLevelProject` — the old `projectPath.getParent().resolve("pom.xml")` would have thrown a raw NPE when releasing from a filesystem root.

**No blockers.** The `instanceof BuildTool` gate is the right shape, `CliException` is the correct exception type here (expected, user-facing abort), no public signature was broken, and the tests stay inside `target`. A few things I'd like to see before merge:

### Should-fix
1. `ReleaseCommandlet:45-48` — the error message collapses *"no build descriptor at all"* and *"build tool found but it cannot release"* into one text. A Gradle user gets told nothing useful.
2. `BuildCommandlet:73` — the new detection is a package-private `static` that `ReleaseCommandlet` reaches across into, and it takes an `IdeContext` parameter although both callers already have `this.context`. The issue itself points at the better home: *"Ideally we should ask the commandlet manager"*.
3. `BuildCommandlet:75` — `findBuildCommandlet` NPEs on a `null` path; `BuildCommandlet` guards before calling, `ReleaseCommandlet` does not.
4. `BuildCommandletTest:142` — the yarn-over-npm fix is the user-visible part of this PR but is only covered at helper level; no test asserts that `ide build` in the yarn workspace actually runs `yarn`, and nothing covers the mvn-vs-npm precedence flip.

### Minor
5. `ReleaseCommandlet:106` — first condition is always true at the only call site.
6. `ReleaseCommandlet:102` — parameter named `buildTool` but typed `LocalToolCommandlet`, while `doRun` has a *different* variable also called `buildTool`.
7. `BuildCommandlet:55` — error message still hardcodes the descriptor list that this PR is de-hardcoding.
8. `CHANGELOG.adoc:9` — the `ide build` precedence fix is user-facing and not mentioned.

### PR hygiene
* The branch is **BEHIND** `main` and needs an update before merge.
* The checklist ticks *"PR and issue(s) have suitable labels"*, but this PR carries no labels and no milestone. Per [DoD.adoc](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc) both should be set before close (the issue itself is labelled fine).

### Manual verification
Green CI is not sufficient for this one. `ide release` drives `git commit`/`git tag`/`git push` and a real deploy build — none of which the mock `GitContextImplMock` exercises end-to-end. Please do run your own testing instructions (2) and (3) on a real Maven project before merge, and note the result in the PR.

Scope against #2178: the primary ask ("`ReleaseCommandlet` should ask the available build commandlets if they apply") is **met**. The *Additional context* part — discovering build commandlets dynamically via the commandlet manager and expressing the yarn/npm priority without a hardcoded list — is **not** addressed; `BUILD_TOOLS` is untouched. That is explicitly framed as "ideally" in the issue, so I am not blocking on it, but please say in the PR whether you consider it out of scope so it can be tracked as a follow-up rather than silently dropped.

cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
```diff
@@ -50,25 +50,37 @@ protected void doRun() {
...
 
+  /**
+   * Detects the applicable build tool for the given {@code buildPath} by {@link LocalToolCommandlet#findBuildDescriptor(Path) querying} the available build
+   * commandlets (in order of priority) for a matching build descriptor (e.g. {@code pom.xml}, {@code build.gradle} or {@code package.json}).
+   *
+   * @param context the {@link IdeContext}.
+   * @param buildPath the {@link Path} to the directory to build.
+   * @return the applicable build {@link LocalToolCommandlet} or {@code null} if no build descriptor was found.
+   */
+  static LocalToolCommandlet findBuildCommandlet(IdeContext context, Path buildPath) {
```
maybeec, line 73: **Should-fix** — placement. Two things about this signature:

1. It takes `IdeContext context` as a parameter although *both* call sites already hold `this.context`. Passing the context around when the owning object already has it injected is the pattern we avoid across the codebase.
2. It makes `ReleaseCommandlet` depend on a package-private `static` inside a sibling *commandlet*. Commandlets are peers dispatched by the same manager, so one reaching into another's internals is exactly the separation-of-concerns smell #2178 set out to remove.

#2178 names the better home directly: *"Ideally we should ask the commandlet manager for all commandlets that are build commandlets and then ask them for the build descriptor."* `CommandletManager` already exposes `getCommandlets()`, so a `LocalToolCommandlet findBuildCommandlet(Path)` there would keep `BUILD_TOOLS` and the priority ordering as an implementation detail of the manager, and both callers collapse to:

```java
LocalToolCommandlet commandlet = this.context.getCommandletManager().findBuildCommandlet(projectPath);
```

That would also give `Intellij#importRepository` (`Intellij.java:158-172`) — which today runs a *third* copy of this "iterate build tools, first descriptor wins" loop over its own `BUILD_TOOL_TO_IJ_TEMPLATE` map — somewhere to converge later.

If you would rather not touch `CommandletManager` in this PR, the minimum I would ask for is dropping the `context` parameter and making this an instance method, with `ReleaseCommandlet` obtaining the commandlet the normal way (`getCommandlet(BuildCommandlet.class)`). Not blocking on the exact home — but the reach-across plus redundant parameter should go.

cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
```diff
@@ -50,25 +50,37 @@ protected void doRun() {
...
+
+    for (Class<? extends LocalToolCommandlet> toolClass : BUILD_TOOLS) {
```
maybeec, line 75: **Should-fix** — null contract. `Mvn.findBuildDescriptor` does `directory.resolve(POM_XML)`, so a `null` `buildPath` produces a raw `NullPointerException` from inside the loop rather than a clean CLI error.

`BuildCommandlet#doRun` guards this at line 49 *before* calling, but `ReleaseCommandlet#doRun` now calls `findBuildCommandlet(this.context, this.context.getCwd())` as its very first use of the cwd with no such guard. Previously `git.hasUntrackedFiles(projectPath)` was the first consumer, so this is not a regression you introduced — but extracting the shared helper is precisely the moment to fix the asymmetry, rather than leaving each caller to remember.

Moving the check into the helper covers both callers and keeps `doRun` shorter:

```java
static LocalToolCommandlet findBuildCommandlet(IdeContext context, Path buildPath) {

  if (buildPath == null) {
    throw new CliException("Missing current working directory!");
  }
  for (Class<? extends LocalToolCommandlet> toolClass : BUILD_TOOLS) {
```

The existing `testBuildWithNoCwd` keeps passing, and `ide release` with no cwd stops NPE-ing. See [coding-conventions.adoc § Avoid catching NPE](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc) — the rule there is to check for `null` explicitly instead of letting an NPE happen, and constructing an exception with a full stack collect is the expensive path we should not take for a condition a single `if` covers.

cli/src/test/java/com/devonfw/tools/ide/commandlet/BuildCommandletTest.java
```diff
@@ -118,4 +124,22 @@ void testBuildWithNoBuildFile() {
...
+  void testFindBuildCommandlet() {
+
+    IdeTestContext context = newContext(PROJECT_BUILD);
+    Path workspace = context.getWorkspacePath();
+
+    assertThat(BuildCommandlet.findBuildCommandlet(context, workspace.resolve("mvn"))).isInstanceOf(Mvn.class);
+    assertThat(BuildCommandlet.findBuildCommandlet(context, workspace.resolve("gradle"))).isInstanceOf(Gradle.class);
+    assertThat(BuildCommandlet.findBuildCommandlet(context, workspace.resolve("npm"))).isInstanceOf(Npm.class);
+    // both npm and yarn match package.json, but yarn.lock is present so yarn must take precedence over npm
+    assertThat(BuildCommandlet.findBuildCommandlet(context, workspace.resolve("yarn"))).isInstanceOf(Yarn.class);
```
maybeec, line 142: **Should-fix** — coverage gap on the behaviour that actually changed for users.

This asserts the *helper* returns `Yarn`, which is good, but the previously broken behaviour was one level up: because the old loop in `doRun` had no `break`, a yarn project was really built by **npm**. Nothing here asserts that `ide build` in `workspaces/main/yarn` now logs `yarn run build`. The existing `testNpmBuildWithProvidedArguments` shows the shape (`IdeLogEntry.ofInfo("npm start test")`); a yarn equivalent would pin the fix so a future reordering of `BUILD_TOOLS` cannot silently undo it. I appreciate that needs `_ide/urls/yarn` + repository fixtures in the `build` test project, so it is more work than one line — but it is the assertion that proves the bug is dead.

The second uncovered flip is quieter and worth at least a helper-level assertion here, since it costs nothing: a directory containing **both** `pom.xml` and `package.json`. Old code ran `npm` with `MVN_BUILD_OPTS` (`npm clean install`); new code correctly picks `Mvn`. Adding a `mvn-and-npm` fixture folder and one more line:

```java
// a polyglot project must be built by the highest-priority tool, not the last one that matches
assertThat(BuildCommandlet.findBuildCommandlet(context, workspace.resolve("mvn-and-npm"))).isInstanceOf(Mvn.class);
```

would document the priority contract that `BUILD_TOOLS` now carries.

cli/src/main/java/com/devonfw/tools/ide/commandlet/ReleaseCommandlet.java
```diff
@@ -94,14 +99,15 @@ private boolean warnIfFork(GitContext git, Path projectPath) {
...
 
-  private boolean isTopLevelProject(Path projectPath) {
+  private boolean isTopLevelProject(LocalToolCommandlet buildTool, Path projectPath) {
 
-    // returns false in case there's no pom.xml present or if parent directory has a pom.xml
-    return Files.exists(projectPath.resolve("pom.xml"))
-        && !Files.exists(projectPath.getParent().resolve("pom.xml"));
+    // top-level if a build descriptor is present here but not in the parent directory
+    Path parent = projectPath.getParent();
+    return (buildTool.findBuildDescriptor(projectPath) != null)
```
maybeec, line 109: **Minor** — the first operand is dead. At the only call site (line 57) `commandlet` came out of `findBuildCommandlet(this.context, projectPath)`, which returns a commandlet *only* when `findBuildDescriptor(projectPath) != null`. So this re-runs a filesystem `Files.exists` check whose answer is already known to be `true`.

The method reduces to the question it is actually asking:

```suggestion
    // top-level if the build descriptor found here is not also present in the parent directory
    Path parent = projectPath.getParent();
    return (parent == null) || (buildCommandlet.findBuildDescriptor(parent) == null);
```

(uses the renamed parameter from the comment above). Non-blocking — the current code is correct, just doing redundant I/O.

cli/src/main/java/com/devonfw/tools/ide/commandlet/ReleaseCommandlet.java
```diff
@@ -94,14 +99,15 @@ private boolean warnIfFork(GitContext git, Path projectPath) {
     return false;
   }
 
-  private boolean isTopLevelProject(Path projectPath) {
+  private boolean isTopLevelProject(LocalToolCommandlet buildTool, Path projectPath) {
```
maybeec, line 105: **Minor** — naming. This parameter is called `buildTool` but is typed `LocalToolCommandlet`, while `doRun` binds a *different* variable also named `buildTool` — of type `BuildTool` (line 46) — to the very same instance. Two names for one object, and the shared name is the one that does *not* match this type. Reading `isTopLevelProject` in isolation, `buildTool` suggests you are holding the `BuildTool` interface, which is the type that cannot answer `findBuildDescriptor`.

```suggestion
  private boolean isTopLevelProject(LocalToolCommandlet buildCommandlet, Path projectPath) {
```

(and the two usages in the body). [coding-conventions.adoc § Naming](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc) — "always use short but speaking names"; here the name actively points at the wrong abstraction.

cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
```diff
@@ -50,25 +50,37 @@ protected void doRun() {
...
-        commandlet = toolCommandlet;
-        if (args.isEmpty()) {
-          String variableName = commandlet.getName().toUpperCase(Locale.ROOT) + "_BUILD_OPTS";
-          args = getDefaultToolOptions(variableName);
-        }
-      }
-    }
+    LocalToolCommandlet commandlet = findBuildCommandlet(this.context, buildPath);
     if (commandlet == null) {
       throw new CliException("Could not find build descriptor - no pom.xml, build.gradle, or package.json found!");
```
maybeec: **Minor** — not introduced by you, but this PR is the one that de-hardcodes build-tool knowledge, so it stands out now. This message enumerates the descriptors by hand and has already drifted: `Gradle.findBuildDescriptor` also accepts `build.gradle.kts`, and it says nothing about `yarn.lock`. Every future entry in `BUILD_TOOLS` has to remember to edit this string, which is exactly the coupling the new helper removes elsewhere.

Since `findBuildCommandlet` already walks the tools, the message could be derived instead of restated — or, cheapest, drop the enumeration:

```suggestion
      throw new CliException("Could not find a build descriptor in " + buildPath + " - no supported build tool detected.");
```

Entirely optional and safe to leave for a follow-up.

CHANGELOG.adoc
```diff
@@ -6,6 +6,7 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/2178[#2178]: Make ReleaseCommandlet independent of specific build commandlet
```
maybeec, line 16: **Minor** — the entry only advertises the `ReleaseCommandlet` refactoring, but the most visible effect of this PR for existing users is on `ide build`: yarn projects were being built with npm, and polyglot repos were being built with the wrong tool *and* the wrong default options. That is a user-facing bugfix and per [DoD.adoc](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc) belongs in the changelog, otherwise nobody upgrading will connect a changed `ide build` behaviour to this issue.

```suggestion
* https://github.com/devonfw/IDEasy/issues/2178[#2178]: Make ReleaseCommandlet independent of specific build commandlet and fix `ide build` using npm instead of yarn
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
hohwille: The `BuildCommandlet` is abstract and cannot be instantiated.
For 1. this could be moved to CommandletManager, or keep the IdeContext and move it to `BuildTool` better following SoC.
For 2. I would simply make the method public or you add a public method to `CommandletManager` that delegates to this method with limited visibility - but according to my experience with limited Java visibilities without using modules (JPMS) this is not really making it better.

cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
hohwille: If we already return `null` if nothing was found, we can even return `null` if `buildPath` was `null` since we cannot find anything in a non-existent path.

cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
hohwille: Shouldn't the method be renamed to `findBuildTool` and return type be `BuildTool`?
IMHO the current API is confusing.

cli/src/test/java/com/devonfw/tools/ide/commandlet/BuildCommandletTest.java
laert-ll: I added the polyglot fixture and assertion, thanks for the suggestion.

Regarding the yarn end-to-end test, I think the best decision would be to skip it. The loop that caused the bug no longer exists. To test it end-to-end one would have to fake a whole yarn install in the test project, which is not worth it in my opinion. However feel free to let me know if you disagree.

cli/src/main/java/com/devonfw/tools/ide/commandlet/BuildCommandlet.java
laert-ll: I moved it to `CommandletManager` as a public `findBuildTool(Path)`. The `IdeContext` parameter is gone now and the `BUILD_TOOLS` list is private inside the manager, so `ReleaseCommandlet` doesn't have to reach into `BuildCommandlet` anymore.

I left the return type as `LocalToolCommandlet`. Only `Mvn` implements `BuildTool`, so returning that would mean gradle, npm and yarn never get found. It also wouldn't compile, since `BuildCommandlet` calls `getName()` and `runTool()` and those aren't on the interface.

hohwille (approved):
@laert-ll thanks for your PR. This looks great now and is ready for merge. 👍 
I will create a follow up to prevent the typing and avoiding the casts.
Update: Done as #2420

Issue #2278, opened 2026-08-06
Restructure main-view.fxml into seperate layout files


laim2003:
### Task summary

Right now, the main_view.fxml file is a considerably big file with lots of contents. As the UI grows and grows, it becomes harder and harder to maintain the UI file this way. Therefore, we should try to divide the UI into smaller files (or reusable components where it makes sense), in order to make solving merge conflicts etc. easier.

### Additional context

FXML has an `<include>` tag that allows to include external layout files in another layout file.

PR #2306, opened 2026-08-11
#2296: Refactor ToolCommandlet to determine InstalledEdition and Version in a Single Cached Lookup

### LINKED ISSUE #2296: Improve support to get installed version and edition
### Task summary

This issue is created from review of PR #2244 
The concrete problem we identified is that Docker commandlet is still not able to correctly determine the installed version (e.g. in case Rancher Desktop edition is installed).
A first step (e.g. as subtask and according PR) would be to support a List of values for getWindowsRegistryAppName and try them all sequentially until the first hit was found).

However, thinking this even further, we could think about the entire process of how we determine both the installed edition and version. IMHO both information are closely related and in most cases require the same kind of tests that could be "expensive" (take several seconds).
For some commandlets we already introduced the usage of `CachedValue` for getting the installed version.
I would suggest a general refactoring:
* Introduce a simple record `EditionAndVersion` that only contains `String edition` and `VersionIdentifier version`.
* I was thinking if we can reuse `ToolEditionAndVersion` but that is a slightly different animal as it also has `version` as `GenericVersionRange` as well as a `resolvedVersion` as `VersionIdentifier`.
* Now add a method `public EditionAndVersion getInstalledEditionAndVersion()` to `ToolCommandlet`.
* We can already implement this method using a `CachedValue` delegating to a `computeInstalledEditionAndVersion()` (see pattern from `PackageManagerBasedLocalToolCommandlet`).
* Now we can implement `getInstalledVersion()` and `getInstalledEdition()` to delegate to the new method returning the requested result (without causing NPE of course). Make the methods final so we reveal compile errors in all overriding methods.
* Implement `computeInstalledEditionAndVersion()` for `LocalToolCommandlet`: The way we resolve the installed edition before this change is via resolving the software link target. If we already do this to get the edition, we can already get the version in the same way. Only if there is no such link target (if `isIgnoreSoftwareRepo()` is true or we have a devonfw-ide legacy project), we can keep the fallback to read the version file and guess the edition as the tool name. Here we should also integrate with issue #2284
* Now in `GlobalToolCommandlet` we could change that instead of returning a `String` for `getWindowsRegistryAppName` we could return a `Map<String, String>` that maps an edition to the registry app name. For simplicity we could keep the old method that can be implemented when there is only one edition. But for docker we could directly implement the new method returning `Map.of("docker", "Docker Desktop", "rancher", "Rancher Desktop")`.
* Then we could iterate over the map and search in the registry for the iterated value. If we find a match from the registry, we also know the edition from the key of the map entry.

### Additional context

_No response_

krystynaShatkovska:
### Implemented changes:

   This PR refactors the way IDEasy determines the installed edition and version of tools. Previously, `getInstalledEdition()` and `getInstalledVersion()` were separate methods
   that often performed redundant expensive lookups (registry queries, process execution, etc.). The new combined approach computes both values in a single call.

   - Introduced the record `EditionAndVersion(String edition, VersionIdentifier version)` to hold both values together
   - Added `getInstalledEditionAndVersion()` with `CachedValue` to `ToolCommandlet` — delegates to `computeInstalledEditionAndVersion()`
   - Made `getInstalledVersion()` and `getInstalledEdition()` `final` — they now delegate to the combined method
   - Implemented `computeInstalledEditionAndVersion()` in `LocalToolCommandlet` (resolves edition + version via software link target)
   - `GlobalToolCommandlet`: `getWindowsRegistryAppNames()` returns `Map<String, String>` mapping edition → registry app name
   - `Docker`: returns `Map.of(\"docker\", \"Docker Desktop\", \"rancher\", \"Rancher Desktop\")` — enabling correct detection of Rancher Desktop edition
   - Implemented `computeInstalledEditionAndVersion()` in `LocalToolCommandlet` (resolves edition + version via software link target)
   - `GlobalToolCommandlet`: `getWindowsRegistryAppNames()` returns `Map<String, String>` mapping edition → registry app name
   - `Docker`: returns `Map.of(\"docker\", \"Docker Desktop\", \"rancher\", \"Rancher Desktop\")` — enabling correct detection of Rancher Desktop edition
   - Updated all relevant subclasses (IdeasyCommandlet, KubeCtl, DelegatingToolCommandlet, PackageManagerBasedLocalToolCommandlet, NodeBasedCommandlet)
   - Added cache invalidation after installation via `invalidateInstalledEditionAndVersion()`
   - Fixed pre-existing GUI compilation issue (`IdeContext.findProjects` was package-private instead of public)

   Backward compatibility is maintained via deprecated hooks (`computeInstalledEdition()`, `computeInstalledVersion()`, `getInstalledVersionDeprecated()`,
   `getInstalledEditionDeprecated()`).
   
   ---
   
   ### Testing instructions
   
   1. Run \`mvn clean test\` — all 901 tests (CLI: 845, GUI: 56) pass with BUILD SUCCESS
   2. Test Docker edition detection: install Rancher Desktop and run \`ideasy docker info\` — should correctly report edition as \"rancher\" instead of \"docker\"
   3. Test global tool detection on Windows: run \`ideasy docker install\` and verify the registry lookup correctly identifies the installed edition
   
   ---
   
   ### Checklist for this PR
   
   - [x] When running \`mvn clean test\` locally all tests pass and build is successful
   - [x] PR title is of the form \`#«issue-id»: «brief summary»\`
   - [x] Issue is set to \`In Progress\` and assigned to you
   - [x] You have added the issue implemented by your PR in CHANGELOG.adoc
   - [x] You have not changed any dependency in \`pom.xml\` files
   - [x] You have formulated clear instructions on how to test your contribution under \"Testing instructions\"" 2>&1


CHANGELOG.adoc
laert-ll, line 1: You accidentally deleted some other changelog lines during your edit. Make sure to add the new line underneath the `026.08.002` line.

cli/src/main/java/com/devonfw/tools/ide/tool/kubectl/KubeCtl.java
```diff
@@ -28,15 +29,19 @@ public KubeCtl(IdeContext context) {
...
+      return super.computeInstalledEditionAndVersion();
     }
 
     List<String> outputs = this.context.newProcess().runAndGetOutput(this.tool, "version", "--client");
     String singleLineOutput = String.join("\n", outputs);
-    return resolveVersionWithPattern(singleLineOutput, KUBECTL_VERSION_PATTERN);
+    VersionIdentifier version = resolveVersionWithPattern(singleLineOutput, KUBECTL_VERSION_PATTERN);
+    if ((version == null) && isCommandAvailable("kubectl")) {
+      version = null;
+    }
```
laert-ll: `if ((version == null) && …) { version = null; }` assigns null to null. This is a no-op.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -366,7 +366,7 @@ protected void postExtract(Path extractedDir) {
   }
 
   @Override
-  public VersionIdentifier getInstalledVersion() {
+  protected VersionIdentifier getInstalledVersionDeprecated() {
```
laert-ll, line 385: The issue wanted edition and version resolved together in one pass here, but they're still looked up separately. Could we override `computeInstalledEditionAndVersion()` to do both at once?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -608,7 +612,8 @@ protected ToolInstallation createToolInstallation(Path rootDir, VersionIdentifie
   protected ToolInstallation createToolInstallation(Path rootDir, Path linkDir, Path binDir, VersionIdentifier version, boolean newInstallation,
       EnvironmentContext environmentContext, boolean additionalInstallation) {
 
-    // do not copy the version file into macOS .app bundles: changing the bundle after codesigning breaks the seal.
+    // Invalidate cached edition/version so that subsequent calls reflect the new installation
```
laert-ll, line 615: Looks like the macOS comment got removed by accident (if I'm not missing something).

```suggestion
    // Invalidate cached edition/version so that subsequent calls reflect the new installation
    // do not copy the version file into macOS .app bundles: changing the bundle after codesigning breaks the seal.
```

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -115,10 +121,16 @@ public VersionIdentifier getInstalledVersion() {
         LOG.error("Couldn't get installed version of " + this.getName());
       }
 
-      return parsedVersion;
+      return new EditionAndVersion("desktop", parsedVersion);
```
laert-ll: On Windows, super already worked out the edition from the registry, but we throw it away and hardcode "desktop". Could we keep super's edition and use one consistent name (the map uses "docker")?

cli/src/main/java/com/devonfw/tools/ide/tool/kubectl/KubeCtl.java
```diff
@@ -28,15 +29,19 @@ public KubeCtl(IdeContext context) {
   }
 
   @Override
-  public VersionIdentifier getInstalledVersion() {
+  protected EditionAndVersion computeInstalledEditionAndVersion() {
```
laert-ll: This override is never called. KubeCtl is a `DelegatingToolCommandlet<Docker>` and `DelegatingToolCommandlet` overrides the public `getInstalledEditionAndVersion()` (line 49) to delegate to Docker. Because `getInstalledVersion` is now final and goes through that public method, this `computeInstalledEditionAndVersion()` hook is unreachable, so `kubectl` reports Docker's version instead of `kubectl version --client`. On the main this worked because KubeCtl overrode the public getter directly. Please make delegating tools route through the compute hook (or override the public method here) and add a KubeCtl test asserting the version comes from `kubectl version --client`.

laert-ll (commented):
Thanks for your work! The overall direction is nice, and the EditionAndVersion with a single cached lookup is exactly what the issue was after. I've left a few inline comments on some details that need some improvement.

The main thing I noticed is that the branch seems to accidentally include some unrelated work (the GUI changes from #2278 and some changelog changes as well). Make sure to rebase to main to clean this up and fix the merge conflicts. Apart from that I would also suggest adding some tests for this.

Other than that, thanks for your contribution again!

cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
```diff
@@ -436,7 +436,7 @@ default List<Path> findProjects() {
    * @param ideRoot the IDE root containing the IDEasy projects.
    * @return the paths of all detected IDEasy projects.
    */
-  static List<Path> findProjects(Path ideRoot) {
+  public static List<Path> findProjects(Path ideRoot) {
```
laert-ll: I think this is accidental and not related to the PR, no?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -827,7 +866,57 @@ public boolean isInstalled() {
...
+
+  /**
+   * @deprecated Override {@link #computeInstalledEditionAndVersion()} instead.
+   * @return the installed edition.
+   */
+  @Deprecated
+  protected String getInstalledEditionDeprecated() {
+
+    return null;
+  }
```
hohwille, line 913: do we still need this or was that only and intermediate state during refactoring?

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -827,7 +866,57 @@ public boolean isInstalled() {
...
+
+  /** Invalidates the cached installed edition and version so the next call to {@link #getInstalledEditionAndVersion()} recomputes the result. */
+  protected void invalidateInstalledEditionAndVersion() {
+
+    this.installedEditionAndVersion = null;
```
hohwille, line 918:
```suggestion
    if (this.installedEditionAndVersion != null) {
      this.installedEditionAndVersion.invalidate();
    }
```

BTW: I would even suggest to make `installedEditionAndVersion` final and initialize it in the constructor since the creation is cheap and does not compute the value. Then you can skip such `if` statements and also the lazy init in getInstalledEditionAndVersion() is not needed any more. BTW: In the GUI we will have more concurrency and initialization in constructor is always thread-safe while lazy init in getter could cause race condition.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -366,7 +366,23 @@ protected void postExtract(Path extractedDir) {
...
+  protected EditionAndVersion computeInstalledEditionAndVersion() {
+
+    Path toolPath = getToolPath();
+    if (isToolNotInstalled(toolPath)) {
+      return null;
+    }
+    // Resolve edition and version from a single tool-path lookup (one pass) instead of two separate lookups.
+    String edition = getInstalledEdition(toolPath);
+    VersionIdentifier version = getInstalledVersion(toolPath);
+    if ((edition == null) && (version == null)) {
```
hohwille, line 378: Just for the record: having an edition here but no version seems like an inconsistent state to me.
Hopefully never happens but wouldn't this be more consistent?

```suggestion
    if (version == null) {
```

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -96,54 +99,65 @@ protected List<PackageManagerCommand> getInstallPackageManagerCommands() {
...
 
   private VersionIdentifier getDockerDesktopVersionLinux() {
 
     String dockerDesktopVersionLinuxCommand = "apt list --installed | grep docker-desktop | awk '{print $2}'";
-    String output = this.context.newProcess().runAndGetSingleOutput("bash", "-lc", dockerDesktopVersionLinuxCommand);
-    return super.resolveVersionWithPattern(output, DOCKER_DESKTOP_LINUX_VERSION_PATTERN);
+    // Log a warning and return null (instead of throwing) when the command produces no usable output, e.g. when
+    // Docker Desktop is not installed via apt.
+    String output = this.context.newProcess().runAndGetSingleOutput(IdeLogLevel.WARNING, "bash", "-lc", dockerDesktopVersionLinuxCommand);
+    return (output != null) ? super.resolveVersionWithPattern(output, DOCKER_DESKTOP_VERSION_PATTERN) : null;
```
hohwille, line 145: Not your fault - was already like this before but prefixing method calls with `super` should only be used when overridden as otherwise this is only causing confusion:

```suggestion
    return (output != null) ? resolveVersionWithPattern(output, DOCKER_DESKTOP_VERSION_PATTERN) : null;
```

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -96,54 +99,65 @@ protected List<PackageManagerCommand> getInstallPackageManagerCommands() {
...
-  private VersionIdentifier getRancherDesktopClientVersion() {
+  private VersionIdentifier getDockerDesktopVersionMac() {
 
-    String output = this.context.newProcess().runAndGetSingleOutput("rdctl", "version");
-    return super.resolveVersionWithPattern(output, RDCTL_CLIENT_VERSION_PATTERN);
+    String dockerDesktopVersionMacCommand = "plutil -extract CFBundleShortVersionString raw /Applications/Docker.app/Contents/Info.plist";
+    // Log a warning and return null (instead of throwing) when the command produces no usable output, e.g. when
+    // Docker Desktop is not installed at /Applications/Docker.app.
+    String output = this.context.newProcess().runAndGetSingleOutput(IdeLogLevel.WARNING, "bash", "-lc", dockerDesktopVersionMacCommand);
+    return (output != null) ? super.resolveVersionWithPattern(output, DOCKER_DESKTOP_VERSION_PATTERN) : null;
```
hohwille, line 154:
```suggestion
    return (output != null) ? resolveVersionWithPattern(output, DOCKER_DESKTOP_VERSION_PATTERN) : null;
```

cli/src/main/java/com/devonfw/tools/ide/tool/kubectl/KubeCtl.java
```diff
@@ -28,15 +29,16 @@ public KubeCtl(IdeContext context) {
   }
 
   @Override
-  public VersionIdentifier getInstalledVersion() {
+  public EditionAndVersion getInstalledEditionAndVersion() {
```
hohwille, line 32: Shouldn't this happen in `computeInstalledEditionAndVersion()`?
If you make `getInstalledEditionAndVersion()` final the compiler will help you to find such things.

hohwille (approved):
@krystynaShatkovska thanks for your PR. You got the story right and made your way perfectly through this. Great that you added missing tests for docker and kubectl. Maybe your mockito approach is worth sharing and presenting to the team. Great job 👍 

I left some comments for rework.

Follow up (not part of this PR):
Once comments are resolved and this is merged, we should have the foundation to rework our `.ide.software.version` issues. For software installed via repository we can then resolve the version together with the edition from the symlink. A fallback to `.ide.software.version` could only happen for devonfw-ide legacy compatibility if no (sym)link was found.
For other tools like Python we could simply add some code that determines the version from cli (e.g. `python --version`) and auto-recovers a lost .ide.software.version` file.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
krystynaShatkovska: Yes, you are right it was the migration bridge from the two separate lookups to the single cached one. I finished that migration and removed this

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
krystynaShatkovska: Thank you for the suggestion. I went with the full version. installedEditionAndVersion is now final, created in the constructor, so I could drop the lazy-init in getInstalledEditionAndVersion() and the null-check in invalidateInstalledEditionAndVersion(). This also removes the lazy-init race you pointed out for the GUI.

cli/src/main/java/com/devonfw/tools/ide/tool/kubectl/KubeCtl.java
krystynaShatkovska: Yes, you are right, fixed. The kubectl logic now in computeInstalledEditionAndVersion() (it still falls back to super i.e. the Docker delegate when kubectl is not available), and getInstalledEditionAndVersion() is now final ToolCommandlet, so the compiler rejects any future bypass.

hohwille (approved):
@krystynaShatkovska thanks for your update. Now everything is perfect and ready for merge. 👍 
However, there are two merge conflicts preventing the merge.
The first one looks not like something I should resolve in the browser.
Could you have a look and resolve the conflicts? Thanks 🙏

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -260,26 +261,34 @@ public String getWindowsRegistryAppName() {
...
+
   @Override
-  public VersionIdentifier getInstalledVersion() {
+  protected EditionAndVersion computeInstalledEditionAndVersion() {
+
     if (this.context.getSystemInfo().isLinux()) {
-      return getNativePackageVersion();
+      // on Linux global tools are typically installed via the package manager of the OS
+      VersionIdentifier version = getNativePackageVersion();
+      return (version != null) ? new EditionAndVersion(null, version) : null;
```
hohwille: Here we set the edition to `null`.
This will result in `getInstalledEdition()` to always return `null` what implies the tool is not installed.
My suggestion would be the following:
1. Start with a pragmatich fix:
```suggestion
      return (version != null) ? new EditionAndVersion(this.tool, version) : null;
```
2. Please create a ticket to come up with a better solution for `GlobalToolCommandlet.computeInstalledEditionAndVersion`. E.g. for Linux we now have the new solution for `NativePackage` - here we could add the edition so that `getNativePackageVersion()` method can be refactored to also determine the edition. For Windows we have a similar concept to support multiple editions in a Map that maps to the app name. The PR for mac is still in progress and I have not reviewed it yet but it should also support something like this so with the follow up story all should finally come together.

If you apply the suggestion and link the new ticket, we can resolve this and finally merge.

hohwille (approved):
@krystynaShatkovska thanks for resolving the conflicts 👍 
Unfortunately we currently have some parallel stories going on that slightly depend/influence each other.
Can you have a look at my review comment so we can complete the merge? 🙏

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
krystynaShatkovska: thank you for your recommendations, I have done all of them

hohwille:
FYI: This is hard to test e2e since there is no direct feature exposed but just an internal refactoring done.
Testing all tools regarding getting installed version and edition, could be the perfect e2e test but that is also quite some effort, I cannot invest today.
I tested `get-version --installed` and `get-edition --installed` for several common local tools and it worked fine.
Also for docker as global tool things look sane:
```
$ ide get-edition docker
rancher

$ ide get-version docker
1.24.0
```

My Rancher Desktop shows me the same information in the about popup.
Feel free to re-test more if you have some time left, but I will already set QA approved label.

PR #2095, opened 2026-06-30
#2040: Fix workspace ComboBox reset on project switch

KarimALotfy:
### Implemented changes:

- Workspace `onAction` was re-registered inside `setWorkspaceComboBox()`
  on every project selection. Clearing the items mid-setup fired the old
  handler with a stale/null workspace, causing a `FileNotFoundException`
  error dialog or a `NullPointerException` that aborted `addAll()`.
- Handler is now registered once in `initialize()` via `this::onWorkspaceSelected`.

---

### Testing instructions
- Open GUI by running `AppLauncher`
- Select a project → workspace dropdown populates, prompt text shown
- Switch to a different project → workspace resets to prompt text, buttons re-disabled
- Select a workspace in new project → no error dialog, correct context
  

---

### Checklist for this PR

Make sure everything is checked before merging this PR. For further info please also see
our [DoD](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc).

- [ ] When running `mvn clean test` locally all tests pass and build is successful
- [ ] PR title is of the form `#«issue-id»: «brief summary»` (e.g. `#921: fixed setup.bat`). If no issue ID exists, title only.
- [ ] PR top-level comment summarizes what has been done and contains link to addressed issue(s)
- [ ] PR and issue(s) have suitable labels
- [ ] Issue is set to `In Progress` and assigned to you *or* there is no issue (might happen for very small PRs)
- [ ] You followed all [coding conventions](https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc)
- [ ] You have added the issue implemented by your PR in [CHANGELOG.adoc](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc) unless issue is labeled
  with `internal`
- [ ] You have formulated clear instructions on how to test your contribution under "Testing instructions"

Issue #2214, opened 2026-07-27
Overhaul language selection logic


laim2003:
### Actual behavior

Currently, when selecting a language, the following logic is used:
```
  private void reloadMainView() {

    try {
      Parent reloadedRoot = loadMainView();
      this.root = reloadedRoot;
      if (this.primaryStage != null && this.primaryStage.getScene() != null) {
        this.primaryStage.getScene().setRoot(reloadedRoot);
      }
    } catch (IOException e) {
      LOG.error("Failed to reload main view after locale change", e);
    }
  }
```

This causes the whole window to be reloaded when switching the language, which in turn causes a lot of following issues with the view logic, like the project selection being reset and the workspace selection as well.

### Reproduce

1. Launch the GUI
2. Select a project/workspace
3. See how the selection gets reset once you switch the language

### Expected behavior

We should use a different architecture for managing the language, that follows the KISS principle and simply loads the correct language when starting the application. We should persist the language selection over multiple sessions (maybe even include this into core ideasy).

### IDEasy status

```bash
Updates are available for the settings repository. If you want to apply the latest changes, call "ide update"
IDE_ROOT is set to ~/projects
IDE_HOME is set to ~/projects/project
Your version of IDEasy is 2026.07.002.
Your are using the latest stable version of IDEasy and no update is available.
Your operating system is windows(10.0)@x64 [Windows 11@amd64]
You are online.
Found bash executable at: C:/Program Files/Git/usr/bin/bash.***
Found git executable at: C:/Program Files/Git/***/git.***
Your settings are not up-to-date, please run 'ide update'.
Successfully completed ide (status)

Please use (git-)bash (integrated in Windows Terminal) for full IDEasy support:
https://github.com/devonfw/IDEasy/blob/main/documentation/advanced-tooling-windows.adoc#tabs-for-shells
```

PR #2309, opened 2026-08-11
#2040: Fix buggy workspace selection

### LINKED ISSUE #2040: Workspace selection buggy
### Actual behavior

For some reason, the workspace selection is buggy, when running the gui via `ide gui`. When selecting a project, and then switching to another workspace other than "main", the seletion field becomes empty, and when reselecting the drop-down, all entries are duplicated.

### Reproduce

1. Upgrade to latest snapshot `ide upgrade --mode=snapshot`
2. Run `ide gui`
3. Select a project, then select a workspace other than main.

### Expected behavior

Should simply select the workspace as intended.

### IDEasy status

```bash
IDE_ROOT is set to /Users/lukasfaber/projects
IDE_HOME is set to /Users/lukasfaber/projects/IDEasy
Your version of IDEasy is 2026.06.001-06_17_03-SNAPSHOT.
Your operating system is mac(26.1)@arm64 [Mac OS X@arm64]
You are online.
Found bash executable at: /bin/bash
Found git executable at: /usr/bin/git
Your settings are up-to-date.
```

### Related/Dependent issues

#1987

### Comments/Hints

What is very weird about this, is the fact, that this only happens when running the GUI via `ide gui` after downloading a snapshot version. I could not reproduce this behavior when running AppLauncher.main() from my IDE; in this case the selector works absolutely fine. In order to solve issues like this faster, the new `--enableLogging` flag from #1987 could be very helpful.

samuelkos17:
### Implemented changes:

* For more details look at #2095 

---

### Testing instructions

1. For more details look at #2095 

---

### Checklist for this PR

  with `internal`


samuelkos17:
I've taken this PR since i encountered this bug while working on #2214. This fix works perfectly fine, I merged the current main-branch into it and tested the fix. I think this counts as a Team Review so I'll move it to In Review.

gui/src/main/java/com/devonfw/ide/gui/MainController.java
```diff
@@ -131,6 +132,21 @@ private void initialize() {
 
     setProjectsComboBox();
     initLanguageComboBox();
+    selectedWorkspace.setOnAction(this::onWorkspaceSelected);
+  }
+
+  @FXML
```
maybeec: **Minor** - `@FXML` here is dead metadata. It only matters for members that `FXMLLoader` injects or invokes, and `main-view.fxml` does not reference `onWorkspaceSelected` anywhere; the handler is wired programmatically one line above via `setOnAction(this::onWorkspaceSelected)`. Leaving the annotation on suggests a wiring that does not exist.

Two clean options; the second is more consistent with how this very view wires its other handlers:

1. Drop `@FXML` and keep the programmatic registration.
2. Keep `@FXML`, drop the `setOnAction(...)` line from `initialize()`, and wire it declaratively like the four IDE buttons already are (`onAction="#openAndroidStudio"` etc., `main-view.fxml:145/186/228/268`):

```xml
        <ComboBox fx:id="selectedWorkspace"
          disable="true"
          onAction="#onWorkspaceSelected"
          ... />
```

*Why:* one wiring convention per view means a reader can answer "what handles this control?" from the FXML alone instead of grepping the controller.

gui/src/main/java/com/devonfw/ide/gui/MainController.java
```diff
@@ -131,6 +132,21 @@ private void initialize() {
...
+  private void onWorkspaceSelected(ActionEvent actionEvent) {
+
+    String workspaceName = selectedWorkspace.getValue();
+    if (workspaceName == null) {
+      return;
+    }
+    updateContext(selectedProject.getValue(), workspaceName);
+    androidStudioOpen.setDisable(false);
```
maybeec: **Should-fix** - this four-line block now exists twice (here enabling, and at the end of `setWorkspaceComboBox()` disabling). Add a fifth IDE tomorrow and both sites have to be found and updated; miss one and you get a button that is enabled with no context behind it.

Extract it once:

```java
  private void setIdeButtonsDisabled(boolean disabled) {

    this.androidStudioOpen.setDisable(disabled);
    this.eclipseOpen.setDisable(disabled);
    this.intellijOpen.setDisable(disabled);
    this.vsCodeOpen.setDisable(disabled);
  }
```

and call `setIdeButtonsDisabled(false);` here.

*Why:* [coding-conventions.adoc](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc) - duplicated logic is duplicated maintenance; the fix is to move it up, not to copy it. (Note the `this.` qualifier on the fields, per section *Usage of this* - helpful precisely in web-based reviews like this one, where you cannot hover to see whether `eclipseOpen` is a field or a local.)

gui/src/main/java/com/devonfw/ide/gui/MainController.java
```diff
@@ -234,15 +250,12 @@ private void setWorkspaceComboBox() {
 
     selectedWorkspace.getItems().clear();
     selectedWorkspace.getItems().addAll(workspaces);
+    selectedWorkspace.setValue(null);
```
maybeec: **Minor / non-blocking** - consider clearing the selection *before* mutating the items rather than after:

```java
    selectedWorkspace.setValue(null);
    selectedWorkspace.getItems().clear();
    selectedWorkspace.getItems().addAll(workspaces);
```

*Why:* between `clear()` and this line the ComboBox can still hold the previous project's workspace name. If the newly added list happens to contain an equally-named workspace (very common - every project has `main`), the selection model can re-match it during `addAll(...)` and fire the action handler with a non-`null` value, so `updateContext(newProject, oldWorkspaceName)` runs and switches the context, and then this line blanks the ComboBox again, leaving the context pointing at a workspace the UI shows as unselected.

The `null` guard in `onWorkspaceSelected` cannot catch that case because the value is not `null`. Clearing first means there is simply no stale value to re-match, which is a stronger guarantee than relying on JavaFX's selection-sync ordering. Not blocking - I have not reproduced this, it is a hardening of the same window the PR is already closing.

maybeec (commented):
Thanks for picking this up and carrying #2095 over the finish line :+1: The diagnosis here is good, and the fix attacks the actual root cause rather than papering over the symptom.

Worth spelling out, because it explains the odd symptom in #2040: the old code re-registered the handler at the *end* of `setWorkspaceComboBox()`, so on the **second** project selection the **previous** handler was still attached while `getItems().clear()` / `addAll(...)` mutated the list. It fired with a stale workspace name, `updateContext(...)` hit `FileNotFoundException` in [`GuiStateManager.switchContext`](https://github.com/devonfw/IDEasy/blob/main/gui/src/main/java/com/devonfw/ide/gui/context/GuiStateManager.java#L70), and `errorDialog.showAndWait()` then spun up a **nested JavaFX event loop in the middle of that mutation** - which is almost certainly how you end up with an empty selection field and a duplicated item list. Registering once in `initialize()` plus the `null` guard closes exactly that window. Good fix.

**No blockers.** Everything below is Should-fix or Minor.

---

## Should-fix

### 1. `CHANGELOG.adoc` entry is missing

The checkbox is unchecked in the PR description, and I confirmed there is no `2040` entry in `CHANGELOG.adoc`. #2040 is labelled `GUI`, not `internal`, and the bug shipped to users (the reporter hit it on `2026.06.001-...-SNAPSHOT`), so per [DoD.adoc](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc) it belongs in the next open release section. On current `main` that section is `== 2026.08.002` (milestone 49) and it is still empty:

```adoc
== 2026.08.002

* https://github.com/devonfw/IDEasy/issues/2040[#2040]: Fixed buggy workspace selection in the GUI
```

*Why:* the CHANGELOG is the release-notes source for end users; if it is not written while the change is fresh, nobody reconstructs it at release time.

### 2. Commit message points at the wrong issue

The single functional commit is:

```
#1936:fix workspace ComboBox reset on project switch
```

`#1936` is *"Improve localization of the GUI"* (closed, unrelated). GitHub will auto-link this commit to that issue. Two things per [commit.adoc](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/commit.adoc): the issue id must be the one actually addressed, and the format is `#<issue-id>: <describe your change>` - note the space after the colon.

```
#2040: fix workspace ComboBox reset on project switch
```

*Why:* the `#id:` prefix is not decoration, it is the mechanism that ties the commit to the issue in GitHub's timeline. A wrong id silently attaches this change to unrelated history.

### 3. No regression test, although the harness for one already exists

This is the part I would most like to see added. `gui/src/test/java/com/devonfw/ide/gui/AppBaseTest.java` is already a headless TestFX test of *this exact controller*, and already covers neighbouring behaviour (`testIdeOpenButtonsEnabledWhenWorkspaceSelected`, `testWorkspaceComboboxEnabledEnabledWhenProjectSelected`). Per [AGENTS.md](https://github.com/devonfw/IDEasy/blob/main/AGENTS.md) a bugfix should start from a test that reproduces the bug, and here that is roughly a dozen lines:

```java
  @Test
  public void testWorkspaceSelectionIsResetWhenProjectChanges() {

    interact(() -> selectedProject.getSelectionModel().select("project-1"));
    interact(() -> selectedWorkspace.getSelectionModel().select("main"));

    interact(() -> selectedProject.getSelectionModel().select("project-2"));

    // no stale selection, no duplicated items, buttons back to disabled
    ...
  }
```

Two caveats so this is not a trap:

* `FakeProjectFolderStructureHelper` currently creates only a `main` workspace per project, so it needs a second workspace folder to exercise the "workspace other than main" path from the issue.
* `AppBaseTest` static-imports `org.testfx.assertions.api.Assertions.assertThat`; for the item-list assertion you will want AssertJ (`doesNotHaveDuplicates()`), so qualify it or import it explicitly - AGENTS.md mandates AssertJ for assertions.

*Why:* without it, the next refactoring of `setWorkspaceComboBox()` silently reopens exactly this bug, and it is a bug that only manifests on the *second* project switch - the kind nobody re-checks manually.

### 4. Verify in the environment where the bug actually reproduces

The reporter was explicit in #2040: *"I could not reproduce this behavior when running `AppLauncher.main()` from my IDE; in this case the selector works absolutely fine."* The testing instructions inherited from #2095 say *"Open GUI by running `AppLauncher`"*, i.e. the one environment where the bug never showed. Green CI does not cover this either.

Please confirm before merge that this was verified via `ide gui` on a snapshot build (ideally macOS, as reported). If it was already tested that way, just say so in the PR and this is resolved.

### 5. Branch is behind `main`

GitHub reports `BEHIND`. Per [DoD.adoc](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc), the feature branch has to be up-to-date with `main` before merge. `Update branch` should suffice, `mergeable` is clean.

### 6. Issue #2040 has no milestone

Milestones must be set before issues/PRs get closed (DoD.adoc). Also worth closing the loop on #2095 by linking it here as superseded.

---

## Minor

Inline comments cover: the `@FXML` annotation on `onWorkspaceSelected`, the duplicated four-button enable/disable blocks (2 sites), and the ordering of `setValue(null)`.

One more, **pre-existing and not introduced by you**, so purely optional while you are in this method. `setWorkspaceComboBox()` still has:

```java
    } catch (NotDirectoryException e) {
    }
```

Per [coding-conventions.adoc, section "Catching and handling Exceptions"](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc) the cause is kept (good) but the message is not: `throw new IllegalStateException("Failed to read workspaces for project " + selectedProject.getValue(), e)` makes the stacktrace self-explaining. Entirely non-blocking.

gui/src/main/java/com/devonfw/ide/gui/MainController.java
```diff
@@ -234,15 +250,12 @@ private void setWorkspaceComboBox() {
...
 
-    selectedWorkspace.setOnAction(actionEvent -> {
-      updateContext(selectedProject.getValue(), selectedWorkspace.getValue());
-
-      androidStudioOpen.setDisable(false);
-      eclipseOpen.setDisable(false);
-      intellijOpen.setDisable(false);
-      vsCodeOpen.setDisable(false);
-    });
+    androidStudioOpen.setDisable(true);
```
maybeec: **Should-fix** - second occurrence of the duplicated four-button block (see the comment on the enabling side). Replace with:

```java
    setIdeButtonsDisabled(true);
```

gui/src/main/java/com/devonfw/ide/gui/MainController.java
samuelkos17: I've deleted the @FXML tag

maybeec (commented):
## Re-review of `9eae2c4`

Thanks for turning this around so quickly - **every inline finding from the previous round is resolved**, and cleanly. Still **no blockers**. What remains is entirely the process/verification half of the review, not the code.

### Resolved

| Previous finding | Status |
| --- | --- |
| `@FXML` on `onWorkspaceSelected` was dead metadata | Fixed - annotation dropped, handler stays wired programmatically in `initialize()`. |
| Duplicated four-button enable block (`MainController.java:146`) | Fixed - extracted to `setIdeButtonsDisabled(boolean)`, called with `false`. |
| Duplicated four-button disable block (`MainController.java:255`) | Fixed - now `setIdeButtonsDisabled(true)`. |
| `setValue(null)` ordering vs. item mutation | Fixed - `setValue(null)` now runs *before* `clear()`/`addAll(...)` (`MainController.java:247-249`), so there is no stale value left for the selection model to re-match. |
| Branch behind `main` | Fixed - GitHub now reports `behind_by: 0`, `mergeStateStatus: CLEAN`, all four checks green. |

The extracted helper also uses the `this.` field qualifier, which is what the conventions ask for. Nice.

One detail worth noting so it does not get lost: `ComboBox.setValue(null)` itself fires an `ActionEvent`, so the new `null` guard in `onWorkspaceSelected` is not just defensive - it is load-bearing for the reordering above. Do not let a future cleanup remove it as "unreachable".

---

### Still open

#### 1. `CHANGELOG.adoc` entry - **Should-fix**

The PR still touches exactly one file (`gui/src/main/java/com/devonfw/ide/gui/MainController.java`), and the `== 2026.08.002` section on `main` is still empty. #2040 is labelled `GUI`, not `internal`, and the bug shipped to users, so per [DoD.adoc](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc) it belongs in the next open release section:

```adoc
== 2026.08.002

* https://github.com/devonfw/IDEasy/issues/2040[#2040]: Fixed buggy workspace selection in the GUI
```

*Why:* the CHANGELOG is the release-notes source for end users; nobody reconstructs it at release time if it is not written while the change is fresh. This is also the one unchecked box in the PR description.

#### 2. Commit message points at the wrong issue - **Should-fix**

The functional commit is still `#1936:fix workspace ComboBox reset on project switch`. #1936 is *"Improve localization of the GUI"* (closed, unrelated), so GitHub auto-links this change into unrelated history. Per [commit.adoc](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/commit.adoc) it should be `#2040: fix workspace ComboBox reset on project switch` (note the space after the colon).

Since the branch already contains merge commits, an interactive rebase is awkward - if this PR is merged with **squash**, setting the squash commit title to `#2040: Fix buggy workspace selection` resolves it just as well. Either is fine; please just do not merge it with the `#1936` title intact.

#### 3. No regression test - **Should-fix**

Unchanged: `gui/src/test/java/com/devonfw/ide/gui/AppBaseTest.java` is a headless TestFX test of this exact controller and already covers neighbouring behaviour (`testIdeOpenButtonsEnabledWhenWorkspaceSelected`, `testWorkspaceComboboxEnabledEnabledWhenProjectSelected`), so the harness cost is zero.

This is the finding I would most like to see closed, because the refactoring in this round makes it *cheaper*, not harder: `setIdeButtonsDisabled(boolean)` is now a single observable state transition, so a test can assert on the second project switch that (a) `selectedWorkspace.getValue()` is `null`, (b) the item list has no duplicates, and (c) all four IDE buttons are disabled again.

Caveats from last time still apply: `FakeProjectFolderStructureHelper` only creates a `main` workspace per project, so it needs a second workspace folder to exercise the "workspace other than `main`" path from the issue; and `AppBaseTest` static-imports TestFX's `assertThat`, so qualify AssertJ explicitly for `doesNotHaveDuplicates()` ([AGENTS.md](https://github.com/devonfw/IDEasy/blob/main/AGENTS.md) mandates AssertJ).

*Why:* this bug only manifests on the *second* project switch - exactly the kind nobody re-checks manually. Coveralls confirms the point: "No coverable lines changed in this PR."

#### 4. Verification in the environment where the bug reproduces - **Should-fix**

Still unanswered. The reporter in #2040 was explicit that it did *not* reproduce when running `AppLauncher.main()` from the IDE, yet that is the environment named in the testing instructions. Green CI does not cover it either.

You wrote earlier that you "merged the current main-branch into it and tested the fix" - could you say *how*? If that was `ide gui` on a snapshot build (ideally macOS, as reported), just confirm it in a comment and this is closed.

#### 5. Milestones and issue links - **Should-fix**

Both #2040 and this PR still have no milestone. Per DoD.adoc a milestone must be set before close - `2026.08.002` (milestone 49) matches the CHANGELOG section above. Also still worth linking #2095 here as superseded so that PR can be closed with a trail.

---

### Minor (unchanged, non-blocking)

`setWorkspaceComboBox()` still has `throw new RuntimeException(e)` for the `NotDirectoryException` (`MainController.java:243-245`). Pre-existing, not introduced by you, so entirely optional - but per [coding-conventions.adoc](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc), section *Catching and handling Exceptions*, the cause is kept but the message is not. `throw new IllegalStateException("Failed to read workspaces for project " + selectedProject.getValue(), e)` makes the stacktrace self-explaining.

---

**Summary:** code is good to go. Items 1, 2 and 5 are a few minutes of work; item 4 is one sentence if you already did it; item 3 is the only real remaining effort, and it is the one that keeps this bug from coming back.

samuelkos17:
Thanks for the Re-Review. I've added the Changelog entry and a regression test for the changes made in this PR. I've tested the bug in the local dev build (build-local-dev.sh) running `ide gui` on Windows 11. Regarding the wrong commit message: I think PRs are usually squashed during merging into main so i guess this will fix this problem.

CHANGELOG.adoc
```diff
@@ -6,6 +6,9 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/2040[#2040]: Fixed buggy workspace selection in the GUI
+
+
```
```suggestion
```

hohwille (approved):
@samuelkos17 thanks for fixing this bug. 👍

samuelkos17:
Sucessfully tested with 2026.08.002-08_26_03-SNAPSHOT on Windows.

PR #2315, opened 2026-08-11
#2206: Fix IDEasy title not properly shown on macOS

### LINKED ISSUE #2206: IDEasy title not properly shown on macOS
### Actual behavior

IDEasy displays "java" as the application title on macOS.
Image

### Reproduce

Run `AppLauncher.main()`

### Expected behavior

We should show the simple title of the product: "IDEasy"

### IDEasy status

```bash
(This status is based on the PR 2161 by @oanding-blrng)
IDE_ROOT is set to ~/projects
Your version of IDEasy is SNAPSHOT.
Your version of IDEasy is SNAPSHOT but version SNAPSHOT is available. Please run the following command to upgrade to the latest version:
ide upgrade
Your operating system is mac(26.5.2)@arm64 [Mac OS X@arm64]
You are online.
Found bash executable at: /bin/bash
Found git executable at: /usr/bin/git
```

### Related/Dependent issues

#2161

### Comments/Hints

I'm guessing the issue is somewhat similar to the issue with the IDEasy icon not being displayed on macOS.

Ali-Shariati-Najafabadi:
### Implemented changes:

* On macOS, `ide gui` now launches the JavaFX process through a symlink named `IDEasy` that points to the real `java` executable, instead of exec-ing `java` directly. macOS derives the Dock/menu bar title from the launched executable's filename (this GUI has no native app bundle), so this makes the title read "IDEasy" instead of "java". No-op on Windows/Linux.
* Note: an earlier approach of setting `System.setProperty("apple.awt.application.name", "IDEasy")` in `App.java` was tried and verified (by building and running the GUI locally) to have **no effect**, since this GUI is pure JavaFX (Glass toolkit) and that property is only read by AWT's toolkit.

---

### Testing instructions

1. On macOS, run `ide gui` (or build the `gui` module and run `com.devonfw.ide.gui.AppLauncher` via the same `exec:exec` Maven invocation used by the `gui` commandlet).
2. Check the Dock and the menu bar (⌘+Tab app switcher) while the GUI is running: the title now shows "IDEasy" instead of "java".
3. Unit tests `com.devonfw.tools.ide.tool.gui.GuiTest` cover the executable resolution: a symlink named `IDEasy` is created and used on macOS, while Windows/Linux keep using the plain `java` executable unchanged.

---

### Checklist for this PR

- [x] PR title is of the form `#«issue-id»: «brief summary»`.


laim2003 (commented):
Thanks for your PR! 🙂 For some reason though, the title in the Dock still shows up as "java": 

Image

Ali-Shariati-Najafabadi:
> Thanks for your PR! 🙂 For some reason though, the title in the Dock still shows up as "java":
> 
> Image

I've pushed a fix. Tested locally and the Dock now correctly shows "IDEasy". Could you pull the latest commit and recheck on your end?

CHANGELOG.adoc
```diff
@@ -47,6 +47,7 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/2292[#2292]: Questions appear twice
 * https://github.com/devonfw/IDEasy/issues/825[#825]: Add commandlet for GC Log Analyzer
 * https://github.com/devonfw/IDEasy/issues/1938[#1938]: GUI not launchable outside of project context
+* https://github.com/devonfw/IDEasy/issues/2206[#2206]: IDEasy title not properly shown on macOS
```
laim2003: move this to 2026.08.02

cli/src/test/java/com/devonfw/tools/ide/tool/gui/GuiTest.java
```diff
@@ -0,0 +1,91 @@
...
+    return new ToolInstallation(binDir.getParent(), binDir.getParent(), binDir, VersionIdentifier.of("25.0.0"), false);
+  }
+
+  @Test
+  void testGetGuiExecutableOnMacCreatesAppBundleWithSymlinkNamedIDEasy() throws Exception {
+
+    // arrange
+    WindowsSymlinkTestHelper.assumeSymlinksSupported();
+    IdeTestContext context = newContext(PROJECT_BASIC, null, true);
+    ((AbstractIdeTestContext) context).setSystemInfo(SystemInfoMock.MAC_ARM64);
```
laim2003: The cast `((AbstractIdeTestContext) context)` is redundant, as IdeTestContext extends AbstractIdeTestContext

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -111,4 +112,59 @@ protected void doRun() {
...
+  }
+
+  private static void writeFile(Path file, String content) {
+
+    try {
+      Files.writeString(file, content);
+    } catch (IOException e) {
+      throw new IllegalStateException("Failed to write file: " + file, e);
+    }
+  }
```
laim2003:
```suggestion
```
Redundant since we can use `FileAccess`'s `writeFileContent()`

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -111,4 +112,59 @@ protected void doRun() {
...
+  String getGuiExecutable(ToolInstallation javaInstallation) {
+
+    if (!this.context.getSystemInfo().isMac()) {
+      return "java";
+    }
+    Path javaExecutable = javaInstallation.binDir().resolve("java");
+    Path contentsDir = this.context.getTempPath().resolve("IDEasy.app").resolve("Contents");
+    Path launcher = contentsDir.resolve("MacOS").resolve("IDEasy");
+    this.context.getFileAccess().mkdirs(launcher.getParent());
+    writeFile(contentsDir.resolve("Info.plist"), GUI_INFO_PLIST);
```
laim2003: Instead of creating our own helper method, we should be able to just use `writeFileContent()` from our exisitng `FileAccess` class.

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -111,4 +112,59 @@ protected void doRun() {
...
+
+  private static final String GUI_INFO_PLIST = """
+      <?xml version="1.0" encoding="UTF-8"?>
+      <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+      <plist version="1.0">
+      <dict>
+        <key>CFBundleExecutable</key>
+        <string>IDEasy</string>
+        <key>CFBundleIdentifier</key>
+        <string>%s</string>
```
laim2003:
```suggestion
        <string>com.devonfw.tools.ideasy.gui</string>
```

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -111,4 +112,59 @@ protected void doRun() {
           "Failed to launch the GUI. If maven reports issues with dependency resolution, check whether the maven M2 repo is enabled in your project.", e);
     }
   }
+
+  private static final String GUI_APP_BUNDLE_ID = "com.devonfw.tools.ideasy.gui";
```
laim2003:
```suggestion
```
If we hardcode the APP_BUNDLE_ID anyways?, why dont we just include it in the GUI_INFO_PLIST string?

laim2003 (changes requested):
Hi, thanks for the PR! Tested it on my mac and it works perfectly! Just added some thoughts that may improve this PR even more 😄

laim2003 (approved):
Looks good and still works👍

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -134,4 +135,48 @@ static List<String> buildMvnArgs(Path installationPath, Path pomPath) {
...
+  String getGuiExecutable(ToolInstallation javaInstallation) {
+
+    if (!this.context.getSystemInfo().isMac()) {
+      return "java";
+    }
+    Path javaExecutable = javaInstallation.binDir().resolve("java");
+    Path contentsDir = this.context.getTempPath().resolve("IDEasy.app").resolve("Contents");
+    Path launcher = contentsDir.resolve("MacOS").resolve("IDEasy");
+    this.context.getFileAccess().mkdirs(launcher.getParent());
+    this.context.getFileAccess().writeFileContent(GUI_INFO_PLIST, contentsDir.resolve("Info.plist"));
```
hohwille, line 178: You are correct that we are missing this plist file for MacOS.
IMHO it should be part of the installation package that will be signed in the future.
Therefore, we should not create it on the fly when the GUI is launched.
Also this way we avoid having internal knowledge about MacOS Application package structure in the GUI commandlet where it does not really belong.
But since you are creating the entire IDEasy.app in a temporary directory, this seems like a quickfix.
So my pragmatic suggestion to go forward is that you create a new cleanup ticket, that we create a more clean solution to properly build and install a real MacOS application and then remove this hack/quickfix.
Paste the issue link here and then mark this thread as resolved so we can then merge this quickfix but do not forget the cleanup later.

cli/src/test/java/com/devonfw/tools/ide/tool/gui/GuiTest.java
```diff
@@ -30,6 +35,76 @@ void restoreVersion() {
...
+
+  @Test
+  void testGetGuiExecutableOnMacCreatesAppBundleWithSymlinkNamedIDEasy() throws Exception {
+
+    // arrange
+    WindowsSymlinkTestHelper.assumeSymlinksSupported();
+    IdeTestContext context = newContext(PROJECT_BASIC, null, true);
+    context.setSystemInfo(SystemInfoMock.MAC_ARM64);
+    ToolInstallation javaInstallation = newJavaInstallation(context);
+    Gui gui = new Gui(context);
```
hohwille, line 59: Can you align with @laim2003 on testing GUI stuff properly?
A real GUI usually has its own GUI specific context implementation.
Maybe you agree that we can simply reuse the mocked test context also for the GUI but I fear that this will turn out to be wrong later.
I would at least suggest to introduce an abstract class as a shim for GUI tests that need context (either just name it `AbstractGuiTest` or `AbstractIdeGuiContextTest`.
If fine for now then just extend it from `AbstractIdeContextTest` and have an empty body.
But if we later then might have hundreds of such tests and need to change how `nextContext` has to behave in the GUI space, then we already have a central place where to do that without refactoring all these hundreds of JUnits.

hohwille (approved):
@Ali-Shariati-Najafabadi thanks for your PR. Assuming you have tested your approach on MacOS this looks good and ready for merge to me. Nice job👍 
However, I left few remarks to ensure we do not cause hidden technical debt.

cli/src/test/java/com/devonfw/tools/ide/tool/gui/GuiTest.java
laim2003: I don't think we actually have a GUI-specific test context yet. Noted📝

hohwille: I also wanted to know if it makes sense to introduce a base class for GUI context based tests already but it seems we have to do that later as cleanup...

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
hohwille: I created cleanup issue #2527 and will resolve this thread so we can finally merge.

PR #2321, opened 2026-08-12
#2286: Fix SystemPath.findBinary to search extraPathEntries

### LINKED ISSUE #2286: SystemPath.findBinary ignores extraPathEntries, so withPathEntry has no effect on binary resolution
### Actual behavior

`ProcessContext.withPathEntry(Path)` reads as "add this directory to the PATH", but it only affects the PATH exported to the child process. It has no effect on the binary resolution IDEasy performs itself.

`ProcessContextImpl.run` merges the extra entries and then resolves the executable (`ProcessContextImpl.java:188` and `:192`). `SystemPath.findBinary` iterates `tool2pathMap` (`SystemPath.java:287`) and `paths` (`:293`), but never `extraPathEntries`. That field is assigned at `SystemPath.java:177` and read in exactly one place, `toString()` at `:346`.

A caller that provides a tool via `withPathEntry` and then expects IDEasy to resolve that tool therefore fails with `CreateProcess error=2`, even though the directory containing the binary was passed in.

### Reproduce

Add this test to `cli/src/test/java/com/devonfw/tools/ide/common/`. It builds a `SystemPath` the same way `ProcessContextImpl.run` does at `:188` and then resolves a binary that exists in the directory just handed in.

```java
IdeTestContext context = newContext(PROJECT_BASIC);
Path binDir = context.getIdeHome().resolve("scratch-bin");
Files.createDirectories(binDir);
Files.writeString(binDir.resolve("faketool.cmd"), "@echo hi");

// empty PATH and no software folder, so tool2pathMap and paths stay empty
SystemPath base = new SystemPath(context, "", null, null, ';', List.of());
SystemPath merged = base.withPath(null, List.of(binDir));

assertThat(merged.toString()).contains(binDir.toString());          // passes
assertThat(merged.findBinary(Path.of("faketool"))).isNotEqualTo(Path.of("faketool")); // fails
```

The same `SystemPath` instance exports `binDir` on the PATH but returns `faketool` unresolved. In a real run this surfaces as `Cannot run program "«tool»": CreateProcess error=2` once `ToolCommandlet.runTool` hands the bare name to `ProcessBuilder`.

The two empty collections are what makes it visible. With `IDE_HOME` set, `collectToolPath` fills `tool2pathMap` from the project `software` folder and the lookup succeeds there instead, which is why this stays hidden in normal use. It was found in #2279, where the `gui` commandlet could not resolve `mvn` without `IDE_HOME`.

### Expected behavior

A directory passed to `withPathEntry` is considered by `SystemPath.findBinary`, so that a tool provided this way can be executed.

### IDEasy status

```bash
IDE_ROOT is set to C:\projects
Your version of IDEasy is 2026.08.001-SNAPSHOT
```

### Related/Dependent issues

Found while fixing #2279

### Comments/Hints

#2279 works around this locally by registering the bin directory in the tool map, the way `LocalToolCommandlet.java:127` does after a fresh installation. That is fine for a single call site but does not scale, and `LocalToolCommandlet.doInstallStep` returns early for an already installed tool (`:107`), so the registration at `:127` cannot be relied upon either.

Letting `findBinary` also search `extraPathEntries` would make the two consistent. Since `findBinary` sits in the path of every tool invocation, the change deserves its own review and a look at the intended precedence between `tool2pathMap`, `paths` and `extraPathEntries`.

krystynaShatkovska:
### Implemented changes:

  `withPathEntry()` adds directories to the PATH exported to child processes, but `SystemPath.findBinary()` never searched `extraPathEntries`. This caused binaries added this way to
  fail with `CreateProcess error=2`.

  - Added `extraPathEntries` as the first search location in `findBinary()` — matching the PATH precedence order used by `toString()` (`extraPathEntries` → `tool2pathMap` → `paths`)
  - Added `testFindBinaryFindsBinaryInExtraPathEntries()` to verify the fix

  ---

  ### Testing instructions

  1. Run `mvn clean test` — all tests (CLI: 848, GUI: 56, URL-Updater: 72) pass with BUILD SUCCESS
  2. The new test creates a directory with a fake binary, adds it via `withPath()`, and verifies `findBinary()` resolves it correctly

  ---

  ### Checklist for this PR

  - [x] You have not changed any dependency in `pom.xml` files

### DIFF
--- a/cli/src/test/java/com/devonfw/tools/ide/common/SystemPathTest.java
+++ b/cli/src/test/java/com/devonfw/tools/ide/common/SystemPathTest.java
@@ -1,5 +1,7 @@
 package com.devonfw.tools.ide.common;
 
+import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
@@ -144,14 +146,14 @@ class SystemPathTest extends AbstractIdeContextTest {
   }
 
   @Test
-  void testFindBinaryFindsBinaryInExtraPathEntries() throws java.io.IOException {
+  void testFindBinaryFindsBinaryInExtraPathEntries() throws IOException {
     // arrange
     IdeTestContext context = newContext(PROJECT_BASIC);
     Path binDir = context.getIdeHome().resolve("scratch-bin");
-    java.nio.file.Files.createDirectories(binDir);
+    Files.createDirectories(binDir);
     // create a plain binary name (no extension) so it works on both Linux and Windows
     Path fakeToolFile = binDir.resolve("faketool");
-    java.nio.file.Files.writeString(fakeToolFile, "@echo hi");
+    Files.writeString(fakeToolFile, "@echo hi");
 
     // empty PATH and no software folder, so tool2pathMap and paths stay empty
     SystemPath base = new SystemPath(context, "", null, null, ';', List.of());
@@ -162,6 +164,28 @@ class SystemPathTest extends AbstractIdeContextTest {
     Path resolved = merged.findBinary(Path.of("faketool"));
     assertThat(resolved).isNotEqualTo(Path.of("faketool"));
     assertThat(resolved).isEqualTo(fakeToolFile);
+  }
+
+  @Test
+  void testFindBinaryExtraPathEntriesWinsOverTool2pathMap() throws IOException {
+    // arrange - create two directories, both with a binary named "mytool"
+    IdeTestContext context = newContext(PROJECT_BASIC);
+    Path toolDir = context.getIdeHome().resolve("software-tool");
+    Files.createDirectories(toolDir);
+    Files.writeString(toolDir.resolve("mytool"), "tool-version");
+
+    Path extraDir = context.getIdeHome().resolve("extra-path");
+    Files.createDirectories(extraDir);
+    Files.writeString(extraDir.resolve("mytool"), "extra-version");
+
+    // tool2pathMap has "toolDir", extraPathEntries has "extraDir"
+    SystemPath base = new SystemPath(context, "", null, toolDir, ';', List.of());
+    SystemPath merged = base.withPath(null, List.of(extraDir));
+
+    // assert - findBinary should return the one from extraPathEntries, NOT from tool2pathMap
+    Path resolved = merged.findBinary(Path.of("mytool"));
+    assertThat(resolved).isEqualTo(extraDir.resolve("mytool"));
+    assertThat(resolved).isNotEqualTo(toolDir.resolve("mytool"));
   }
 
   @Test
--- a/cli/src/test/java/com/devonfw/tools/ide/common/SystemPathTest.java
+++ b/cli/src/test/java/com/devonfw/tools/ide/common/SystemPathTest.java
@@ -169,23 +169,26 @@ class SystemPathTest extends AbstractIdeContextTest {
   @Test
   void testFindBinaryExtraPathEntriesWinsOverTool2pathMap() throws IOException {
     // arrange - create two directories, both with a binary named "mytool"
+    // tool2pathMap collects subdirectories of softwarePath, so mytool must be a subdirectory
     IdeTestContext context = newContext(PROJECT_BASIC);
     Path toolDir = context.getIdeHome().resolve("software-tool");
     Files.createDirectories(toolDir);
-    Files.writeString(toolDir.resolve("mytool"), "tool-version");
+    Path mytoolInToolDir = toolDir.resolve("mytool");
+    Files.createDirectories(mytoolInToolDir);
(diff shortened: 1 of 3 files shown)


cli/src/test/java/com/devonfw/tools/ide/common/SystemPathTest.java
```diff
@@ -143,6 +143,27 @@ void testFindBinaryFindsNothingWithoutFilter() {
     assertThat(result).isEqualTo(test);
   }
 
+  @Test
+  void testFindBinaryFindsBinaryInExtraPathEntries() throws java.io.IOException {
```
Ali-Shariati-Najafabadi: Minor: fully-qualified java.io.IOException / java.nio.file.Files (here and lines 151, 154) instead of imports, doesn't match the rest of the file. Mind adding proper imports?

cli/src/test/java/com/devonfw/tools/ide/common/SystemPathTest.java
```diff
@@ -143,6 +143,27 @@ void testFindBinaryFindsNothingWithoutFilter() {
...
+    Path fakeToolFile = binDir.resolve("faketool");
+    java.nio.file.Files.writeString(fakeToolFile, "@echo hi");
+
+    // empty PATH and no software folder, so tool2pathMap and paths stay empty
+    SystemPath base = new SystemPath(context, "", null, null, ';', List.of());
+    SystemPath merged = base.withPath(null, List.of(binDir));
+
+    // assert
+    assertThat(merged.toString()).contains(binDir.toString());
+    Path resolved = merged.findBinary(Path.of("faketool"));
```
Ali-Shariati-Najafabadi, line 164: This checks that extraPathEntries gets found, but not that it actually wins over tool2pathMap when both have the same name, which is the precedence claim in the description. Worth a test for that too.

Ali-Shariati-Najafabadi (commented):
Logic looks right, extraPathEntries now gets checked before tool2pathMap, matching the order toString() already uses. Tested it against the actual bug path in ProcessContextImpl.run(), ran the full suite locally, all green.

Issue [#2286](https://github.com/Ali-Shariati-Najafabadi/IDEasy/issues/2286) isn't labeled internal, so it needs a CHANGELOG.adoc entry per the DoD.

cli/src/test/java/com/devonfw/tools/ide/common/SystemPathTest.java
krystynaShatkovska: yes, sure, thank you for your recommendation.

cli/src/test/java/com/devonfw/tools/ide/common/SystemPathTest.java
krystynaShatkovska: great idea, i will add one. thank you for your review

cli/src/test/java/com/devonfw/tools/ide/common/SystemPathTest.java
```diff
@@ -143,6 +145,49 @@ void testFindBinaryFindsNothingWithoutFilter() {
...
+    assertThat(resolved).isEqualTo(fakeToolFile);
+  }
+
+  @Test
+  void testFindBinaryExtraPathEntriesWinsOverTool2pathMap() throws IOException {
+    // arrange - create two directories, both with a binary named "mytool"
+    IdeTestContext context = newContext(PROJECT_BASIC);
+    Path toolDir = context.getIdeHome().resolve("software-tool");
+    Files.createDirectories(toolDir);
+    Files.writeString(toolDir.resolve("mytool"), "tool-version");
```
Ali-Shariati-Najafabadi: Thanks for the updates, all good now.
One thing left on the new test though: testFindBinaryExtraPathEntriesWinsOverTool2pathMap writes mytool as a plain file under toolDir, but tool2pathMap only picks up subdirectories (see collectToolPath, it does Files.isDirectory(child) before registering anything). So toolDir/mytool never actually makes it into tool2pathMap and it stays empty. The test still passes, but only because extraPathEntries finds it, not because it beat tool2pathMap. To really test the precedence you intended, mytool needs to be a subdirectory (with the binary inside it), so both tool2pathMap and extraPathEntries actually have a competing entry.

hohwille (approved):
@krystynaShatkovska thanks for your PR. Straight forward fix with solid JUnit test - nice job 👍 
BTW: The end-user will not have any clue what this story will mean. It would have made mode sense to mark the story as `internal`. When I do not create the story, I miss such details and do not want to get in the way. I am just trying to raise the awareness in the team so you can understand for yourself what should belong into the CHANGELOG and what not.
For today, I want to get some PRs merged rather than blocking it due to such detailed discussions...

krystynaShatkovska:
Succesfully tested on Windows on 27.08.2026

Issue #1145, opened 2025-03-18
CVE: suggest version update to user


hohwille:
If #1144 is implemented (PR created and merged) an add-on improvement will be to suggest version updates to the user:
In case the tool is installed directly (and not indirectly as dependency e.g. via `ide install <tool>`) and we are not in batch mode the following will happen:

* for each CVE we check if the latest version is not affected by it
* if so we additionally determine the nearest version to the configured version that is also not affected (from our list of available versions)
* we aggregate this for all CVEs found and then ask the user if he wants to update the version.

Example:
```
The tool {} in version {} is affected by {} CVE(s) logged above.
The nearest version {} as well as the latest version {} has no CVEs / is only affected by {} CVE(s).
Which version do you want to use?
Option 1: current (keep current version - WARNING insecure)
Option 2: nearest
Option 3: latest
```

PR #2330, opened 2026-08-13
#2313: Fixed python installation failing

### LINKED ISSUE #2313: Python installation fails with wired behaviour
### Actual behavior

```
$ ide python install 3.14.0
No CVEs found for version 0.12.3 of tool uv.
Successfully installed uv in version 0.12.3 replacing previous version 0.11.26 of uv at D:\projects\IDEasy\software\uv
Found 1 CVE(s) for version 3.14.7 of tool python:
CVE-2026-15308 with severity 7.5 and affected versions: [(,3.15.0)]
https://nvd.nist.gov/vuln/detail/CVE-2026-15308
Which version do you want to install?
Option 1: current (3.14.7 - unsafe)

No CVEs found for version 0.12.3 of tool uv.
Creating backup by moving D:\projects\IDEasy\software\python to D:\projects\IDEasy\backups\2026\08\11\17-19-59_python\IDEasy\software\python
Successfully installed python in version 3.14.7 replacing previous version 3.14.6 of python at D:\projects\IDEasy\software\python
D:\projects\IDEasy\software\python\bin\python.exe: can't open file 'D:\\projects\\IDEasy\\software\\install': [Errno 2] No such file or directory
Running command 'D:\projects\IDEasy\software\python\bin\python.exe' with arguments 'install' '3.14.0'
failed with exit code 2!
Logfile can be found at D:\projects\_ide\logs\2026\08\11\IDEasy-ide-python-17-19-50.log
Running command 'D:\projects\IDEasy\software\python\bin\python.exe' with arguments 'install' '3.14.0'
failed with exit code 2!

Error: IDEasy failed with exit code 2
```

### Reproduce

```
ide python install 3.14.0
```

### Expected behavior

Either this version does not exist and I should get the according message or that version exists then it should be successfully installed.

### IDEasy status

```bash
You are not inside an IDE project: D:/projects
Your version of IDEasy is 2026.08.001-08_11_11-SNAPSHOT.
You are online.
```

### Related/Dependent issues

#2152

### Comments/Hints

The behaviour is wired:
When I asked to install `3.14.0` why is it then asking about `3.14.7` but in the end still installing `3.14.0` what finally fails?

samuelkos17:
### Implemented changes:

* CVE Version Suggestion Bug:
  * Root Cause: In `ToolCommandlet.cveCheck()`, when a safer version above `resolvedVersion` was found and assigned to `latest`, `nearest` was reset to null but `nearestVulnerabilities` was left stale. This caused subsequent versions with anequal vulnerability profile to incorrectly claim the nearest slot via a flawed else if branch that compared against outdated CVE data.
  * Fix:
    - Reset `nearestVulnerabilities = latestVulnerabilities` whenever nearest is reset to null to prevent stale comparisons.
    - Corrected the nearest assignment condition from `version.isGreater(resolvedVersion)` to `!version.isGreater(resolvedVersion)`  (nearest should only track versions below resolvedVersion, not above it).
* Python Installation .venv Conflict (found while reproducing the bug from #2313):
  * Root Cause: When installing a new Python version, `uv venv --python` creates a .venv directory in the software folder before it gets renamed to the target python/ folder. If a previous installation or interrupted install left a stale .venv directory behind, uv venv would fail with "A virtual environment already exists", because only the python/ target was backed up, not the intermediate .venv.
  * Fix: Added a check in `Python.performToolInstallation()` to delete any existing .venv directory before invoking `uv venv`.

---

### Testing instructions

1. Checkout to this branch, run `build-local-dev.sh` and run `ide install python 3.14.0`.
2. Two options `3.14.0` and `3.14.7` should be printed. Select `3.14.0`.
3. Python should be successfully installed.

---

### Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -742,12 +742,13 @@ protected VersionIdentifier cveCheck(ToolInstallRequest request) {
...
             latest = ToolVersionChoice.ofLatest(toolEditionAndVersion, latestVulnerabilities);
             nearest = null;
+            nearestVulnerabilities = latestVulnerabilities;
           } else {
             nearestVulnerabilities = newVulnerabilities;
             nearest = ToolVersionChoice.ofNearest(toolEditionAndVersion, nearestVulnerabilities);
           }
         } else if (newVulnerabilities.isSaferOrEqual(nearestVulnerabilities)) {
-          if (newVulnerabilities.isSafer(nearestVulnerabilities) || version.isGreater(resolvedVersion)) {
+          if (newVulnerabilities.isSafer(nearestVulnerabilities) || !version.isGreater(resolvedVersion)) {
```
maybeec: **Blocker** — this inversion drops the documented "nearest safe upgrade" suggestion.

`getVersions()` returns the versions sorted **descending** (`PythonRepository.computeSortedVersions()` sorts with `Comparator.reverseOrder()`, and the default URL repository behaves the same). So this loop walks from the newest version down to the oldest and passes `resolvedVersion` somewhere in the middle. The `|| version.isGreater(resolvedVersion)` you replaced is exactly what lets `nearest` keep stepping *down* through equally safe versions **above** `resolvedVersion` until it lands on the closest one — i.e. the minimal upgrade. That is the behaviour `documentation/security.adoc` documents:

```
$ ide install mvn 3.6.2
...
Option 2: nearest (3.8.1 - safe)
Option 3: latest (3.9.9 - safe)
```

`3.8.1` is *greater* than the requested `3.6.2`. With `!version.isGreater(...)` no version above the requested one can become `nearest` any more, so that option disappears.

Reproduced on this branch with the existing fixture by setting `INTELLIJ_VERSION=2022.3` in `cli/src/test/resources/ide-projects/intellij/project/settings/ide.properties` (2022.3 is unsafe, 2022.3.2 is safe and one patch level above it) and running `CveCheckInstallTest#testInstallToolNearest`:

`main`:
```
Option 1: current (2022.3 - unsafe)
Option 2: nearest (2022.3.2 - safe)
Option 3: latest (2025.1.1.1 - safe)
```

this branch:
```
Option 1: current (2022.3 - unsafe)
Option 2: latest (2025.1.1.1 - safe)
```

The safe one-patch-level upgrade is gone and the only alternative left is a three-year jump.

The two changes in this method are coupled — reverting only this line while keeping line 745 makes `testInstallToolNearest` fail with `nearest == null`, which is presumably why you changed it. See my comment on line 745 for a variant that fixes the reported problem without losing this suggestion.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -742,12 +742,13 @@ protected VersionIdentifier cveCheck(ToolInstallRequest request) {
             latestVulnerabilities = newVulnerabilities;
             latest = ToolVersionChoice.ofLatest(toolEditionAndVersion, latestVulnerabilities);
             nearest = null;
+            nearestVulnerabilities = latestVulnerabilities;
```
maybeec: **Should-fix** — good catch that `nearestVulnerabilities` survives the `nearest = null` reset and then acts as a comparison baseline for a candidate that was already discarded. That part is a genuine bug. But raising the baseline to `latestVulnerabilities` also changes what `nearest` means: `documentation/security.adoc` describes it as the nearest version "that has no **or less** CVEs" — measured against the *current* version, not against `latest`. After this line, once a completely safe `latest` exists, a partially safer version can never be suggested again.

Verified on this branch with the same fixture plus one extra CVE (severity 5.0) on `2022.3.2` in `cli/src/test/resources/urls/intellij/security.json`, configured version 2023.3.3 (severity 5.0):

* `main`: `Option 2: nearest (2022.3 - unsafe)` — max severity 1.0, i.e. genuinely "less CVEs"
* this branch: no `nearest` option at all

What I believe you actually want is: never offer a `nearest` that is **no safer than the current version** — which is precisely the "why is it asking about 3.14.7?" from #2313. That can be stated directly, right after the loop, leaving both conditions as they are on `main`:

```java
    if ((nearest != null) && !nearest.vulnerabilities().isSafer(currentVulnerabilities)) {
      nearest = null; // never suggest a version that does not reduce the CVEs
    }
    if ((latest == null) && (nearest == null)) {
```

I ran this locally with lines 745 and 751 restored to their `main` state: all three tests in `CveCheckInstallTest` stay green and the `INTELLIJ_VERSION=2022.3` scenario from my other comment still offers `nearest (2022.3.2 - safe)`. Please validate it against your python 3.14 case before adopting it — and either way, a test pinning the expected options would keep this from regressing again.

cli/src/main/java/com/devonfw/tools/ide/tool/python/Python.java
```diff
@@ -53,6 +53,10 @@ protected void performToolInstallation(ToolInstallRequest request, Path installa
       fileAccess.backup(installationPath);
     }
     Path softwarePath = installationPath.getParent();
+    Path venvPath = softwarePath.resolve(".venv");
```
maybeec: **Should-fix** — `".venv"` is now used twice in this class (here and in `renameVenvFolderToPython`). Per `documentation/contributing/coding-conventions.adoc` § Constants, a literal used in multiple places should be a `static final` constant, so a rename or typo cannot make the two sites disagree:

```java
  /** The folder created by {@code uv venv} inside the software folder before it is renamed to the python installation. */
  private static final String FOLDER_VENV = ".venv";
```

and then `softwarePath.resolve(FOLDER_VENV)` in both places — or even better a small private `getVenvPath(Path softwarePath)` helper so the location is defined exactly once.

cli/src/main/java/com/devonfw/tools/ide/tool/python/Python.java
```diff
@@ -53,6 +53,10 @@ protected void performToolInstallation(ToolInstallRequest request, Path installa
...
+    if (Files.exists(venvPath)) {
+      fileAccess.delete(venvPath);
+    }
```
maybeec: **Should-fix** — the `Files.exists` guard is redundant here, and slightly weaker than what `delete` does itself. `FileAccessImpl.delete` already skips non-existing paths, and it checks with `NOFOLLOW_LINKS`:

```java
  public void delete(Path path) {
    if (!Files.exists(path, LinkOption.NOFOLLOW_LINKS)) {
      LOG.trace("Deleting {} skipped as the path does not exist.", path);
      return;
    }
```

`Files.exists(venvPath)` *follows* links, so a dangling `.venv` symlink/junction left behind by an interrupted install evaluates to `false` and would not be removed — while `uv venv` would still refuse to create it, i.e. exactly the failure you are fixing would survive in that case. Reusing `FileAccess` as-is is shorter and more robust:

```suggestion
    fileAccess.delete(venvPath);
```

maybeec (changes requested):
Thanks for digging into #2313, and especially for tracking down the stale `.venv` while reproducing it 👍 The root-cause write-up in the PR description made this easy to follow, and the stale `nearestVulnerabilities` you found is a real defect.

I checked out the branch and compared the CVE suggestion behaviour against `main` by running `CveCheckInstallTest` on the existing `intellij` fixture with different configured versions. Unfortunately the change to the `nearest` condition regresses behaviour that is explicitly documented in `documentation/security.adoc`, so I have to request changes here.

### Blocker

* `ToolCommandlet.java:751` — `!version.isGreater(resolvedVersion)` removes the "nearest safe **upgrade**" suggestion. `documentation/security.adoc` documents `ide install mvn 3.6.2` offering `Option 2: nearest (3.8.1 - safe)`, and `3.8.1` is *greater* than the requested `3.6.2`. Reproduced below in the inline comment.

### Should-fix

* `ToolCommandlet.java:745` — resetting `nearestVulnerabilities` to `latestVulnerabilities` raises the bar for `nearest` from "less CVEs than **current**" (as documented) to "as safe as **latest**", so partially-safer versions are no longer suggested once a fully safe `latest` exists. Inline comment contains an alternative that fixes the reported problem without either side effect, verified against the existing tests.
* **No regression test.** Per `AGENTS.md` / DoD a bug fix starts with a failing test. Both fixes are testable with what is already there:
  * CVE logic: `CveCheckInstallTest` + the `intellij` fixture already contains safe and unsafe versions above and below the configured one; a second test project with a different `INTELLIJ_VERSION` is enough to pin the `nearest` semantics (that is exactly how I reproduced the regression).
  * `.venv` cleanup: `PythonTest.testInstallOnIntelMacResolvesVersionFromUvNotIdeUrls` shows the uv interaction is mockable, so a test that pre-creates `software/.venv` and asserts the install succeeds (and the folder is gone) is feasible.
* **`CHANGELOG.adoc` entry missing** (the checkbox is unticked in the PR description). Both defects shipped in released versions — CVE suggestions since [#1145](https://github.com/devonfw/IDEasy/issues/1145), so this is user-facing and needs an entry under `== 2026.08.001`, unless the issue gets the `internal` label.
* `Python.java:57` — redundant `Files.exists` guard, see inline comment.
* `Python.java:56` — `".venv"` is now a literal in two places in this class; `coding-conventions.adoc` § Constants asks for a `static final` constant.
* **Scope vs. the issue.** #2313 reproduces with `ide python install 3.14.0` (which runs the python binary with the arguments `install 3.14.0` — that is what the `can't open file '...\software\install'` in the issue log comes from), while the testing instructions here use `ide install python 3.14.0`. Please confirm what the reporter's exact command does after this PR, or state in the issue that the command was mistyped and which part of the reported behaviour is actually fixed.
* **Manual QA.** The uv-driven python installation is not exercised end-to-end in CI, so please verify the stale-`.venv` scenario manually (interrupt an install, then re-install) on Windows and at least one Unix OS before merging.

### Housekeeping

* The branch is `BEHIND` main — please update it before merging.
* Neither the PR nor issue #2313 has a milestone set.
* Non-blocking, pre-existing and outside the diff: `Python.java:31` declares `private final VersionIdentifier PYTHON_MIN_VERSION` — a constant name on a non-static field initialized at its declaration (`coding-conventions.adoc` § Constants and § Field Initialization). Nice to clean up while you are in this file.

samuelkos17:
@maybeec thanks for the Review! You were totally right with the problem of the nearest version. I've implemented your suggestion and added the changelog entry.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -742,18 +742,22 @@ protected VersionIdentifier cveCheck(ToolInstallRequest request) {
             latestVulnerabilities = newVulnerabilities;
             latest = ToolVersionChoice.ofLatest(toolEditionAndVersion, latestVulnerabilities);
             nearest = null;
+            //nearestVulnerabilities = latestVulnerabilities;
```
maybeec: **Blocker** - out-commented code. `documentation/contributing/coding-conventions.adoc` (Clean Code) says explicitly *"Do not push out-commented code"*, with [RSPEC-125](https://rules.sonarsource.com/java/RSPEC-125/) as the rationale: a commented-out statement carries no compiler or test coverage, so nobody notices when it rots, and every later reader has to re-derive whether it was disabled on purpose or by accident. Git history already preserves the variant you tried.

```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -742,18 +742,22 @@ protected VersionIdentifier cveCheck(ToolInstallRequest request) {
...
           } else {
             nearestVulnerabilities = newVulnerabilities;
             nearest = ToolVersionChoice.ofNearest(toolEditionAndVersion, nearestVulnerabilities);
           }
         } else if (newVulnerabilities.isSaferOrEqual(nearestVulnerabilities)) {
-          if (newVulnerabilities.isSafer(nearestVulnerabilities) || version.isGreater(resolvedVersion)) {
+          if (newVulnerabilities.isSafer(nearestVulnerabilities) || /*!*/version.isGreater(resolvedVersion)) {
```
maybeec: **Blocker** - same rule, but this one is more dangerous than a plain commented-out line: `/*!*/` disables an operator *inside a live boolean expression*. Deleting three characters silently flips the condition back and removes the documented "nearest safe upgrade" suggestion again (`documentation/security.adoc`: `ide install mvn 3.6.2` offering `Option 2: nearest (3.8.1 - safe)`), and no reviewer scanning this line would spot it. Please restore the plain condition:

```suggestion
          if (newVulnerabilities.isSafer(nearestVulnerabilities) || version.isGreater(resolvedVersion)) {
```

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
```diff
@@ -742,18 +742,22 @@ protected VersionIdentifier cveCheck(ToolInstallRequest request) {
...
             nearest = ToolVersionChoice.ofNearest(new ToolEditionAndVersion(toolEdition, version), newVulnerabilities);
           }
           nearestVulnerabilities = newVulnerabilities;
         }
       }
     }
+    if ((nearest != null) && !nearest.vulnerabilities().isSafer(currentVulnerabilities)) {
+      nearest = null; // never suggest a version that does not reduce the CVEs
```
maybeec, line 758: This is the right place for the fix and I confirmed it behaves as intended: with the `intellij` fixture copied to `target` and its `security.json` rewritten so the only allowed version above the configured one carries the same max severity, `main` prints `Option 2: nearest (2024.2.6 - unsafe)` while this branch offers only `current` and `latest`. That is exactly the #2313 symptom - and exactly why this line deserves a test; see `testInstallToolWithoutNearestIfNotSaferThanCurrent` in the review body.

cli/src/main/java/com/devonfw/tools/ide/tool/python/Python.java
```diff
@@ -30,6 +30,8 @@ public class Python extends LocalToolCommandlet {
 
   private final VersionIdentifier PYTHON_MIN_VERSION = VersionIdentifier.of("3.8.2");
 
+  private static final String VENV_FOLDER = ".venv";
```
maybeec: **Minor** - thanks for extracting the constant. Two small things: static constants belong above instance fields (this one sits below `PYTHON_MIN_VERSION`), and a short JavaDoc saves the next reader from having to find `renameVenvFolderToPython` to learn that this folder is an intermediate created by `uv venv`, not the installation itself.

```suggestion
  /** The folder created by {@code uv venv} inside the software folder before it is renamed to the python installation. */
  static final String VENV_FOLDER = ".venv";
```

(package-private so a regression test can reuse it instead of repeating the literal - see the review body.)

maybeec (changes requested):
Thanks for turning this around so quickly, and for taking the alternative from the last round seriously. I re-checked out the branch and re-ran the tests: the CVE behaviour is correct again - `INTELLIJ_VERSION=2022.3` offers `nearest (2022.3.2 - safe)` as documented in `documentation/security.adoc`, and the new post-loop guard does suppress the useless suggestion from #2313. `CHANGELOG.adoc`, the milestone and the merge with `main` are all sorted now.

One correction on my side: I wrote `== 2026.08.001` for the changelog entry last time. The milestone is `release:2026.08.002`, so where you put it is the right place.

### Status of the previous review

Addressed:

* **Blocker** `ToolCommandlet.java:751` - the `!` inversion is effectively reverted, the "nearest safe upgrade" suggestion is back.
* **Should-fix** `ToolCommandlet.java:745` - `nearestVulnerabilities` reset reverted, post-loop guard adopted.
* **Should-fix** `Python.java` - redundant `Files.exists` guard removed, `.venv` literal extracted into `VENV_FOLDER`.
* **Should-fix** `CHANGELOG.adoc` entry added (please also tick the checkbox in the PR description).
* **Housekeeping** - branch merged with `main` (`MERGEABLE`/`CLEAN`), milestone set on both PR and issue.

Still open - details below.

### Blocker

* **Out-commented code in `ToolCommandlet.java` (lines 745 and 751).** Both reverts were done by commenting out rather than deleting. `documentation/contributing/coding-conventions.adoc` (Clean Code section) states explicitly: *"Do not push out-commented code"*. The `/*!*/` on line 751 is the worse of the two - a disabled operator sitting inside a live boolean expression is one careless edit away from silently reintroducing exactly the regression we just removed, and it hides from the next reader that this line is the load-bearing part of the documented `nearest` semantics. Please delete both; git history is what keeps the old variant.

### Should-fix

* **Still no regression test.** This was the main open item last round and it is unchanged: the diff touches no test at all. Per `AGENTS.md` and `documentation/contributing/DoD.adoc` a bug fix starts with a failing test - otherwise nothing stops the next refactoring of `cveCheck()` from reintroducing either defect, which is exactly how we got here. I did the work to show it is cheap; the proposal below is verified, not hypothetical.

* **Scope vs. issue #2313.** Unchanged: the issue reproduces with `ide python install 3.14.0` (which runs the python binary with the arguments `install 3.14.0` - that is where `can't open file '...\software\install'` in the reported log comes from), while the testing instructions here use `ide install python 3.14.0`. There is still no comment on the issue clarifying this. Please either confirm what the reporter's exact command does after this PR, or note on the issue that the command was mistyped and which part of the reported behaviour is actually fixed - otherwise we close #2313 without knowing whether the reporter's symptom is gone.

* **Manual QA.** Unchanged: the uv-driven python installation is not exercised end-to-end in CI. Please confirm in the PR that you interrupted an install and re-installed successfully on Windows and at least one Unix OS.

### Change proposal (verified locally)

I wrote the three missing tests and validated each one both ways - every test below **fails without** the corresponding production change and **passes with** it. Verified on this branch on Windows via `mvn -Dtest=CveCheckInstallTest test` and `mvn -Dtest=PythonTest test`.

Two notes that cost me some time and will save you the same:

1. `newContext(...)` already resolves the variables, so writing `settings/ide.properties` afterwards has no effect unless you call `context.reload()`.
2. The shared `intellij` CVE fixture cannot reproduce the #2313 end state: every version below the configured one is safer, so a legitimate `nearest` always overwrites the bogus one. Copying `src/test/resources/urls` into the project's `target` folder and rewriting `security.json` there keeps the git-tracked fixture untouched (`junit-testing.adoc`: never write into `src/...`) and makes the defect reproducible.

`cli/src/test/java/com/devonfw/tools/ide/tool/CveCheckInstallTest.java`:

```diff
 import com.devonfw.tools.ide.context.AbstractIdeContextTest;
 import com.devonfw.tools.ide.context.IdeTestContext;
+import com.devonfw.tools.ide.io.FileAccess;
+import com.devonfw.tools.ide.io.FileCopyMode;
 import com.devonfw.tools.ide.tool.intellij.Intellij;
 import com.devonfw.tools.ide.tool.repository.ToolRepository;
@@
   protected static final Path URLS_PATH = Path.of("src/test/resources/urls");
   private static final String PROJECT_INTELLIJ = "intellij";
+
+  private static final String SECURITY_JSON_EQUAL_SEVERITY = """
+      {
+        "issues": [
+          {
+            "id": "CVE-TEST-EQUAL",
+            "severity": 5.0,
+            "versions": [
+              "(0,2025.1.1)"
+            ]
+          },
+          {
+            "id": "CVE-TEST-HIGH",
+            "severity": 9.0,
+            "versions": [
+              "[2025.1.1]"
+            ]
+          }
+        ]
+      }
+      """;
   private final IdeTestContext context = newContext(PROJECT_INTELLIJ);
@@ (appended at the end of the class)
+  /**
+   * Install Intellij with a configured version for which a version above it is the nearest safe upgrade. Ensures the documented "nearest" suggestion (see
+   * {@code documentation/security.adoc}) also offers versions greater than the configured one.
+   */
+  @Test
+  void testInstallToolNearestSuggestsMinimalSafeUpgrade() {
+    //arrange
+    IdeTestContext context = newContext(PROJECT_INTELLIJ);
+    context.setUrlsPath(URLS_PATH);
+    context.getFileAccess().writeFileContent("JAVA_VERSION=17.0.10_7\nINTELLIJ_VERSION=2022.3\n", context.getSettingsPath().resolve("ide.properties"));
+    context.reload();
+    Intellij commandlet = new Intellij(context);
+    context.setAnswers("nearest");
+
+    //act
+    commandlet.install();
+
+    //assert
+    assertThat(context.getSoftwarePath().resolve("intellij").resolve(IdeTestContext.FILE_SOFTWARE_VERSION)).exists().hasContent("2022.3.2");
+  }
+
+  /**
+   * Install Intellij where the only allowed version above the configured one has exactly the same CVEs. Such a version must never be offered as "nearest"
+   * since it does not reduce the vulnerabilities - this is the bug reported in <a href="https://github.com/devonfw/IDEasy/issues/2313">#2313</a>.
+   */
+  @Test
+  void testInstallToolWithoutNearestIfNotSaferThanCurrent() {
+    //arrange
+    IdeTestContext context = newContext(PROJECT_INTELLIJ);
+    Path urlsPath = context.getIdeRoot().resolve("cve-urls");
+    FileAccess fileAccess = context.getFileAccess();
+    fileAccess.mkdirs(urlsPath);
+    fileAccess.copy(URLS_PATH, urlsPath, FileCopyMode.COPY_TREE_OVERRIDE_TREE);
+    fileAccess.writeFileContent(SECURITY_JSON_EQUAL_SEVERITY, urlsPath.resolve("urls").resolve("intellij").resolve("security.json"));
+    context.setUrlsPath(urlsPath.resolve("urls"));
+    Intellij commandlet = new Intellij(context);
+    context.setAnswers("current");
+
+    //act
+    commandlet.install();
+
+    //assert
+    assertThat(context).logAtInteraction().hasMessageContaining("latest (2025.1.1.1 - safe)");
+    assertThat(context).logAtInteraction().hasNoMessageContaining("nearest (");
+  }
```

* `testInstallToolNearestSuggestsMinimalSafeUpgrade` pins the behaviour the first version of this PR broke. Against the original `!version.isGreater(...)` variant it errors out (no `nearest` option to answer); against the current branch it is green.
* `testInstallToolWithoutNearestIfNotSaferThanCurrent` pins the actual #2313 fix. Against `main` it fails with `Found unexpected log entry: Option 2: nearest (2024.2.6 - unsafe)` - a suggestion with exactly the same max severity as the configured version, i.e. the "why is it asking about 3.14.7?" symptom. Against the current branch it is green.

`cli/src/test/java/com/devonfw/tools/ide/tool/python/PythonTest.java`:

```diff
+  /**
+   * Test that a stale {@code .venv} folder left behind by a previously interrupted installation is removed before {@code uv venv} is invoked, see
+   * <a href="https://github.com/devonfw/IDEasy/issues/2313">#2313</a>.
+   */
+  @Test
+  public void testInstallRemovesStaleVenvFolder(WireMockRuntimeInfo wireMockRuntimeInfo) {
+
+    // arrange
+    IdeTestContext context = newContext(PROJECT_UV, wireMockRuntimeInfo);
+    context.setSystemInfo(SystemInfoMock.MAC_X64);
+    Path staleMarker = context.getSoftwarePath().resolve(".venv").resolve("stale.txt");
+    context.getFileAccess().writeFileContent("stale", staleMarker, true);
+    Python python = context.getCommandletManager().getCommandlet(Python.class);
+
+    // act
+    python.install();
+
+    // assert
+    assertThat(context.getSoftwarePath().resolve("python").resolve("stale.txt")).doesNotExist();
+    assertThat(context.getSoftwarePath().resolve(".venv")).doesNotExist();
+    assertThat(context).logAtSuccess().hasMessageContaining("Successfully installed python");
+  }
```

The marker file is what makes this a real test rather than a tautology: without your `fileAccess.delete(venvPath)` the mocked `uv venv` populates the *existing* `.venv`, so the stale content is moved into `software/python` and `stale.txt` survives the install. Against `main` the test fails on exactly that; against the current branch it is green.

Feel free to adjust naming or fold the two CVE tests into one - the important part is that both defects are pinned. Note that `staleMarker` repeats the `".venv"` literal, so you may want to make `VENV_FOLDER` package-private and reuse it in the test.

### Minor (non-blocking)

* `Python.java:33` - `VENV_FOLDER` is a static constant declared *after* the instance field `PYTHON_MIN_VERSION`, and it has no JavaDoc. See inline comment.
* `Python.java:31` - pre-existing, mentioned last time and still open: `private final VersionIdentifier PYTHON_MIN_VERSION` is a constant name on a non-static field initialized at its declaration (`coding-conventions.adoc`, sections Constants and Field Initialization). Nice to clean up while you are in this file.
* The `CHANGELOG.adoc` checkbox in the PR description is still unticked although the entry exists.

cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java
samuelkos17: Thanks for pointing that out! I've forgot to delete it before pushing, my bad

samuelkos17:
Thanks for the Re-Review and the proposed tests. Regarding your questions:

This is the output now, when i run `ide install python 3.14.0`: 
Option 1: current (3.14.0 - unsafe)
Option 2: nearest (3.14.6 - unsafe)
Option 3: latest (3.14.7 - unsafe)

I've also installed & uninstalled Python successfully on Windows and WSL.

cli/src/main/java/com/devonfw/tools/ide/tool/python/Python.java
```diff
@@ -28,7 +28,10 @@ public class Python extends LocalToolCommandlet {
 
   private static final Logger LOG = LoggerFactory.getLogger(Python.class);
 
-  private final VersionIdentifier PYTHON_MIN_VERSION = VersionIdentifier.of("3.8.2");
+  private final VersionIdentifier pythonMinVersion = VersionIdentifier.of("3.8.2");
```
hohwille: Indeed this was wrong before, but wasn't it obvious that this is a constant?

```suggestion
  private static final VersionIdentifier PYTHON_MIN_VERSION = VersionIdentifier.of("3.8.2");
```

cli/src/main/java/com/devonfw/tools/ide/tool/python/Python.java
```diff
@@ -28,7 +28,10 @@ public class Python extends LocalToolCommandlet {
...
+
+  /** The folder created by {@code uv venv} inside the software folder before it is renamed to the python installation. */
+  protected static final String VENV_FOLDER = ".venv";
```
hohwille: I do not see that we need to access this from sub-classes since there are none and will be none.
If you only want to make it visible for the JUnit test, use default visibility.
Otherwise simply declare it public so it could also be accessed from `Uv` or other classes in the future...
```suggestion
  static final String VENV_FOLDER = ".venv";
```

cli/src/main/java/com/devonfw/tools/ide/tool/python/Python.java
```diff
@@ -44,15 +47,19 @@ public Python(IdeContext context) {
   protected void performToolInstallation(ToolInstallRequest request, Path installationPath) {
 
     VersionIdentifier resolvedVersion = request.getRequested().getResolvedVersion();
-    if (resolvedVersion.compareVersion(PYTHON_MIN_VERSION).isLess()) {
-      throw new CliException("Python version must be at least " + this.PYTHON_MIN_VERSION);
+    if (resolvedVersion.compareVersion(pythonMinVersion).isLess()) {
+      throw new CliException("Python version must be at least " + this.pythonMinVersion);
```
hohwille:
```suggestion
    if (resolvedVersion.compareVersion(PYTHON_MIN_VERSION).isLess()) {
      throw new CliException("Python version must be at least " + this.PYTHON_MIN_VERSION);
```

cli/src/main/java/com/devonfw/tools/ide/tool/python/Python.java
```diff
@@ -44,15 +47,19 @@ public Python(IdeContext context) {
...
 
     FileAccess fileAccess = this.context.getFileAccess();
     if (Files.exists(installationPath)) {
       fileAccess.backup(installationPath);
     }
     Path softwarePath = installationPath.getParent();
+    Path venvPath = softwarePath.resolve(VENV_FOLDER);
+
+    fileAccess.delete(venvPath);
+
```
hohwille, line 62: Nothing wrong here. If I understand it correctly, `uv.installPython` might fail if the `.venv` folder was already present (e.g. from a previous incomplete setup that was aborted or failed).
Shouldn't this be considered as a bug of `installPython` method so it should be fixed there?
We can also keep it as is - I just want to understand and challenge this...

hohwille (approved):
@samuelkos17 thanks for your PR. Great analysis, fix and test improvement 👍 
@maybeec thanks for your excellent review support.
I left some minor suggestions to keep our code clean but then this is ready to merge.
Functionally all is already perfect and I also tested it and can confirm it fixes the bug.

cli/src/main/java/com/devonfw/tools/ide/tool/python/Python.java
samuelkos17: Yes, you understood that correctly. I've encountered this issue while reproducing the bug. And yeah you're right, since this specifically concerns Uv it should be understood as a bug of the `installPython()` method of `Uv.java`. However, moving the bugfix there would require a few more changes than if we just keep it this way, what do you say should I move it there or keep it here?

hohwille: Thanks for your explanation. Lets keep it like that, resolve this and merge. 👍

samuelkos17:
Sucessfully tested with 2026.08.002-08_26_03-SNAPSHOT on Windows.

PR #2335, opened 2026-08-14
#1695: Clone settings to temporary directory, analyse, and then move

### LINKED ISSUE #1695: Clone settings to temporary directory, analyse, and then move

laim2003:
### Implemented changes:

* When creating a new project (or updating a project with a broken settings directory), the settings or code-settings directory passed through `ide create projectname <<git-url>>` is pulled to a temporary directory. There, we check whether the repository is a valid settings or code-settings repository and if the check succeeds, we move it to the final location depending on the repo type (and symlink in case of a `code-settings` repo).
* Removed the `--code` flag of `ide create` as we now automatically detect the type of the repository.

---

### Testing instructions

1. Checkout the state of this PR.

#### Create commandlet

3. Run `ide create` using the configurable running executions in intellij, so you dont accidentally test the create commandlet from the latest release, but the state from this PR. For the parameters: Pick a random project name. For the `git-url` parameter see below:
4. Try once using the default settings repo. In this case, check if a new project has been created, which has the settings repo contents correctly placed under IDE_HOME/settings. The creation should succeed!
5. Try once using a code-settings-mixed repo. As a git url, you can use: https://github.com/laim2003/code-settings-repo. The creation should succeed. Important: in the IDE_HOME project folder, a symlink with the name "settings" should be created. It should correctly point to the folder: `IDE_HOME/workspaces/main/[repository name, in our case "code-settings-repo"]/settings`
6. Try once using an invalid repository, for example, the default devonfw/IDEasy repository (this one is neither a settings or a code-settings repository). The project creation should now fail with an error message! Also, the creation process should not continue after the error!

#### Update commandlet
In order to test the update commandlet, you need to create your own settings repository. 

1. On GitHub, create a new repository and during the creation process, when GitHub prompts you (normally the first step), select the devonfw/ide-settings template.
2. Run the `ide create` commandlet (again with the IntelliJ run configuration) and as the `git-url` use your newly created settings repo. It should succeed normally.
3. After that succeeded, make an edit to the settings repo. You can do that directly on GitHub (do *not* make that edit in your settings repo locally, this is important to test the functionality!). For example, update the ide.properties and change the `NODE_VERSION=v18*` to `v23*`, and commit the changes.
4. Now locally, in the newly created project, try running a commandlet, for example `ide eclipse`.
5. It is important to now check that the console prints "Updates are available for the settings repo".
6. Run `ide update` (again, using the run configuration in intellij), it should succeed.
7. As a final test, now delete the contents of your settins folder in your newly created project.
8. Now, run `ide update` again. It should detect the empty folder and prompt you to input a new Git Url. You can use the same URL you used before. This step should also succeed.

---

### Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -236,60 +231,78 @@ private GitUrl getOrAskSettingsUrl() {
...
+  /**
+   * We pull the settings repo from the remote into a temporary folder to perform health checks.
+   */
+  private void pullAndCheckIntegrity(GitUrl gitUrl) {
+    GitContext gitContext = this.context.getGitContext();
+    Path tempProjectPath = this.context.getTempPath().resolve(IdeContext.FOLDER_PROJECTS).resolve(this.context.getProjectName());
+
+    gitContext.pullOrClone(gitUrl, tempProjectPath);
+
+    checkIntegrityAndMove(tempProjectPath, gitUrl.getProjectName());
```
laim2003, line 243: Maybe it makes sense that the same way we use a new step for cloning the settings repo in general, in order to create a better user experience, we should also use a new step for the integrity check here? So it shows up as a separate step in the CLI? Just a thought.

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -167,7 +168,7 @@ protected void updateSettings() {
       LOG.info("Skipping git pull in settings due to code repository. Use --force-pull to enforce pulling.");
       return;
     }
-    this.context.newStep(getStepMessage()).run(() -> updateSettingsInStep(codeRepository));
+    this.context.newStep(getStepMessage()).run(() -> updateSettingsInStep(codeRepository), true);
```
hohwille, line 171: Do we really want to rethrow here?
This will make the process fail if any kind of exception is happening in `updateSettingsInStep` like e.g. "git pull" fails due to error of custom git server.
Currently I can still continue installing software updates then.
This will change now.
IMHO you want to archive that the further processing stops and fails if specific guardrails did not meet like the settings repo could not be cloned or it is not containing any settings.
Thinking this over and digging into `step.run` implementation, I would suggest the following:
* Currently we either handle or re-throw all exceptions there.
* I would suggest to add a method `CliException.isForceRethrowInStep()` that by default returns `false`.
* Extend implementation of `step.run` to also consider this method so we always re-throw in case that method returns `true`.
* Create a `CliRethrowException` (or `CliFatalException`) that overrides this new method returning `true`.
* Explicitly throw this new exception in `updateSettingsInStep` if no valid settings could be established.

BTW: We could even consider (in a separate PR if `CliAbortException` should also cause this effect (of `isForceRethrowInStep()`).

cli/src/main/java/com/devonfw/tools/ide/commandlet/AbstractUpdateCommandlet.java
```diff
@@ -236,60 +231,78 @@ private GitUrl getOrAskSettingsUrl() {
...
-        repository = IdeContext.DEFAULT_SETTINGS_REPO_URL;
-      }
-    }
-    return repository;
+  /**
+   * We pull the settings repo from the remote into a temporary folder to perform health checks.
+   */
+  private void pullAndCheckIntegrity(GitUrl gitUrl) {
+    GitContext gitContext = this.context.getGitContext();
+    Path tempProjectPath = this.context.getTempPath().resolve(IdeContext.FOLDER_PROJECTS).resolve(this.context.getProjectName());
```
hohwille, line 239: It would be safer to use `FileAccess.createTempDir` that guarantees us to produce a unique empty directory.
Otherwise you should ensure the directory is empty and not some leftover from a previous try is still present there (FileAccess.delete followed by FileAccess.mkdirs).

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryUtil.java
```diff
@@ -0,0 +1,49 @@
+package com.devonfw.tools.ide.git.repository;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import com.devonfw.tools.ide.context.IdeContext;
+import com.devonfw.tools.ide.environment.EnvironmentVariables;
+
+/// Utility class for IDEasy settings/code repositories
```
hohwille, line 9: Use JavaDoc instead of inline comment.
```suggestion
/** Utility class for IDEasy settings/code repositories. */
```

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryUtil.java
```diff
@@ -0,0 +1,49 @@
...
+      return RepositoryType.UNKNOWN;
+    }
+
+    if (Files.exists(repositoryPath.resolve(EnvironmentVariables.DEFAULT_PROPERTIES))
+        || Files.exists(repositoryPath.resolve(EnvironmentVariables.LEGACY_PROPERTIES))) {
+      return RepositoryType.SETTINGS;
+    } else if (gitProjectName != null
+        && Files.exists(repositoryPath.resolve(IdeContext.FOLDER_SETTINGS))
+        && getRepositoryType(repositoryPath.resolve(IdeContext.FOLDER_SETTINGS), gitProjectName) == RepositoryType.SETTINGS) {
+      return RepositoryType.CODE_SETTINGS_COMBINED;
```
hohwille, line 31: Couldn't this unconstrainted recursion lead to accepting stuff like `settings/settings/ide.properties`?
I would at least ensure we do not run into recursion again:

```suggestion
    } else if (gitProjectName != null
        && Files.exists(repositoryPath.resolve(IdeContext.FOLDER_SETTINGS))
        && getRepositoryType(repositoryPath.resolve(IdeContext.FOLDER_SETTINGS), null) == RepositoryType.SETTINGS) {
      return RepositoryType.CODE_SETTINGS_COMBINED;
```

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryUtil.java
```diff
@@ -0,0 +1,49 @@
...
+    /// Git Repository is a code repository.
+    CODE,
+    /// Git Repository is a settings repository.
+    SETTINGS,
+    /// A combined code & settings repository contains both the settings-folder and the code within the workspace folder.
+    CODE_SETTINGS_COMBINED,
+    /// The type of the repository could not be determined.
+    UNKNOWN
+  }
+}
```
hohwille, line 49: 1. also here use JavaDoc instead of inline comments (BTW: inline comments in Java only have 2 slashes not 3).
2. in general avoid nested types. It fully makes sense to have this enum as a top-level type in its own class. Esp. since it is used outside of this `RepositoryUtil`.

hohwille (commented):
@laim2003 thanks for taking over and bringing this on the right track. 👍 
I already left some comments to give early feedback since this story was already in end-review at least twice and we are turning in too many circles.
Please also note that if `ide create` fails then also no project shall even be created what is currently not the case.
My idea for this was that the update of the settings would be changed such that it returns a more complex result object that allows us to represent the following states:
* settings were already present and are valid (and maybe have been successfully updated)
* settings were cloned to temporary directory and are valid
* in all cases the information what type of repository we have (as `RepositoryType`).

Other cases currently lead to an exception but could also be represented in that state if we prefer to take control of throwing exception outside.

Then the operation to move a potential temporary settings or "code with settings" repo to the right place can be in a separate method.
This allows the `update` flow to call both methods sequentially but the `create` flow to call the first one initially, then if that succeeded actually create the project and populate it, updating the context and only then call the 2nd method to move the repo into the right place.
Does that make sense or do you have a better idea?

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryUtil.java
hohwille: Just got it now:
https://docs.oracle.com/en/java/javase/25/javadoc/using-markdown-documentation-comments.html#GUID-42FB039A-A551-4616-8981-BAF46B5138E7

laim2003:
@hohwille there is still some cleanup necessary, i did a bit of refactoring...

laim2003:
Cases that have been explicitly tested:

- [x] Create with invalid URL (yxz)
- [x] Create with non git url (www.google.com)
- [x] Create with invalid repo (settings)
- [x] Create with valid git url (code-settings)
- [x] Create with valid git url (settings repo)
- [x] Create with default url using "-" as a parameter

Update:
- [x] Update settings repo; was valid before, but new version is invalid (asks user, doenst automatically apply the newest settings)
- [x] Update settings repo; valid update

Fix: 
- [x] Conf folder is sometimes generated

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryUtil.java
```diff
@@ -0,0 +1,57 @@
...
+public class RepositoryUtil {
+
+  /**
+   * Checks whether the given git repository is a settings repository, a combined settings and code repository, or a typical code repository. A combined code
+   * and settings repository is detected by a top-level {@code settings} folder that itself is a valid settings folder.
+   *
+   * @param repositoryPath the {@link Path} to the repository to check.
+   * @return the {@link RepositoryType} of the repository.
+   */
+  public static RepositoryType getRepositoryType(Path repositoryPath, GitContext gitContext) {
```
hohwille: 1. Consider moving this factory method directly into `RepositoryType` omitting this `RepositoryUtil` class.
2. I would either not pass the `GitContext` here at all or you should check it consistent. IMHO the `repositoryPath` must be a git-repository so the condition `gitContext.isGitRepo(repositoryPath)` must be true in order to get any reasonable result other than `UNKOWN`. IMHO there are two separate checks: a. Some directory is a git repository or not what is already answered by `GitContext.isGitRepo` and b, is some directory looking like a direct settings folder or more like something that contains a `settings` sub-folder with that abilities.
3. The `RepositoryType` also holds a constant `PLAIN_CODE` that seems never to be used. I would remove it, but if you want to keep it, it should also be covered here. However, I have some doubts if that and also the naming really makes sense. The term "code repository" was invented by IDEasy and if a git-repository does not contain what we call "settings" we cannot assume anything about its content - it may be code, it can also be infrastructure or it could be recipies, poems, whatever.

cli/src/main/java/com/devonfw/tools/ide/commandlet/update/settings/SettingsUpdater.java
```diff
@@ -0,0 +1,332 @@
...
+   * aborted. Here, if a new version is available, we clone the new version into a temporary folder and perform health checks. If the cloned, new version is
+   * valid, we call git update in the existing settings folder.
+   */
+  private SettingsHealthCheckResult checkSettingsPresent(Path settingsPath, RepositoryType repositoryType) {
+
+    try {
+      //Get Git url of existing settings, clone newest version of them to temp dir
+      GitUrl gitUrl = GitUrl.of(this.context.getGitContext().retrieveGitUrl(settingsPath));
+      Path tempDir = cloneRepoToTempDir(gitUrl);
+      RepositoryType clonedType = RepositoryUtil.getRepositoryType(tempDir, this.context.getGitContext());
```
hohwille: I am totally not getting this:
You determine the Git URL of an already cloned git repo and clone it again creating another copy of it.
Why? IMHO this does not make sense. What would be the difference compared to a "git pull"?
Please also note that combined settings-with-code repositories can be very big - my former project had a cloned size of ~20 GB. I do not want another cloned copy in temp "for fun" ;)

cli/src/main/java/com/devonfw/tools/ide/commandlet/update/AbstractUpdateCommandlet.java
```diff
@@ -158,138 +162,81 @@ private void setupConf(Path template, Path conf) {
...
+      Step healthCheckStep = this.context.newStep("Performing settings health check");
+      SettingsHealthCheckResult healthCheckResult;
+      healthCheckResult = healthCheckStep.call(() -> {
+        SettingsHealthCheckResult _healthCheckResult = settingsUpdater.checkSettings(this.context.getSettingsPath());
+        SettingsHealthCheckStatus status = _healthCheckResult.status();
+
+        if (status == null) {
+          throw new CliException("Health check on settings failed due to unknown error - the settings have not been updated");
+        } else if (status == SettingsHealthCheckStatus.SETTINGS_INVALID) {
+          throw new CliException("The settings health check failed: " + _healthCheckResult.errorMessage());
```
hohwille: Extract to method with speaking name, see 4. of:
https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc#lambdas-and-streams

cli/src/main/java/com/devonfw/tools/ide/commandlet/update/AbstractUpdateCommandlet.java
```diff
@@ -158,138 +162,81 @@ private void setupConf(Path template, Path conf) {
...
-        LOG.info("'-' was found for settings repository, the default settings repository '{}' will be used.", IdeContext.DEFAULT_SETTINGS_REPO_URL);
-        repository = IdeContext.DEFAULT_SETTINGS_REPO_URL;
-      }
-    }
-    return repository;
-  }
+        boolean onlyPull = healthCheckResult.status() == SettingsHealthCheckStatus.SETTINGS_VALID && healthCheckResult.isExistingProject();
+        SettingsUpdateResult settingsUpdateResult = settingsUpdater.applySettings(onlyPull,
+            healthCheckResult.temporarySettingsDirectory());
+        if (settingsUpdateResult == null) {
```
hohwille: Same here:
https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/coding-conventions.adoc#lambdas-and-streams

cli/src/main/java/com/devonfw/tools/ide/commandlet/update/settings/SettingsHealthCheckStatus.java
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Status of the settings {@link SettingsUpdater#checkSettings(Path)}  health check} describing what {@link SettingsUpdater#applySettings(boolean, Path)} has to
+ * do.
+ */
+public enum SettingsHealthCheckStatus {
+  /** The settings repository was cloned to a temporary directory and is valid - it can be moved to its final location. */
+  SETTINGS_VALID,
+  /** The settings repository is invalid */
+  SETTINGS_INVALID
+}
```
hohwille, line 14: What is the additional information of this status?
IMHO it is fully redundant and could be derived from `RepositoryType`.

cli/src/main/java/com/devonfw/tools/ide/commandlet/update/settings/SettingsHealthCheckResult.java
```diff
@@ -0,0 +1,42 @@
...
+    return new SettingsHealthCheckResult(status, repositoryType, temporarySettingsDirectory, null, isExistingProject);
+  }
+
+  /**
+   * @param repositoryType the {@link RepositoryType} of the settings that are already present.
+   * @param errorMessage the reason why the settings could not be updated.
+   * @param temporarySettingsDirectory path to the temporary folder this health check was performed on.
+   * @return a {@link SettingsHealthCheckResult} for a failed but recoverable health check.
+   */
+  public static SettingsHealthCheckResult failed(RepositoryType repositoryType, String errorMessage, Path temporarySettingsDirectory,
```
hohwille: should be named `ofFailed` as it is also a factory method but an alternative to happy-path with regular `of` method.

hohwille (commented):
@laim2003 thanks for your improvements. Getting there👍 
I would suggest that we end review ping-pong and do a meeting where we complete the last points together with screensharing and collaborative coding.

cli/src/main/java/com/devonfw/tools/ide/commandlet/update/settings/SettingsHealthCheckStatus.java
hohwille: It seems unused now. Can you delete it?

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
```diff
@@ -55,42 +51,51 @@ public boolean isIdeHomeRequired() {
...
+    Path newProjectPath = getNewProjectPath();
     LOG.info("Creating new IDEasy project in {}", newProjectPath);
-    if (!this.context.getFileAccess().isEmptyDir(newProjectPath)) {
+    FileAccess fileAccess = this.context.getFileAccess();
+    if (!fileAccess.isEmptyDir(newProjectPath)) {
       this.context.askToContinue("Directory {} already exists. Do you want to continue?", newProjectPath);
-    } else {
-      this.context.getFileAccess().mkdirs(newProjectPath);
+      fileAccess.backup(newProjectPath);
     }
```
hohwille, line 60: The old behaviour was to create the new project still in the existing directory.
You now decided to backup it and do a fresh start.
This makes sense to me but then you should also change the question asked to the user so he understands what he is deciding:
```
Do you want to move the project to the backup and create it from scratch?
```

CHANGELOG.adoc
```diff
@@ -18,6 +19,7 @@ The full list of changes for this release can be found in https://github.com/dev
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/1695[#1695]: Project creation logic extended with health checks and removed `--code` flag.
```
```suggestion
```

documentation/settings.adoc
```diff
@@ -50,6 +51,18 @@ But we do not want to forget the following advantage:
 Anyhow you can still create feature branches in standalone settings repositories to manage such scenarios and follow KISS and trunk-based development so you more or less avoid such problems.
 However, if you are in a monolithic project with complex release branches you may consider using the "settings in code repository" approach.
 
+== Health check
+
+Whenever `IDEasy` clones or updates your settings it first clones the git repository into a temporary directory and performs a health check on it:
+
+* the given git URL has to be valid,
+* cloning the repository has to succeed,
+* and the repository has to be a settings repository or a combined code and settings repository (see link:#code-repository[above]).
```
hohwille:
```suggestion
* and the repository has to be a settings repository or a combined code and settings repository (see xref:code-repository[above]).
```

cli/src/test/resources/code-settings/pom.xml
```diff
@@ -0,0 +1 @@
+<project>code</project>
```
hohwille: Test projects should aim to be consistent.
If you really need to add a `pom.xml` then add a valid POM file. We later have maintenance issues if some test later will use it (e.g. some feature change may trigger a maven build or whatever).
Otherwise just delete it in case it is not needed.

cli/src/test/java/com/devonfw/tools/ide/step/StepTest.java
```diff
@@ -130,4 +131,60 @@ void testInvalidUsageErrorSuccess() {
...
+  void testCallRethrowsForcedError() {
+
+    // arrange
+    IdeTestContext context = newContext(PROJECT_BASIC, "project", false);
+    Step step = context.newStep("Test-Step");
+    // act & assert
+    assertThatThrownBy(() -> step.call(() -> {
+      throw new CliFatalException("fatal error");
+    }, () -> "fallback")).isInstanceOf(CliFatalException.class).hasMessage("fatal error");
+    assertThat(step.isFailure()).isTrue();
```
hohwille, line 193: wouldn't it make sense here to assert that the result is not `fallback`?

cli/src/main/java/com/devonfw/tools/ide/commandlet/update/AbstractUpdateCommandlet.java
```diff
@@ -63,6 +60,7 @@ public abstract class AbstractUpdateCommandlet extends Commandlet {
   /** {@link FlagProperty} for skipping the setup of git repositories. */
   public final FlagProperty skipRepositories;
 
+  //TODO: If this is only used for the case of code-settings repos, why have that property here and not in UpdateCommandlet?
```
hohwille: I hope this is meanwhile clear and understood.
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/commandlet/update/settings/SettingsHealthCheckStatus.java
laim2003: Sorry, seems like I missed that.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CreateCommandlet.java
laim2003: applied👍

cli/src/test/resources/code-settings/pom.xml
laim2003: Will clean this up. I just realized anyways that this folder should probably be under resources/ide-projects/code-settings and not resources/code-settings, right?

cli/src/test/java/com/devonfw/tools/ide/step/StepTest.java
laim2003: Makes sense, I did not think of that. Applied👍

cli/src/main/java/com/devonfw/tools/ide/commandlet/update/AbstractUpdateCommandlet.java
laim2003: yes it is, this is an older todo I forgot to remove...

cli/src/test/java/com/devonfw/tools/ide/step/StepTest.java
laim2003: I do not know why this doesn't show as outdated, the change is clearly there when I select "All Commits" in the review changes tab.

hohwille (approved):
@areinicke / @laim2003 thanks for making this story come true. This was heavily underestimated by me but with all your effort and patience we now created this great solution. Awesome job 🥇

laim2003:
Succesfully retested on Windows and Mac on 16.09.2026

PR #2352, opened 2026-08-19
#2351: Fixed GUI not launching with local dev build

### LINKED ISSUE #2351: GUI not launching with local dev build
### Actual behavior

When I try to run `ide gui` after running the `build-local-dev.sh` script, it fails with the following error:

[ERROR] Failed to execute goal on project ide-gui-launcher: Could not resolve dependencies for project com.devonfw.tools.IDEasy:ide-gui-launcher:pom:$[project.version]
[ERROR] dependency: com.devonfw.tools.IDEasy:ide-gui:jar:$[project.version] (compile)
[ERROR]         Could not find artifact com.devonfw.tools.IDEasy:ide-gui:jar:$[project.version] in central (https://repo.maven.apache.org/maven2)
[ERROR] 
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DependencyResolutionException


### Reproduce

1. Run the `build-local-dev.sh` script on any branch that's up-to-date with main.
2. Open a shell and try to run `ide gui`.

### Expected behavior

The GUI should open normally.

### IDEasy status

```bash
IDE_ROOT is set to C:\Users\sakostiu\projects
IDE_HOME is set to C:\Users\sakostiu\projects\IDEasy
Your version of IDEasy is 2026.08.002-SNAPSHOT.
Your version of IDEasy is 2026.08.002-SNAPSHOT but version 2026.08.002-20260819.030935-4 is available. Please run the following command to upgrade to the latest version:
ide upgrade                                                                                                                                                                                                                          
You are online.
Your settings are up-to-date.
```

### Related/Dependent issues

#1987

### Comments/Hints

The problem is the way #1987 changed the launcher pom's dependency on ide-gui.

samuelkos17:
### Implemented changes:

* Problem 1: Unresolved `$[project.version]` in the installed `gui/pom.xml`
  * Cause: The script copied the raw cli/src/main/package directory into the local-dev installation, bypassing Maven's resource filtering. As a result the launcher gui/pom.xml kept the literal $[project.version] placeholder instead of the resolved version, so ide gui tried to resolve a non-existent artifact and failed.
  * Fix: copy the filtered build output (cli/target/package) instead of the raw source. This matches how the release assembly assembles the package.
* Problem 2: GUI resolved from the wrong source
  * Cause: build-local-dev.sh only built the CLI and never built the GUI. But ide gui doesn't build the GUI from source — it runs Maven against the installed gui/pom.xml, which declares a dependency on the ide-gui snapshot and resolves it from the local maven repository. So ide gui launched whatever ide-gui snapshot happened to be cached/remote (e.g. from another branch), not the one built here.
  * Fix: install the gui module (and the CLI, so ide-cli resolves) into the local maven repository.
* Additional cleanup:
  * Add `clean` so stale build output from another branch/checkout can't leak into the local-dev installation.
  * Use `-f <pom>` for both Maven invocations instead of cd, so the build is independent of the caller's working directory.

---

### Testing instructions

1. Run `./build-local-dev.sh`
2. Open a shell and run `ide gui`. The GUI should open.

---

### Checklist for this PR

  with `internal`


krystynaShatkovska:
Hi, i checked fixes, they work, well done. The only thing i want to recommend is to do both steps (the build script and ide gui) inside the same ide shell. That way both use the same cache folder. Because maven keeps downloaded libraries in a local folder, and which folder it uses is decided by an environment variable (M2_REPO) that is only set inside an ide shell (via conf/mvn/settings.xml). So if the two steps happen in different shells, the fix can silently not work. Then ide gui doesn't find your local GUI and quietly falls back to the remote snapshot. The GUI opens anyway but so it looks like the fix works, but it's launching the wrong build.

samuelkos17:
@krystynaShatkovska did you test that and experienced that problem or are you just saying that it could happen? If you did experience this, then the fix isn't really that useful

krystynaShatkovska:
Yes, I tested it. The fix works with a local install, ide gui launches the locally built GUI, not the remote one. About next, the two shell types really point Maven at two different folders (ide shell to conf/mvn/repository, plain shell to ~/.m2). If the build runs in a plain shell and ide gui in an ide shell, the two use different repos, so ide gui silently falls back to the remote GUI. So maybe you can think about adding just one line in the testing instructions: run both steps in the same ide shell, so a tester on a fresh shell doesn't get a false

samuelkos17:
@krystynaShatkovska I couldn't reproduce your problem, could you please explain step by step how you've encountered this issue? I've run the `build-local-dev.sh` script in a normal shell and when i opened a new shell everything worked perfectly fine. I also did the same using the ide shell and everything worked perfectly fine.

krystynaShatkovska:
@samuelkos17 I think we have just misunderstood each other. You tested matching shells, but the bug only shows when you build in one shell type and run ide gui in the other. Then it silently uses the remote snapshot, not your build. I will explain. There are two separate local Maven folders, and they don't automatically know about each other:
M2_REPO (conf/mvn/repository). This is the one ide gui always reads from.
And ~/.m2. This is the default mvn uses unless you're inside an IDE shell.

Local build (build-local-dev.sh) does mvn install, which drops newly built GUI into whichever folder that shell uses:
If build in an IDE shell then lands in M2_REPO (conf/mvn/repository) and ide gui finds it.
If build in a normal shell then lands in Folder  ~/.m2 and ide gui looks in  M2_REPO (conf/mvn/repository) , finds nothing .

When it finds nothing, and because version is a -SNAPSHOT, the GUI launcher quietly adds -U (force-update) and just downloads the remote snapshot from the internet instead. So the GUI opens, but it s not your code. That's the bug. Maybe in your test you always built and launched in the same kind of shell (normal ->normal or IDE->IDE), which always agree on the folder. Could you try my case?

How I proved it:
Firstly, I found the real rule, like ide gui is hardwired to read M2_REPO and  ~/.m2 has zero ide-gui jars, but M2_REPO has four. Different folders, confirmed.
Secondly, i ran the exact launcher command from Gui.java against two isolated folders, trying to reproduce the broken case. installed a "local build" jar into a plain-shell folder (~/.m2). And then I ran ide gui's real resolution (-f .../gui/pom.xml + -U) against the IDE folder (M2_REPO), which had no local build. So in a result, it resolved to the remote timestamped jar (...-20260821.031237-5.jar), not a local one.

samuelkos17:
Thanks again for the details, the problem was not the shells but that you had to be in the same project directory, otherwise it wouldn't work. This could've been merged anyways and documented, but I believe that wouldn't be right so I've reworked the gui handling in the local dev build:
When the installation is a local-dev one (.ide.software.version = local-dev-version), Gui.java resolves the GUI offline from a self-contained repository (<install>/.m2) that build-local-dev.sh populates with ide-gui, its dependencies, and the launcher's exec plugin. Stable/normal installs are unaffected. Added a fast unit test (GuiTest) and updated the docs/diagram.

I've tested every combination we discussed and it always worked.

krystynaShatkovska:
Thank you for your rework. It is really a good fix. Making the local-dev install self-contained and launching offline removes the only path by which the remote Sonatype snapshot could be selected, so ide gui now works from any folder. I alsoverified 3 new GuiTest cases pass, and the two seed commands produce a self-contained .m2 that resolves the base -SNAPSHOT ide-gui jar offline with no remote snapshot present. Just two things that are not blocking but just some cosmetic: 1. the execute() catch error message still references the old "M2 repo enabled in your project" story and should point at the self-contained repo (In Gui.execute, the catch block still says "check whether the maven M2 repo is enabled in your project." That was the old M2_REPO story. Now, for a local-dev install, the realistic failure is an incomplete self-contained .m2 (like script interrupted before the seed steps). The message should point at the self-contained repo / re-running the build.). 2. and buildMvnArgs recomputes installationPath.resolve("gui/pom.xml") even though execute already computed pomPath and verified it.

samuelkos17:
@krystynaShatkovska thanks for the suggestions! I've added them and will move this to In Review now 😄

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -28,6 +29,9 @@ public class Gui extends Commandlet {
 
   private static final Logger LOG = LoggerFactory.getLogger(Gui.class);
 
+  /** The value of {@link IdeContext#FILE_SOFTWARE_VERSION} written by the local-dev build script to mark a local-dev installation. */
+  private static final String LOCAL_DEV_VERSION = "local-dev-version";
```
hohwille: IMHO also the local dev version should be indicated as SNAPSHOT version like it is before merging this PR.
Otherwise IDEasy will think that it is an official release version and does not treat it as snapshot.
For SNAPSHOT versions e.g. `ide status` will warn me that I am currently not working on an official release.
This feature might still be helpful if I forgot to upgrade from a local dev version before I start reporting bugs that may not officially exist.

How about this?

```suggestion
  static final String LOCAL_DEV_VERSION = "local-dev-version";
```

BTW: Removing the private visibility allows access from the JUnit avoiding duplication.

cli/src/test/java/com/devonfw/tools/ide/tool/gui/GuiTest.java
```diff
@@ -0,0 +1,98 @@
...
+import com.devonfw.tools.ide.context.IdeContext;
+import com.devonfw.tools.ide.version.IdeVersion;
+
+/**
+ * Test of {@link Gui}.
+ */
+class GuiTest extends Assertions {
+
+  /** Version marker written into a local-dev installation by {@code build-local-dev.sh}. */
+  private static final String LOCAL_DEV_VERSION = "local-dev-version";
```
hohwille: Now you can remove this and just reference `Gui.LOCAL_DEV_VERSION`.

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
```diff
@@ -94,21 +130,36 @@ protected void doRun() {
...
+  /**
+   * @param installationPath the IDEasy installation directory.
+   * @return {@code true} if the installation is a local-dev installation (marked via {@link IdeContext#FILE_SOFTWARE_VERSION}), {@code false} otherwise.
+   */
+  private static boolean isLocalDevInstallation(Path installationPath) {
+
+    if (installationPath == null) {
+      return false;
+    }
+    Path versionFile = installationPath.resolve(IdeContext.FILE_SOFTWARE_VERSION);
```
hohwille: We are currently reworking our way to determine the current installed edition and version of tools.
Maybe for several scenarios this file (IdeContext.FILE_SOFTWARE_VERSION) might even disapear.
Therefore it would be good to avoid adding more and more places that have knowledge about it.
IMHO this entire method could be simplified to this:
```
return IdeVersion.getVersionString().equals(Gui.LOCAL_DEV_VERSION);
```

BTW: The LOCAL_DEV_VERSION is IMHO not really a feature of `Gui` so I would consider it slightly misplaced there.
I would instead suggest to have this constant in `IdeVersion` and you could even provide a `boolean isLocalDevVersion()` method there just like `isUndefined()` and `isSnapshot()`.

documentation/contributing/diagrams/gui/gui-launch-sequence.puml
```diff
@@ -36,6 +36,14 @@ note over guiCmd
   - Class: AppLauncher
 end note
 
+note right of guiCmd
+  For a local-dev installation
+  (.ide.software.version = local-dev-version)
```
hohwille: Spreading too deep implementation details makes documentation hard to maintain.
```suggestion
```

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
hohwille: Another thought: Is it really required to have a static version for this?
This way we are loosing the information from what release version this build was created from.
Esp. in cases where I want to create an (internal) bug issue and include `ide -p status` this can be confusing if I look at the ticket after a long time and cannot relate it to a release.

We could just have a boolean method `IdeVersion.isLocalDev()` that can give us the information from matching the version format (endsWith("-SNAPSHOT") + no timestamp included).

hohwille (approved):
@samuelkos17 thanks for this story and PR. You created an elaborate solution including the local repository that you populate during the build with all required artifacts. Great job 👍 

My major suggestion for improvement is to still keep the current release version included in the local dev build as it is currently due to several aspects and effects I tried to motivate in my review comments.
In my last comment I suggested the easiest solution to finally address this.
Please have a look.

BTW: Why do you think this PR is urgently required as you said today? To launch and test the GUI I can just start it in Intellij. Do not get me wrong. This feature is great for testing changes of the GUI launcher logic itself under realistic conditions but such changes are IMHO very rare.

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
samuelkos17: This is actually just a marker to identify if it's a local-dev-build, I didn't change anything about the naming of the local-dev-version (it's still identified as a SNAPSHOT version). `ide -p status` after running `build-local-dev.sh`:

IDE_ROOT is set to ~/projects
Your version of IDEasy is 2026.08.002-SNAPSHOT.
Your version of IDEasy is 2026.08.002-SNAPSHOT but version 2026.08.002-20260825.030842-7 is available. Please run the following command to upgrade to the latest version:
ide upgrade                                                                                                                                                                                                                          
You are online.

samuelkos17:
@hohwille thanks for the hint on IdeContext.FILE_SOFTWARE_VERSION. I've reworked the local dev build to change the version name instead while following the snapshot logic (appending -LOCAL-DEV). I also removed the details in `gui-launch-sequence.puml`. 

The reason I think this is important is that I've worked on a issue once (not sure if it was Team Review or if I did the work) but the problem only accured when running `ide gui` from the CLI and IMHO we should always test our GUI features using the local dev build since this resembles what the end-user will use and won't hide problems that IntelliJ-magic might hide.

cli/src/main/java/com/devonfw/tools/ide/tool/gui/Gui.java
hohwille: OK, my mistake, thanks for explanation. All great then. 👍

hohwille (approved):
@samuelkos17 thanks for addressing all review feedback. Nice that you also fixed typos in existing log messages 😄 
I would have preferred to still have "SNAPSHOT" in the local dev version number to distinguish it from fixed release version numbers, but for our release this is fine. We can still create a mini PR later if that turns out to be an issue.
Merging now.

samuelkos17:
Succesfully tested with 2026.08.002-DEV-BUILD on Windows.

PR #2362, opened 2026-08-24
#2345: Fix DOCKER_EDITION ignored on Linux

### LINKED ISSUE #2345: DOCKER_EDITION ignored on Linux
### Actual behavior

On Linux DOCKER_EDITION=docker seems to be completely ignored and we only support rancher-desktop but not docker-desktop.
Maybe this was done longe time ago when docker-desktop was not available for Linux, but this is not the case any more:
https://docs.docker.com/desktop/setup/install/linux/

### Reproduce

1. ide set-edition docker docker
2. ide install docker

### Expected behavior

docker desktop gets installed also on Linux.

### IDEasy status

```bash
You are not inside an IDE project: D:/projects
Your version of IDEasy is 2026.08.001.
You are online.
```

### Related/Dependent issues

#2258

### Comments/Hints

_No response_

Hiepiscus:
### This PR fixes #2345

On Linux, setting `DOCKER_EDITION=docker` had no effect because the Docker Desktop URL metadata did not contain a Linux download URL. As a result, Docker Desktop could not be selected and Rancher Desktop was installed instead.

While adding Linux support, the Docker Desktop URL updater also had to be adjusted because the previous release notes URL no longer provided the expected content.

#### Docker Desktop installation on Linux
Docker Desktop requires a different installation than Rancher Desktop. Rancher Desktop is available as a package from its configured package repository and can therefore be installed directly by its package name.

image

Docker Desktop, however is distributed as a separate Debian package. According to the [official Docker Desktop](https://docs.docker.com/desktop/setup/install/linux/ubuntu/#install-docker-desktop) , the Docker package repository must first be configured, the Docker Desktop `.deb` package must then be downloaded separately, and finally the local package must be installed using:

image

### Implemented changes:

* Add Linux support for `DockerDesktopUrlUpdater`
* Enable `Docker.java` to resolve the `docker` edition on Linux when `DOCKER_EDITION=docker` is configured.
* Update the Docker Desktop release notes URL from `docs.docker.com/desktop/release-notes` to `docs.docker.com/desktop/release-notes.md`.
* Adapt the regular expression to extract the Docker Desktop build code from the Markdown release notes.
* Use the extracted build code to generate the Linux download URL.
* Download the selected Docker Desktop Debian package through the configured `ToolRepository`.
* Add a dedicated package manager command to install the downloaded `.deb` file
* Keep `docker-desktop` as the native package name for version detection and uninstallation.

---

### Testing instructions
1. Create a local `urls-status` directory and clone the `ide-urls-status` repository:
`git clone https://github.com/devonfw/ide-urls-status.git <path-to-ide-urls-status>`
2. Run the DockerDesktopUrlUpdater locally by passing the paths to your local ide-urls and ide-urls-status repositories:
`UpdateInitiator <path-to-ide-urls> <path-to-ide-urls-status> PT1H docker`
4. Verify that the generated Docker Desktop metadata contains a Linux download URL.
5. Commit the generated metadata in the local ide-urls repository so that it remains available when IDEasy installs Docker Desktop:
```bash
cd <path-to-ide-urls>
git add .
git commit -m "Add Docker Desktop metadata for Linux"
```
7. On a Linux system, run:
   * Run `ide set-edition docker docker`
   * Run `ide install docker`
8. Verify that Docker Desktop is selected and installed instead of Rancher Desktop.
   * `dpkg-query -W docker-desktop`
10. Uninstall Docker Desktop and verify docker-desktop was removed:
    * `ide uninstall docker`
    * `dpkg-query -W docker-desktop`

---

### Checklist for this PR

  with `internal`


Hiepiscus (commented):
The default native package installation is implemented in `GlobalToolCommandlet#getInstallPackageManagerCommands`. It retrieves the native packages defined by the tool and passes the resolved IDEasy version to each package installation:

https://github.com/devonfw/IDEasy/blob/f145266dc5eab4f207007bf7e393374ddd786dd6/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java#L201-L204

This behavior works for packages that are installed from a package repository such as Rancher Desktop. The package name and resolved version can be combined to create a version specific package.

For APT, `NativePackageManager#getPackageSpec` creates this package specification by appending the resolved version and a wildcard to the package name:
https://github.com/devonfw/IDEasy/blob/f145266dc5eab4f207007bf7e393374ddd786dd6/cli/src/main/java/com/devonfw/tools/ide/tool/NativePackageManager.java#L75-L84

For a regular repository package, this produces a valid package:
`rancher-desktop=1.20.0*`

Docker Desktop requires a different installation because it is distributed as a separate Debian package. Applying the default version logic to a downloaded local file would produce an invalid APT argument:
`/tmp/docker-desktop-amd64.deb=4.34.0*`

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -93,6 +114,33 @@ protected List<NativePackage> getNativePackages() {
...
   }
 
+  @Override
+  protected List<PackageManagerCommand> getInstallPackageManagerCommands(VersionIdentifier resolvedVersion) {
+    if (!EDITION_DOCKER.equals(getConfiguredEdition())) {
+      return super.getInstallPackageManagerCommands(resolvedVersion);
+    }
+
+    ToolRepository toolRepository = this.context.getDefaultToolRepository();
+    Path downloadedDeb = toolRepository.download(this.tool, EDITION_DOCKER, resolvedVersion, this);
```
Hiepiscus: `getInstallPackageManagerCommands` overrides the default installation only for the docker edition. The implementation uses the configured ToolRepository to download the Debian package matching the resolved Docker Desktop version

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -93,6 +114,33 @@ protected List<NativePackage> getNativePackages() {
...
+        List.of(
+            "sudo install -m 0755 -d /etc/apt/keyrings",
+            "sudo curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc",
+            "sudo chmod a+r /etc/apt/keyrings/docker.asc",
+            "echo \"deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] "
+                + "https://download.docker.com/linux/ubuntu $(. /etc/os-release && echo \\\"$VERSION_CODENAME\\\") stable\" | "
+                + "sudo tee /etc/apt/sources.list.d/docker.list > /dev/null",
+            "sudo apt update"
+        ),
+        List.of()
```
Hiepiscus: Using the path returned by `ToolRepository` ensures that installs the exact artifact resolved through the generated ide-urls metadata instead of downloading the latest unversioned Docker Desktop package.

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -93,6 +114,33 @@ protected List<NativePackage> getNativePackages() {
...
+    );
+    return List.of(dockerDesktopInstallPackage.install(null));
```
Hiepiscus: The package is installed with install(null), which is intentional. It prevents `NativePackageManager#getPackageSpec` from appending the resolved version to the local .deb path (e.g. `/tmp/docker-desktop-amd64.deb` instead of `/tmp/docker-desktop-amd64.deb=4.34.0*`).

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -65,6 +70,22 @@ private String detectContainerRuntime() {
...
+              List.of("docker-desktop"),
+              List.of("--allow-downgrades"),
+              List.of(),
+              List.of(
+                  "sudo rm -f /etc/apt/sources.list.d/docker.list",
+                  "sudo rm -f /etc/apt/keyrings/docker.asc"
+              )
+          )
+      );
+    }
```
Hiepiscus, line 103: The existing Rancher Desktop installation remains unchanged.

The regular NativePackage definition for the Docker edition is still required. This definition uses the installed package name docker-desktop, rather than the temporary download path. The package name is needed for package-related operations after installation, especially uninstallation. This allows to create the correct uninstall command:
`sudo apt -y autoremove --purge docker-desktop`

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -93,6 +114,33 @@ protected List<NativePackage> getNativePackages() {
...
+            "sudo chmod a+r /etc/apt/keyrings/docker.asc",
+            "echo \"deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] "
+                + "https://download.docker.com/linux/ubuntu $(. /etc/os-release && echo \\\"$VERSION_CODENAME\\\") stable\" | "
+                + "sudo tee /etc/apt/sources.list.d/docker.list > /dev/null",
+            "sudo apt update"
+        ),
+        List.of()
+    );
+    return List.of(dockerDesktopInstallPackage.install(null));
+  }
```
JoelAdbu: IMHO we should not download the package as part of getPackageManagerCommands(). It feels a bit unexpected to perform a side effect while merely resolving the package manager commands.

Instead, we could download the .deb earlier as part of the installation flow (for example by overriding doInstall()), store the downloaded path, and then pass that path through getNativePackages() just like we do for other native packages.

Something along these lines:

```suggestion
  private Path downloadedDebPackageForDocker;

 @Override
  protected ToolInstallation doInstall(ToolInstallRequest request) {
    if (EDITION_DOCKER.equals(getConfiguredEdition())) {
      downloadDebPackageStepAndSetPackagePath(request.getRequested().getResolvedVersion());
    }
    return super.doInstall(request);
  }

  private void downloadDebPackageStepAndSetPackagePath(VersionIdentifier resolvedVersion) {
    ToolRepository toolRepository = this.context.getDefaultToolRepository();
    this.downloadedDebPackageForDocker = toolRepository.download(this.tool, EDITION_DOCKER, resolvedVersion, this);
  }

@Override
protected List<NativePackage> getNativePackages() {

    if (EDITION_DOCKER.equals(getConfiguredEdition())) {
      return List.of(
          new NativePackage(
        NativePackageManager.APT,
        List.of(downloadedDebPackageForDocker.toString()),
        List.of("--allow-downgrades"),
        List.of(
            "sudo install -m 0755 -d /etc/apt/keyrings",
            "sudo curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc",
            "sudo chmod a+r /etc/apt/keyrings/docker.asc",
            "echo \"deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] "
                + "https://download.docker.com/linux/ubuntu $(. /etc/os-release && echo \\\"$VERSION_CODENAME\\\") stable\" | "
                + "sudo tee /etc/apt/sources.list.d/docker.list > /dev/null",
            "sudo apt update"
        ),
        List.of()
    );
    }

return List.of(
new NativePackage(
NativePackageManager.ZYPPER,
@@ -93,6 +114,33 @@
);
}
```
This would keep getNativePackages() focused on describing the package installation and avoid mixing package resolution with downloading logic.

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
JoelAdbu, line 103: IMHO we should not download the package as part of getPackageManagerCommands(). It feels a bit unexpected to perform a side effect while merely resolving the package manager commands.

Instead, we could download the .deb earlier as part of the installation flow (for example by overriding doInstall()), store the downloaded path, and then pass that path through getNativePackages() just like we do for other native packages.

Something along these lines:


```suggestion
 private Path downloadedDebPackageForDocker;

 @Override
  protected ToolInstallation doInstall(ToolInstallRequest request) {
    if (EDITION_DOCKER.equals(getConfiguredEdition())) {
      downloadDebPackageStepAndSetPackagePath(request.getRequested().getResolvedVersion());
    }
    return super.doInstall(request);
  }

  private void downloadDebPackageStepAndSetPackagePath(VersionIdentifier resolvedVersion) {
    ToolRepository toolRepository = this.context.getDefaultToolRepository();
    this.downloadedDebPackageForDocker = toolRepository.download(this.tool, EDITION_DOCKER, resolvedVersion, this);
  }

@Override
protected List<NativePackage> getNativePackages() {

    if (EDITION_DOCKER.equals(getConfiguredEdition())) {
      return List.of(
          new NativePackage(
        NativePackageManager.APT,
        List.of(downloadedDebPackageForDocker.toString()),
        List.of("--allow-downgrades"),
        List.of(
            "sudo install -m 0755 -d /etc/apt/keyrings",
            "sudo curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc",
            "sudo chmod a+r /etc/apt/keyrings/docker.asc",
            "echo \"deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] "
                + "https://download.docker.com/linux/ubuntu $(. /etc/os-release && echo \\\"$VERSION_CODENAME\\\") stable\" | "
                + "sudo tee /etc/apt/sources.list.d/docker.list > /dev/null",
            "sudo apt update"
        ),
        List.of(
                  "sudo rm -f /etc/apt/sources.list.d/docker.list",
                  "sudo rm -f /etc/apt/keyrings/docker.asc"
        )
    );
    } 
```

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
JoelAdbu: If the previous changes (https://github.com/devonfw/IDEasy/pull/2362/changes#r3871801542) have already been applied, these lines should no longer be necessary:


```suggestion

```

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
JoelAdbu: IMHO the best way to handle this issue is to extend NativePackage so that it can also support package artifacts (e.g. downloaded .deb or .rpm files). e.g.

``` java
private final List<String> optionalNativePackageArtifactPaths;
````
This would require updating the existing constructor and fixing the resulting compilation issues.

If the list is empty, installation continues to use the package names returned by getPackages(), which is the current behavior.

If artifact paths are provided, NativePackageManager.install(...) should use those paths instead. This would allow us to support tools such as Docker Desktop that require installation from a downloaded package artifact while keeping the existing package-based installation mechanism unchanged.

something like this in NativePackageManager.java

``` java
/**
   * @param version the version of the package to install.
   * @param nativePackage the {@link NativePackage} to install.
   * @return the {@link PackageManagerCommand} to install the given {@link NativePackage} including its {@link NativePackage#getSetupCommands()} setup commands.
   */
  public PackageManagerCommand install(NativePackage nativePackage, String version) {
    verifyPackageManager(nativePackage);
    List<String> commands = new ArrayList<>(nativePackage.getSetupCommands());
    StringBuilder command = new StringBuilder(SUDO).append(' ').append(getBinaryName());
    for (String option : nativePackage.getExtraInstallOptions()) {
      command.append(' ').append(option);
    }
    command.append(' ').append(this.installCommand);
    if (nativePackage.getOptionalNativePackageArtifactPath().isEmpty()) {
      for (String pkg : nativePackage.getPackages()) {
        command.append(' ').append(getPackageSpec(pkg, version));
      }
    } else {
      for (String nativePackageArtifactPath : nativePackage.getOptionalNativePackageArtifactPath()) {
        command.append(' ').append(nativePackageArtifactPath);
      }
    }
    commands.add(command.toString());
    return new PackageManagerCommand(this, commands);
  }
````

It would also be nice if you could extend the existing tests to cover this new functionality.

File: cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java [hunk 6]
```diff
@@ -65,12 +74,41 @@ private String detectContainerRuntime() {
...
+                  "sudo chmod a+r /etc/apt/keyrings/docker.asc",
+                  "echo \"deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] "
+                      + "https://download.docker.com/linux/ubuntu $(. /etc/os-release && echo \\\"$VERSION_CODENAME\\\") stable\" | "
+                      + "sudo tee /etc/apt/sources.list.d/docker.list > /dev/null",
+                  "sudo apt update"
+              ),
+              List.of(
+                  "sudo rm -f /etc/apt/sources.list.d/docker.list",
+                  "sudo rm -f /etc/apt/keyrings/docker.asc"
+              ),
```
JoelAdbu, line 99: I tested this locally and noticed that after running ide uninstall docker, dpkg -l | grep docker still shows a leftover Docker Desktop package entry with status ic (configuration files only).

Therefore, I would extend the cleanup commands as follows:

```suggestion
              List.of(
                  "sudo rm -f /etc/apt/sources.list.d/docker.list",
                  "sudo rm -f /etc/apt/keyrings/docker.asc",
                  "sudo dpkg --purge docker-desktop"
              ),
```
This ensures that any remaining package metadata and configuration files are removed as well.

Hiepiscus: Thanks for checking this. I retested the uninstall flow without sudo dpkg --purge docker-desktop and verified the result using dpkg -l | grep docker. In my tests, no leftover Docker Desktop entries remained after running ide uninstall docker, so the additional purge command was not necessary. Therefore, I reverted the change.

PR #2370, opened 2026-08-25
#1628: Support for OS specific CVEs

### LINKED ISSUE #1628: Support for OS specific CVEs
### Feature idea

The idea of #103 is that we extract a minimised subset of the CVE database for IDEasy in our `ide-urls`.
This way IDEasy end-users do not have the need to download the CVE database (what takes very long without API token due to rate limits). From the database we need less than 1% of the data for the tools we actually manage.
A CVE from the CVE DB (`Vulnerability` in the owasp lib) can have a large list of `VulnerableSoftware` objects what extends `Cpe` with some version range information.
We only search for the tools we manage and then for each `Vulnerability` we reduce the data to only the CPE matching our tool (e.g. some CVE can be about a library having a critical vulnerability but has entries for all products and their versions containing that library).

All this is fine and makes sense, but there is one catch I noticed during fixing and cleanup of the involved code:
CPEs can match only for specific operating systems/platforms (e.g. Windows, Linux, Mac).
In our minimised model, we cannot express this since for every tool + edition we have one `security.json` file with and array of issues (CVEs). Each such issue has:
* an ID
* a severity as decimal number in the range `(0,10.0]` (technically the base score of the CVSS)
* an array of `VersionRange` for the affected versions

Example:
https://github.com/devonfw/ide-urls/blob/3f6a2c39b0029d60ca82bd09044d2c4641ffa840/terraform/terraform/security.json#L1-L11

We would need to extend this model to also include versions conditional on the operating system:
E.g. 
```
{
    "id" : "CVE-xxxx-yyyy",
    "severity" : 5.0,
    "versions" : [ "(,1.0.0)" ],
    "windows" : [ "[1.0,0, 1.0.1]", "[2.0.0, 2.0.8]" ],
    "linux" : [ "[2.0.0, 2.0.5]" ],
    "mac" : [ "[2.0.0, 2.0.6]" ]
  }
```

New/unknown properties are ignored by IDEasy, so we can add them without breaking existing processing.
Future versions of IDEasy could then combine the "global" `versions` property with the one for the current OS of the end-user running IDEasy.
Maybe a more flexible and future proof approach would be to have a generic object/map for conditions instead of fixed properties:
 ```
{
    "id" : "CVE-xxxx-yyyy",
    "severity" : 5.0,
    "versions" : [ "(,1.0.0)" ],
    "conditions": {
      "windows" : [ "[1.0,0, 1.0.1]", "[2.0.0, 2.0.8]" ],
      "windows_arm64" : [ "[2.0.9, 2.0.9]" ],
      "linux" : [ "[2.0.0, 2.0.5]" ],
      "mac" : [ "[2.0.0, 2.0.6]" ]
    }
  }
```

### Additional context

This story is not easy to implement.
Already writing the code to build a minimised `versions` array by merging and combining `VersionRange` objects was rather tricky (connected loosely such as `[1.0,1.1]` with `[1.2,1.3]` to `[1.0,1.3]` and again `(,2.0)` with `[1.0,1.3]` and `[1.5,1.5]` to `(,2.0)` but not merging `[1.0.0,1.1.0]` with `[1.2.0,1.3.0]` together).

### PR DESCRIPTION
**Pull request by laert-ll**
### Implemented changes:

`security.json` CVE model extended so that a single CVE can express affected version ranges that only apply on specific operating systems.

* Added an optional `conditions` map to the `Cve` model.
* `CveJsonDeserializer` / `CveJsonSerializer` read and write `conditions`. It is only written when present, so existing `security.json` files remain unchanged.
* `Cve.isAffected(version, os)` combines the global `versions` with the current OS's conditional ranges; `Cve.merge(...)` also merges the per-OS conditions.
* Added CHANGELOG entry.

---

### Testing instructions

Run `mvn clean test`

Manual (single machine):

1. For a tool you can install, edit `<IDE_ROOT>/urls/<tool>/<edition>/security.json` (create if absent) and add a CVE whose global `versions` do NOT match the version you'll install, but whose `conditions` for your current OS (`windows`/`linux`/`mac`) do:
   ```json
   { "issues": [ { "id": "CVE-TEST-OS", "severity": 9.0,
     "versions": ["(0,0.0.1)"], "conditions": { "linux": ["[<version>]"] } } ] }
2. Set <TOOL>_VERSION=<version> in ide.properties and run ide install <tool> CVE-TEST-OS is reported (matched only via the OS condition).
3. Change the key to a different OS and re-run the CVE is no longer reported.

---

### Checklist for this PR

  with `internal`

### CODE REVIEW

File: cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/CveJsonSerializer.java [hunk 1]
```diff
@@ -17,5 +20,15 @@ protected void serializeProperties(Cve cve, JsonGenerator jgen, SerializerProvid
     jgen.writeNumberField(Cve.PROPERTY_SEVERITY, cve.severity());
     jgen.writeFieldName(Cve.PROPERTY_VERSIONS);
     writeArray(cve.versions(), jgen);
+    Map<String, List<VersionRange>> conditions = cve.conditions();
+    if (!conditions.isEmpty()) {
+      jgen.writeFieldName(Cve.PROPERTY_CONDITIONS);
+      jgen.writeStartObject();
+      for (Map.Entry<String, List<VersionRange>> condition : conditions.entrySet()) {
+        jgen.writeFieldName(condition.getKey());
+        writeArray(condition.getValue(), jgen);
```
Ali-Shariati-Najafabadi, line 29: If an OS entry in `conditions` is explicitly `null` (e.g. `"windows": null` in security.json), it loads fine but blows up here with an NPE the moment the file gets re-saved/normalized, since `writeArray` iterates the list without a null check. Tested it locally, confirmed the crash. Worth guarding against in `readConditions` (either reject or default to an empty list) rather than passing the null through.

cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/CveJsonDeserializer.java
```diff
@@ -39,16 +43,35 @@ public void setProperty(String property, JsonParser p, DeserializationContext ct
...
+        }
         default -> {
           super.setProperty(property, p, ctxt);
         }
       }
     }
 
+    private Map<String, List<VersionRange>> readConditions(JsonParser p) throws IOException {
+
+      if (p.getCurrentToken() != JsonToken.START_OBJECT) {
```
Ali-Shariati-Najafabadi, line 57: Everywhere else in this deserializer (`readArray`, `readValue`) an unexpected token throws `IllegalStateException`. Here we just silently return `null` if `conditions` isn't an object. Feels inconsistent - a malformed `conditions` block would just get dropped without any error, which seems risky for security data.

cli/src/test/java/com/devonfw/tools/ide/url/model/file/json/CveTest.java
```diff
@@ -1,17 +1,35 @@
...
+
+    // act + assert
+    assertThat(cve.isAffected(VersionIdentifier.of("0.9.0"), OperatingSystem.LINUX)).isTrue();
+    assertThat(cve.isAffected(VersionIdentifier.of("2.0.6"), OperatingSystem.WINDOWS)).isTrue();
+    assertThat(cve.isAffected(VersionIdentifier.of("2.0.6"), OperatingSystem.LINUX)).isFalse();
+    assertThat(cve.isAffected(VersionIdentifier.of("2.0.6"), OperatingSystem.MAC)).isFalse();
+    assertThat(cve.isAffected(VersionIdentifier.of("2.0.6"), null)).isFalse();
+  }
+
   @Test
```
Ali-Shariati-Najafabadi, line 33: Could we get a test for `merge()` actually combining `conditions` from both sides? Right now `mergeConditions()` is only hit indirectly through the round-trip test, which never calls `merge()`. Given how fiddly the version-range merging already is, I'd want that path covered directly (overlapping ranges, one-sided keys, etc.).

cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/Cve.java
```diff
@@ -17,20 +20,62 @@
...
+   * @param version the {@link VersionIdentifier} to check.
+   * @param os the current {@link OperatingSystem} (may be {@code null}).
+   * @return {@code true} if the given {@link VersionIdentifier} is affected by this CVE on the given {@link OperatingSystem}, {@code false} otherwise.
+   */
+  public boolean isAffected(VersionIdentifier version, OperatingSystem os) {
+
+    if (contains(this.versions, version)) {
+      return true;
+    }
+    return (os != null) && contains(this.conditions.get(os.toString()), version);
```
Ali-Shariati-Najafabadi, line 66: Minor: the OS key lookup is case-sensitive (`os.toString()` is always lowercase). A typo like `"Windows"` in a hand-written security.json would just silently never match, no warning. Not blocking, just easy to miss.

cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/Cve.java
```diff
@@ -17,20 +20,62 @@
  * @param severity the severity in the range from (0,10.0] where 10.0 is most critical.
```
Ali-Shariati-Najafabadi, line 20: Nice implementation of the conditions map from #1628. One thing worth a doc note: the issue mentioned arch-specific keys like `windows_arm64` as a possible future extension - right now `isAffected` only checks `os.toString()` so those wouldn't actually be evaluated even though they'd parse fine. Probably fine to punt on for now, just flagging so nobody assumes it works.

Ali-Shariati-Najafabadi (changes requested):
Pulled the branch and ran the full test suite locally plus checkstyle. Build is green and the core idea (isAffected/merge/JSON round-trip) works and is tested. Found one actual crash bug and a few smaller things worth looking at before merging, left inline.

cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/Cve.java
```diff
@@ -49,7 +94,21 @@ public Cve merge(Cve issue) {
...
     }
-    return new Cve(this.id, this.severity, newVersions);
+    return new Cve(this.id, this.severity, newVersions, mergeConditions(issue.conditions));
+  }
+
+  private Map<String, List<VersionRange>> mergeConditions(Map<String, List<VersionRange>> other) {
+
+    if (this.conditions.isEmpty() && other.isEmpty()) {
+      return Map.of();
+    }
```
hohwille: Wouldn't this make more sense here?

```suggestion
    if (this.conditions.isEmpty()) {
      return other;
    }
```

cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/Cve.java
```diff
@@ -49,7 +94,21 @@ public Cve merge(Cve issue) {
...
+    Map<String, List<VersionRange>> newConditions = new TreeMap<>();
+    this.conditions.forEach((os, ranges) -> newConditions.put(os, new ArrayList<>(ranges)));
+    other.forEach((os, ranges) -> {
+      List<VersionRange> newRanges = newConditions.computeIfAbsent(os, key -> new ArrayList<>());
+      ranges.forEach(range -> mergeVersionRage(newRanges, range));
+    });
+    return newConditions;
```
hohwille: This `Cve` is a record and should stay immutable:
```suggestion
    return Map.copyOf(newConditions);
```

cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/Cve.java
```diff
@@ -49,7 +94,21 @@ public Cve merge(Cve issue) {
...
+    return new Cve(this.id, this.severity, newVersions, mergeConditions(issue.conditions));
+  }
+
+  private Map<String, List<VersionRange>> mergeConditions(Map<String, List<VersionRange>> other) {
+
+    if (this.conditions.isEmpty() && other.isEmpty()) {
+      return Map.of();
+    }
+    Map<String, List<VersionRange>> newConditions = new TreeMap<>();
+    this.conditions.forEach((os, ranges) -> newConditions.put(os, new ArrayList<>(ranges)));
```
hohwille: Can be simplified and made more efficient:

```suggestion
    Map<String, List<VersionRange>> newConditions = new TreeMap<>(this.conditions);
```

hohwille (approved):
@laert-ll thanks for your PR. Great work, well done 👍 
I actually had to read my own story again to do the review properly.
Only then I remembered why we had to do this so complicated instead of just having an `os` property with an array of OSes that the CVE applies to. But in reality a CVE can apply versions for all OS but other versions only for particular OS. With this more complex design we can represent all the states we need to express the reality that actually is complex.

I left some review comments which are all small constructive suggestions to make the code cleaner or nicer but nothing I found was wrong in any way.
IMHO we can apply all suggestions and then merge this PR.
Great that you already fixed the merge method so completing the story to also use this new feature in the security updater process is now very easy.

p.s.: You could have passed the `SystemInfo` instead of just `OperatingSystem` since then we do not have to refactor anything in the future if we might also need to support CVE versions that are even specific to the combination of OS + architecture. But since this is not needed for now, we can leave that as is and refactor only in case we ever run into such demand.

cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/Cve.java
laert-ll: I applied the changes, however `new TreeMap<>(this.conditions)` is a shallow copy, since the `List` values are shared with `this.conditions`. This is why I also made sure to copy each OS's `List` into a fresh `ArrayList` inside the loop before merging, otherwise `mergeVersionRage` would mutate the original record's lists in place.

https://github.com/devonfw/IDEasy/blob/4251b614a7bd77cbe612b4d4a1cadd5be2ce332a/cli/src/main/java/com/devonfw/tools/ide/url/model/file/json/Cve.java#L105-L111

hohwille: We could still improve since I now understood that the `Map.copyOf` should actually go to the constructor instead.
Dealing with nested collections and immutability is quite painful in Java.
Feel free to create a follow up story but ready for merge.

laert-ll:
Retested successfully with `2026.09.002-09_12_06-SNAPSHOT`:

```
ide -f install helm 3.18.0
Found 5 CVE(s) for version 3.18.0 of tool helm:
CVE-2025-53547 with severity 8.6 and affected versions: [(,3.17.4), [3.18.0,3.18.4)]
https://nvd.nist.gov/vuln/detail/CVE-2025-53547
CVE-2025-55198 with severity 6.5 and affected versions: [(,3.18.5)]
https://nvd.nist.gov/vuln/detail/CVE-2025-55198
CVE-2025-55199 with severity 6.5 and affected versions: [(,3.18.5)]
https://nvd.nist.gov/vuln/detail/CVE-2025-55199
CVE-2026-35206 with severity 4.4 and affected versions: [(,3.20.2), [4.0.0,4.1.4)]
https://nvd.nist.gov/vuln/detail/CVE-2026-35206
CVE-2026-63308 with severity 6.5 and affected versions: [(,4.2.3]]
https://nvd.nist.gov/vuln/detail/CVE-2026-63308
Which version do you want to install?
Option 1: current (3.18.0 - unsafe)
```

The vulnerability is detected, reported with severity and affected version ranges, and the safe/unsafe install prompt works. Backward compatibility with existing security.json files is confirmed.

Note: I could not exercise the new OS-specific conditions filtering, because no security.json in ide-urls uses that field yet, so there is no data to trigger it. Only the general CVE flow was verifiable.

Issue #1868, opened 2026-04-28
Revision of the functionality of global tool commandlets


jakozian:
### Feature idea

As an IDEasy user i want to use global tools (so non-portable, machine-wide installations) as i am already doing with local tools. That means using the IDEasy cli to wrap the tools functionality (if possible and useful).

**At the moment** there are only two global tools with pgadmin and docker (Rancher-/Docker-Desktop) and i can only install and maybe run them if i am lucky. Things like `get-version`, `get-edition`, `uninstall` and `install` are not handled or mal-handled. 

Instead of hard coding these functionalities into the respective sub class of `GlobalToolCommandlet` it would be useful for the future to find generic implementations for each OS and think about how the global commandlets can be revisioned. Some tools may have individual ways and even better ways than generic solutions but nonetheless the workflow of global tools has to be reconsidered.

### Additional context

https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/tool/pgadmin/PgAdmin.java
https://github.com/devonfw/IDEasy/blob/main/cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java

PR #2376, opened 2026-08-26
#2252: add uninstall support for globally installed tools on macOS

### LINKED ISSUE #2252: Provide uninstall support for globally installed tools (macOS)
### Task summary

Global tools installed on macOS should support uninstallation where possible and provide useful guidance when automatic removal is not feasible.

The solution should aim to offer a consistent user experience for uninstalling globally installed tools on macOS.

### Additional context

This task is part of issue #399 and focuses on macOS.

Unlike Linux package managers or Windows uninstall registrations, macOS does not provide a single standardized mechanism for uninstalling applications. Different installation methods may require different removal approaches.

The implementation should investigate which uninstall scenarios can be handled automatically and where user guidance is required. The goal is to provide a reusable solution that minimizes tool-specific uninstall logic while still giving users clear and actionable feedback.

Ali-Shariati-Najafabadi:
### Implemented changes:

* On macOS, `GlobalToolCommandlet.uninstall()` now tries, in order: uninstall via Homebrew (formula/cask) using the existing `NativePackageManager` package-manager infrastructure, then removing a known `*.app` bundle from the Applications folder, and only if both fail, logs actionable manual-uninstall guidance instead of the previous generic error.
* Added `NativePackageManager.BREW` / `BREW_CASK` so Homebrew reuses the same tested install/uninstall command derivation as the Linux package managers (APT/YUM/ZYPPER/DNF).
* Fixed the shared package manager command builder to only prepend `sudo` for package managers that need it (Homebrew must never run as root).
* Wired up `Docker` and `PgAdmin` (the two existing `GlobalToolCommandlet`s) with their Homebrew cask name and known macOS `.app` bundle name, mirroring the existing `getWindowsRegistryAppName()` pattern.

---

### Testing instructions

1. `mvn -Dtest=GlobalToolCommandletTest,NativePackageManagerTest,NativePackageTest,UninstallCommandletTest test` in the `cli` module.
2. Or run the full `cli` module test suite with `mvn test`.

---

### Checklist for this PR

- [x] You have added the issue implemented by your PR in CHANGELOG.adoc unless issue is labelled with `internal`
- [x] You have not changed any dependency in `pom.xml` files or otherwise if runtime dependencies changed, you have updated our LICENSE.asciidoc


CHANGELOG.adoc
```diff
@@ -28,6 +28,7 @@ Release with new features and bugfixes:
 * https://github.com/devonfw/IDEasy/issues/2219[#2219]: Add unpack commandlet
 * https://github.com/devonfw/IDEasy/issues/2251[#2251]: Provide generic uninstall support for globally installed tools (windows)
 * https://github.com/devonfw/IDEasy/issues/1135[#1135]: Fix PowerShell env variable initialization on Windows by sourcing functions from the PowerShell profile
+* https://github.com/devonfw/IDEasy/issues/2252[#2252]: Add uninstall support for globally installed tools on macOS
```
laim2003: Move up

laim2003 (approved):
Following your testing instructions every test is green on my mac 🙂

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
```diff
@@ -310,8 +316,61 @@ public void uninstall() {
...
+   * Uninstalls this tool on macOS. Unlike Linux, macOS has no single standardized package manager, so we try the best-effort options in order and finally
+   * fall back to giving the user actionable guidance if nothing could be done automatically.
+   */
+  private void uninstallMac() {
+    if (runWithPackageManager(false, getUninstallPackageManagerCommands(), NativePackageAction.UNINSTALL)) {
+      return;
+    }
+    Path appBundle = findMacApplicationBundle();
+    if (appBundle != null) {
+      this.context.getFileAccess().delete(appBundle);
```
hohwille, line 336: Isn't this a privileged operation on MacOS since at least mac os 12 (Monterey)?
I observed that my apps all moved to `/Applications` and this has some sort of access protection so regular processes (and viruses, worms, malware) cannot write or delete there.
I saw your code below but IMHO we would never reach it since we cause an exception here...
Correct me if I am wrong...

cli/src/main/java/com/devonfw/tools/ide/tool/NativePackageManager.java
```diff
@@ -61,7 +72,16 @@ public static NativePackageManager extractPackageManager(String command) {
...
 
-    return name().toLowerCase();
+    return this.binaryName;
+  }
+
+  /**
+   * @return {@code true} if commands of this {@link NativePackageManager} need to be run with {@code sudo} (root permissions), {@code false} otherwise (e.g.
+   *     for {@link #BREW}/{@link #BREW_CASK} that must never be run as root).
+   */
+  public boolean needsSudo() {
```
hohwille, line 82: I know JavaBeans conventions are stupid but following them this getter would need to be named `isSudo()`.
Maybe rename field to `needSudo` and getter to `isNeedSudo()`... WDYT?

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
```diff
@@ -92,10 +92,17 @@ protected List<NativePackage> getNativePackages() {
...
-        )
+        ),
+        new NativePackage(NativePackageManager.BREW_CASK, List.of("docker"))
     );
   }
 
+  @Override
+  public String getMacApplicationName() {
+
+    return "Docker";
```
hohwille, line 103: See also #2345 - seems to be the same for Mac.
Not a task of your story and PR, but it would be great if we can somehow manage to fully resolve #1868 and finally make global tool commandlets fully functional and new sub-classes maintainable.
I think we are on the perfect way and have achieved a lot already but the way to get this finally done is much longer than what I expected.

hohwille (approved):
@Ali-Shariati-Najafabadi thanks for your PR. Great improvement and JUnit coverage. 👍 
I left some comment. Let's see what we can just close as is or where we should do a small improvement before we finally merge.

cli/src/main/java/com/devonfw/tools/ide/tool/docker/Docker.java
hohwille: This is mainly FYI. It would be lovely if you could create a subtask of #1868 for it, comment a link here and then resolve this. 🙏

Ali-Shariati-Najafabadi: Created #2468 as a subtask of #1868 to track this properly (also covers the same gap you found on Linux in #2345). Resolving this thread.

cli/src/main/java/com/devonfw/tools/ide/tool/NativePackageManager.java
Ali-Shariati-Najafabadi: Done — renamed to needSudo / isNeedSudo() as suggested, in 49fedb10.

cli/src/main/java/com/devonfw/tools/ide/tool/GlobalToolCommandlet.java
Ali-Shariati-Najafabadi: you're right, good catch. fixed it so a delete failure falls back to the guidance message instead of blowing up with a stack trace. added a test for it too, see 49fedb10

hohwille (approved):
@Ali-Shariati-Najafabadi thanks for perfectly addressing all review comments 👍 
Ready for merge now.

shodiBoy1:
Retested on macOS (main).

Followed the testing instructions:

```bash
cd cli
mvn test -Dtest=GlobalToolCommandletTest,NativePackageManagerTest,NativePackageTest,UninstallCommandletTest
```

42 tests, BUILD SUCCESS. All green.

PR #2378, opened 2026-08-27
#2372: Add --retention-delay option to ide cleanup

### LINKED ISSUE #2372: Implement --retention-delay option for ide cleanup (stale files in updates, _ide/tmp and   ~/Downloads/ide)
### Task summary

Sub-task of #1580. The base functionality of `ide cleanup` (scanning the software repository + all projects, summary report, confirmation, deletion of unused tool versions) is already implemented and merged via #1953 and #2230. This sub-task adds the remaining age-based part of the feature: an option to delete files and folders that are not used and not older than a configurable delay.

### Additional context

_No response_

krystynaShatkovska:
### Implemented changes:

This PR makes the --retention-delay option of ide cleanup also apply to installed tool versions. It takes an ISO-8601 duration (e.g. P30D, PT2H30M; default 1 year). It now controls two things:
  - stale files (not modified within that period) in $IDE_HOME/updates, $IDE_ROOT/_ide/tmp and ~/Downloads/ide; and
  - installed tool versions: a version no project links to is now only deleted if it has not been modified for longer than that period. An unused version that is still within the period is kept, and the running IDEasy installation (the target of _ide/installation) is always kept even if it is old and unused.

  After deletion it deletes empty folders (but keeps the scanned roots). Invalid durations are rejected with a clear error.

  Note: with the default (no flag), an unused tool version installed within the last year is now kept rather than deleted. Use a
  smaller --retention-delay (or backdate the folder) to see an unused version get removed.

 ### Testing instructions

  Automated tests

  Run the test class, it covers all the core cases (including the new tool-version retention cases: an old unused version is
  deleted, a fresh unused version is kept, and the running installation is kept regardless of age):

  mvn -pl cli test -Dtest=CleanupCommandletTest

  Manual test

  ide cleanup scans three roots for stale files: $IDE_HOME/updates (only inside a project), $IDE_ROOT/_ide/tmp, and the download cache ~/Downloads/ide. It also marks unused installed tool versions for deletion, subject to the same retention delay. The manual test must be run against a throwaway IDE root, not your real one, for two reasons:
  - The download cache lives under user home (~/Downloads/ide), which is independent of IDE_ROOT. So throwaway IDE_ROOT alone would still scan (and delete from) real ~/Downloads/ide.
  - IDE_ROOT is only used when no real IDE home is found by walking up from the working directory. If you run from inside your IDE tree, the real home is found and the IDE_ROOT override is ignored.

  So you must isolate both the IDE root and the user home, and run from a working directory outside your IDE tree. Using exec-ide.sh  / mvn exec:exec does not achieve this (the forked process's working directory is the project), so the steps below invoke java directly.

  Set the throwaway location once (any path outside your IDE tree, Windows example shown):

  $THROW = a fresh temp directory of your choice, e.g. %TEMP%\ide-cleanup-manual
  $RUN = a sibling empty directory, e.g. %TEMP%\ide-cleanup-run

  Build the classpath (from the repo root; regenerates cli/cp.txt with the CLI's dependency jars):

  mvn -pl cli dependency:build-classpath -Dmdep.outputFile=cli/cp.txt

  Create the throwaway IDE root - a stale file, a fresh file, an old unused tool version, a fresh unused tool version, a "running" IDEasy version, and a link from _ide/installation to that running version (this is the guard under test; the running version is backdated so the test proves the guard overrides staleness):

  Windows (PowerShell - uses a directory junction, no admin needed):

  $T = "C:\Users\<you>\AppData\Local\Temp\ide-cleanup-manual" # your $THROW
  if (Test-Path $T) { Remove-Item $T -Recurse -Force }
  New-Item -ItemType Directory -Path "$T\_ide\tmp" -Force | Out-Null
  New-Item -ItemType Directory -Path "$T\_ide\software\maven\ideasy\default\2026.08.002" -Force | Out-Null # running install
  New-Item -ItemType Directory -Path "$T\_ide\software\maven\ideasy\default\1.0" -Force | Out-Null # unused, old
  New-Item -ItemType Directory -Path "$T\_ide\software\maven\ideasy\default\3.0" -Force | Out-Null # unused, fresh
  New-Item -ItemType Directory -Path "$T\Downloads\ide" -Force | Out-Null
  Set-Content "$T\_ide\tmp\stale.bin" "x" -Force
  Set-Content "$T\_ide\tmp\fresh.bin" "x" -Force
  (Get-Item "$T\_ide\tmp\stale.bin").LastWriteTime = (Get-Date).AddDays(-31)
  (Get-Item "$T\_ide\software\maven\ideasy\default\1.0").LastWriteTime = (Get-Date).AddDays(-31) # old + unused -> deleted
  (Get-Item "$T\_ide\software\maven\ideasy\default\2026.08.002").LastWriteTime = (Get-Date).AddDays(-31) # old, but running -> kept
  New-Item -ItemType Junction -Path "$T\_ide\installation" -Target "$T\_ide\software\maven\ideasy\default\2026.08.002" | Out-Null
  New-Item -ItemType Directory -Path "$env:TEMP\ide-cleanup-run" -Force | Out-Null # empty cwd, outside the IDE tree

  macOS / Linux (bash - uses a symlink):

  T=$(mktemp -d) # your $THROW
  mkdir -p "$T/_ide/tmp" "$T/_ide/software/maven/ideasy/default/2026.08.002" \
           "$T/_ide/software/maven/ideasy/default/1.0" \
           "$T/_ide/software/maven/ideasy/default/3.0" "$T/Downloads/ide"
  echo x > "$T/_ide/tmp/stale.bin"; echo x > "$T/_ide/tmp/fresh.bin"
  touch -d "31 days ago" "$T/_ide/tmp/stale.bin" # (GNU; macOS: touch -t)
  touch -d "31 days ago" "$T/_ide/software/maven/ideasy/default/1.0" # old + unused -> deleted
  touch -d "31 days ago" "$T/_ide/software/maven/ideasy/default/2026.08.002" # old, but running -> kept
  ln -s "$T/_ide/software/maven/ideasy/default/2026.08.002" "$T/_ide/installation"
  mkdir -p "$TMPDIR/ide-cleanup-run"

  Run cleanup, fully isolated to the throwaway root. IDE_ROOT and user.home are both pointed at $THROW, so your real IDE_ROOT, your real ~/Downloads/ide, and your real software are untouched. Run from the empty working directory (not from inside your IDE tree).
  Answer yes twice (first = terms of use, second = "Do you want to continue?"):

  Windows (Git Bash):

  REPO='/c/Users/<you>/projects/IDEasy/workspaces/main/IDEasy'   # your repo root (MSYS path)
  T='C:\Users\<you>\AppData\Local\Temp\ide-cleanup-manual'       # your $THROW (Windows path)
  cd "$(cygpath -w "$TEMP/ide-cleanup-run")"                      # empty cwd, outside the IDE tree
  export IDE_ROOT="$T"
  CLASSES_WIN=$(cygpath -w "$REPO/cli/target/classes")
  CP="$CLASSES_WIN;$(tr -d '\n' < "$REPO/cli/cp.txt")"
  printf 'yes\nyes\n' | java -Duser.home="$T" -cp "$CP" com.devonfw.tools.ide.cli.Ideasy cleanup --retention-delay=P30D

  macOS / Linux:

  classes="$(pwd)/cli/target/classes"; cp_file="$(pwd)/cli/cp.txt"
  cd "$TMPDIR/ide-cleanup-run" # empty dir, outside the IDE tree
  export IDE_ROOT="$T"
  CP="$classes:$(tr -d '\n' < "$cp_file")" # note: ':' separator on unix
  printf 'yes\nyes\n' | java -Duser.home="$T" -cp "$CP" com.devonfw.tools.ide.cli.Ideasy cleanup --retention-delay=P30D

  Before it deletes anything, confirm isolation: the IDE prints its resolved document path and IDE_ROOT is set to .... Both must point at your $THROW, not your real IDE root. If it prints your real root, stop - the isolation failed and it will act on your real installation.

  Confirm the expected results (substitute your $THROW):

  $THROW/_ide/tmp/stale.bin → deleted (stale)
  $THROW/_ide/tmp/fresh.bin → kept (fresh)
  $THROW/_ide/software/maven/ideasy/default/1.0 → deleted (unused AND older than 30 days)
  $THROW/_ide/software/maven/ideasy/default/3.0 → kept (unused, but not older than 30 days)
  $THROW/_ide/software/maven/ideasy/default/2026.08.002 → kept (running install, guarded — kept even though it is old)

  Windows:

  Test-Path "$T\_ide\tmp\stale.bin" # expected: False
  Test-Path "$T\_ide\tmp\fresh.bin" # expected: True
  Test-Path "$T\_ide\software\maven\ideasy\default\1.0" # expected: False
  Test-Path "$T\_ide\software\maven\ideasy\default\3.0" # expected: True
  Test-Path "$T\_ide\software\maven\ideasy\default\2026.08.002" # expected: True

  macOS / Linux:

  test -e "$T/_ide/tmp/stale.bin"; echo $? # expected 1 (False)
  test -e "$T/_ide/tmp/fresh.bin"; echo $? # expected 0 (True)
  test -e "$T/_ide/software/maven/ideasy/default/1.0"; echo $? # expected 1 (False)
  test -e "$T/_ide/software/maven/ideasy/default/3.0"; echo $? # expected 0 (True)
  test -e "$T/_ide/software/maven/ideasy/default/2026.08.002"; echo $? # expected 0 (True)

  Verify invalid --retention-delay values are rejected (same isolated setup; each must print Invalid value ... for --retention-delay and stop before doing anything):

  PT6M10D # months not allowed in a time-based duration
  P0D # must be positive
  -P30D # must be positive
  abc # not ISO-8601

  (Use the same java -Duser.home="$T" -cp "$CP" ... invocation as step 3, with each value substituted for P30D.)

  Clean up:

  rm -rf <your $THROW>

---

### Checklist for this PR

  with `internal`
- [ ] You have not changed any dependency in `pom.xml` files or otherwise if runtime dependencies changed, you have updated our [LICENSE.asciidoc](https://github.com/devonfw/IDEasy/blob/main/documentation/LICENSE.adoc)


cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -49,21 +63,59 @@ protected void doRun() {
...
+      return DEFAULT_RETENTION_DELAY;
+    }
+    try {
+      return Duration.parse(value);
+    } catch (DateTimeParseException e) {
+      throw new CliException(
+          "Invalid value '" + value + "' for --retention-delay. Please provide a time-based ISO-8601 duration such as P30D or PT2H30M.",
+          e);
+    }
+  }
```
samuelkos17, line 122: Duration.parse() legally accepts negative and zero values. This means it won't throw the CliException and later down the line in isStale() every file under all roots becomes stale, which leads to every file being deleted. You should add a positivity check here.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -356,4 +408,148 @@ private int deleteFolder(Path path) {
...
+    }
+
+    for (Path child : this.context.getFileAccess().listChildren(folder, child -> true)) {
+      if (Files.isDirectory(child)) {
+        discoverStaleFilesRecursive(child, retentionDelay, staleFiles);
+      } else if (isStale(child, retentionDelay)) {
+        staleFiles.add(child);
+      }
+    }
+  }
```
samuelkos17, line 456: Files.isDirectory(child) follows links. If a symlink inside any root (e.g. ~/Downloads/ide/work -> C:\data) makes the scan descend outside the roots and delete the target's stale files, which would be a huge problem. Furthermore a self-referencing link causes unbounded recursion. You need to check for link children here and skip these. This also destroyed my IDEasy installation and I'm not sure how yours didn't get destroyed when you tested your work.

cli/src/main/resources/nls/Help.properties
```diff
@@ -13,6 +13,7 @@ cmd.claude=Tool commandlet for Claude Code CLI.
 cmd.claude.detail=Claude Code CLI is a command-line interface for interacting with the Claude AI assistant. Detailed documentation can be found at https://code.claude.com/docs/en/overview
 cmd.cleanup=Commandlet to clean up the IDEasy installation by uninstalling all unused tools.
 cmd.cleanup.detail=This will remove any installed tools that are currently not in use by an IDEasy project. Before anything is deleted you are asked for confirmation. Run "ide -b -f cleanup" to skip the confirmation.
```
samuelkos17, line 15: This needs to be updated to match the new functionality.

cli/src/main/resources/nls/Help_de.properties
```diff
@@ -13,6 +13,7 @@ cmd.claude=Werkzeug Kommando für Claude Code CLI.
 cmd.claude.detail=Claude Code CLI ist ein KI-gestützter Programmierassistent, der über die Befehlszeile ausgeführt wird. Detaillierte Dokumentation ist zu finden unter https://code.claude.com/docs/de/overview
 cmd.cleanup=Werkzeug zum Aufräumen der IDEasy-Installation durch Deinstallieren aller ungenutzten Werkzeuge.
 cmd.cleanup.detail=Dies wird alle installierten Werkzeuge entfernen, die derzeit von keinem IDEasy-Projekt verwendet werden. Bevor etwas gelöscht wird, wirst du um Bestätigung gebeten. Führe "ide -b -f cleanup" aus, um die Bestätigung zu überspringen.
```
samuelkos17, line 15: This needs to be updated to match the new functionality.

cli/src/main/resources/nls/Help_de.properties
```diff
@@ -13,6 +13,7 @@ cmd.claude=Werkzeug Kommando für Claude Code CLI.
...
+cmd.cleanup.opt.--retention-delay=die Aufbewahrungsdauer von Dateien in den Ordnern 'updates', '_ide/tmp' und 'Downloads/ide', d.h. Dateien, die innerhalb dieses Zeitraums nicht modifiziert wurden, werden als veraltet gelöscht. Eine zeitbasierte ISO-8601-Dauer (z. B. 'P30D' für 30 Tage oder 'PT2H30M' für 2 Stunden und 30 Minuten). Standardmäßig 1 Jahr (365 Tage), wenn nicht angegeben.
```
samuelkos17:
```suggestion
cmd.cleanup.opt.--retention-delay=Die Altersgrenze von Dateien in den Ordnern 'updates', '_ide/tmp' und 'Downloads/ide', d.h. Dateien, die innerhalb dieses Zeitraums nicht modifiziert wurden, werden als veraltet gelöscht. Eine zeitbasierte ISO-8601-Dauer (z. B. 'P30D' für 30 Tage oder 'PT2H30M' für 2 Stunden und 30 Minuten). Standardmäßig 1 Jahr (365 Tage), wenn nicht angegeben.
```

Furthermore the folders you mention here are correct for Windows and Linux, however on macOS there are somewhere else, maybe just remove the path descriptions and describe the folders?

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -49,21 +63,59 @@ protected void doRun() {
...
 
     Step step = this.context.newStep("Identify unused software");
     step.run(() -> discoverUnusedSoftware(installedSoftware), true);
 
     logSoftwareToBeDeleted(installedSoftware.getTools());
 
-    if (hasSoftwareToDelete(installedSoftware.getTools())) {
+    List<Path> staleRoots = new ArrayList<>();
+    List<Path> staleFiles = new ArrayList<>();
+    if (this.context.getIdeHome() != null) {
```
samuelkos17: This if-guard prevents any clean-up from happening, however _ide/tmp and the download cache cleanup would work when IDE_HOME is null. I'm not sure if this intentional, but you might want to change that if it's not.

cli/src/main/resources/nls/Help.properties
```diff
@@ -13,6 +13,7 @@ cmd.claude=Tool commandlet for Claude Code CLI.
...
+cmd.cleanup.opt.--retention-delay=the retention period of files in the 'updates', '_ide/tmp' and 'Downloads/ide' folders, i.e. files that were not modified within this period are deleted as stale. A time-based ISO-8601 duration (e.g. 'P30D' for 30 days or 'PT2H30M' for 2 hours and 30 minutes). Defaults to 1 year (365 days) if not provided.
```
samuelkos17:
```suggestion
cmd.cleanup.opt.--retention-delay=The retention period of files in the 'updates', '_ide/tmp' and 'Downloads/ide' folders, i.e. files that were not modified within this period are deleted as stale. A time-based ISO-8601 duration (e.g. 'P30D' for 30 days or 'PT2H30M' for 2 hours and 30 minutes). Defaults to 1 year (365 days) if not provided.
```

Furthermore the folders you mention here are correct for Windows and Linux, however on macOS there are somewhere else, maybe just remove the path descriptions and describe the folders?

cli/src/test/java/com/devonfw/tools/ide/commandlet/CleanupCommandletTest.java
```diff
@@ -1,16 +1,22 @@
 package com.devonfw.tools.ide.commandlet;
 
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
```
samuelkos17, line 3:
```suggestion
```

dead code

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -49,21 +63,59 @@ protected void doRun() {
 
     LOG.debug("Start cleanup commandlet");
 
+    Duration retentionDelay = getRetentionDelay();
```
samuelkos17, line 66: You might want to rename this since you already have a StringProperty called retentionDelay in the class.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
```diff
@@ -356,4 +408,148 @@ private int deleteFolder(Path path) {
...
+
+    if (staleFiles.isEmpty()) {
+      LOG.info("No stale files older than {} will be deleted.", retentionDelay);
+    } else {
+      for (Path staleFile : staleFiles) {
+        LOG.info("\t - {} will be deleted", staleFile);
+      }
+      LOG.info("Summary: {} stale file(s) older than {} will be deleted.", staleFiles.size(), retentionDelay);
+    }
+  }
```
samuelkos17, line 561: You might want to format the duration human-readably.

samuelkos17 (changes requested):
Thanks for adding the --retention-dely to the cleanup commandlet. I tried to follow your testing steps, however the commands you provided didn't work out for me. I've manually moved the files to `/_ide/tmp` though and then ran the cleanup command and it worked!
While reviewing I found some problems that could lead to issues and that really need to get addressed before moving this to In Review. You can find them in the Comments here.
Besides that I still have on recommendation:
documentation/tmp.adoc line 18 needs to be updated according to the new functionality.

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
krystynaShatkovska: You're right that Duration.parse() accepts P0D and -P1D, and with either of them isStale() would flag every file under every root as stale and delete them all. I added a check right after parsing. if the value is zero or negative it now throws a CliException with clear message, so the failure happens up front instead of silently deleting everything. Added a test covering P0D and -P1D

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
krystynaShatkovska: tthanks for flagging that it took out your install. Files.isDirectory() follows the link, so the scan would walk into a link target outside the roots (deleting its stale files) and a self-referencing link would loop forever. The scan now skips any child that Files.isSymbolicLink() reports as a link. It neither recurses into it nor deletes it. I added a test also, stale file directly under a scanned root is deleted, but a stale file reached only through a symlinked directory is left alone. That's the exact class of bug that destroyed your setup

cli/src/main/resources/nls/Help.properties
krystynaShatkovska: Updated. I also took your second point, the paths 'updates', '_ide/tmp', 'Downloads/ide' are only correct on Windows and Linux and sit somewhere else on macOS, so I removed the literal paths and describe the folders generically ("the IDEasy updates, temporary and download cache folders"). Same change in the German file. Is it suitable?

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
krystynaShatkovska: It wasn't intentional. Thank you for locating the problem. Guarding the whole stale-file step on getIdeHome() != null meant the _ide/tmp and download-cache roots (which live under IDE_ROOT/user home and are available even outside a project) were silently skipped whenever IDE_HOME was null. I moved the null-check down into getStaleFileRoots(), where only the updates root actually needs it. Now the other two roots are cleaned regardless of IDE_HOME

cli/src/test/java/com/devonfw/tools/ide/commandlet/CleanupCommandletTest.java
krystynaShatkovska: I double checked this one. I don't think it's actually dead. Line 3 is import static ...assertThatThrownBy and it's used in testCleanupRejectsInvalidRetentionDelay (the assertThatThrownBy(cleanup::run)... call). Removing it would break compilation. If you were pointing at something else that I missed, let me know which line

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
krystynaShatkovska: Done

cli/src/main/java/com/devonfw/tools/ide/commandlet/cleanup/CleanupCommandlet.java
krystynaShatkovska: Done. Also added a small formatDurationHumanReadable() helper so the summary reads No stale files older than 365 day(s) will be deleted. instead of the raw PT8760H

samuelkos17 (commented):
Thanks for resolving my comments! I've tested the changes but came across another issue:
`ide cleanup` now wants to delete all my IDEasy installations including the one i'm currently using, the --retention-delay can have any value for that to happen, so i guess it doesn't work there. Does this problem also occur on ur end?

samuelkos17 (changes requested):
Thanks for implementing a guard to keep the current IDEasy installation from deleting, however while looking at your new changes I realized that the retention-delay has no impact on installed tool versions (claude, intellij, uv, ideasy). Even when you run `ide cleanup` with retention-delay set to any positive value the command wants to delete all installed tool versions that no project links to. I'm not quite sure if this intentional, but judging by what the issue description says I think it's not intentional and that you need to add the retention-delay checks there aswell, but please correct me if I'm wrong.

samuelkos17 (approved):
Thanks for also adding the retention-delay to unused software! I think everything is fine now so I'll move this to In Review 😄

PR #2212, opened 2026-07-27
#2190: IDEasy destroys python installation

vivu001:
### Implemented changes:

* Added missing-version-file recovery for preserved local tools.
* Restored the version marker instead of reinstalling Python.
* Kept the existing Python installation intact.

---

### Testing instructions

1. Run [./build-local-dev.sh](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/run-ideasy-locally.adoc#run-ideasy-locally-as-native-binary-using-the-local-dev-script) to build the native image.
2. If Python is not already installed, run `ide install python` once (or `Spyder`) and delete the file `$[IDE_HOME]/software/python/.ide.software.version`.
3. Run command
```
ide install python
```
Result:
```
$ ide install python
Tool python is missing version file in C:\Users\xxx\projects\IDEasy\software\python\.ide.software.version
No CVEs found for version 0.12.0 of tool uv.
Tool python is missing version file in C:\Users\xxx\projects\IDEasy\software\python\.ide.software.version
Found 1 CVE(s) for version 3.14.6 of tool python:
CVE-2026-15308 with severity 7.5 and affected versions: [(,3.15.0)]
https://nvd.nist.gov/vuln/detail/CVE-2026-15308
Could not find any other version resolving your CVEs.
Please keep attention to this tool and consider updating as soon as security fixes are available.
Please note that by selecting an unsafe version to install, you accept the risk to be attacked.
Which version do you want to install?
Which version do you want to install?
Option 1: current (3.14.6 - unsafe)

No CVEs found for version 0.12.0 of tool uv.
Version 0.12.0 of tool uv is already installed
Version file missing at C:\Users\xxx\projects\IDEasy\software\python\.ide.software.version - restoring it for tool python
Successfully installed python in version 3.14.6 at C:\Users\xxx\projects\IDEasy\software\python
```
4. Confirm that IDEasy restored `$IDE_HOME/software/python/.ide.software.version`.
---

### Checklist for this PR

Make sure everything is checked before merging this PR. For further info please also see
our [DoD](https://github.com/devonfw/IDEasy/blob/main/documentation/contributing/DoD.adoc).

- [x] When running `mvn clean test` locally all tests pass and build is successful
- [x] PR title is of the form `#«issue-id»: «brief summary»` (e.g. `#921: fixed setup.bat`). If no issue ID exists, title only.
- [x] PR top-level comment summarizes what has been done and contains link to addressed issue(s)
- [x] PR and issue(s) have suitable labels
[...]

PR #2384, opened 2026-08-27
#2190: IDEasy destroys my python installation

### LINKED ISSUE #2190: IDEasy destroys my python installation
### Actual behavior

I do not exactly know how to reproduce this but our python support is still buggy.
Under some circumstances python tooling messes up our IDEasy specific makers (most likely something seems to delete `software/python/.ide.tool.version`).
As a result IDEasy considers `python` installation as corrupted and simply deletes it:
```
$ ide spyder
...
Deleting corrupted installation at D:\projects\xxx\software\python
Successfully installed python in version 3.14.6 at D:\projects\xxx\software\python
...
```

This is very undesired when I installed tons of dependencies and libraries (via `pip` and `uv sync`).

### Reproduce

1. Cause the problematic situation (maybe manually delete `software/python/.ide.tool.version` but we should also figure out if `uv` itself or `python` or `pip` is removing it under some circumstances).
2. Run an IDEasy command that installs python e.g. `ide install python` or `ide spyder`.

### Expected behavior

The existing Python installation is reused and preserved - only the new libs/tools will be added/installed on top.

### IDEasy status

```bash
Your version of IDEasy is 2026.07.002-07_21_05-SNAPSHOT.
You are online.
Your settings are up-to-date.
```

### Related/Dependent issues

#1442

### Comments/Hints

If we cannot prevent some uv/python tooling from removing our `.ide.tool.version` we would need to redesign the way where and how we store the installed version of a local tool.

### LINKED ISSUE #2284: Restore .ide.software.version from actual installed Python version
### Task summary

Based on the proposal in the [review](https://github.com/devonfw/IDEasy/pull/2212#discussion_r3728810442), IDEasy should recover the real installed version when `.ide.software.version` is missing instead of recreating it from `resolvedVersion`.

### Additional context

Suggestion for PR improving `LocalToolCommandlet.installTool`:

* We could change the logic here to use `getInstalledVersion(Path)`.
* This method already logs a warning if the version file is missing
* We can add another protected method `getInstalledVersionWithMissingVersionFile(Path toolPath)` that can determine the version if software repo is used by resolving the link and taking the name of the leaf-folder but otherwise return `null`.
* If that method returns non null, then we recreate the version file repairing the situation.
* Finally you can override the new method in `Python` and try to get the installed version via `python --version`.

This is a generic proposal that should fix our issue with Python but could also help for other tools.

UPDATE: See https://github.com/devonfw/IDEasy/pull/2306#pullrequestreview-4995719328 for how this story may be implemented properly after that PR is merged.

Paras14:
### Implemented changes:

* `LocalToolCommandlet` now really writes the `.ide.software.version` file when it is missing. The branch added in #2212 logged "restoring it" but never called `writeVersionFile`, so the file stayed missing and every following command showed the warning again and asked for a CVE version selection.
* Added an overridable hook `detectInstalledVersion(installationPath, resolvedVersion)`. The default returns the requested version. `Python` overrides it and determines the version that is really installed, first from `pyvenv.cfg` (only if it contains a patch version, because `uv` writes just `3.13` for its own interpreters) and otherwise by calling `python --version`. This prevents writing a version file that claims a version the virtual environment does not have, which can happen because `uv sync` may recreate the venv with a different python version.
* If the detected version differs from the requested one, the regular installation continues. It backs up the existing installation instead of deleting it.
* Invalidates the cached installed edition and version from #2296, because the version file is changed outside of `createToolInstallation`.
* Added a regression test to `PythonTest` that fails without these changes. The mocked python binary in the `uv` test project now answers `--version`.

---

### Testing instructions

1. Run `mvn clean test` in the `cli` module. The new test `PythonTest.testInstallRestoresMissingVersionFileAndPreservesPackages` covers this fix. It installs python, adds a file to `site-packages`, deletes `.ide.software.version` and installs again. It asserts that the version file is restored with the correct version, that the installed packages are preserved and that no "Deleting corrupted installation" happens.
2. To see the test catch the bug, revert `LocalToolCommandlet.java` and `Python.java` to `main` and run `PythonTest` again. It fails because the version file is not restored.

Note for reviewers: the test pins the OS to `SystemInfoMock.LINUX_X64`, because the mocked python payload in the `uv` test project has a Unix style `bin` folder that collides with the `bin` to `Scripts` symlink created on Windows. Therefore the Windows specific parts of `detectInstalledVersion` (bin path resolution, `.exe` suffix, parsing the real interpreter output) are not covered by automated tests.

---

### Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -208,10 +221,17 @@ public ToolInstallation installTool(ToolInstallRequest request) {
           LOG.warn("Deleting corrupted installation at {}", installationPath);
           fileAccess.delete(installationPath);
         } else {
-          // Version file missing but tool allows this - restore it and preserve installation
-          LOG.warn("Version file missing at {} - restoring it for tool {}", toolVersionFile, this.tool);
-          // Restore the missing file
-          return createToolInstallation(installationPath, resolvedVersion, false, processContext, additionalInstallation);
+          // Version file is missing but tool allows this - restore the file and preserve the installation.
+          VersionIdentifier detectedVersion = detectInstalledVersion(installationPath, resolvedVersion);
```
hohwille: I thought that we would simply use the new infrastructure of #2296 and integrate that directly in `computeInstalledEditionAndVersion`. This way you will also fix the `ide get-version python` currently returns a wrong result if the version file is lost.
Also the `.ide.software.version` file is actually a kind of hack inherited from devonfw-ide (legacy file is `.devonfw.software.version`.
My plan is to get more away from this on the long run.
The installed edition of regular local tools is actually determined from the link target of the software folder in the software repository and the version can be determined in the same way.
I would therefore suggest, that we simply rely on `getInstalledVersion()` what is already used in the installation routine.

See here:
https://github.com/devonfw/IDEasy/blob/740bba7169436ef3a2e619631317647529ba7c43/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java#L369-L409

So Python can simply override getInstalledVersion() and if `null` is returned from super call, simply add fallbacks (what is currently in `detectInstalledVersion`).

CHANGELOG.adoc
```diff
@@ -6,6 +6,7 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/2190[#2190]: IDEasy destroys my python installation
```
hohwille: I opened the next release. Please shift up.

hohwille (approved):
@Paras14 thanks for your PR. The way how you detect the python version is perfectly implemented. Test covers the feature nicely and according python mocked script adoption makes sense for testing. 👍 
I left some review suggestion how I thought we should connect #2296 with this to fix the problem properly. Please have a look,

hohwille (approved):
@Paras14 thanks for your update. Now ready for merge. 👍

Hiepiscus:
Successfully tested with 2026.09.002-09_12_06-SNAPSHOT on windows 11.

All PythonTest test cases run successfully.

PR #2390, opened 2026-08-31
#2193: add alternatives and dependencies to auto-completion entries

### LINKED ISSUE #2193: Support for conditional auto-completion
### Feature idea

After #2192 is implemented, we should add support to introduce dependencies between `CompletionEntry` objects.
That means a `CompletionEntry` can optionally have another `CompletionEntry` as a dependency.

Lets look at a concrete use-case for mvn(d):

We should somehow group together goal options like
* `exec:java`
* `exec:exec`

Like with synonyms it does not make sense to have both together, however logically they are not synonyms since they do something else so they should both be rather separate instances of `CompletionEntry` but both with a `alternative` relation on each other (hint: cannot be designed immutable since we have a cycle here).
So if the `alternative` is already present the according `CompletionEntry` cannot be selected.

Further we define dependencies:
* `-Dexec.mainClass=` has a dependency on `exec:java`
* `-Dexec.args=` has a dependency on `exec:java` or `exec:exec`

This example should shape the design how we extend `CompletionEntry` and potential convenience methods in ``AutoCompletionRegistry`.

### Additional context

After rethinking this, it is IMHO easier to put all features into `CompletionEntry` without separation into sub-classes so far.

Hiepiscus:
### This PR fixes #2193 

This PR extends the auto-completion model with support for alternatives and dependencies between `CompletionEntry` objects.

### Implemented changes:

* Extended `CompletionEntry` with alternative relationships.
  * Alternatives remain separate completion entries because they are different commands.
  * The relationship is symmetric, so adding one entry as an alternative links both entries.
  * Once one alternative or one of its synonyms has already been provided, the other entry is no longer suggested.
* Extended `CompletionEntry` with dependency groups.
  * Dependencies use OR logic within one group.
  * Multiple dependency groups use AND logic.
  * A completion entry is only suggested when every configured dependency group is satisfied.
  * Candidates and their synonyms are both considered when checking whether a dependency is satisfied.
* Addd convenience methods to `AutoCompletionRegistry`.
  * `addAlternative(String, String)` connects two previously registered completion entries.
  * `addDependency(String, List<String>)` registers an OR-dependency group for a previously registered entry.
  * Both methods fail with an `IllegalStateException` if a referenced candidate has not been registered with `add(String)`.
* Changed Maven completion.
  * `exec:java` and `exec:exec` are registered as alternatives.
  * `-Dexec.mainClass=` depends on `exec:java`.
  * `-Dexec.args=` depends on either `exec:java` or `exec:exec`.
* Changed `CompleteTest`:
  * Add tests for alternative candidates and dependency completion filtering. 

---

### Testing instructions

1. Run `build-local-dev.sh`
2. In the console type `ide mvn exec:exec` and use the auto completion with `<tab><tab>`
3. Verify that the output should not contain `exec:java`
4. Type `ide mvn -D` and use `<tab><tab>`
5. Verify that `-Dexec.mainClass=` or `-Dexec.args=` is not in the output
6. Type `ide mvn exec:java -D` and use `<tab><tab>
7. Verify that the output should contain `-Dexec.mainClass=` or `-Dexec.args=`
8. Type `ide mvn exec:exec -D` and use `<tab><tab>`
9. Verify that the output should contain `-Dexec.args=` and does not contain `-Dexec.mainClass=`.

---

### Checklist for this PR

  with `internal`


samuelkos17 (commented):
Great work on adding alternatives and dependencies to the auto-completion entries! I've followed your testing instructions and everything worked fine. On Step 9 you wrote that the output should **only** contain `-Dexec.args=` but I guess you meant it should contain `-Dexec.args=` but not `-Dexec.mainClass=`. If that's true then everything worked as intended and you can move it to In Review :smile:. Also you might have to change the Changelog entry to the next release but i couldn't flag it since this branch isn't up-to-date with main.

Hiepiscus:
> On Step 9 you wrote that the output should only contain -Dexec.args= but I guess you meant it should contain -Dexec.args= but not -Dexec.mainClass=

@samuelkos17 Thanks for the review! You are correct. I meant that the output should contain -Dexec.args= but not -Dexec.mainClass=. I have updated the test instructions accordingly.

cli/src/main/java/com/devonfw/tools/ide/completion/AutoCompletionRegistry.java
```diff
@@ -56,5 +55,61 @@ public void complete(String arg, CompletionCandidateCollector collector,
...
 
+  /**
+   * Registers two already-added candidates as alternatives to each other, so that once one is provided on the command line,
+   * the other is no longer be suggested.
+   *
+   * @param candidate1 the text of the first candidate (must have been added via {@link #add(String)} before).
+   * @param candidate2 the text of the second candidate (must have been added via {@link #add(String)} before).
+   * @throws IllegalStateException if either candidate has not been registered via {@link #add(String)}.
+   */
+  public void addAlternative(String candidate1, String candidate2) {
```
hohwille: Wouldn't the following API make more sense?
```
public CompletionEntry add(String candidate, String... alternatives) { ... }
```

Then you cannot use the API wrong and get an exception.
Further, alternatives IMHO need to be linked bi-directional.
In your example with 2 candidates:
```
entry1.addAlternative(entry2);
entry2.addAlternative(entry1);
```

However, I would consider to create a shared `Set<String>` of all names (candidate + alternatives) and attach it to all `CompletionEntry` instances created here what is IMHO easier and more efficient.
If you always add the candidate to that set of alternatives - even if no alternatives are given - you could always use the alternatives to check if already provided:
```
return this.synonyms.stream().anyMatch(alreadyProvided::contains);
```

To clarify my thoughts:
Technically a synonym and an alternative is not exactly the same thing.
A synonym has the same effect so "mvn -T 1C ..." and "mvn --threads 1C ..." is just a different way of writing the same thing and will have the same effect.
An alternative has a different effect so "mvn exec:exec ..." does something other than "mvn exec:java ...".
However, for our auto-completion we IMHO do not need to distinguish these two.
If one of them is already present, it does not make sense to suggest the other.
Therefore I would actually suggest to remove the `synonyms` field and just go with `Set<String> alternatives`.
Also remove the `add(String, String)` method that will then be taken over by this new method.
Does that all make sense to you?

cli/src/main/java/com/devonfw/tools/ide/completion/AutoCompletionRegistry.java
```diff
@@ -56,5 +55,61 @@ public void complete(String arg, CompletionCandidateCollector collector,
...
+  }
+
+  /**
+   * Registers a dependency for {@code candidate}: it is only suggested once at least one of {@code depends} has already been provided on the command line.
+   *
+   * @param candidate the text of the dependent candidate (must have been added via {@link #add(String)} before).
+   * @param depends the texts of the candidates of which at least one must already be provided (OR semantics).
+   * @throws IllegalStateException if {@code candidate} or any of {@code depends} has not been registered via {@link #add(String)}.
+   */
+  public void addDependency(String candidate, List<String> depends) {
```
hohwille: If the `add` methods return the created `CompletionEntry` then you could directly attach dependencies to it.
```
CompletionEntry execMainClass = registry.add("-Dexec.mainClass=");
execMainClass.addDependency"exec:java");
```
Or even more compact in fluent style:
```
registry.add("-Dexec.mainClass=").addDependency("exec:java");
```
Since "exec:java" is already a alternative of "exec:exec", we do not have to list it again as dependency. 

Maybe I am missing something but I do not understand why you modelled dependencies as `List<List<String>>`. I read your implementation so all elements in such a sub-list have to be met.
However, I do not see the use-case for this.
Which scenario do you have in mind here?
I would use a simple `Set<String>` instead. Most entries will either have no dependency at all or only have a single dependency. I don't know if we actually need more than a single dependency but it is always good to think further for future scenarios so that `Set<String>` makes sense..

cli/src/main/java/com/devonfw/tools/ide/completion/AutoCompletionRegistry.java
```diff
@@ -56,5 +55,61 @@ public void complete(String arg, CompletionCandidateCollector collector,
...
+   * @param candidate the candidate to find.
+   * @return the {@link CompletionEntry} whose {@link CompletionEntry#getCandidate() candidate} matches, or {@code null} if not found.
+   */
+  private CompletionEntry findEntry(String candidate) {
+    for (CompletionEntry entry : this.entries) {
+      if (entry.getCandidate().equals(candidate)) {
+        return entry;
+      }
+    }
+    return null;
```
hohwille: IMHO it seems about time to refactor `List<CompletionEntry> entries` to `Map<String, CompletionEntry> entryMap`. We need many lookups also for dependencies and alternatives.
Lookup is O(1) then instead of O(n).

cli/src/main/java/com/devonfw/tools/ide/completion/CompletionEntry.java
```diff
@@ -63,4 +88,68 @@ public void complete(String arg, CompletionCandidateCollector collector, Propert
...
+    }
+
+    if (!this.alternatives.contains(alternative)) {
+      this.alternatives.add(alternative);
+    }
+
+    if (!alternative.alternatives.contains(this)) {
+      alternative.alternatives.add(this);
+    }
+  }
```
hohwille: Alternatives could even be immutable (using `Set.of(...)`) since all need to be provided in a single `add` method call.
Immutable design is always superior when applicable since it prevents complex state bugs at later runtime.

cli/src/main/java/com/devonfw/tools/ide/completion/CompletionEntry.java
```diff
@@ -48,8 +61,20 @@ public void addSynonym(String synonym) {
   public void complete(String arg, CompletionCandidateCollector collector, Property<?> property, Commandlet commandlet) {
 
     Set<String> alreadyProvided = collector.getAlreadyProvided();
-    if (alreadyProvided != null && (alreadyProvided.contains(this.candidate) || synonyms.stream().anyMatch(alreadyProvided::contains))) {
-      return;
+    if (alreadyProvided != null) {
```
hohwille: do we really have to handle that this is `null`? Actually we control the implementation of `CompletionCandidateCollector` and can guarantee that `getAlreadyProvided()` never returns `null`.
Don't get me wrong a `null` check for robustness is always good, but if you always have this extra if you also add an extra level of indentation to all nested ifs and statements making the code more complex and harder to read.
A method like getAlreadyProvided() returning a `Set` should never return `null` but then return `Set.of()`.
If we follow that style, we can simplify and remove this extra null check...

cli/src/main/java/com/devonfw/tools/ide/completion/AutoCompletionRegistry.java
hohwille: Another example:
```
    if (candidate.startsWith(arg)) {
      collector.add(candidate, "", property, commandlet);
    }

    for (String synonym : synonyms) {
      if (synonym.startsWith(arg)) {
        collector.add(synonym, "", property, commandlet);
      }
    }
```

Can then be simplified to:
```
    for (String alternative : this.alternatives) {
      if (alternative.startsWith(arg)) {
        collector.add(alternative, "", property, commandlet);
      }
    }
```

hohwille (changes requested):
@Hiepiscus thanks for your PR. Your implementation supports all needed features and covers them in a test. Great work 👍 
However, I challenge the implementation design and always try to think further how we can solve things more elegant. IMHO the existing `synonym` feature should be replaced by the new `alternatives` feature and the `dependencies` are IMHO overengineered (and usually I am the one who tends to overengineer and aim for a most generic design, but KISS also has its value 😄).
Please have a look at my review comments.

Hiepiscus:
@hohwille Thanks for the feedback! I simplified the design by removing synonyms and refactored the lookup to use a Map.
Regarding `List<List<String>>` my original thought was to handle AND groups, where a candidate should only be suggested if multiple prerequisites are met. However, I agree that there is currently no use case for this and it just adds unnecessary complexity.

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MavenCommandlet.java
```diff
@@ -75,7 +71,8 @@ protected void initAutoCompletionRegistry(AutoCompletionRegistry registry) {
     registry.add("-Dstyle.color=");
     registry.add("-Duser.dir=");
     registry.add("-Duser.home=");
+    registry.add("exec:java", "exec:exec");
+    registry.add("-Dexec.mainClass=").addDependency("-Dexec.mainClass=", "exec:java");
+    registry.add("-Dexec.args=").addDependency("-Dexec.args=", "exec:java", "exec:exec");
```
hohwille: I am not getting why an entry gets its own candidate as dependency.
Is that wrong or do I miss something?

```suggestion
    registry.add("-Dexec.mainClass=").addDependency("exec:java");
    registry.add("-Dexec.args=").addDependency("exec:java", "exec:exec");
```

hohwille (approved):
@Hiepiscus thanks for your rework. Now everything looks perfect except for one last thing I would like to clarify 👍 
Can you have a look if I am missing something or what we can do about that?

cli/src/main/java/com/devonfw/tools/ide/tool/mvn/MavenCommandlet.java
Hiepiscus: You're right, the candidate should not be its own dependency. I forgot to remove these.

Hiepiscus:
Successfully tested with 2026.09.002-09_22_07-SNAPSHOT on Windows 11

I followed the test instruction and verifed that auto completion works as expected.

`ide mvn -D` no -Dexec.mainClass or -Dexec.args is suggested.
`ide mvn exec:java` the output contains -Dexec.args and -Dexec.mainClass
`ide mvn exec:exec` the output contains -Dexec.args

PR #2395, opened 2026-09-01
#2381: node npm bash isolation poc

### LINKED ISSUE #352: Design for tools like node and python that hack their own installation

### LINKED ISSUE #2381: node + npm in pristine software repo with per-project isolation (Bash, launchers as functions)
### Feature idea

Related also to issue #352

**Goal**
Prove the concept with a minimal, Bash-only increment. Install node and npm as pristine, versioned installations in the shared software repository, and route each project's global npm packages into a per-project directory so projects no longer collide. Launchers are implemented as Bash functions (like the existing claude function in cli/src/main/package/functions), not as shim files.

**Scope**
- node + npm installed into the software repository under softwareRepositoryPath/<tool>/<version>/ (pristine layout), sharing one copy of a version across projects.
- Bash function per tool (node, npm) in functions that, inside a project, resolves the project's configured version and runs that binary from the repo; outside a project, falls back to the system binary (mirrors claude()).
- global npm packages installed to a per-project prefix (e.g. under the project dir), so npm i -g in one project does not write into the shared node install.
- works in Bash / POSIX.

**Acceptance criteria**
1. node and npm are installed into the pristine software repository, not ignored (isIgnoreSoftwareRepo no longer needed for the POC path).
2. two projects (A and B) on the same node version but different npm versions both run correctly from Bash, each resolving its own npm.
3. running npm i -g <package> in project A lands the package in project A's directory and does not modify the shared node install or  project B's packages (this is the #352 scenario).
4. node/npm resolve via the Bash function in functions. Typing the command outside any project still works (falls back to the system binary).

### Additional context

_No response_

krystynaShatkovska:
### Implemented changes:

Make node and npm pristine, independently versioned tools (Option A). Each now installs into the
    shared software repository as its own versioned folder (e.g. _ide/software/default/node/node/<ver>,
    _ide/software/default/npm/npm/<ver>) and is linked per-project — instead of node silently owning a
    bundled copy of npm.
   - SystemPath — new getToolPathsInResolutionOrder() hoists the npm tool's bin to the front of the
      resolution order (dedup by tool key, not path) and is used by both findBinary and the PATH string
      builder. This guarantees the configured npm/npx deterministically shadow node's bundled ones in
      every consumer.
   - Node — removed isIgnoreSoftwareRepo() == true (node is now a pristine versioned tool in the
      shared repo) and removed the npm config set prefix post-install hack (npm manages its own prefix
      now).
   - Npm — reparented to extend LocalToolCommandlet directly (a first-class standalone tool, not a
      child of node). setEnvironment pins npm_config_prefix to a per-project <IDE_HOME>/.npm-global and
      puts .npm-global/bin on PATH, so global npm packages are isolated per project (fixes #352). It
      resolves its own repo via _ide/urls/npm/npm/; npm-based packages (yarn/corepack/pnpm) still
      resolve via the npm repository.
  - Npm.findBuildDescriptor — restored (the reparenting dropped the package.json recognition it
      previously inherited from NodeBasedCommandlet), so ide build still dispatches npm projects.
  - Shell integration. Added node/npm wrappers to functions (bash) and functions.ps1 (PowerShell) so
    that inside a project node/npm re-route through ide (using the configured version) and outside a
    project fall back to the system tool. The PowerShell wrappers use the Get-Command … Application,
    ExternalScript + & $cmd.Source pattern (copied from the existing claude wrapper) to avoid the
    function recursing into itself.
  - Test fixtures & tests. Pristine npm fixtures (npm tool repo + urls + npx stub) added for the
    node/npm-based projects; yarn/corepack npm stubs rewritten to model the pristine .npm-global
    global-install behavior; NpmTest/CorepackTest/YarnTest/EnvironmentCommandletTest updated
    accordingly, plus a new NpmTest asserting npm_config_prefix points at the per-project .npm-global.

  ---

### Testing instructions

  Automated (primary):
cd cli
mvn -o clean test  ->expect green. The node/npm suite (NpmTest, NpmJsTest, NpmJsVersionTest, NpmJsVersionsTest, NpmJsDistTest, YarnTest, CorepackTest, CdkTest, NestTest, NgTest, TaskTest, BuildCommandletTest, EnvironmentCommandletTest, SystemPathTest) all pass (14 classes, 51 tests).
Note: the full suite is gated by CI (build-pr.yml, ubuntu-latest) as the authoritative check. A local full run on a small machine may be killed by resource limits without indicating a regression.

  Manual (verify the pristine behavior on a real machine):
  1. ide set node 18 then ide install node. Confirm it lands at _ide/software/default/node/node/18.<…>
     (shared repo), and node software shows it.
  2. ide set npm 9 then ide install npm. Confirm an independent _ide/software/default/npm/npm/9.<…>
     folder (npm no longer comes from node).
  3. Inside a project: run npm --version and npx. Both must resolve to the pristine npm, not the copy bundled with node. Do NOT rely on the versio  number alone, node bundles its own npm with a similar-looking version. Check the resolved path (where npm on Windows, which npm on Linux/Git-Bash) points at _ide/software/default/npm/npm/<ver>/bin/…, and that the version equals the npm you set. Also confirm npm_config_prefix points at <IDE_HOME>/.npm-global.
     - Regression guard: open _ide/software/default/npm/npm/<ver>/bin/npm.cmd (and .ps1, and bin/npm on Linux). The shim must now launch `node …\npm-cli.js` (flat layout), NOT `node_modules\npm\bin\npm-cli.js`. Old broken behavior: Windows = MODULE_NOT_FOUND (exit 1); Linux/Git-Bash = silently ran node's bundled npm.
  4. npm i -g <pkg>. Confirm the package lands under <IDE_HOME>/.npm-global/bin, not inside the node
     installation (per-project isolation, #352).
  5. In a package.json project with no yarn.lock, ide build still dispatches to npm.
  6. Outside a project (no IDE_HOME), node/npm in your shell fall through to the system tools (bash +
     PowerShell).

---

### Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
```diff
@@ -205,6 +207,29 @@ private void collectToolPath(Path softwarePath) {
...
+    }
+
+    for (Map.Entry<String, Path> entry : this.tool2pathMap.entrySet()) {
+      if (!TOOL_NPM.equals(entry.getKey())) {
+        orderedPaths.add(entry.getValue());
+      }
+    }
+
+    return orderedPaths;
+  }
```
hohwille, line 231: Why do we really need this? Or in other words: What breaks without this?
The class `SystemPath` is generic and fundamental - we should avoid adding knowledge about specific tools in this class as much as possible.

cli/src/main/java/com/devonfw/tools/ide/tool/node/Node.java
```diff
@@ -33,32 +27,9 @@ public Node(IdeContext context) {
...
-    // inside npm we have to guarantee that we will not trigger an installation (again) causing an infinity loop
-    // we would love to get this clean but node and npm are already flawed forcing us to do such hacks...
-    Npm npm = this.context.getCommandletManager().getCommandlet(Npm.class);
-    PackageManagerRequest packageManagerRequest = new PackageManagerRequest("config", this.tool).addArg("config").addArg("set").addArg("prefix")
-        .addArg(getToolPath().toString());
-    ProcessResult result = npm.runPackageManager(packageManagerRequest, true);
-    if (result.isSuccessful()) {
-      IdeLogLevel.SUCCESS.log(LOG, "Setting npm config prefix to: {} was successful", getToolPath());
-    }
-  }
```
hohwille, line 51: This was a kind of bugfix to make `node` work properly. Why was this removed? Is that now done in our shim?

cli/src/main/java/com/devonfw/tools/ide/tool/npm/Npm.java
```diff
@@ -66,7 +94,105 @@ public Path getOrCreateNpmConfigUserConfig() {
...
+    FileAccess fileAccess = this.context.getFileAccess();
+    boolean windows = this.context.getSystemInfo().isWindows();
+    repairShim(fileAccess, bin.resolve("npm.cmd"), cmdShim(NPM_CLI_JS));
+    repairShim(fileAccess, bin.resolve("npx.cmd"), cmdShim(NPX_CLI_JS));
+    repairShim(fileAccess, bin.resolve("npm.ps1"), psShim(NPM_CLI_JS));
+    repairShim(fileAccess, bin.resolve("npx.ps1"), psShim(NPX_CLI_JS));
+    if (!windows) {
+      repairShim(fileAccess, bin.resolve("npm"), posixShim(NPM_CLI_JS));
+      repairShim(fileAccess, bin.resolve("npx"), posixShim(NPX_CLI_JS));
+    }
```
hohwille, line 145: Better like this or am I missing something?
```suggestion
    if (windows) {
      repairShim(fileAccess, bin.resolve("npm.cmd"), cmdShim(NPM_CLI_JS));
      repairShim(fileAccess, bin.resolve("npx.cmd"), cmdShim(NPX_CLI_JS));
      repairShim(fileAccess, bin.resolve("npm.ps1"), psShim(NPM_CLI_JS));
      repairShim(fileAccess, bin.resolve("npx.ps1"), psShim(NPX_CLI_JS));
    } else {
      repairShim(fileAccess, bin.resolve("npm"), posixShim(NPM_CLI_JS));
      repairShim(fileAccess, bin.resolve("npx"), posixShim(NPX_CLI_JS));
    }
```

hohwille (commented):
@krystynaShatkovska thanks for your PR. Your approach looks sane and good to me. 👍 
Just from the review, I cannot tell how solid this will work or not, but here are my comments. Sorry for the questions - most likely we can resolve the comments when answered. Usually it is not a good style to ask questions via review comments but in this case I want to ensure we will not miss something.
I will additionally pull and test your changes but I do not have the time for that today.

cli/src/main/java/com/devonfw/tools/ide/common/SystemPath.java
krystynaShatkovska: Good point, I agree SystemPath should stay generic. It has to live here because the PATH user's shell actually gets is built by SystemPath the shell-export path has no per-tool hook to say "put me first." And node installs as a flat folder (no bin/) that already contains its own npm/npx/corepack, and the tool map is a HashMap, so without an order it's random which npm wins. I turned the single string into a named, documented PATH_PRECEDENCE_TOOLS list so it reads as a policy, not a special case, and added a test that npm beats node. Hope this will be suitable solution

cli/src/main/java/com/devonfw/tools/ide/tool/node/Node.java
krystynaShatkovska: It's removed on purpose, and it's not in the shim. That install-time npm config set prefix only existed because node bundled npm, and it pointed the global prefix at the shared node folder, which is exactly the leak #352 is about. Now npm is a standalone tool and sets its own prefix at runtime: Npm.setEnvironment sets npm_config_prefix to a per-project .npm-global. So it moved from install-time to runtime, and from the shared node dir to per-project. node itself isunaffected because node.exe doesn't read npm's prefix

cli/src/main/java/com/devonfw/tools/ide/tool/npm/Npm.java
krystynaShatkovska: You are right. The npm package (the .tgz we download on install) ships all the shims but each platform only runs its own, so I now only repair the ones for this platform. And I have applied your if/else.

PR #2398, opened 2026-09-01
#2366: support variables and expressions in repository properties

### LINKED ISSUE #2366: Support for variables and expressions in repository URL
### Feature idea

As an IDEasy maintainer, I would like to be able to configure the fork of the current user as repository.
Since the `git_url` is containing the name of the github user I would like to make use of variables and expressions in a git URL.
Then combining this feature together with #2168 and #989 / #2165 we can then get what we need:
IDEasy.properties:
```
path=IDEasy
git_url=https://github.com/@ask-variable("GIT_USER", "Please enter your github login:", "home", "$[USER]")/IDEasy.git
git_remote=upstream:https://github.com/devonfw/ide-settings.git
git_branch=main
eclipse=import
active=true
```

### Additional context

Feature #2168 is already available but PR #2282 needs to be merged before this story can be implemented.

quando632:
### Implemented changes:

* Every value of a repository `.properties` file is now resolved via the `EnvironmentVariables` hierarchy, so [variables](https://github.com/devonfw/IDEasy/blob/main/documentation/variables.adoc) (`$[VAR]`) and expressions (`@function(...)`) can be used in `git_url`, `git_remote`, `path` and every other property.
* Resolution is done centrally in `RepositoryProperties.doGetProperty(..)` and happens **before** any validation, so it also works for validated properties such as `path` (whose `PATH_PATTERN` would otherwise reject the `[` and `]` of an unresolved `$[VAR]`).
* Legacy support is disabled on purpose (`legacySupport=false`), so the `${...}` syntax of e.g. maven properties in a `build_cmd` is passed through untouched.
* The `active` flag is now evaluated **before** the remaining properties are read, so an expression like `@ask-variable` never asks the user for a repository that is skipped anyway. For this, `RepositoryConfig.loadProperties(RepositoryProperties)` was added as an overload; the existing `loadProperties(Path, IdeContext)` is kept and delegates to it.
* Documented the new capability in `documentation/repository.adoc` and added the CHANGELOG entry.
* Resolved values are cached per `RepositoryProperties` instance. Resolving can ask the user, so a property read twice would ask twice - the `active` flag is read by `RepositoryCommandlet` and again by `RepositoryConfig`.
* When iterating all repositories, a properties file whose expression is malformed is logged and skipped instead of aborting the setup of every other repository. `CliAbortException` is rethrown so an abort by the user is never swallowed; an explicitly requested single repository still reports the failure.

This is the last building block for the use-case described in #2366 - configuring the fork of the current user as repository:

```properties
path=IDEasy
git_url=https://github.com/@ask-variable("GIT_USER", "Please enter your github login:", "home", "$[USERNAME]")/IDEasy.git
git_remote=upstream:https://github.com/devonfw/IDEasy.git
git_branch=main
import=eclipse
active=true
```

> [!NOTE]
> #989 / PR #2282 is merged, `main` has been merged into this branch, and the `@ask-variable` example above was verified end-to-end (see testing instructions). This PR needs no expression specific code itself: #2282 registers the `ExpressionParser` inside `AbstractEnvironmentVariables.resolveRecursive`, i.e. *below* `EnvironmentVariables.resolve(..)`, which is what this PR calls.

> [!NOTE]
> **Known limitation, deliberately not addressed here.** All properties of an active repository are resolved in the first loop of `doRun`, while the "repository already exists / was already setup" checks happen later in `doImportRepository`. A `@ask-secret(..)` in a `git_url` is never persisted by design, so it would be asked on every run even for a repository that is then skipped. `@ask-variable(..)` - the use-case of this issue - persists its answer and is therefore asked only once. Fixing this would require resolving the URL lazily and thus restructuring `RepositoryConfig`, which is out of scope for this story.

### Testing instructions

1. Create or edit a repository properties file in your settings, e.g. `settings/repositories/test.properties`:
   ```properties
   path=repos/$[GIT_USER]
   git_url=https://github.com/$[GIT_USER]/IDEasy.git
   git_remote=upstream:https://github.com/devonfw/IDEasy.git
   active=true
   ```
2. Define the variable, e.g. in `$IDE_HOME/conf/ide.properties`: `GIT_USER=«your-github-login»`
3. Run `ide repository setup test` (or `ide update`). The repository is cloned from `https://github.com/«your-github-login»/IDEasy.git` into `workspaces/main/repos/«your-github-login»`, and `git remote -v` inside the clone shows the additional `upstream` remote.
4. Now remove `GIT_USER` from `ide.properties` and run again: a warning `Undefined variable $[GIT_USER] in '«...»/test.properties'` is logged and the placeholder is kept, so the subsequent `git clone` fails with the unresolved URL visible in the error. An undefined variable in a repository URL is a configuration error and is reported as such rather than being silently swallowed.
5. Set `active=false` and run `ide repository` (without `--force-repositories`): only `Skipping repository test because it is not active, ...` is logged - no resolution and thus no question is triggered for a repository that is skipped.
6. Verify that the legacy `${...}` syntax is *not* resolved - IDEasy must pass it through untouched, since it is common in build commands (maven properties). Use a variable that **is defined** (e.g. the `GIT_USER` from step 2); an undefined one would stay unresolved either way and would prove nothing. Put both syntaxes side by side:
   ```properties
   build_cmd=mvn -Downer=${GIT_USER} -Duser=$[GIT_USER] clean install
   ```
   Run the setup and look at the step that is logged *before* the build starts:
   ```
   Start: Build repository via: mvn -Downer=${GIT_USER} -Duser=quando632 clean install
   ```
   Only the `$[..]` form is replaced, `${..}` stays literal. Note this is about the value IDEasy passes on - the `.properties` file itself is never modified by IDEasy.
   If you would rather not trigger a real maven build, use `path=repos/${GIT_USER}` instead: the log then shows `Invalid path repos/${GIT_USER} from property path of «...»` (the path pattern rejects `{`/`}`) and the repository falls back to its id as directory name. Had the legacy syntax been resolved, it would have cloned into `repos/quando632` without any warning.
   This behaviour is also pinned by the unit test `RepositoryPropertiesTest.testLegacyCurlySyntaxIsNotResolved`.
7. Replace the `git_url` with the `@ask-variable(..)` example above and remove `GIT_USER` everywhere. On the first run you are asked `Please enter your github login: [«your-os-user»]` (the default comes from `$[USERNAME]`), the answer is written to `~/.ide/ide.properties`, and the fork is cloned. On a second run you are not asked again.

---

### Checklist for this PR

- [x] When running `mvn clean test` locally all tests pass and build is successful (`mvn verify -pl '!documentation'` → BUILD SUCCESS, `ide-cli`: 959 tests, 0 failures, 0 errors)
- [x] You followed all [coding conventions](https://github.com/devonfw/IDEasy/blob/main/documentation/coding-conventions.adoc) (`mvn spotless:apply` and `mvn checkstyle:check` are clean for all changed files)

### DIFF
--- a/cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java
@@ -90,17 +90,20 @@ public class RepositoryCommandlet extends Commandlet {
 
   private RepositoryConfig prepareActiveRepository(Path repositoryFile, boolean forceMode) {
 
-    RepositoryConfig config = RepositoryConfig.loadProperties(repositoryFile, this.context);
-    if (config == null) {
-      return null;
-    }
-    if (!config.active()) {
+    // the active flag is evaluated before the remaining properties are read: reading a property resolves variables and expressions in its value, so an
+    // expression like @ask-variable would otherwise ask the user for a repository that is skipped anyway
+    RepositoryProperties properties = new RepositoryProperties(repositoryFile, this.context);
+    if (!properties.isActive()) {
       if (forceMode) {
-        LOG.info("Setup of repository {} is forced, hence proceeding ...", config.id());
+        LOG.info("Setup of repository {} is forced, hence proceeding ...", properties.getId());
       } else {
-        LOG.info("Skipping repository {} because it is not active, use --force-repositories to setup all repositories ...", config.id());
+        LOG.info("Skipping repository {} because it is not active, use --force-repositories to setup all repositories ...", properties.getId());
         return null;
       }
+    }
+    RepositoryConfig config = RepositoryConfig.loadProperties(properties);
+    if (config == null) {
+      return null;
     }
     // prepare workspace creation for correct resolution of *
     List<String> workspaces = config.workspaces();
--- a/cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java
@@ -11,6 +11,8 @@ import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.devonfw.tools.ide.cli.CliAbortException;
+import com.devonfw.tools.ide.cli.CliException;
 import com.devonfw.tools.ide.commandlet.Commandlet;
 import com.devonfw.tools.ide.context.IdeContext;
 import com.devonfw.tools.ide.git.GitContext;
@@ -74,7 +76,7 @@ public class RepositoryCommandlet extends Commandlet {
       boolean forceMode = this.context.isForceMode() || this.context.isForceRepositories();
       Map<Path, RepositoryConfig> repositoryConfigMap = new HashMap<>(propertiesFiles.size());
       for (Path propertiesFile : propertiesFiles) {
-        RepositoryConfig config = prepareActiveRepository(propertiesFile, forceMode);
+        RepositoryConfig config = prepareActiveRepositoryIgnoringInvalid(propertiesFile, forceMode);
         if (config != null) {
           repositoryConfigMap.put(propertiesFile, config);
         }
@@ -85,6 +87,27 @@ public class RepositoryCommandlet extends Commandlet {
           importRepository(config);
         }
       }
+    }
+  }
+
+  /**
+   * Like {@link #prepareActiveRepository(Path, boolean)} but ignores a single invalid repository instead of aborting the setup of all others. Resolving a
+   * property value may fail for a malformed expression, and one broken file in the repositories folder must not prevent every other repository from being
+   * set up. Only used when iterating all repositories - if the user explicitly requested a single repository, the failure is reported instead.
+   *
+   * @param repositoryFile the {@link Path} to the repository properties file.
+   * @param forceMode - {@code true} to setup the repository even if it is not active, {@code false} otherwise.
+   * @return the {@link RepositoryConfig} or {@code null} if the repository shall be skipped.
+   */
+  private RepositoryConfig prepareActiveRepositoryIgnoringInvalid(Path repositoryFile, boolean forceMode) {
+
+    try {
+      return prepareActiveRepository(repositoryFile, forceMode);
+    } catch (CliAbortException e) {
+      throw e; // the user deliberately aborted - never swallow this
+    } catch (CliException e) {
+      LOG.error("Ignoring repository {} because its configuration is invalid: {}", repositoryFile, e.getMessage());
+      return null;
     }
   }
 
--- a/cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
@@ -47,6 +47,8 @@ final class RepositoryProperties {
 
   private final Properties properties;
 
+  private final IdeContext context;
+
   private boolean invalid;
 
   /**
@@ -56,7 +58,7 @@ final class RepositoryProperties {
    * @param context the {@link IdeContext}.
    */
   public RepositoryProperties(Path file, IdeContext context) {
-    this(file, context.getFileAccess().readProperties(file));
+    this(file, context.getFileAccess().readProperties(file), context);
   }
 
   /**
@@ -64,9 +66,19 @@ final class RepositoryProperties {
    * @param properties the actual {@link Properties} loaded from the file.
    */
   RepositoryProperties(Path file, Properties properties) {
+    this(file, properties, null);
+  }
+
+  /**
+   * @param file the {@link Path} to the properties file.
+   * @param properties the actual {@link Properties} loaded from the file.
+   * @param context the {@link IdeContext} used to resolve variables and expressions or {@code null} to disable resolution.
+   */
+  RepositoryProperties(Path file, Properties properties, IdeContext context) {
     super();
     this.file = file;
     this.properties = properties;
+    this.context = context;
   }
 
   /**
@@ -120,6 +132,11 @@ final class RepositoryProperties {
 
   private String doGetProperty(String name, String legacyName) {
 
+    return resolve(doGetRawProperty(name, legacyName));
+  }
+
+  private String doGetRawProperty(String name, String legacyName) {
+
     String value = this.properties.getProperty(name);
     if (value != null) {
       return value;
@@ -140,6 +157,23 @@ final class RepositoryProperties {
       value = getLegacyProperty(legacyName, name);
     }
     return value;
+  }
+
+  /**
+   * Resolves {@link com.devonfw.tools.ide.variable.VariableSyntax#SQUARE variables} and expressions in the given property value. This is done centrally here so
+   * that it applies to every property and happens before any validation (e.g. of a {@link #getPath() path}) that would otherwise reject the syntax of an
+   * unresolved value.
+   *
+   * @param value the raw property value or {@code null}.
+   * @return the given value with variables and expressions resolved.
+   */
+  private String resolve(String value) {
+
+    if ((value == null) || (this.context == null)) {
+      return value;
+    }
+    // legacy support is disabled on purpose: the ${...} syntax is common in build commands (e.g. maven properties in build_cmd) and must not be touched
+    return this.context.getVariables().resolve(value, this.file, false);
   }
 
   private static boolean isEmpty(String value) {
--- a/cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
+++ b/cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
@@ -4,8 +4,10 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.regex.Pattern;
@@ -49,6 +51,9 @@ final class RepositoryProperties {
 
   private final IdeContext context;
 
+  /** Cache of already resolved property values so that resolving never happens twice - see {@link #doGetProperty(String, String)}. */
+  private final Map<String, String> resolvedProperties;
+
   private boolean invalid;
 
   /**
@@ -79,6 +84,7 @@ final class RepositoryProperties {
     this.file = file;
     this.properties = properties;
     this.context = context;
+    this.resolvedProperties = new HashMap<>();
   }
 
   /**
@@ -132,7 +138,14 @@ final class RepositoryProperties {
 
   private String doGetProperty(String name, String legacyName) {
 
-    return resolve(doGetRawProperty(name, legacyName));
+    // the result is cached since resolving may have side-effects: an expression like @ask-variable asks the user, so reading the same property twice
+    // (e.g. the active flag, that is read by RepositoryCommandlet and again by RepositoryConfig) must not ask twice.
+    if (this.resolvedProperties.containsKey(name)) {
+      return this.resolvedProperties.get(name);
+    }
+    String value = resolve(doGetRawProperty(name, legacyName));
+    this.resolvedProperties.put(name, value);
+    return value;
   }
 
   private String doGetRawProperty(String name, String legacyName) {
--- a/cli/src/test/java/com/devonfw/tools/ide/commandlet/RepositoryCommandletTest.java
+++ b/cli/src/test/java/com/devonfw/tools/ide/commandlet/RepositoryCommandletTest.java
@@ -87,6 +87,23 @@ class RepositoryCommandletTest extends AbstractIdeContextTest {
   }
 
   @Test
+  void testInactiveRepositoryIsNotResolved() {
+
+    // arrange
+    IdeTestContext context = newContext(IdeContext.FOLDER_REPOSITORY);
+    Properties properties = createDefaultProperties();
+    properties.setProperty("git_url", "https://github.com/$[UNDEFINED_VARIABLE]/" + TEST_GIT_REPO + ".git");
+    RepositoryCommandlet rc = context.getCommandletManager().getCommandlet(RepositoryCommandlet.class);
+    saveProperties(context, properties);
+    // act
+    rc.run();
+    // assert
+    assertThat(context).logAtInfo().hasMessage("Skipping repository test because it is not active, use --force-repositories to setup all repositories ...");
+    // an inactive repository must not be resolved at all so that an expression like @ask-variable never asks the user for a repository that is skipped
+    assertThat(context).logAtWarning().hasNoMessageContaining("Undefined variable $[UNDEFINED_VARIABLE]");
+  }
+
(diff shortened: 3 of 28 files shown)


cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java
```diff
@@ -88,20 +90,44 @@ protected void doRun() {
...
+   */
+  private RepositoryConfig prepareActiveRepositoryIgnoringInvalid(Path repositoryFile, boolean forceMode) {
 
-    RepositoryConfig config = RepositoryConfig.loadProperties(repositoryFile, this.context);
-    if (config == null) {
+    try {
+      return prepareActiveRepository(repositoryFile, forceMode);
+    } catch (CliAbortException e) {
+      throw e; // the user deliberately aborted - never swallow this
+    } catch (CliException e) {
```
JoelAdbu, line 108: We should also consider the special case where resolving a circular variable reference throws an IllegalStateException. Otherwise, one invalid repository could still terminate the entire batch setup.

```suggestion
} catch (CliAbortException | IllegalStateException e) {
```
This would fix the repository batch setup case. However, the root cause seems to be in AbstractEnvironmentVariables.resolveRecursive(): a circular variable reference is a user configuration error and should therefore probably result in a CliException, not an IllegalStateException. Catching IllegalStateException here could potentially hide unrelated programming errors from the same code path.

cli/src/test/java/com/devonfw/tools/ide/commandlet/RepositoryCommandletTest.java
```diff
@@ -86,6 +86,63 @@ void testSetupAllRepositoriesInactive() {
...
+    good.setProperty("active", "true");
+    saveProperties(context, good);
+    RepositoryCommandlet rc = context.getCommandletManager().getCommandlet(RepositoryCommandlet.class);
+    // act
+    rc.run();
+    // assert
+    assertThat(context.getIdeHome().resolve(IdeContext.FOLDER_WORKSPACES).resolve(TEST_WORKSPACE).resolve(TEST_REPO)).isDirectory();
+    assertThat(context).logAtError().hasMessageContaining("Invalid template expression");
+  }
+
```
JoelAdbu, line 146: We could also add a test to cover this edge case.

```suggestion
  @Test
  void testCircularVariableInOneRepositoryDoesNotPreventOtherRepositories() {
    IdeTestContext context = newContext(IdeContext.FOLDER_REPOSITORY);

    context.getVariables().getByType(EnvironmentVariablesType.CONF).set("REPO_A", "$[REPO_B]");
    context.getVariables().getByType(EnvironmentVariablesType.CONF).set("REPO_B", "$[REPO_A]");

    Properties broken = createDefaultProperties();
    broken.setProperty("active", "true");
    broken.setProperty("path", "broken-repo");
    broken.setProperty("git_url", "https://github.com/@ask-variable('REPO_A')/broken.git");
    saveProperties(context, broken, "broken.properties");

    Properties good = createDefaultProperties();
    good.setProperty("active", "true");
    saveProperties(context, good);

    RepositoryCommandlet rc = context.getCommandletManager().getCommandlet(RepositoryCommandlet.class);

    rc.run();

    Path goodRepository = context.getIdeHome()
        .resolve(IdeContext.FOLDER_WORKSPACES)
        .resolve(TEST_WORKSPACE)
        .resolve(TEST_REPO);

    assertThat(goodRepository).isDirectory();

  }

```

JoelAdbu (commented):
Well done! I tested it locally and everything works fine. Please consider the minor suggestions below to ensure all edge cases are covered.

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java
quando632: I fixed it in the last commit.

Reproduced first: with `REPO_A=$[REPO_B]` / `REPO_B=$[REPO_A]` the run aborts with `IllegalStateException: Reached maximum recursion resolving $[REPO_A] ...` and the second, valid repository is not set up.

`resolveRecursive` now throws `CliException` for the maximum recursion. With that, `prepareActiveRepositoryIgnoringInvalid` needs no change at all the existing `catch (CliException)` covers it, and we avoid catching `IllegalStateException`, which is exactly the masking risk you described.

Checked that nothing relies on the old type: the only other `IllegalStateException` assertions in tests are the ambiguous-merge one in `XmlMergerTest:88` and a helper in `DirectoryMergerTest:155`, both different throw sites. Side effect is that a circular reference now surfaces as a clean CLI error in the configurator/merger too, instead of a stacktrace.

One note on the suggestion block itself: line 108 is the `catch (CliException e)`, so applying it literally gives `error: exception CliAbortException has already been caught` (the rethrow above already catches it) and would drop the `CliException` handling.

cli/src/test/java/com/devonfw/tools/ide/commandlet/RepositoryCommandletTest.java
quando632: Added as `testCircularVariableDoesNotPreventOtherRepositories`, close to your version.
Two changes: dropped the `@ask-variable('REPO_A')` wrapper, since a plain `$[REPO_A]` in the `git_url` triggers the same recursion more directly, and added an assertion on the logged error so the test also pins that the repository is reported rather than silently skipped.

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java
JoelAdbu: Thanks for fixing this. Regarding your note, you're correct, that was a typo on my side. Anyway, the fix in AbstractEnvironmentVariables was the right way to address it, so thank you.

hohwille: Just for the record. Switching the recursion error to `CliException` now prevents that a stracktrace gets printed by default (unless in debug mode). For bug-reports the stacktrace might still be helpful, but just a tiny detail not preventing merge.

hohwille, line 108: wouldn't our step infrastructure be much better here?
This seems to be the exact pattern we typically solve with steps what also gives better user-feedback.

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
```diff
@@ -47,6 +49,11 @@ final class RepositoryProperties {
 
   private final Properties properties;
 
+  private final IdeContext context;
```
hohwille, line 52: Nothing wrong but challenging the design:
It was already questionable before that `IdeContext` was passed to the constructor.
But adding it to this simple data object seems not a really good design.
Typically properties, configs, etc. should focus on their data and remain rather stupid objects not being connected to context and other state.
You could also pass the context as argument to get properties methods and resolve.
Or you could even keep the entire resolving out here leaving this as a stupid data object and add the resolving part to the method "transforming" these properties object to RepositoryConfig.

documentation/repository.adoc
```diff
@@ -54,3 +54,26 @@ If omitted no build is triggered.
 |`link`|CSV of `«link»[=«path»]`|(optional) Comma separated list of symbolic links to link the repository into the workspace top-level folder. For a real-world use-case of this feature see link:ai.adoc[AI usage].
 |`active`|`true`|(optional) If set to `false` the repository is skipped during the link:setup.adoc[setup].
 |===
+
+== Variables and expressions
+
+Every value in a repository `.properties` file may contain link:variables.adoc[variables] using the `$[«VARIABLE»]` syntax as well as link:configurator.adoc[expressions] using the `@«function»(...)` syntax.
```
hohwille, line 60:
```suggestion
Every value in a repository `.properties` file may contain link:variables.adoc[variables] using the `$[«VARIABLE»]` syntax as well as link:configurator.adoc#functions[expression functions] using the `@«function»(...)` syntax.
```

documentation/repository.adoc
```diff
@@ -54,3 +54,26 @@ If omitted no build is triggered.
...
+path=IDEasy
+git_url=https://github.com/@ask-variable("GIT_USER", "Please enter your github login:", "home", "$[USERNAME]")/IDEasy.git
+git_remote=upstream:https://github.com/devonfw/IDEasy.git
+git_branch=main
+import=eclipse
+active=true
+```
+
+On the first setup the user is asked for the github login, the answer is stored as `GIT_USER` in `~/.ide/ide.properties` (`home`) and reused on every following run.
+The default `$[USERNAME]` is the login name of the current user on Windows - on Linux and macOS use `$[USER]` instead.
```
hohwille, line 75: If you already found out this flaw that on Windows it is `$USER` and on Linux/mac it is `$USERNAME` then this approach is flawed since it is not working cross-platform.
IDEasy aims to provide a cross-platform solution.
Therefore we should not document a flaw but instead create an issue to solve the flaw.
I would suggest that we make USER variable also available on Windows and use it here instead of `$[USERNAME]`.

documentation/repository.adoc
```diff
@@ -54,3 +54,26 @@ If omitted no build is triggered.
...
+
+NOTE: A repository with `active=false` is skipped before its properties are read, so a question from `@ask-variable` is never asked for a repository that is not set up.
+
+NOTE: The legacy `${«VARIABLE»}` syntax is intentionally *not* resolved here so that e.g. maven properties in a `build_cmd` are passed through untouched.
```
hohwille, line 79: Please avoid too much documentation duplications in different places since this makes it hard to maintain.
Also this is not exactly true since there is a feature to enable this syntax for backward compatibility.

```suggestion
```

hohwille (approved):
@quando632 thanks for your PR. You properly implemented and tested the story well. 👍 
I left some comments for improvement. Please have a look.

hohwille (approved):
~~@quando632 thanks for your PR. You properly implemented and tested the story well. 👍 
I left some comments for improvement. Please have a look.~~

Review was submitted twice due to a GitHub error.

documentation/repository.adoc
hohwille: Acutally it is so trivial to fix this problem that it is not worth creating a new issue for it.
It is less effort if you just add `USER` to `IdeVariables` with fallback to USERNAME to make this work and fix the problem.

documentation/repository.adoc
hohwille: OK, got it. In your case, you are explicitly passing `false` so it is always disabled.
I understand your point but isn't crisp documentation more valuable?
If you say variables are supported via `$[...]` then why explicitly saying that `${...}` is not supported?
We already have long documentation and users tend to have little time to read all this.
IMHO we can still remove this to avoid information overload. There are so many things you could write that we are not doing or not supporting so lets better document what we support.

quando632: I Agreed. Documenting what we support rather than what we do not is the better rule

documentation/repository.adoc
quando632: Applied. The anchor works, `=== Functions` is at `configurator.adoc:71`.

documentation/repository.adoc
quando632: Done. `USER` is in `IdeVariables` now with `USERNAME` as fallback, the example uses `$[USER]` and I added a row to `variables.adoc`.

Small thing on the direction: it is the other way round. Windows sets `USERNAME`, Linux and macOS set `USER`.

There was one catch. `EnvironmentVariablesSystem.getFlat(..)` blocks every variable registered in `IdeVariables.VARIABLES`, only `PATH`, `IDE_ROOT` and `HOME` are excluded. So just registering `USER` hid the real `USER` env variable on Linux and macOS, and the Windows fallback would have won everywhere. The test showed it immediately (`expected: "unix-login" but was: "windows-login"`). I added `USER` to that exception list, since it is an OS variable like `HOME` and `PATH`. Tell me if you prefer keeping the list at three entries and reading the env directly in the fallback instead.

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java
quando632: I checked this. Steps are clearly the pattern here. But there is a catch in this spot.

`Step.run(Runnable, boolean)` catches `RuntimeException | Error` and with `rethrow=false` it swallows everything, including `CliAbortException`. My rethrow is currently the only place in the production code that handles `CliAbortException` at all. With a plain step, a user who answers "no" to an `@ask-variable` question would get asked again for every following repository, and the `ProcessResult.ABORT` exit code would be lost.

Options I see:

1. keep the try/catch
2. use a step but keep the rethrow around it
3. make `Step.run` never swallow `CliAbortException`

I think 3 is the correct fix, but it touches shared code, so probably a separate issue. For now I would keep 1. What do you prefer?

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
quando632: Fair point. The class uses the context for two things only: reading the file in the public constructor and `getVariables().resolve(..)`.

Moving the resolving into the transformation to `RepositoryConfig` does not work though. `RepositoryCommandlet` reads `isActive()` before the other properties on purpose, so an `@ask-variable` in a `git_url` never asks about a repository that gets skipped. If resolving happened only in the transformation, `active` would be read unresolved. The value cache is there for the same reason, resolving has side effects, so reading a property twice must not ask twice.

What about passing a `UnaryOperator<String>` resolver instead of the `IdeContext`? Then the data object only holds a function, no context and no state, and the caching and the ordering stay as they are. The test constructor would just pass `UnaryOperator.identity()`.

Does that work for you?

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java
quando632: Small correction: in this path there is no stacktrace at all, also not with `--debug`, because the handler only logs `e.getMessage()`. I can pass the exception to the log call if you want it for bug reports.

hohwille (approved):
@quando632 thanks for the update. Now perfect and ready for merge. 👍

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryProperties.java
hohwille: It is already working as-is. It is just a small design flaw that is "not so nice".
I would suggest that we do not block this PR any further since everything is functionally correct.
Feel invited to create a issue or simply a PR for later cleanup.
We could even consider not supporting active property to use variables.
Also it would be possible to change the using code to do the transformation providing the context and then only working with the resulting config instead of using the properties object at all that is IMHO only an intermediate thing with the only purpose to encapsulate the parsing of the properties file.

cli/src/main/java/com/devonfw/tools/ide/git/repository/RepositoryCommandlet.java
hohwille: When we discussed the ability that a CliException can trigger rethrow automatically via the new boolean method, I also suggested that CliAbortException should do that just like CliFatalException since abort is thrown if the user decided to stop further processing.
Please also consider this for a cleanup issue / PR.

quando632:
Retest on SNAPSHOT `2026.09.002-09_16_07-SNAPSHOT` on Windows 11 passed

PR #2429, opened 2026-09-08
#2409: cleanup of expression functions

### LINKED ISSUE #2409: Cleanup of expression functions
### Task summary

This is a follow up of #989 from PR #2282:
The following aspects shall be cleaned up and addressed from the review (see original review comments in PR #2282):
R1: revisit/remove new IdeContext methods: `addSecretVariable` and `addSecretValue`: 
> Great that you care about security of credentials. I appreciate that very much.
However, I see some problems here:
> 1. If the function was not invoked this method gets never called and hence we have no clue what variables are secrets.
> 2. If we really want to protected secrets, we IMHO need to store them encrypted (see how maven is doing that or [spring-boot with jasypt spring-boot starter](https://github.com/devonfw/devon4j/blob/master/documentation/spring/guide-spring-configuration.asciidoc#password-encryption) is solving it). This is a complex thing to implement and should deserve its own story if we ever need to support it. The other alternative would be not to persist secrets what would be quite simple. This works fine for templates that are instantiated only once (e.g. in `settings/templates` or in `workspace/setup`). For the moment, I would go for KISS and think about this way.
> 3. Sometimes it is easier and more pragmatic to decide for a convention than trying to implement a bullet proof solution: We can use the convention that secret variables have to follow specific naming rules e.g. start or end with one of `SECRET`, `PASSWORD`, or `API_KEY`. Then we can avoid this complex runtime state of secret variable names registered in `IdeContext`. Also you could even enforce that the function `ask-secret` only accepts variables following this naming pattern. Then we could also detect secret values already when we create our environment and reading the variables from the convention.

R2: Change from individual argument handling to generic approach: Instead of checking argument count and checking + parsing every argument from a `List` we should do something like we do with Commandlets and Properties.
> All correct, but maybe we can approve (in a cleanup story after merge by someone else):
> 1. Simple idea: Create subclasses of `ExpressionFunction` like `UnaryExpressionFunction`, `BinaryExpressionFunction`, `TenerayExpressionFunction` that already implement `apply` verify the maximum number of arguments and then call a dedicated `apply` method variant with the proper number of arguments.
> 2. Advanced idea: In `ExpressionFunctionManager` we only register factories by function name and the factory then creates the `ExpressionFunction` that then becomes stateful and has a list of `Property` arguments just like a `Commandlet`. Then we can even define defaults and keep the parsing from String to dedicated enums and types out of the concrete function implementation.

R3: Remove individual documentation of every function for maintainability. Instead link to JavaDoc via docjar (that also allows to find documentation state for an older release):
> This is great but I fear that this way it will be hard to keep the documentation in sync with the code.
Maybe we can use docjar and just link to the latest JavaDoc of the functions package.
If we later follow my suggestions to express function arguments more structural with generic parsing support rather than manual handling in apply method implementation the JavaDoc itself would be fully sufficient to lookup what is needed. Otherwise we can also just put a link to the package in the GitHub source-code.

R4: Improve test (ExpressionParserTest) - avoid pointless mock but test the real thing since there is no external influence to mock:
> I understand that you want to create a plain unit-test here, but maybe it would make more sense to test this without mocking by testing it via EnvironmentVariablesResolved so you do not need this mock?

### Additional context

_No response_

Ali-Shariati-Najafabadi:
### Implemented changes:

* Replace the `addSecretVariable`/`addSecretValue` runtime registration with a naming convention: a variable is treated as secret if its name ends with `SECRET`, `PASSWORD`, or `API_KEY` (`IdeContext.isSecretVariableName`). `@ask-secret` now rejects a non-compliant variable name.
* Introduce `UnaryExpressionFunction`, `BinaryExpressionFunction`, and `QuaternaryExpressionFunction` so `PathFunction`, `IfOsFunction`, and `AskFunction` no longer check the argument count and index into the argument list themselves.
* Replace the per-function documentation in `configurator.adoc` with a link to the function package so the docs cannot drift out of sync with the JavaDoc.

---

### Testing instructions

1. Run `mvn -Dtest=ExpressionParserTest,EnvironmentVariablesTest,DirectoryMergerExpressionTest test` in the `cli` module - all tests pass, including the new `testAskSecretRequiresCompliantVariableName`.
2. Run `mvn clean test` in the `cli` module - no new failures compared to `main`.

---

### Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/environment/EnvironmentVariablesMap.java
```diff
@@ -38,9 +38,11 @@ public String getFlat(String name) {
...
       LOG.trace("{}: Variable {} is undefined.", getSource(), name);
     } else {
-      // register the value before it is logged so that a secret variable read from an existing ide.properties is
-      // masked as well, not only a value that was just entered by the user
-      this.context.addSecretValue(name, value);
+      // mask the value before it is logged if its variable name follows the naming convention for secrets, so a
+      // secret variable read from an existing ide.properties is masked as well, not only a value just entered by the user
+      if (IdeContext.isSecretVariableName(name)) {
+        this.context.addSecret(value);
+      }
```
quando632, line 45: Masking moves from opt-in registration to a name convention, and since `EnvironmentVariablesSystem` extends this class, plain OS variables are included. With `SECRET_MIN_LENGTH=3` and the substring `replace` in `AbstractIdeContext.formatArgument`, a short value in any `*_PASSWORD` variable is blanked out wherever it appears, so `dev` would turn `C:\projects\devonfw` into `C:\projects\********onfw`. Raising `SECRET_MIN_LENGTH` (`AbstractIdeContext:187`), or matching whole tokens instead of any substring, would keep the convention without that side effect.

documentation/configurator.adoc
```diff
@@ -70,43 +70,15 @@ Functions may also be nested, so an argument can itself contain a call of anothe
...
-----
+Each function is a small class implementing `ExpressionFunction` and registered by name in `ExpressionFunctionManager`.
+For the up-to-date list of available functions and their arguments, defaults and examples, see the JavaDoc of the
+link:https://github.com/devonfw/IDEasy/tree/main/cli/src/main/java/com/devonfw/tools/ide/expression/function[function package] instead of a second copy of
+that information here, so the documentation cannot drift out of sync with the code:
+
+* `@path` normalizes a path (e.g. to convert `$[IDE_HOME]/software/node/node.exe` to a form usable on the current operating system).
+* `@ask-variable` and `@ask-secret` ask the user for the value of a variable (in plain text, respectively with masked input) unless it is already defined.
+* `@if-windows`, `@if-mac`, `@if-linux` and `@if-unix` insert their single argument only if the current operating system matches and otherwise resolve to the
+empty string, e.g. to select an OS specific value inline: `$[IDE_HOME]/software/mvn/bin/mvn@if-windows('.bat')@if-unix('.sh')`.
```
quando632, line 81: The short list no longer says what the arguments are, and the removed section was the only place that documented them: line 88 still refers to "the chosen configuration location" and the example on line 103 uses `@path('...', native)`, but neither the locations nor the modes are described anywhere now. One line per function keeps the page self-contained without reintroducing a full copy.

```suggestion
* `@path` normalizes a path (e.g. to convert `$[IDE_HOME]/software/node/node.exe` to a form usable on the current operating system).
The optional 2nd argument is `unix` (the default, slashes) or `native` (backslashes on Windows).
* `@ask-variable` and `@ask-secret` ask the user for the value of a variable (in plain text, respectively with masked input) unless it is already defined.
The optional further arguments are the question, the configuration location to persist to (`settings`, `workspace`, `conf` or `home`, default `conf`, `@ask-variable` only) and a default value (`null` for none).
* `@if-windows`, `@if-mac`, `@if-linux` and `@if-unix` insert their single argument only if the current operating system matches and otherwise resolve to the
empty string, e.g. to select an OS specific value inline: `$[IDE_HOME]/software/mvn/bin/mvn@if-windows('.bat')@if-unix('.sh')`.
```

documentation/configurator.adoc
```diff
@@ -70,43 +70,15 @@ Functions may also be nested, so an argument can itself contain a call of anothe
...
-* `@if-windows`, `@if-mac`, `@if-linux` and `@if-unix` insert their single argument only if the current operating system matches and otherwise resolve to the empty string.
-This allows to select an OS specific value inline.
-+
-[source]
-----
-$[IDE_HOME]/software/mvn/bin/mvn@if-windows('.bat')@if-unix('.sh')
-----
+Each function is a small class implementing `ExpressionFunction` and registered by name in `ExpressionFunctionManager`.
+For the up-to-date list of available functions and their arguments, defaults and examples, see the JavaDoc of the
+link:https://github.com/devonfw/IDEasy/tree/main/cli/src/main/java/com/devonfw/tools/ide/expression/function[function package] instead of a second copy of
```
quando632, line 75: nit: the target is labelled "JavaDoc" but the link points at the sources on GitHub, pinned to branch `main`.

```suggestion
For the up-to-date list of available functions and their arguments, defaults and examples, see the sources of the
link:https://github.com/devonfw/IDEasy/tree/main/cli/src/main/java/com/devonfw/tools/ide/expression/function[function package] instead of a second copy of
```

quando632 (commented):
Nice cleanup. The base classes remove the repeated argument-count checks without changing behaviour: `ExpressionParser.apply` validates min/max before dispatching, so the direct `args.get(0)` is safe, and `PathFunction`'s `mode == null -> MODE_UNIX` matches the old `args.size() > 1` branch. No leftover references to `addSecretVariable`, `addSecretValue` or `secretVariables` anywhere in cli, gui, url-updater or security.

I built the `cli` module and ran `ExpressionParserTest`, `EnvironmentVariablesTest` and `DirectoryMergerExpressionTest`: 45 tests, 0 failures.

quando632 (approved):
All three comments are addressed: `SECRET_MIN_LENGTH` is now 8, the `@path` and `@ask-*` arguments are documented again, and the link label is fixed. `testShortSecretValueIsNotMaskedToAvoidCorruptingUnrelatedLogOutput` pins the case down nicely.

I re-ran `ExpressionParserTest`, `EnvironmentVariablesTest`, `DirectoryMergerExpressionTest`, `ClaudeTest`, `PrivacyUtilTest`, `EditionSetCommandletTest` and `VersionSetCommandletTest` on d1941711: 65 tests, 0 failures. CI on the head commit is green.

One note for a later story, not blocking this PR: with the threshold at 8, a secret shorter than 8 characters is no longer masked at all, while the previous threshold of 3 still masked it. Matching whole tokens instead of any substring would keep both properties.

cli/src/main/java/com/devonfw/tools/ide/context/IdeContext.java
```diff
@@ -301,20 +302,29 @@ default String askForSecret(String message) {
...
+   * @return {@code true} if the given variable name follows the naming convention for secret variables, {@code false} otherwise.
    */
-  void addSecretValue(String name, String value);
+  static boolean isSecretVariableName(String name) {
+
+    if (name == null) {
+      return false;
+    }
+    String upperName = name.toUpperCase(Locale.ROOT);
+    return upperName.endsWith("SECRET") || upperName.endsWith("PASSWORD") || upperName.endsWith("API_KEY");
```
hohwille: In the story I wrote "starts or ends with" but maybe even easier do it like this:

```suggestion
    return upperName.contains("SECRET") || upperName.contains("PASSWORD") || upperName.contains("API_KEY");
```

Issue #2433, opened 2026-09-08
Consider adding an EventBus to the new GUI


samuelkos17:
### Task summary

In order to get a cleaner implementation of the GUI we consider adding an EventBus/Eventlistener to it so that e.g. the TabFactory doesn't have to access the TabPane from the MainWindow.

### Additional context

#2401

samuelkos17:
Will be merged with #2401

Issue #2400, opened 2026-09-01
Rework Console to MVVM


samuelkos17:
### Task summary

Rework Console to MVVM

Issue #2107, opened 2026-07-03
Add documentation for contributing to the GUI


laim2003:
### Task summary

A more extensive documentation on what is necessary to succesfully contribute to the UI. See the sub-issues for all the tasks required to be done.

PR #2437, opened 2026-09-09
#2401: Rework Mainwindow

### LINKED ISSUE #2401: Rework MainWindow
### Task summary

The MainWindow is currently way too overloaded and needs to be changed to match MVVM.

### Additional context

_No response_

samuelkos17:
### Implemented changes:

This is quite a huge PR that completely reworked our GUI architecture, the most important changes:
* Reworked the GUI from "MVC" to MVVM
* Split the MainController into single modules (NavigationPanel, IdeLauncherTab)
* Introduced Tabs
* Introduced an EventBus
* Removed any logic from the .fxml files
* Introduced Services & Factories
* Refactored the folder structure of the GUI
  * Regarding the terminology of the new folder structure:
    * Factory: Creates ViewModels & Views while containing the necessary services to do so. Initialized in App
    * Service: Encapsulates business logic and access to data or functionality independently of the UI. Initialized in App
    * Control: A self-contained UI component of the MainWindow. Extracted frm MainWindowView to improve maintainability and seperation of concerns, where introducing a dedicated View/ViewModel split would not be worthwile.
    * View: The actual UI representation without business logic. Its responsibility is to display data nd information from the ViewModel through bindings.
    * ViewModel: Acts as a bridge between the View and the Services/Factory/Data layer. It maintains the current state and contains the UI-related logic.

The Refactoring-Work isn't done with this PR though, every functionality we had before (besides the language selection) is also present in the new GUI, but the GUI architecture is still not clean enough and there are a few hacks/smells in the code that still need to be reworked. In order to not bloat this PR even further the next tasks are listed here:
* Rework/Reintroduce language selection #2214 
* Implement Alert & Dialog Factorys #2439 
* Rework the console handling #2400
* Update GUI Docu #2107, 
* Implement an EventBus/EventListeners, especially to remove the weird TabPane attaching in the TabFactory #2433 

---

### Testing instructions

1. Run `ide gui` using the local dev build and test if everything works.

---

### Checklist for this PR

  with `internal`


gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
```diff
@@ -0,0 +1,96 @@
...
+      return existing;
+    }
+
+    Tab tab = new Tab(this.nlsService.get(titleKey), content);
+    tab.setUserData(titleKey);
+    tab.setClosable(true);
+    this.tabPane.getTabs().add(tab);
+    return tab;
+  }
+
```
laim2003: I think I might have an idea how we can optimize this. Instead of composing it like this, we could also have a TabComponent interface/abstract class that has the method getView() (which returns a node). This is the way UI on android implements reusable components ("fragments"): https://developer.android.com/reference/androidx/fragment/app/Fragment#onCreateView(android.view.LayoutInflater,%20android.view.ViewGroup,%20android.os.Bundle)

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
```diff
@@ -0,0 +1,96 @@
...
+
+  public void openLauncherTab() {
+    IdeLauncherViewModel viewModel = new IdeLauncherViewModel(this.guiStateManager, this.commandletService);
+    open(viewModel.getTabTitleKey(), new IdeLauncherView(viewModel, this.nlsService));
+  }
+
+
+  public Tab open(String titleKey, Node content) {
```
laim2003: `titleKey` seems a bit confusing to me. is `titleKey` a key that identifies a tab (seems to me) or is it the displayed tab title? Maybe lets decide on either "key" or "title". Or did I misunderstand something? 😅
```suggestion
  public Tab open(String key, Node content) {
```

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
```diff
@@ -0,0 +1,96 @@
...
+      this.tabPane.getSelectionModel().select(existing);
+      return existing;
+    }
+
+    Tab tab = new Tab(this.nlsService.get(titleKey), content);
+    tab.setUserData(titleKey);
+    tab.setClosable(true);
+    this.tabPane.getTabs().add(tab);
+    return tab;
+  }
```
laim2003: I think there might be a cleaner way to structure this. Right now the caller has to manually compose both the viewmodel and the view (e.g. in `openLauncherTab()`), and pass a title key that's really already owned by the viewmodel.

Instead, we could introduce a `TabComponent` interface (or abstract class) that encapsulates a tab and owns its own construction:

- `getId()` or `getKey()` - NLS key the tab is identified(?) by (replaces the separate `titleKey` parameter)
- `getView()` - returns the tab's Node, creating it on first call and attaching its viewModel.
- `onBeforeOpenTab()` - a hook run right before `getView()` is called
- `onClose()` - there might already be listeners for this in javafx, so this is just an inspiration
- `onFocus()` - same here?
 
This gives us a small, explicit UI lifecycle. It mirrors how Android handles reusable UI components ("fragments"), where the fragment owns creating its layout and viewmodel. I know I always talk about Android because I'm kind of used to it, but generally, I think just orienting a bit on how Android does thing UI-wise might not be bad for the GUI as they have quite a clean UI concept IMHO: [Fragment.onCreateView](https://developer.android.com/reference/androidx/fragment/app/Fragment#onCreateView(android.view.LayoutInflater,%20android.view.ViewGroup,%20android.os.Bundle)). The caller would then just do `tabFactory.open(someTabComponent)` and stop threading NlsService/viewmodel construction through the call site.

This also allows us to optimize other areas:
- We could simply inject the `ConsoleController` and the `NlsService` into each `TabComponent` via the constructor.

## Example how we implement this: 
```java
public abstract class TabComponent {

  public GuiStateManager guiStateManager;
  public CommandletService commandletService;
  public ConsoleController consoleController;

  public TabComponent(GuiStateManager guiStateManager, CommandletService commandletService, ConsoleController consoleController) {
    this.guiStateManager = guiStateManager;
    this.commandletService = commandletService;
    this.consoleController = consoleController;
  }

  public abstract TabView getView(NlsService nlsService);

  public abstract SimpleStringProperty getTabTitleKey();
}
```
Then the Launcher Tab content would be implemented by creating a custom `LauncherTabComponent`:
```java
public class LauncherTab extends TabComponent {

  public SimpleStringProperty tabTitleKey = new SimpleStringProperty("key_launcher_tab");

  public LauncherTab(GuiStateManager guiStateManager, CommandletService commandletService, ConsoleController consoleController) {
    super(guiStateManager, commandletService, consoleController);
  }

  @Override
  public TabView getView(NlsService nlsService) {
    IdeLauncherViewModel viewModel = new IdeLauncherViewModel(guiStateManager, commandletService);
    IdeLauncherView launcherView = new IdeLauncherView(viewModel, nlsService);

    //do more stuff (possibly)

    return launcherView;
  }

  @Override
  public SimpleStringProperty getTabTitleKey() {
    return this.tabTitleKey;
  }
}
```

Then in `TabFactory` we can use these components like this while completely detaching them from the `TabFactory`:
```java
  public void openLauncherTab() {
    open(new LauncherTab(this.guiStateManager, this.commandletService, this.consoleController));
  }


  public Tab open(TabComponent tabComponent) {

    Tab existing = findByTitleKey(tabComponent.getTabTitleKey().getValue());
    if (existing != null) {
      this.tabPane.getSelectionModel().select(existing);
      return existing;
    }

    Tab tab = new Tab(this.nlsService.get(tabComponent.getTabTitleKey().getValue()), tabComponent.getView(this.nlsService));
    tab.setUserData(tabComponent.getTabTitleKey().getValue());
    tab.setClosable(true);
    this.tabPane.getTabs().add(tab);
    return tab;
  }
```

@JoelAdbu maybe you can give your two cents on this as well

**TL;DR** : Introduce a `TabComponent` abstraction that owns its own view/viewmodel construction and lifecycle hooks. This moves composition logic from the caller into the tab itself, making the API simpler and better aligned with how frameworks like Android handle reusable UI components.

Edit: I see that `TabView.java` and `TabViewModel` already implement a similar pattern in the current code. However, this suggestion keeps the view and viewmodel composition closer together, avoiding the need for the caller to thread dependencies and manually wire them. So basically the use of `TabComponent` just extends what you already added in this PR.

gui/src/main/java/com/devonfw/ide/gui/helper/combobox/ComboBoxViewModel.java
laim2003, line 1: I'm not too sure whether this VM is not creating more complexity than it solves. This seems to me like its adding an unnecessary layer: The whole point of a helper VM is that callers express intent through it. Instead, MainWindowViewModel reaches in and mutates the raw list directly as` getItems()` etc just pass on the Observables.

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
laim2003: Ah. I think this is referring to the NLS key of the title of the tab.

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
```diff
@@ -0,0 +1,96 @@
...
+
+import com.devonfw.ide.gui.context.GuiStateManager;
+import com.devonfw.ide.gui.service.CommandletService;
+import com.devonfw.ide.gui.service.NlsService;
+import com.devonfw.ide.gui.ui.controls.console.ConsoleController;
+import com.devonfw.ide.gui.ui.tab.launcher.IdeLauncherView;
+import com.devonfw.ide.gui.ui.tab.launcher.IdeLauncherViewModel;
+
+
+public class TabFactory {
```
laim2003, line 21:
```suggestion
public class TabManager {
```

Maybe we can discuss whether the `TabFactory ` isn't actually more like a manager class, especially since we have functions like `isOpen()` and `focus()`?

gui/src/main/java/com/devonfw/ide/gui/service/CommandletService.java
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * Sets the action that is run before each commandlet launch. Part of the hack to make the Console auto-show when an IDE is launched, should be removed when
+   * Console is reworked.
+   *
+   * @param preLaunchAction the action to run before launching a commandlet (must not be <code>null</code>).
+   */
+  public void setPreLaunchAction(Runnable preLaunchAction) {
+    this.preLaunchAction = Objects.requireNonNull(preLaunchAction);
+  }
```
laim2003, line 64: Considering that `CommandletService` is instantiated once in App.java and then injected, i think this might cause some issues. Because when multiple sources will try to set their own preLaunchAction, they will override each other as they all use the same `CommandletService`, while not knowing of each other. Maybe we can solve this via a list and `addPreLaunchAction()` etc? I think generally it is very useful to be able to have configurable pre- and post actions though!

gui/src/main/java/com/devonfw/ide/gui/ui/controls/mainwindow/NavigationPanelControl.java
laim2003, line 1: I understand the purpose of the `ComboboxViewModel` now. However, I think introducing a `NavigationPanelViewModel` would lead to a cleaner architecture. While it would likely be a very small view model, I believe the improved alignment of responsibilities would justify the additional class. Right now, the responsibilities are kind of intertwined between the `NavigationPanelControl` and `MainViewModel` but then also the combobox ViewModels, so basically the `MainViewModel` seems like a man in the middle😅

gui/src/main/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowView.java
```diff
@@ -0,0 +1,188 @@
...
+    loader.setControllerFactory(clazz -> clazz == ConsoleController.class ? this.consoleController : null);
+
+    try {
+      loader.load();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+
+    this.tabFactory.attach(this.tabContainer);
+    this.tabFactory.openLauncherTab();
```
laim2003, line 121: If we use the architecture I explained in my review comments in TabFactory, we could simply use this here:
```java
    this.tabFactory.open(new LauncherTab(xyz));
```
This greatly improves the SOC and TabFactory does not need to know of specifically a `LauncherTab`

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
```diff
@@ -0,0 +1,96 @@
...
+    this.tabPane = Objects.requireNonNull(tabPane);
+  }
+
+  /**
+   * Registers an action that runs before any commandlet launched from a tab (e.g. to show the console). This is a hack for the console auto-showing when
+   * starting an IDE and should be removed when the console is reworked.
+   *
+   * @param preLaunchAction the action to run before a commandlet launches.
+   */
+  public void setPreLaunchAction(Runnable preLaunchAction) {
```
laim2003, line 62: By the name of this method you could misunderstand that setting a pre-launch action is an action that is executed before a tab is opened in general, while this actually wires the action to the commandletService.

gui/src/main/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowView.java
```diff
@@ -0,0 +1,188 @@
...
+    // Show the console before launching an IDE (restores the previous behavior).
+    this.tabFactory.setPreLaunchAction(() -> this.viewModel.setConsoleVisible(true));
```
laim2003, line 123: Architecturally, it seems weird for this to be here, since this should be decided by the Tab that launches the commandlets. If we apply the new structure for the tabbing I explained in the comments above, we could potentially completely remove any reference to the `CommandletService` in the `TabFactory` and directly pass the `CommandletService` to the new `LauncherTab` instead. Then the `LauncherTab` can directly itself set the pre-launch action. `LauncherTab` could accept a reference to this `ViewModel` via a constructor parameter, although that also does not seem perfect... After we successfully implemented the EventBus in a later PR, we can also simply send an "OPEN_CONSOLE" event and let the ViewModel handle it directly without passing it as an argument.

gui/src/main/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowView.java
```diff
@@ -0,0 +1,188 @@
...
+      }
+    });
+  }
+
+  private void applyStatusLabelStyle(boolean clickable) {
+
+    if (clickable) {
+      this.statusLabel.setStyle(
+          "-fx-text-fill: blue;"
+              + "-fx-cursor: hand"
```
laim2003, line 189: We should probably add a ticket in the future to implement styling via CSS, i'm not sure if we have done that yet

gui/src/main/java/com/devonfw/ide/gui/ui/tab/launcher/IdeLauncherViewModel.java
```diff
@@ -0,0 +1,41 @@
+package com.devonfw.ide.gui.ui.tab.launcher;
```
laim2003, line 1: See my comments about the new architecture.

gui/src/test/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowViewModelTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+  private MainWindowViewModel viewModel;
+
+  /** Collects what the view model reports instead of the dialog the view would show. */
+  private List<String> reportedErrors;
+
+  @BeforeEach
+  void setUp() throws IOException {
+
+    FakeProjectFolderStructureHelper.createFakeProjectFolderStructure(this.mockIdeRoot);
```
laim2003, line 47: This is probably a left-over from my old tests right? Because generally, we should avoid using this helper method and rather use the fake project testing system that we use in all other components of IDEasy.

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
```diff
@@ -0,0 +1,96 @@
...
+  private final CommandletService commandletService;
+
+  private TabPane tabPane;
+
+
+  public TabFactory(GuiStateManager guiStateManager, NlsService nlsService, CommandletService commandletService, ConsoleController consoleController) {
+    this.guiStateManager = Objects.requireNonNull(guiStateManager);
+    this.nlsService = Objects.requireNonNull(nlsService);
+    this.consoleController = Objects.requireNonNull(consoleController);
+    this.commandletService = commandletService;
```
laim2003:
```suggestion
    this.commandletService = Objects.requireNonNull(commandletService);
```

laim2003 (changes requested):
Thanks for the deep rework! It looks way better already. The architecture is now much cleaner and allows for other features to be built on top. Regarding the TabFactory/general Tab architecture I had some optimization ideas that we should think about. But anyways good work!

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
laim2003: I just realised that my example is not perfect as well, of course `TabComponent `would not hold a class like `CommandletService`. That would rather be implemented by a subclass like `LauncherTabComponent`

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
laim2003: In this case the we may not actually change it, or only like "nlsTitleKey"

gui/src/main/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowView.java
```diff
@@ -0,0 +1,188 @@
...
+  /** Only expand the console when the divider is (nearly) collapsed, so an already open console keeps the size the user chose. */
+  private static final double CONSOLE_COLLAPSED_THRESHOLD = 0.9;
+
+  private static final double PROGRESSBAR_VISIBLE_WIDTH = 150.0;
+
+  @FXML
+  private NavigationPanelControl navigationPanelControl;
+
+  @FXML
+  private TabPane tabContainer;
```
JoelAdbu: we should rename this to:

```suggestion
private TabPane tabPane
```

gui/src/main/java/com/devonfw/ide/gui/App.java
JoelAdbu: I think it would make sense to extract this into a method something like `configureMainWindow()` to improve readability.

gui/src/main/java/com/devonfw/ide/gui/service/CommandletService.java
samuelkos17: This whole preLaunchAction thing is actually just a hack until we figure out what to do with the console, i intented for it to be removed once we have the Console handling optimized

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
samuelkos17: Yes this is also part of the hack i didn't really pay attention to the JavaFocs tbh and will make it more clear

gui/src/main/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowView.java
JoelAdbu: We should rename this to tabPane to stay consistent with the naming used throughout the GUI:
```suggestion
private TabPane tabPane
```

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
laim2003: yeah I now saw that you actually commented that above some of the methods, didn't catch that earlier tbh

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
samuelkos17: Thanks a lot for the suggestion and the work that you put in this Review. TBH i don't really see the issue with creating the ViewModel and the View seperately in the e.g. TabFactory. And I also believe that we can't generalize what each constructor of Views and ViewModel will need in the future. I feel like this would make everything more complicated than it has to be, but maybe I just don't quite understand what the value of this change would be. Could you maybe explain it further?

gui/src/main/java/com/devonfw/ide/gui/helper/combobox/ComboBoxViewModel.java
samuelkos17: I see what you mean, I'll look into it again and then answer this comment.

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
samuelkos17: Yeah I'll change it!

gui/src/main/java/com/devonfw/ide/gui/ui/controls/mainwindow/NavigationPanelControl.java
samuelkos17: My idea was that the MainView basically just acts as a shell that's the "background" of our GUI and every element of it should be handled within the MainViewModel. If that's too confusing then we can change it though.

gui/src/main/java/com/devonfw/ide/gui/service/CommandletService.java
```diff
@@ -0,0 +1,120 @@
...
+      Task<Void> commandletTask = new Task<>() {
+        @Override
+        protected Void call() {
+
+          try {
+            IdeStartContextImpl startContext = new IdeStartContextImpl(IdeLogLevel.INFO, guiLogListener);
+            Path workspacePath = CommandletService.this.guiStateManager.getIdeRootDir()
+                .resolve(CommandletService.this.guiStateManager.getSelectedProject())
+                .resolve(IdeContext.FOLDER_WORKSPACES)
+                .resolve(CommandletService.this.guiStateManager.getSelectedWorkspace());
```
JoelAdbu: We should reuse the context provided by switchContext() instead of rebuilding the workspace path here. This avoids duplicating the path construction logic.
```suggestion
guiStateManager.getCurrentContext().getCwd();
```

gui/src/main/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowView.java
samuelkos17: Yeah this is, as I mentioned in the top comment, not a clean implementation that's why I explicility mentioned the tabfactory-rework in #2433 and already made a cleaner solution while working that out. I guess we could discuss if we would prefer your rework suggestion here or the EventBus solution.

gui/src/main/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowView.java
samuelkos17: Yeah i agree

gui/src/test/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowViewModelTest.java
samuelkos17: I'll look into it

gui/src/main/java/com/devonfw/ide/gui/ui/tab/TabView.java
```diff
@@ -0,0 +1,33 @@
...
+    final FXMLLoader loader = new FXMLLoader(getClass().getResource(fxmlName));
+    loader.setRoot(this);
+    loader.setController(this);
+    loader.setResources(nlsService.getResourceBundle());
+    try {
+      loader.load();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
```
JoelAdbu: IMHO, it would make sense to extract this FXML loading logic into FxHelper, as we already have very similar code in `NavigationPanelControl` and `MainWindowView`.

You could add two helper methods along these lines:

``` java
public static void loadFxml(Node root, String fxmlName, ResourceBundle resources) {

    loadFxml(root, fxmlName, resources, null);
  }

public static void loadFxml(Node root, String fxmlName, ResourceBundle resources, Callback<Class<?>, Object> controllerFactory) {

    FXMLLoader loader = new FXMLLoader(root.getClass().getResource(fxmlName));
    loader.setRoot(root);
    loader.setController(root);
    loader.setResources(resources);
    if (controllerFactory != null) {
      loader.setControllerFactory(controllerFactory);
    }
    try {
      loader.load();
    } catch (IOException e) {
    }
  }
```
This would reduce code duplication and keep the FXML loading pattern consistent across the different views.

After introducing this helper in FxHelper, the following block could be simplified to:

```suggestion
FxHelper.loadFxml(this, fxmlName, nlsService.getResourceBundle());
```
Please apply the same refactoring in the MainWindowView constructor and the NavigationPanelControl constructor as well.

gui/src/main/java/com/devonfw/ide/gui/App.java
JoelAdbu, line 109: I think it would make sense to extract this into a method something like `configureMainWindow()` to improve readability.

gui/src/main/java/com/devonfw/ide/gui/helper/combobox/ComboBoxViewModel.java
samuelkos17: The reason I implemented it is to improve readibility, if we would just write the properties into the MainWindowViewModel we would have 9 extra properties and their Getters in the ViewModel which would bloat it up

gui/src/main/java/com/devonfw/ide/gui/ui/tab/TabView.java
```diff
@@ -0,0 +1,33 @@
...
+  protected void loadFxml(String fxmlName, NlsService nlsService) {
+    final FXMLLoader loader = new FXMLLoader(getClass().getResource(fxmlName));
+    loader.setRoot(this);
+    loader.setController(this);
+    loader.setResources(nlsService.getResourceBundle());
+    try {
+      loader.load();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
```
JoelAdbu: IMHO, it would make sense to extract this FXML loading logic into FxHelper, as we already have very similar code in `NavigationPanelControl` and `MainWindowView`.

You could add two helper methods along these lines:

``` java
public static void loadFxml(Node root, String fxmlName, ResourceBundle resources) {

    loadFxml(root, fxmlName, resources, null);
  }

public static void loadFxml(Node root, String fxmlName, ResourceBundle resources, Callback<Class<?>, Object> controllerFactory) {

    FXMLLoader loader = new FXMLLoader(root.getClass().getResource(fxmlName));
    loader.setRoot(root);
    loader.setController(root);
    loader.setResources(resources);
    if (controllerFactory != null) {
      loader.setControllerFactory(controllerFactory);
    }
    try {
      loader.load();
    } catch (IOException e) {
    }
  }
```
This would reduce code duplication and keep the FXML loading pattern consistent across the different views.

After introducing this helper in FxHelper, the following block could be simplified to:

```suggestion
FxHelper.loadFxml(this, fxmlName, nlsService.getResourceBundle());
```
Please apply the same refactoring in the MainWindowView constructor and the NavigationPanelControl constructor as well.

gui/src/test/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowViewModelTest.java
samuelkos17: Yeah it's still a leftover and thanks for pointing that out! Chhanging this here wouldn't be much of an issue, but in `AppBaseTest` the change would be bigger. The tests work now and to not further bloat this PR I would suggest to move this into a seperate issue "Rework GUI Tests".

gui/src/main/java/com/devonfw/ide/gui/service/CommandletService.java
samuelkos17: This is actually not my work but I agree that we should change this, however I would suggest to use `.getWorkspacePath()` over `.getCwd()`. Is there a reason why you suggested `.getCwd()`?

gui/src/main/java/com/devonfw/ide/gui/ui/tab/TabView.java
samuelkos17: Thanks for the suggestion, I'll implement it. However I have a question, what is the controllerFactory that you mention in your suggestion?

samuelkos17: Hm I'm not sure if this can be applied to MainWindowView and NavigationPanelControl though since they add other stuff to the loader before loading

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
samuelkos17: You're right, the TabFactory currently isn't a factory per se but I intended for it to be one but made a mistake while designing it. It should actually only be a helper Class that only creates new Tabs and doesn't do anything else. The `isOpen` and `focus()` logic should be handled by the MainWindowView. The root cause for this problem is the way that the TabFactory currently handles the TabPane from the MainWindowView, but this will be removed in the future (#2433) so that the TabFactory will actually end up as a TabFactory and not act as a Manager as it currently does.

gui/src/main/java/com/devonfw/ide/gui/service/CommandletService.java
JoelAdbu: Good point. I agree, getWorkspacePath() would be the right method to use here.

gui/src/main/java/com/devonfw/ide/gui/App.java
samuelkos17: Thanks for the hint, I'll impement it

gui/src/main/java/com/devonfw/ide/gui/ui/tab/TabView.java
JoelAdbu: The controllerFactory is the “other stuff” you mentioned. It refers to JavaFX’s `FXMLLoader.setControllerFactory(...)`.

Before loading the FXML, `MainWindowView` configures the loader as follows:
``` java
loader.setControllerFactory(clazz -> clazz == ConsoleController.class ? this.consoleController : null);
```
This is necessary because `MainWindowView.fxml` includes `console.fxml`, which declares its own `fx:controller` of type `ConsoleController`.

Without the controller factory, the `FXMLLoader` would create a new `ConsoleController` using its no-argument constructor. However, that new instance would not contain the dependencies already injected into `this.consoleController`.

The controller factory prevents this by returning the existing, fully initialized `ConsoleController` instance.
Therefore, the four-argument overload:

``` java
loadFxml(root, fxmlName, ressources, controllerFactory);
```

is required by `MainWindowView`. `NavigationPanelControl`, on the other hand, only needs the ressources and can therefore use the three-argument overload.

The corresponding method calls would look as follows:

In `MainWindowView`:

``` java
FxHelper.loadFxml(this, "MainWindowView.fxml", this.nlsService.getResourceBundle(),
        clazz -> clazz == ConsoleController.class ? this.consoleController : null);
```

In `NavigationPanelControl`:

``` java
FxHelper.loadFxml(this, "NavigationPanelControl.fxml", ResourceBundle.getBundle("nls.messages", Locale.getDefault()));
```

In `TabView`:
``` java
FxHelper.loadFxml(this, fxmlName, nlsService.getResourceBundle());
```
The two methods in FxHelper could then be adjusted as follows:

``` java
public static void loadFxml(Node root, String fxmlName, ResourceBundle resources) {
    loadFxml(root, fxmlName, resources, null);
  }
  
  public static void loadFxml(Node root, String fxmlName, ResourceBundle resources, Callback<Class<?>, Object> controllerFactory) {

    final FXMLLoader loader = new FXMLLoader(root.getClass().getResource(fxmlName));
    loader.setRoot(root);
    loader.setController(root);
    loader.setResources(resources);
    if (controllerFactory != null) {
      loader.setControllerFactory(controllerFactory);
    }
    try {
      loader.load();
    } catch (IOException e) {
    }
  }

```

samuelkos17: Hm I think this would just make things more complicated, if we're making specialized loadFxml methods then we could just move them to the Views or keep them the way they are. I would suggest leaving things like they are (I already pushed the update to the loadFXML() method for Tabs), and maybe look back upon this on the future, especially since the problems we have right now are only due to some Refactoring-work not being done (see #2214 and #2400). I think when we are done with this, all loadFxml methods would be the generic version we use for the tabs.

gui/src/main/java/com/devonfw/ide/gui/ui/controls/mainwindow/NavigationPanelControl.java
samuelkos17: After further thinking about this I agree that we should split the NavigationPanelControl in a View and a ViewModel and i'll change it that way

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
laim2003: I think this would allow a much clearer separation of concerns. Let me explain with the `TabFactory` code as the reference.

The problem I see in the current design - look at `openLauncherTab()` as an example:

```java
public void openLauncherTab() {
  IdeLauncherViewModel viewModel = new IdeLauncherViewModel(this.guiStateManager, this.commandletService);
  open(viewModel.getTabTitleKey(), new IdeLauncherView(viewModel, this.nlsService));
}
```

Three things happen at the call site that conceptually belong to the tab itself:
1. The `IdeLauncherViewModel` is constructed here.
2. The `IdeLauncherView` is constructed here (and it wires the viewModel).
3. The tab title is derived from the viewModel (`viewModel.getTabTitleKey()`), yet we still pass it into `open()` as a separate parameter.

With my design the call site just becomes:

```java
public void openLauncherTab() {
  open(new LauncherTab(this.guiStateManager, this.commandletService, this.consoleController));
}
```

`LauncherTab` (or any other tab) then owns:
- its own title key (the component decides what its tab's title should be),
- constructing its own viewModel,
- constructing its own view (and wiring the viewModel into it),
- and any lifecycle hooks (before-open, close, focus).

This keeps the TabFactory (and any other caller) from caring about the individual `[TabName]ViewModel` / `[TabName]View`. It just takes a `TabComponent`, pulls the title and content node out of it, and handles the generic
tab mechanics (find-existing-by-title, create Tab, add to `TabPane`).

IMHO the payoff is modularity: we can launch a tab from anywhere without having to ensure every launch site does the composition the same way. If we end up on the EventBus there will hopefully be a single place that triggers launching a tab, but there can still be edge cases with multiple sources creating a tab, and this makes that case easier to handle. I hope my explanation makes sense😅

With this design im thinking of the ["Tell, Don't Ask" Principle](https://dev.to/thicha0/tell-dont-ask-principle-explained-1gbf)

laim2003: Ok, so I'll quickly read a bit about the Factory pattern... to understand why it makes sense here. Give me a sec, maybe I can make my explanation more precise.

gui/src/main/java/com/devonfw/ide/gui/ui/controls/mainwindow/NavigationPanelControl.java
laim2003: Nice!

laim2003: I cant resolve the comments for some reason btw.

gui/src/main/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowView.java
laim2003: As I said I will read a bit into the Factory Pattern again as I havent used it in a while before making more recommendations here.

gui/src/test/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowViewModelTest.java
laim2003: I'll create the issue.

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
samuelkos17: Ahh okay I understand it now. I think this is a good idea, even though it creates a little bit more work now it will indeed make working with Tabs easier in the future. This can also be nicely combined with the EventBus, i guess I'll just implement it

laim2003: Okay! TBH in the end I wasn't too sure about my recommendation either, but if we implement it either way, it will be fine in my opinion.

laim2003 (approved):
Thanks for your work👍 Just changed this to approve so the merge is not blocked

gui/src/test/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowViewModelTest.java
laim2003, line 1: Did you forget to push here? 😅

laim2003 (approved):
Looks good, thanks

gui/src/test/java/com/devonfw/ide/gui/ui/mainwindow/MainWindowViewModelTest.java
samuelkos17: I forgot to delete the file 😅

gui/src/main/java/com/devonfw/ide/gui/event/TabChangeEvent.java
```diff
@@ -0,0 +1,7 @@
+package com.devonfw.ide.gui.event;
+
+import javafx.scene.control.Tab;
+
+public record TabChangeEvent(Tab tab) {
```
hohwille, line 5: Just thinking but these events will be send from different threads. Should they contain JavaFx objects or should we avoid that? However for a `TabChangeEvent` this seems pragmatic and easy - if we need to process out of GUI thread, we can dispatch it as async task so maybe fine.

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
```diff
@@ -0,0 +1,68 @@
...
+ * of creating a duplicate. The managing part is to be removed in the future.
+ */
+public class TabFactory {
+
+  private final GuiStateManager guiStateManager;
+  private final NlsService nlsService;
+  private final ConsoleController consoleController;
+  private final CommandletService commandletService;
+  private final EventBus eventBus;
+
```
hohwille, line 27: We are adding more and more dependencies in the GUI and have no nice dependency injection yet.
So I am just thinking...
* since GUI is not running in native image and could use reflection - should we introduce some DI framework?
* or should we follow the same pattern and have a GuiContext as "API" extending IdeContext adding the additional dependencies like NlsService, EventBus, etc. continuing our god-object-(anti-)pattern?

gui/src/main/java/com/devonfw/ide/gui/ui/modal/IdeDialog.java
```diff
@@ -1,4 +1,4 @@
-package com.devonfw.ide.gui.modal;
+package com.devonfw.ide.gui.ui.modal;
```
hohwille, line 1: Thanks for cleaning the structure.
My thoughts:
1. for `gui` the subpackge `ui` does not feel so natural since a GUI is actually a UI. How about dividing the `gui` into subpackages like `controler` , `view`, etc.? Maybe I also did not really get the `ui` sub-package since pretty much almost everything is now in `ui` except for fundamentals `App`, `AppLauncher`, etc. So probably read on to my 2nd point.
2. For packaging it is typically best practice to first divide into functional aspects and then only do technical dividing on a deeper level. E.g. assuming we have a feature/dialog to configure software, another to launch IDEs, etc. then IMHO these functional units should be grouped primary together in their own sub-package directly below `gui`. If we then want to divide further, we can create sub-packages for `view`, `controler`, `model`, or whatever we need. But maybe we then also do not need this additional dividing if the functional units dont grow so big and have maybe ~5-10 classes. If they have 50+ classes then further sub-packages are encouraged to give structure and orientation. Looking at your first examples, I like your style to have the "logic" or "functionality" as a prefix of the class-names and the technical aspects as a suffix (`View`, `ViewModel`, `Tab`). For launching/opening an IDE, I would however, consider the `ide` as the more important "thing" for the sub-package than the "launcher" aspect, because "launcher" could also be the launcher of the entire GUI or anything else.
3. We already discussed that `dialog` and `modal` can be confusing terms. My suggestion would be to use the term `popup` for package and class here. Then the fact if this is system or application modal can also change without the need to refactor namings. A "view" that lives inside a tab and groups a specific feature is what I typically call a `dialog` (while a dialog may also be shown in a popup). If we do not agree on this term, we should find another word for that that is neither `dialog` nor `popup`.

gui/src/main/resources/com/devonfw/ide/gui/ui/controls/console/console.fxml
```diff
@@ -1,22 +1,15 @@
...
+<?import javafx.scene.layout.*?>
 <AnchorPane
   xmlns:fx="http://javafx.com/fxml"
   xmlns="http://javafx.com/javafx"
   fx:id="consoleRoot"
-  fx:controller="com.devonfw.ide.gui.console.ConsoleController"
+  fx:controller="com.devonfw.ide.gui.ui.controls.console.ConsoleController"
   prefWidth="Infinity"
-  style="-fx-background-color: #FFFFF; -fx-border-color: #cccccc; -fx-border-width: 1;"
+  style="-fx-background-color: #FFFFFF; -fx-border-color: #cccccc; -fx-border-width: 1;"
```
hohwille, line 12: fine, but we should aim to remove all such styles out of FXML and into CSS on the long run.
For now you can close this thread so we can get this PR merged.

hohwille (approved):
@samuelkos17 thanks for your PR.
I very much like your work and can already see that we achieve the lose-coupling and more modular and maintainable code this way. Great work 👍 
I focused on two things that we should rather get aligned also now and not when tons of screens (aka dialogs) have been implemented (and need to be refactored):
* packaging layout
* dependency management to access technical components (required almost everywhere like e.g. `NlsService` alongside with our famous `IdeContext`. Just think to add another such dependency in the future. You will have to add it to many constructors everywhere causing quite some refactoring that we maybe can avoid.

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
hohwille: I do not really want to promote singleton pattern but maybe for EventBus and NlsService this could also be a pragmatic alternative if you do not like my two alternatives given above.

gui/src/main/resources/com/devonfw/ide/gui/ui/controls/console/console.fxml
samuelkos17: Yes, Lukas and me already talked about this and he created an Issue for changing that #2482

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
samuelkos17: I think introducing a DI framework would be a good addition to our GUI. However I would suggest creating a new issue for that so that we can get this PR done.

gui/src/main/java/com/devonfw/ide/gui/event/TabChangeEvent.java
samuelkos17: I think it's okay to send JavaFX Objects through these events, as long as the listener wraps it's handling in `FxHelper.runFxSafe`, as we currently do. If we ever need a truly non-GUI consumer for these tabs (e.g. a logger that wants to know which tab was requested) then we could just add another event not containing the tab but maybe only the tab title key.

gui/src/main/java/com/devonfw/ide/gui/ui/modal/IdeDialog.java
samuelkos17: Regarding 1:
The `ui` subpackage is intented to contain all classes (Views and their ViewModels) that are really shown to the user, while the other subpackages contain only classes that are part of the GUI but which can't be seen (Factorys, Helpers, the whole Context thing). But if this is making things more complicated or if it's just redundant then I can remove `ui` and keep everything in the `gui` package.

Regarding 2:
I see, so this means that I wouldn't keep a subpackage called `tab` that contains the abstract classes for Tabs and then subpackages for each Tab, but the subpackages would be directly in the `gui` package, while also containing e.g. services that are directly tied to this Tab/view/viewmodel? For the IDE-launcher the package would then be named `ide` and it would contain the IdeLauncherTab, View, and ViewModel and if we would add another tab that changes settings of IDEs before downloading them then their Tab, View and ViewModel would also be created in the `ide` subpackage?  I'm not sure if I understand it correctly so clarification would be really helpful

Regarding 3:
Yes, I agree, I actually just didn't change anything about the whole popup thing but I like the name popup for them!

gui/src/main/java/com/devonfw/ide/gui/factory/TabFactory.java
laim2003: We should probably discuss which framework to use in a separate meeting for the UI in the next days.

gui/src/main/java/com/devonfw/ide/gui/core/factory/TabFactory.java
```diff
@@ -0,0 +1,68 @@
...
+ * of creating a duplicate. The managing part is to be removed in the future.
+ */
+public class TabFactory {
+
+  private final GuiStateManager guiStateManager;
+  private final NlsService nlsService;
+  private final ConsoleController consoleController;
+  private final CommandletService commandletService;
+  private final EventBus eventBus;
+
```
samuelkos17: I'll mark this as resolved since we discussed this today and decided on doing research on DI frameworks and sticking to a god object for now.

gui/src/main/java/com/devonfw/ide/gui/ui/modal/IdeDialog.java
samuelkos17: I refactored the packages as we discussed today.

hohwille (approved):
@samuelkos17 thanks for the update. For me all looks good now and I would consider this as ready to merge. 👍 
Maybe we should wait until the release is through but if nobody else has concerns we should then finally merge to open the road for follow-up changes.

samuelkos17:
@hohwille yes we can wait til the release is done. While updating my review for the docu in #2291 I've also realized that I didn't use the guiStateManager god-object everywhere and still had a few useless parameters in constructors so I cleaned that up and pushed it. It's only a small change and IMHO you don't really need to look at it.

PR #2438, opened 2026-09-09
#2422: cleanup determine installed edition and version and restore software version file

### LINKED ISSUE #2422: Cleanup how to determine installed edition and version and potentially restore software version file
### Task summary

This is a follow up of story #2284 and PR #2384.
First please check how PR #2384 was implemented (see files changed tab)
Status quo of `ToolCommandlet` is that the design is great:
* `getInstalledEdition()` and `getInstalledVersion()` are final and delegate to `getInstalledEditionAndVersion()` what is fine and correct.
* `getInstalledEditionAndVersion()` determines its value from a CachedValue and if it needs to be computed calls `computeInstalledEditionAndVersion()` - also fine and correct.

Now what needs to be improved is the implementation of that in `LocalToolCommandlet`:
https://github.com/devonfw/IDEasy/blob/2befb780b9bb619c9ced5f59b45bc63208bc490d/cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java#L409-L423

* If the software folder for the tool does not exist, it returns `null` - already great
* Now we should try to resolve a potential link (see implementation of `getInstalledEdition`) - if we can resolve such link and it points to our software repository, we can determine both the edition and the version from it in one go. Example `$IDE_HOME/software/«tool» -> $IDE_ROOT/_ide/software/default/«tool»/«edition»/«version»[/...]`
* If that fails (e.g. the tool is not using software repository) we should fall back to another protected method `computeInstalledEditionAndVersionFromLocalSoftwareFolder`.
* This method shall lookup the tool version file. If it exists, take the version from there. 
* If the version file does not exist, we call another protected method `computeInstalledVersionFromLocalSoftwareFolder` that by default returns `null`. This method will then be overridden by `Python` to determine the installed version from CLI.
* Now if that method returned a `VersionIdentifier` and not `null`, we log a warning that the version file was lost and will be restored with the current version and then we restore that version file. This logic is in `LocalToolCommandlet.computeInstalledEditionAndVersionFromLocalSoftwareFolder` and not in `Python` or other specific commands.
* However, if we still got `null` for the version, we return `null` as a final result (do not further try to determine an edition) and assume the tool installation is broken (so it will be reinstalled).
* For the edition, we can only guess in such case. Therefore create another protected method `computeInstalledEditionFromLocalSoftwareFolder` that by default returns `this.tool`. Potentially we can then override this method for special tools.
* Now we can cleanup the changes to the installation routine from PR #2384 since it is not needed any more.
* Also we should improve/fix `Python.readVersionFromInterpreter` since it is hardcoded for Windows only (`exe` suffix cannot work on other platforms). Further, we should reuse the proper `runTool` method (that does not trigger another installation) for better reuse instead of creating our own ProcessContext what might als be buggy:
https://github.com/devonfw/IDEasy/blob/2befb780b9bb619c9ced5f59b45bc63208bc490d/cli/src/main/java/com/devonfw/tools/ide/tool/ToolCommandlet.java#L295

### Additional context

There is also a method `computeInstalledPackageVersion` this seems to be what we want to archive with what I named `computeInstalledVersionFromLocalSoftwareFolder` as I was not aware of this any more.
Please try to align when designing the PR implementation.

JoelAdbu:
### Implemented changes:

* Reworked `LocalToolCommandlet.computeInstalledEditionAndVersion` as described in issue #2422
* Introduced the separation between:
   * `computeInstalledEditionAndVersionFromLocalSoftwareFolder()`
   * `computeInstalledVersionFromLocalSoftwareFolder()`
   * `computeInstalledEditionFromLocalSoftwareFolder()`
* Restored missing software version files centrally in `LocalToolCommandlet` instead of tool-specific installation logic.
* Removed the installation-time workarounds introduced in PR #2384 that are no longer necessary.
* Simplified Python-specific version detection by overriding `computeInstalledVersionFromLocalSoftwareFolder()`. 
* Updated `Python.readVersionFromInterpreter()` to reuse `ToolCommandlet.runTool(..)` and avoid hardcoded `.exe` handling.

---

### Testing instructions

1. Run `mvn clean test` in the `cli` module. The test `PythonTest.testInstallRestoresMissingVersionFileAndPreservesPackages` covers the behavior affected by this refactoring. It installs Python, adds a file to `site-packages`, deletes `.ide.software.version`, and installs again. The test verifies that the version file is restored, installed packages are preserved, and no "Deleting corrupted installation" occurs.

Note for reviewers: this PR mainly restructures how installed edition and version information is determined. The existing test continues to verify that the missing version file is restored and that Python installations are not recreated unnecessarily.

---

### Checklist for this PR

  with `internal`


quando632 (commented):
Splitting `computeInstalledEditionAndVersion` into the three `...FromLocalSoftwareFolder` hooks is a good seam. Python now expresses its special case in one small override instead of intercepting `getInstalledVersion(Path)`, and deriving edition and version from a single `getInstalledSoftwareRepoPath` lookup removes the duplicated path matching of the old `getEdition`. `mvn clean test` passes locally with 1005 tests and no failures.

I have only some small findings

cli/src/main/java/com/devonfw/tools/ide/tool/python/Python.java
```diff
@@ -146,15 +147,9 @@ private VersionIdentifier readVersionFromPyvenvCfg(Path installationPath) {
...
-    Path binaryPath = binPath.resolve(getBinaryName());
-    if (!Files.exists(binaryPath)) {
-      binaryPath = binPath.resolve(getBinaryName() + ".exe");
-    }
-    if (!Files.exists(binaryPath)) {
-      LOG.debug("Python binary does not exist in {}.", binPath);
-      return null;
-    }
-    String output = this.context.newProcess().runAndGetSingleOutput(IdeLogLevel.DEBUG, binaryPath.toString(), "--version");
+    ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.NONE).withPathEntry(binPath);
```
quando632, line 150: Without the previous `Files.exists(binaryPath)` guard the lookup falls back to `tool2pathMap` and the system PATH, because `SystemPath.findBinary` returns the plain name when the venv interpreter is gone. A foreign python is then queried and its version is written into the venv by `restoreMissingVersionFile`. Verified locally: with an empty `software/python/bin`, `getInstalledVersion()` returns a version and writes it to `.ide.software.version`, while with the guard below it returns null and writes nothing, with all `PythonTest` cases still green.

```suggestion
    Path binaryPath = this.context.getPath().findBinary(binPath.resolve(getBinaryName()));
    if (!Files.exists(binaryPath)) {
      LOG.debug("Python binary does not exist in {}.", binPath);
      return null;
    }
    ProcessContext pc = this.context.newProcess().errorHandling(ProcessErrorHandling.NONE).withPathEntry(binPath);
```

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -207,17 +206,10 @@ public ToolInstallation installTool(ToolInstallRequest request) {
...
-          restoreMissingVersionFile(installationPath, installedVersion);
-          if (installedVersion.equals(resolvedVersion)) {
-            return createToolInstallation(installationPath, installedVersion, false, processContext, additionalInstallation);
-          }
-          // the installation on disk is a different version than requested so we continue with the regular installation
-          // that will backup the existing installation before installing the requested version.
+          // Version file missing but tool allows this - restore it and preserve installation
+          LOG.warn("Version file missing at {} - restoring it for tool {}", toolVersionFile, this.tool);
+          // Restore the missing file
+          return createToolInstallation(installationPath, resolvedVersion, false, processContext, additionalInstallation);
```
quando632, line 212: The message announces that the version file is restored, but this branch only calls `createToolInstallation`, which does not write it. Calling `restoreMissingVersionFile` makes the code do what the log says, and it already emits its own warning.

```suggestion
          // version file is missing but tool allows this, so restore it and preserve the installation
          restoreMissingVersionFile(installationPath, resolvedVersion);
          return createToolInstallation(installationPath, resolvedVersion, false, processContext, additionalInstallation);
```

cli/src/main/java/com/devonfw/tools/ide/tool/python/Python.java
```diff
@@ -96,20 +102,15 @@ protected boolean isIgnoreMissingSoftwareVersionFile() {
...
-      if (version == null) {
-        version = readVersionFromInterpreter(toolPath);
-      }
-      if (version != null) {
-        LOG.debug("Determined version {} of python from the installation at {}.", version, toolPath);
-      }
+      version = readVersionFromInterpreter(toolPath);
+    }
+    if (version != null) {
+      Log.debug("Determined version {} of python from the installation of {}. ", version, toolPath);
```
quando632: This uses the `org.jline.utils.Log` imported at line 9, whose `debug(Object...)` signature concatenates its arguments without `{}` substitution and bypasses the slf4j `LOG` of this class, so the braces are logged literally. Switching to `LOG` renders the message correctly, and the import at line 9 can then go.

```suggestion
      LOG.debug("Determined version {} of python from the installation at {}.", version, toolPath);
```

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
```diff
@@ -414,12 +393,54 @@ protected EditionAndVersion computeInstalledEditionAndVersion() {
       return null;
     }
     // Resolve edition and version from a single tool-path lookup (one pass) instead of two separate lookups.
-    String edition = getInstalledEdition(toolPath);
+    Path repoPath = getInstalledSoftwareRepoPath(toolPath, true);
+
+    if (repoPath != null) {
+      VersionIdentifier version = VersionIdentifier.of(repoPath.getFileName().toString());
```
quando632, line 399: The version now comes from the repo folder name instead of the version file, so `ide get-version` reports a version for a software-repo installation whose `.ide.software.version` is missing. Was that intentional, or should the version file stay the primary source here?

JoelAdbu: I believe this is intentional based on Jörg's requirement:

- Now we should try to resolve a potential link (see implementation of getInstalledEdition) - if we can resolve such link and it points to our software repository, we can determine both the edition and the version from it in one go. Example $IDE_HOME/software/«tool» -> $IDE_ROOT/_ide/software/default/«tool»/«edition»/«version»[/...]

So my understanding is that for software-repository installations both edition and version are supposed to be derived from the resolved repository path.

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
JoelAdbu: Good catch. During cleanup I missed the call to restoreMissingVersionFile(). The existing tests did not fail because this specific scenario was not covered yet. I fixed the implementation and added a test case that verifies a missing .ide.software.version file is actually recreated while preserving the installation.

JoelAdbu: Correction: simply calling restoreMissingVersionFile(installationPath, resolvedVersion) here would restore the requested version, not necessarily the version actually present on disk.
The actual installed version is already restored centrally by computeInstalledEditionAndVersionFromLocalSoftwareFolder() when it can be determined. Reaching this branch means that it could not be determined, so the installation should be considered broken and the regular reinstallation flow should continue. I have updated the implementation and the corresponding test accordingly.
I have also added dedicated tests for the central restore path and the reinstallation fallback, so this behavior is now covered directly instead of only via PythonTest

quando632 (approved):
Nice work 👍 It looks good to me now. You only need to resolve the conflicts then it should be ready to merge

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
hohwille: Yes, confirmed. The `.ide.software.version` file is more or less just a hack we introduced to easily recover the installed tool version again.
Since we have established our software repo and already determine the edition from the path in the software repo, it seems consistent to do the same also for the version.
The `.ide.software.version` file becomes obsolete in such case.
If we have software that is not installed via software repo, we still rely on this `.ide.software.version` file,

cli/src/main/java/com/devonfw/tools/ide/tool/LocalToolCommandlet.java
hohwille, line 547: Sorry to add extra stuff for legacy code you never changed but since this is already a cleanup ticket, I would like to avoid creating yet another cleanup ticket for it:
We have extracted a `FileAccess.findAncestor` method exactly for this purpose.
When we created this, my intention was that we replace all "copy&pasted" variants of this with the reuse via that method.
Can you please check if we can replace all this method content with a simple call to `this.context.getFileAccess().findAncestor(installPath, softwareRepoPath, 4)`?

hohwille (approved):
@JoelAdbu thanks for your PR. This is a nice cleanup and now the code looks much better to me. Great job 👍

CHANGELOG.adoc
```diff
@@ -6,6 +6,7 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/2422[#2422]: Cleanup how to determine installed edition and version and restore software version file
```
hohwille: Cleanup stories with internal refactoring have no relevance for end users.

```suggestion
```

hohwille (approved):
@JoelAdbu thanks for your PR. This is looking great and what I was hoping for when I started the idea of this general refactoring of getting the installed edition and version. Great job 👍 Ready for merge.

JoelAdbu:
Successfully retested with 2026.09.002-09_22_07-SNAPSHOT on Windows:

- ide install python installs Python successfully
- Deleted software/python/.ide.software.version
- Re-ran ide install python
- Missing version file was detected and restored correctly
- Existing Python installation was recognized (Version 3.14.7 of tool python is already installed)
- No Deleting corrupted installation message appeared
- ide python --version still returns Python 3.14.7

PR #2447, opened 2026-09-11
#2420: Improve BuildTool support further

### LINKED ISSUE #2420: Improve BuildTool support further
### Task summary

This is a follow up of issue #2178 and PR #2302.
Here we want to address the following improvements:
* Change signature of `CommandletManager.findBuildTool` to return `BuildTool` instead of `LocalToolCommandlet`
* Enrich `BuildTool` to have the required API so either we need to make it a class and solve our typical single-inheritance issue for classes or extract interface API from Commandlet and LocalToolCommandlet and make `BuildTool` extend these new interfaces - whatever seems more appropriate and easy.
* Make all build tools implement/extend `BuildTool` - currently only maven does that but also Gradle, Npm, and Yarn need to be instances of `BuildTool`. To make things easy, you can make the methods of `BuildTool` to throw an appropriate `UnsupportedOperationException` so they do not immediately have all to be implemented and you can even create follow up issues or sub-tasks for that keeping your PR smaller.
* Change `CommandletManagerImpl.findBuildTool` to iterate over all commandlets and check for instanceof `BuildTool` instead of having a hard-coded list of build-tools.
* However, for that to work properly we somehow have to either ensure that Yarn is tested before Npm (or we need to somehow violate SoC and include knowledge about Yarn and `yarn.lock` into `Npm` but that might also be a pragmatic solution to avoid overengineering this with approaches like a priority method or annotation, but also keeping the order of registration in the CommandletManager together with a comment that Yarn has to be registered before Npm would be fine).
* The method `findBuildDescriptor` should be removed from `LocalToolCommandlet` and moved to `BuildTool` being abstract forcing build tools to implement this method.
* Complete the cleanup by avoiding casts to `BuildTool` outside CommandletManager and according error handling (see ReleaseCommandlet)

### Additional context

_No response_

laert-ll:
### Implemented changes:

* `CommandletManager.findBuildTool` now returns `BuildTool` instead of `LocalToolCommandlet`.
*  Replaced the hard-coded `BUILD_TOOLS` list in `CommandletManagerImpl` with dynamic iteration over all commandlets using `instanceof BuildTool` pattern matching.
* Changed `commandletTypeMap` from `HashMap` to `LinkedHashMap` and moved `Npm` registration after `Yarn` so registration order decides detection priority (`Yarn` before `Npm`).
* Removed `findBuildDescriptor` and `findWrapper` from `LocalToolCommandlet` (moved to `BuildTool`)
* Made `Gradle`, `Npm`, and `Yarn` implement `BuildTool`; `Yarn.findBuildDescriptor` only matches when `yarn.lock` is present and falls back to `Npm` otherwise.
* Updated `BuildCommandlet` to use `BuildTool` directly instead of casting through `LocalToolCommandlet`.
* Updated `ReleaseCommandlet` to catch `UnsupportedOperationException` from `getProjectVersion`/`setProjectVersion` and rethrow as a `CliException` with a clear "does not support releasing" message.

---

### Testing instructions

Test projects for points 1 - 5 can be found under `cli/src/test/resources/ide-projects/build/project/workspaces/main/` and for points 6 - 8 under `cli/src/test/resources/ide-projects/release/project/workspaces/main/`.

1. In a Maven project (has pom.xml): run ide build and verify it invokes mvn.
2. In a Gradle project (has build.gradle): run ide build and verify it invokes gradle.
3. In a Node project with only package.json: run ide build and verify it invokes npm.
4. In a Node project with both package.json and yarn.lock: run ide build and verify it invokes yarn, not npm.
5. In a Maven project that also has package.json (polyglot): run ide build and verify it invokes mvn, not npm.
6. In a Gradle project: run ide release and verify it fails with a clear message containing "does not support releasing" rather than a stack trace or cryptic error.
7. In an npm/yarn project: same, it should fail with the same readable message.
8. In a Maven project: run ide release and verify it proceeds normally (no regression).

---

### Checklist for this PR

  with `internal`


cli/src/main/java/com/devonfw/tools/ide/commandlet/ReleaseCommandlet.java
```diff
@@ -57,11 +53,16 @@ protected void doRun() {
...
+    if (!this.context.isForceMode() && !isTopLevelProject(buildTool, projectPath)) {
       throw new CliException("Release has to be performed from the top-level project or using force option.");
     }
 
-    VersionIdentifier currentVersion = buildTool.getProjectVersion(projectPath);
+    VersionIdentifier currentVersion;
+    try {
+      currentVersion = buildTool.getProjectVersion(projectPath);
+    } catch (UnsupportedOperationException e) {
+      throw new CliException("The build tool " + buildTool.getName() + " detected in " + projectPath + " does not support releasing.");
```
majesteSil, line 64: The caught exception is lost here. Per documentation/contributing/coding-conventions.adoc ("Catching and handling Exceptions"), when wrapping and re-throwing the caught exception must be passed as the cause:

 throw new CliException("...", e);

cli/src/main/java/com/devonfw/tools/ide/tool/BuildTool.java
```diff
@@ -11,25 +15,72 @@
  */
 public interface BuildTool {
 
+  Logger LOG = LoggerFactory.getLogger(BuildTool.class);
```
majesteSil, line 18: Having a Logger field in this interface is a bit unusual.  the coding conventions ("Logging") recommend using this.context instead of a LOG field for commandlets. If LOG is only used by the findWrapper default method, could we log via the passed-in IdeContext instead?

cli/src/main/java/com/devonfw/tools/ide/tool/BuildTool.java
```diff
@@ -11,25 +15,72 @@
...
    * Performs a single build-and-deploy (release) build of the project in the current working directory.
    *
    * @param additionalArgs the additional arguments to append to the build command (may be {@link List#isEmpty() empty}).
    * @return the {@link ProcessResult} of the build (allowing the caller to react on {@link ProcessResult#isSuccessful() failures}, e.g. by retrying).
    */
-  ProcessResult buildAndDeploy(List<String> additionalArgs);
+  default ProcessResult buildAndDeploy(List<String> additionalArgs) {
+
+    throw new UnsupportedOperationException();
+  }
```
majesteSil, line 85: This is allowed by the issue, but using UnsupportedOperationException as a capability flag is exception-as-control-flow (callers must catch it). For a follow-up, a nullable query (e.g. tryGetProjectVersion) or an explicit capability check would read more naturally.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CommandletManagerImpl.java
```diff
@@ -106,7 +104,7 @@ public CommandletManagerImpl(IdeContext context) {
 
     super();
     this.context = context;
-    this.commandletTypeMap = new HashMap<>();
+    this.commandletTypeMap = new LinkedHashMap<>();
```
majesteSil, line 107: With HashMap → LinkedHashMap, the iteration order of getCommandlets() / commandletIterator now changes globally.  this also affects command completion and findCommandlet. I assume the insertion order is required for the Yarn-before-Npm priority, but it is a side effect beyond the build tools. Also, the "Yarn must be registered before Npm" constraint now only lives in the CommandletManager javadoc.  a short comment at the registration site (add(new Yarn(...)); add(new Npm(...));) would keep it visible where it is enforced.

cli/src/main/java/com/devonfw/tools/ide/tool/gradle/Gradle.java
```diff
@@ -49,7 +50,7 @@ public String getToolHelpArguments() {
   @Override
   protected void configureToolBinary(ProcessContext pc, ProcessMode processMode) {
     Path gradle = Path.of(getBinaryName());
-    Path wrapper = findWrapper(GRADLE_WRAPPER_FILENAME);
+    Path wrapper = findWrapper(this.context.getCwd(), GRADLE_WRAPPER_FILENAME);
```
majesteSil, line 53: With the new interface default, doesn't this override become a no-op (return null;)? If it adds nothing, it could be removed.

majesteSil:
thank you @laert-ll  for your work on this there a some point where i see some improvments so let me know if you have any further Question on this

majesteSil (changes requested):

cli/src/main/java/com/devonfw/tools/ide/commandlet/ReleaseCommandlet.java
hohwille: We should create issues for all cases where we never implemented these method and throw `UnsupportedOperationException`.
When you have done this, remove throwing `UnsupportedOperationException` and return `null` instead and remove this try catch.

cli/src/main/java/com/devonfw/tools/ide/tool/BuildTool.java
```diff
@@ -11,25 +15,72 @@
...
+   * @param wrapperFileName the name of the wrapper file.
+   * @return the {@link Path} to the wrapper file or {@code null} if none was found.
+   */
+  default Path findWrapper(Path cwd, String wrapperFileName) {
+
+    Path dir = cwd;
+    while ((dir != null) && (findBuildDescriptor(dir) != null)) {
+      Path wrapper = dir.resolve(wrapperFileName);
+      if (Files.exists(wrapper)) {
+        LOG.debug("Using wrapper: {}", wrapper);
```
hohwille: It seems this was the reason for exposing the LOG as public API.
However, this method itself is more a utility for than an API of a `BuildTool`.
Simply move it to a class and make it static so it can be reused to resolve both the review issues in this file.

cli/src/main/java/com/devonfw/tools/ide/tool/BuildTool.java
hohwille: I also agree. Also this should not be the contract of this interface:
If I implement future a `BuildCommandlet`, I should get compile errors for all these methods forcing me to implement them and to think about their implementation.
The fact that we have not yet implemented this properly for all tools is just a temporary and unclean state but should not become part of the API contract.

CHANGELOG.adoc
```diff
@@ -6,6 +6,7 @@ This file documents all notable changes to https://github.com/devonfw/IDEasy[IDE
 
 Release with new features and bugfixes:
 
+* https://github.com/devonfw/IDEasy/issues/2420[#2420]: Improve BuildTool support further
```
hohwille: IMHO this is only an internal cleanup but the end-user will not notice.


```suggestion
```

hohwille (approved):
@laert-ll thanks for your PR. Getting close to our goal👍 
When there are review threads open from team-review, they should be resolved by the author of the PR.
Please address all review threads before we can merge this.

cli/src/main/java/com/devonfw/tools/ide/commandlet/CommandletManagerImpl.java
laert-ll: Thanks for pointing that out, the change was intentional. Since HashMap's order is not deterministic, I switched to LinkedHashMap to make the order explicit and at the same time yarn gets picked before npm.

laert-ll:
Thank you @majesteSil and @hohwille for your feedback. I addressed all the review comments and now it should be ready. I also opened the issues for release support for the remaining build tools (#2550).

PR #2465, opened 2026-09-11
#2423: isRepositoryUpdateAvailable is not covered by any test that reaches the comparison

### LINKED ISSUE #2423: isRepositoryUpdateAvailable is not covered by any test that reaches the comparison
### Task summary

`GitContextImpl.isRepositoryUpdateAvailable(Path)` first guards on `hasUpstream(repository)` and only then compares `rev-parse HEAD` against `rev-parse @{u}`. That comparison is currently never executed in a test.

The only test targeting the method, `GitContextTest.testIsRepositoryUpdateAvailableNoUpdates`, configures no output on the `ProcessContextGitMock`. The mock has no special handling for `rev-parse` or `config`, so `determineCurrentBranch(...)` returns null, `hasUpstream(...)` fails its first guard, and the method returns early. The test asserts `false` and passes, but for the wrong reason: no upstream is simulated, rather than local and remote being equal as its javadoc claims. The `errors` and `outs` lists it builds, including the two identical hashes and the comment "same as remote to simulate no updates", are never passed anywhere.

`GitContextMockTest` looks like it covers this but does not: `GitContextMock` overrides the method with a different implementation that compares `.git/FETCH_HEAD` against HEAD, without `hasUpstream` and without `rev-parse`. Those assertions exercise the test double, not the production code.

Acceptance criteria:

1. A test that simulates branch and upstream configuration plus two identical commit ids and asserts `false`, actually reaching the comparison.
2. A test with two differing commit ids that asserts `true`. No such case exists today.
3. The unused `errors` and `outs` lists are removed from the existing test, and its javadoc matches what it verifies.

### Additional context

Split out of #892, which reported three weak tests in `GitContextTest`. Two of them (`testFetchIfNeededOffline`, `testFetchIfNeededRecentFetchHead`) assert meaningfully by now, so only this gap remains, and it differs from the original description.

The pattern for feeding the mock is already used in the same file at lines 96, 115 and 168:

```java
this.processContext.addOutputMessage(new OutputMessage(false, "test-remote"));
```

One thing to expect while implementing: `ProcessContextGitMock` returns output from a flat list without distinguishing by command, so simulating two `rev-parse` calls with different answers may need a small extension of the mock.

JoelAdbu:
### Implemented changes:

* enhanced `ProcessContextGitMock` to support sequential Git outputs by adding a per-command FIFO (`addCommandOutput`), allowing multiple git invocations within a single test flow to return distinct results
* updated `testIsRepositoryUpdateAvailableNoUpdates` to simulate a configured upstream and identical local/remote commit IDs, ensuring the actual commit comparison is reached
* added `testIsRepositoryUpdateAvailableUpdates` to cover the case where local and remote commit IDs differ and an update is available
*  added additional edge case tests covering missing upstream configuration and `null` commit IDs
* removed unused test data (`errors` / `outs`) and updated the Javadocs to accurately reflect the tested behavior

---

### Testing instructions

1. Open `GitContextTest`
2. Run the newly added tests with coverage.
3. Verify all test pass successfully and verify that `GitContextImpl.isRepositoryUpdateAvailable(Path)` is covered by the test execution.

---

### Checklist for this PR

  with `internal`


cli/src/test/java/com/devonfw/tools/ide/git/GitContextTest.java
```diff
@@ -275,18 +273,31 @@ void testFetchIfNeededRecentFetchHead(@TempDir Path tempDir) throws IOException
...
+   * {@code hasUpstream} guard and reach the commit id comparison.
+   *
+   * @param localCommitId the commit id returned for {@code rev-parse HEAD}.
+   * @param remoteCommitId the commit id returned for {@code rev-parse @{u}}.
+   */
+  private void simulateUpstreamAndCommitIds(String localCommitId, String remoteCommitId) {
+    this.processContext.addCommandOutput(new OutputMessage(false, "master"));
+    this.processContext.addCommandOutput(new OutputMessage(false, "origin"));
+    this.processContext.addCommandOutput(new OutputMessage(false, "refs/heads/master"));
+    this.processContext.addCommandOutput(new OutputMessage(false, localCommitId));
```
quando632: These two tests currently pass for an accidental reason rather than through the failure handling they target: `new OutputMessage(false, null)` produces `exit=0, out=[null]`, so `runGitCommandAndGetSingleOutput` takes the `size == 1` path and returns null only because the queued string is null. Mutation-verified: replacing that method's final `return null` with `return ""` keeps all 16 tests green, so its failure path stays uncovered.

A missing commit id in reality means `git rev-parse` exits with 128, which `ProcessContextGitMock` cannot express yet. Adding `addCommandFailure(int exitCode, String errorMessage)` to the mock and queueing that when `commitId == null` drives the real path (verified locally: it kills the mutation above and the full `cli` suite stays green). Non-blocking, the PR is a clear improvement either way.

quando632 (commented):
Nice catch on the root cause: the old `testIsRepositoryUpdateAvailableNoUpdates` built two `List`s that never reached the mock, so the commit comparison was never executed. The sequential output queue fixes that, and both the no-update and the update path are now genuinely covered. One non-blocking note inline on the "commit id missing" tests.

cli/src/test/java/com/devonfw/tools/ide/git/GitContextTest.java
JoelAdbu: Good catch, thanks! I overlooked that. I've now implemented the suggested approach and updated the tests to exercise the actual failure handling path.

quando632 (approved):
Nice now everything looks good for me. You only need to update the branch with main, then we can move this to In-Review :)

PR #2479, opened 2026-09-15
#2471: install VSCode plugins after the IDE has created the profile

### LINKED ISSUE #2471: VSCode plugin installation fails with "Profile '<name>' not found" when VSCODE_PROFILE_ENABLED is enabled
### Expected behavior

`ide vscode` installs all configured VSCode plugins and then launches VSCode.

### Actual behavior

When `VSCODE_PROFILE_ENABLED` is `true`, **every** plugin installation fails with `Profile '<name>' not found.` and exit code 1. In my project all 21 plugins failed in a row:

```
Start: Install plugin git-graph (1/21)
Running command '...\vscode\1.137.0\bin\code.cmd' with arguments '--new-window' '--profile=ideasy-ai-main'
  '--extensions-dir=D:\projects\ai\plugins\vscode' 'D:\projects\ai\workspaces\main' '--force'
  '--install-extension' 'mhutchie.git-graph'
failed with exit code 1!
Profile 'ideasy-ai-main' not found.
com.devonfw.tools.ide.cli.CliProcessException: ...
        at com.devonfw.tools.ide.tool.vscode.Vscode.installPlugin(Vscode.java:66)
        at com.devonfw.tools.ide.tool.plugin.PluginBasedCommandlet.doInstallPluginStep(PluginBasedCommandlet.java:233)
        ...
        at com.devonfw.tools.ide.tool.plugin.PluginBasedCommandlet.postInstall(PluginBasedCommandlet.java:131)
        at com.devonfw.tools.ide.tool.LocalToolCommandlet.doInstallStep(LocalToolCommandlet.java:128)
Step 'Install plugin git-graph (1/21)' ended with failure.
```

### Root cause

`--profile` only auto-creates a profile on the *open* code path. From `code --help`:

> `--profile <profileName>`  Opens the provided folder or workspace with the given profile and associates the profile with the workspace. **If the profile does not exist, a new empty one is created.**

As soon as `--install-extension` (or `--list-extensions` / `--uninstall-extension`) is passed, VSCode takes the extension-management code path instead: the folder argument is ignored, no window is opened, and the profile is only *looked up*. If it does not exist, VSCode aborts with `Profile '<name>' not found.` and exit code 1.

IDEasy installs the plugins from `postInstall()` → `installPlugins()`, i.e. **before** VSCode is ever launched (see the stack trace above). So on a fresh profile the plugin installation can never succeed — the profile that `Vscode.getProfileName()` computes does not exist yet, and none of the plugin calls create it.

### Steps to reproduce

Deterministic, verified against VSCode 1.137.0 on Windows 11. Reproducible directly against the VSCode CLI, no IDEasy build required:

```bash
# 1) exactly the call IDEasy makes - profile does not exist yet
code.cmd --new-window --profile=ideasy-repro-main --user-data-dir=<tmp> \
         --extensions-dir=<tmp-ext> <folder> --force --install-extension mhutchie.git-graph
# -> Profile 'ideasy-repro-main' not found.      exit=1

# 2) open VSCode once with the same profile (what IDEasy does at the very end of `ide vscode`)
code.cmd --new-window --profile=ideasy-repro-main --user-data-dir=<tmp> \
         --extensions-dir=<tmp-ext> <folder>
# -> exit=0, profile is created

# 3) repeat the exact call from step 1
# -> Extension 'mhutchie.git-graph' v1.30.0 was successfully installed.   exit=0
```

Read-only variant, installs nothing:
`code.cmd --profile=<any-non-existing-name> --list-extensions` → `Profile '<name>' not found.`, exit 1.

Via IDEasy:
1. Set `VSCODE_PROFILE_ENABLED=true`.
2. Make sure the profile `ideasy-<project>-<workspace>` does not exist yet (fresh setup, or profile deleted).
3. Run `ide vscode` → all plugins fail.
4. Run `ide vscode` a second time → all plugins install fine, because step 3 created the profile when it finally opened the window.

### Why this slipped through review and tests

It is a **first-run-only** bug. After `ide vscode` has opened the window once, the profile exists and every following run works. Anyone testing the feature twice never sees it again. It hits every new project setup and everyone who deletes their profile.

`VscodeTest` only asserts on the generated argument list, so it cannot catch this — the arguments are correct, VSCode's behavior for those arguments is the problem.

### Affected versions

Introduced by #2058 / PR #2072 (commit `09c20b86`, merged 2026-09-11) in `Vscode.configureToolArgs()`.
**Not released yet** — #2058 is not in `CHANGELOG.adoc` (latest entry is 2026.09.002), so this would reach users with the upcoming release. Should be fixed before the release.

### Suggested directions

- Create the profile explicitly before installing plugins (e.g. one `--profile=<name>` open/no-op call, or by creating the profile entry up front), instead of relying on a side effect of the launch path.
- Alternatively install plugins *after* the IDE launch, or fall back to the `--user-data-dir` path for plugin installation.
- Independently: `VSCODE_PROFILE_ENABLED` defaults to `Boolean.FALSE` in `IdeVariables` but is missing from the `VARIABLES` collection, so it is not listed by `ide env`/documentation tooling. Worth fixing together with this.

### Related

- #2058 / PR #2072 — introduced the `--profile` handling

quando632:
### Implemented changes:

* `Vscode` no longer installs the plugins before VS Code has created its profile.

VS Code only creates a named profile when it *opens a window*. From `code --help`:

> `--profile <profileName>`  **Opens** the provided folder or workspace with the given profile […] If the profile does not exist, a new empty one is created.

As soon as `--install-extension` (or `--list-extensions` / `--uninstall-extension`) is passed, VS Code takes the extension management code path instead: the folder argument is ignored, no window is opened, and the profile is only *looked up*. If it does not exist, VS Code aborts with `Profile '<name>' not found.` and exit code 1.

Since the plugins are installed from `postInstall()`, i.e. before VS Code is ever launched, every plugin installation failed on a fresh profile while `VSCODE_PROFILE_ENABLED=true`.

`installPlugins(...)` now checks whether the profile already exists. If it does not, the plugins are deferred and installed after the IDE has been launched, once the profile shows up (checked once per second for at most 20 s, with an info message while IDEasy waits). The profile is only considered missing if VS Code names it on stderr (`Profile '<name>' not found.`). Any other failure of VS Code is reported with its output instead: before the launch the plugins are installed as usual so the real error surfaces, after the launch the wait stops immediately. If the profile already exists, every run after the first, so the established order is kept, so the plugins are active right away.

* The plugins are installed into a newly created profile even if their marker files already exist.

The plugin marker files live per IDEasy project while the plugins of VSCode belong to a profile. A profile that VSCode has just created is empty, so the marker files cannot tell whether a plugin is present in it. Enabling the toggle in an existing project therefore left the user with an IDE without a single plugin: the new profile is empty, but every plugin was skipped as already installed, IDEasy reported success and asked for a window reload that changed nothing.

`PluginBasedCommandlet` now offers `isForcePluginInstallation()` as an extension point for the marker file check and reports how many plugins were installed. `Vscode` overrides it while installing into a profile it has just created, and only asks for the reload when something was actually installed. The default behaviour of all other IDE commandlets is unchanged.

Also corrected a code comment claiming that a profile isolates authentication. It does not: VS Code keeps auth sessions in the OS keyring and shares them across all profiles (see the analysis in #2058).

---

### Testing instructions

**Verified end to end** against a real VS Code 1.137.0 on Windows 11, with `VSCODE_PROFILE_ENABLED=true`, a profile that did not exist yet and no plugin marker files. The observed order of `code.cmd` invocations was:

1. `--list-extensions` — fails, profile does not exist, plugin installation is deferred
2. launch without extension arguments — VS Code opens and creates the profile
3. `--list-extensions` — still failing on the first poll
4. `--list-extensions` — succeeds on the second poll
5. `--force --install-extension …` for every plugin

Result: **39 of 39 plugins installed, 0 failures.** Before this change the same scenario failed 39 out of 39 times with `Profile '...' not found.`

A second run was done with the marker files deliberately left in place, which is what a user enabling the toggle in an existing project actually has: **38 plugins installed, 0 failures, 0 skipped because of a marker file**, and the freshly created profile ended up with 62 extensions. Before this change that exact scenario produced a profile with **0** extensions while IDEasy reported success.

Step 4 below was not exercised in those runs; the "profile already exists" path was only observed separately in another workspace.

1. Automated tests:

```bash
mvn -pl cli test -Dtest=VscodeTest
```

Six cases are covered:
* `testPluginsAreInstalledBeforeLaunchByDefault`: toggle disabled, established order unchanged
* `testPluginsAreInstalledAfterLaunchIfProfileEnabled`: toggle enabled, profile missing, plugins installed after the launch
* `testPluginsAreInstalledBeforeLaunchIfProfileAlreadyExists`: toggle enabled, profile present, established order kept
* `testPluginsAreInstalledIntoNewProfileDespiteExistingMarkerFiles`: plugins installed into the new profile although their marker files exist
* `testPluginsAreNotDeferredIfVscodeFailsForAnotherReason`: VS Code fails for another reason, plugins are not deferred and the cause is reported
* `testDeferredPluginsAreSkippedIfVscodeFailsAfterLaunch`: VS Code fails after the launch, waiting stops on the first unexpected failure

2. Reproduce the underlying VS Code behaviour without IDEasy (any VS Code, here 1.137.0):

```bash
# profile does not exist -> fails, no window is opened
code --profile=some-not-existing-name --list-extensions
# -> Profile 'some-not-existing-name' not found.   exit 1
```

3. End to end with a fresh profile:
   * set `VSCODE_PROFILE_ENABLED=true` in `$IDE_HOME/conf/ide.properties`
   * make sure the profile `ideasy-«project»-«workspace»` does not exist yet
   * keep your existing plugin marker files under `$IDE_HOME/.ide/plugin.vscode.*` — this is the realistic case when enabling the toggle in an existing project
   * run `ide vscode`

   Expected: VS Code starts, then all active plugins are installed and a message asks for a window reload. Verify that the new profile really contains them:

```bash
code --profile=ideasy-«project»-«workspace» --extensions-dir=$IDE_HOME/plugins/vscode --list-extensions
```

   Before this PR every plugin failed with `Profile '...' not found.`, and with the previous commit alone the profile would have stayed empty because of the marker files.

4. Run `ide vscode` a second time.

   Expected: unchanged behaviour, nothing is reinstalled, no reload message.

---

### Checklist for this PR

**On the two unchecked items:**

* *CHANGELOG*: deliberately omitted. The DoD states that a bugfix to a story that was never released does not need a CHANGELOG entry, and #2058 is still behind the feature toggle and intentionally not in the CHANGELOG until the pilot succeeds.


MeShehi:
Thanks for the PR, the tests are clearly passing so no issues there. I was wondering on the choice of waiting 30 times, 1 seconds each, is there a particular reason for this choice or is it just a "hunch"? I wonder if it could be lowered or if there is an argument to increasing it. Just a curiosity/nitpick, and also asking in case there are other similar cases in IDEasy. After the changelog addition I would still move to Reviews

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -82,24 +97,129 @@ public boolean installPlugin(ToolPluginDescriptor plugin, Step step, ProcessCont
...
+   */
+  private boolean isProfileAvailable(ProcessContext pc) {
+
+    // there is no CLI command to create or query a profile, but every profile bound call fails while the profile does not exist
+    try {
+      return runTool(pc, ProcessMode.DEFAULT_CAPTURE, List.of("--list-extensions")).isSuccessful();
+    } catch (CliProcessException e) {
+      LOG.debug("The VS Code profile {} does not exist yet.", getProfileName(), e);
+      return false;
+    }
```
laert-ll: This method runs `code --list-extensions` and checks if it failed, and in that case it treats every crash as   a case of the profile not existing yet, however the cause can be different things (a real crash, missing binary, broken VS Code).

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -82,24 +97,129 @@ public boolean installPlugin(ToolPluginDescriptor plugin, Step step, ProcessCont
...
+
+    return Boolean.TRUE.equals(IdeVariables.VSCODE_PROFILE_ENABLED.get(this.context));
+  }
+
+  @Override
+  protected boolean isForcePluginInstallation() {
+
+    // A profile that VSCode has just created is empty. Our plugin marker files live per project while the plugins of VSCode belong to a profile, so the marker
+    // files cannot tell whether a plugin is present in the new profile. Without this the user ends up with an empty IDE. See issue #2471.
+    return this.installingIntoNewProfile || super.isForcePluginInstallation();
```
laert-ll, line 123: `installingIntoNewProfile` is set true whenever plugins are being installed into a freshly created profile.

Suggestion: instead of reinstalling everything, could we check the actual plugin list VS Code has and only install what's missing?

laert-ll (commented):
Thanks for your work. I'd also add to what @MeShehi said about the 30 second wait, I think that might be a bit too much, especially if it stays hanging with no clear message to the user.

Otherwise looks good to me :)

quando632:
Thanks both for the review! @MeShehi, the 30 x 1 s was an upper bound picked by feel, not a measured value. In my end to end runs the profile showed up on the second check, roughly 2 s after launch. The count based loop could also take much longer than 30 s, since every check spawns `code --list-extensions`, which itself takes about a second. It is now a wall clock deadline of 20 s, and as @laert-ll suggested, the user gets an info message while IDEasy waits instead of a silent hang. There is no comparable wait loop elsewhere in IDEasy, `Thread.sleep` does not appear anywhere else in `cli/src/main/java`.

Regarding the CHANGELOG: I left it out on purpose, see the note at the end of the PR description. #2058 is still behind the `VSCODE_PROFILE_ENABLED` toggle and unreleased, and the DoD says a fix to an unreleased story needs no entry. Happy to add one if you see it differently.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
quando632: Good catch, agreed. The check now runs with its own process context without error handling and only treats a failure as a missing profile if VS Code's stderr names our profile (`Profile '<name>' not found.`). I match only the quoted profile name since the message itself may be localized. Any other failure is reported as a warning including VS Code's output. Before the launch the plugins are then installed as usual so the real error surfaces, and after the launch the wait stops immediately instead of polling until the timeout. This also fixes a side effect I had missed: every failed check used to be logged as an error, since the plugin process context uses `THROW_CLI`. Covered by `testPluginsAreNotDeferredIfVscodeFailsForAnotherReason` and `testDeferredPluginsAreSkippedIfVscodeFailsAfterLaunch`.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
quando632: I like the idea, but in this PR it would not change the outcome: the flag is only set while installing into a profile that VS Code has just created, and such a profile is empty, so every plugin is missing anyway. Checking the real extension list instead of the marker files would be more robust in general, for example when a user resets the profile or the profile creation times out while old marker files exist. However, it changes how `PluginBasedCommandlet` decides what is installed for all IDEs, including id matching and pinned versions. I would prefer to do that in a follow-up issue, if that is fine with you. Shall I open one?

PR #2490, opened 2026-09-16
#607: test that IDEA_PROPERTIES and STUDIO_PROPERTIES are set on the process context

### LINKED ISSUE #607: Test for environment variables of process contexts
With PR #578 we have added the functionality to set the `IDEA_PROPERTIES` and `STUDIO_PROPERTIES` to the process context that starts intellij and androidstudio respectively, instead of writing their values to `ide.properties` files at installation and reading them again when running the tools. 

We want to test for their presence in the respective intellij and androidstudio tests. There might be further tools setting environment variables at run time (e.g. due to evaluated dependencies), that may also be tested.  

In order to test this, we would need to mock the `ProcessContext`, run the corresponding tool and then verify that the environment variable was set in the mocked `ProcessContext`.

majesteSil:
This PR fixes #607

### Implemented changes:

* Added a test `IntellijTest#testSetEnvironmentSetsIdeaProperties` verifying that `setEnvironment` puts `IDEA_PROPERTIES` (pointing to `idea.properties` in the workspace) into the environment context.
* Added a test `AndroidStudioTest#testSetEnvironmentSetsStudioProperties` verifying that `setEnvironment` puts `STUDIO_PROPERTIES` (pointing to `studio.properties` in the workspace) into the environment context.

Both use a `RecordingEnvironmentContext` to capture the variables set on the process context, as requested in the issue.

---

### Testing instructions

Run the two new tests in the `cli` module:

1. `mvn -Dtest='IntellijTest#testSetEnvironmentSetsIdeaProperties' test`
2. `mvn -Dtest='AndroidStudioTest#testSetEnvironmentSetsStudioProperties' test`

Both pass (verified locally, BUILD SUCCESS).

---

### Checklist for this PR

- [x] You have added the issue implemented by your PR in [CHANGELOG.adoc](https://github.com/devonfw/IDEasy/blob/main/CHANGELOG.adoc) unless issue is labelled with `internal` *(issue #607 is labelled `internal`, so no entry)*
- [x] You have not changed any dependency in `pom.xml` files


laert-ll:
Thanks for your work, the tests are clear and pass for me locally.

One suggestion I would add is to add a test that starts the tool. The tests as they are written now call `setEnvironment` directly. If I understand the issue correctly, we need to start the tool and then check the variable on the process.

My suggestion would be to change the mock scripts so they write the variable to a file, e.g. `echo "$IDEA_PROPERTIES" > ideaproperties`. Then the existing `testIntellijRun` and `testAndroidStudioRun` can check that file.

majesteSil:
Thanks for the review @laert-ll !
Your reading of the issue is fair, and I want to make the trade-off i made explicit:

1. The variable is defined in one place, and that's what we test. IDEA_PROPERTIES is set only in Intellij.setEnvironment() and STUDIO_PROPERTIES only in AndroidStudio.setEnvironment(),  the only place a realistic regression can occur (wrong key/path). The tests assert the exact value against the workspace path.

2. The process bridge is generic and already covered. ProcessContext extends EnvironmentContext, and run() builds the process via setEnvironmentOfInstalledTools(pc) and hands that same context to the ProcessBuilder. The same plumbing that carries PATH, IDEA_HOME, etc. The existing ...Run tests already prove variables reach the spawned process (mock scripts write to a file, tests assert it). If the bridge broke, every variable on every tool would break and those tests would catch it.

So the unit tests pin the tool-specific definition (the real risk), and the generic plumbing is left to the existing run coverage, the same pattern ClaudeTest and SpyderTest use.

If you  see the process-level assertion as part of #607's acceptance criteria, I'm glad to add it the way you suggested (mock script writes echo "$IDEA_PROPERTIES" > ideaproperties, testIntellijRun/testAndroidStudioRun assert the file).

laert-ll:
Thanks for the explanation, I see now what you mean. I'll push it to In Reviews then, looks good to me!

hohwille (approved):
@majesteSil thanks for your PR and improving the tests 👍 Ready for merge.

Issue #2174, opened 2026-07-20
VSCode: store metadata under $IDE_HOME/.ide/vscode/<workspace> + migration


quando632:
### Feature idea

As a VSCode user, I want VSCode's user-data to be stored outside of my workspace under `$IDE_HOME/.ide/vscode/«workspace»/config` so that my workspace stays clean and independent of the IDE being used.

### Additional context

Part of #2142. Builds on #2172 (`getIdeMetadataPath()`).

* `Vscode.configureToolArgs`: point `--user-data-dir` to `getIdeMetadataPath().resolve("config")` instead of `«workspace»/.vscode/.userdata`. The opened folder (`getWorkspacePath()`) stays unchanged. This is a clean, flag-based relocation without the `.idea` convention problem of the JetBrains IDEs.
* **Migration**: move any existing `.vscode/.userdata` of every workspace to the new location (idempotent), registered in `IdeMigrator`.

`IdeMigrator` requires strictly ascending migration versions — coordinate the migration version with #2173 depending on merge order.

PR #2526, opened 2026-09-21
#2509: merge VSCode user settings template into --user-data-dir

### LINKED ISSUE #2509: VSCode user settings from the settings repository are ignored since user-data moved to $IDE_HOME/.ide
### Actual behavior

Since #2174, which will ship with the upcoming release 2026.09.002, the user level VSCode settings provided by the settings repository are not applied. This happens with the default configuration, i.e. without `VSCODE_PROFILE_ENABLED`, and silently: no error or warning is shown.

With the default [devonfw/ide-settings](https://github.com/devonfw/ide-settings/blob/main/vscode/workspace/update/.vscode/.userdata/User/settings.json) this means:

* **Telemetry is enabled again for every IDEasy user.** The template disables 7 telemetry settings (VSCode, crash reporter, GitLens, Red Hat, REST Client, Code Runner, GitHub Pull Requests); VSCode's default is to send telemetry.
* **`update.mode: none` is not applied**, so VSCode checks for updates on its own again although IDEasy manages its version. Whether it can also update the installation managed by IDEasy was not checked.

Project specific settings repositories are affected the same way, e.g. formatter configuration placed in this file.

### Expected behavior

The settings from `vscode/workspace/*/.vscode/.userdata/User/settings.json` in the settings repository are active in VSCode, as they were before #2174.

### Root cause

#2174 moved VSCode's `--user-data-dir` from `«workspace»/.vscode/.userdata` to `$IDE_HOME/.ide/vscode/«workspace»/config` and migrated the existing folder once (`Mig202609002`).

The settings repositories still provide the template as `.vscode/.userdata/User/settings.json`. The workspace merge mirrors the template structure, so it keeps writing to `«workspace»/.vscode/.userdata/User/settings.json`, a file VSCode no longer reads. Nothing in the code maps this path to the new location.

As a side effect `.vscode/.userdata` is created in every workspace again, which is exactly what #2142 intended to avoid.

### Steps to reproduce

1. Create a new workspace and run `ide vscode` in it with the default configuration.
2. Compare
   * `«workspace»/.vscode/.userdata/User/settings.json` — written by the IDEasy merge
   * `$IDE_HOME/.ide/vscode/«workspace»/config/User/settings.json` — read by VSCode, as passed via `--user-data-dir`

Verified on a fresh workspace with the current `main`: the merge writes all settings of the template (8 in the test project) to `«workspace»/.vscode/.userdata/User/settings.json`, creating `.vscode/.userdata` in the workspace; `$IDE_HOME/.ide/vscode/«workspace»/config/User/settings.json`, the file VSCode reads, does not exist. None of the settings from the settings repository are active.

### Suggested directions

* Map the template path `.vscode/.userdata` to `getIdeMetadataPath().resolve("config")` during the workspace merge. This keeps all existing settings repositories working unchanged.
* Changing the path in the settings repositories instead would break every custom settings repository.
* In both cases, remove the `.vscode/.userdata` folders the merge has created in the workspaces.

### Related

* #2174 / #2142 — moved the VSCode user-data out of the workspace
* #2058 — with `VSCODE_PROFILE_ENABLED` the user settings are not applied either, for a different reason: a named profile reads its own `settings.json` from `%APPDATA%\Code\User\profiles\«id»\`, which IDEasy does not write. Fixing this issue alone does not fix that case.

quando632:
### This PR fixes #2509

Since #2174 VSCode is started with `--user-data-dir=$IDE_HOME/.ide/vscode/«workspace»/config`, but the settings repositories still provide the user settings template as `vscode/workspace/*/.vscode/.userdata/User/settings.json`. The workspace merge kept writing it to `«workspace»/.vscode/.userdata`, which VSCode no longer reads. As a result, the user settings were silently ignored (e.g. telemetry was on again and `update.mode: none` was not applied), and `.vscode/.userdata` was recreated in every workspace.

### Implemented changes:

* `IdeToolCommandlet`: new hook `getWorkspaceRedirects(Path)` so an IDE can merge template sub-folders to a different location than the workspace. The default redirects nothing.
* `DirectoryMerger`: new `merge(..., Map<Path, Path> redirects)` overload that applies these redirects while traversing the templates. The existing `merge` delegates to it with an empty map.
* `Vscode`:
  * Redirects `«workspace»/.vscode/.userdata` to the user-data folder passed via `--user-data-dir`. Existing settings repositories keep working unchanged. All other `.vscode` templates are still merged into the workspace.
  * Before configuring the workspace, removes a leftover `«workspace»/.vscode/.userdata` folder:
    * If the user-data folder does not exist yet, the leftover is moved there.
    * Otherwise it is backed up to `$IDE_HOME/backups`, so nothing is deleted.
  * This also covers SNAPSHOT users who already ran `Mig202609002`, since a migration does not run twice.
* Tests in `VscodeTest` for the redirected merge and for both cleanup cases.

Not in scope: with `VSCODE_PROFILE_ENABLED` the user settings are still not applied (#2058, different cause).

No CHANGELOG entry: the bug was introduced by #2174, which has not been released yet (2026.09.002), so per DoD it is not listed separately.

---

### Testing instructions

1. Build IDEasy from this branch and use a project whose settings contain `vscode/workspace/update/.vscode/.userdata/User/settings.json` (e.g. the default [ide-settings](https://github.com/devonfw/ide-settings)).
2. Run `ide vscode` in a workspace.
3. Verify:
   * `«workspace»/.vscode/.userdata` is **not** created.
   * `$IDE_HOME/.ide/vscode/«workspace»/config/User/settings.json` contains the settings from the template.
   * In VSCode, `telemetry.telemetryLevel` is `off` and `update.mode` is `none`.
4. Create a folder `«workspace»/.vscode/.userdata` manually and run `ide vscode` again. The folder is removed from the workspace and appears under `$IDE_HOME/backups/…`.

---

### Checklist for this PR

  with `internal` (not needed: fixes an unreleased regression of #2174, see DoD)


cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -89,6 +95,50 @@ private String getProfileName() {
...
+  private void cleanupLegacyUserData() {
+
+    Path legacyUserData = this.context.getWorkspacePath().resolve(LEGACY_USER_DATA);
+    if (!Files.isDirectory(legacyUserData)) {
+      return;
+    }
+    FileAccess fileAccess = this.context.getFileAccess();
+    Path userData = getUserDataPath();
+    if (Files.exists(userData)) {
+      LOG.info("Removing obsolete VSCode user-data folder {} from workspace since VSCode uses {}", legacyUserData, userData);
```
hohwille: Should hopefully never happen but if it does, it should rather be a warning since something went wrong.
```suggestion
      LOG.warn("Removing obsolete VSCode user-data folder {} from workspace since VSCode uses {}", legacyUserData, userData);
```
But I agree with your backup strategy since we do not want to merge both directory tries in this edge-case.
However, if the user has done valuable settings for VScode over time they might be lost.
Since there is still a backup, he could still recover manually...

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -89,6 +95,50 @@ private String getProfileName() {
...
+    }
+  }
+
+  @Override
+  protected Map<Path, Path> getWorkspaceRedirects(Path workspaceFolder) {
+
+    // the settings still provide the user settings template in the legacy location inside the workspace
+    return Map.of(workspaceFolder.resolve(LEGACY_USER_DATA), getUserDataPath());
+  }
+
```
hohwille, line 141: The `LEGACY_USER_DATA` is a Path containing of 2 folders while your merger adoption seems to handle only the current filename.
I would rather fix this without changing DirectoryMerger by changing the 3 path arguments passed to it.
So you could simply change the visibility of this method and override it for VSCode and Intellij/IdeaBasedIdeCommandlet:
https://github.com/devonfw/IDEasy/blob/b9eb9b35685ceba1bda1ec63729776d0f21cc1e5/cli/src/main/java/com/devonfw/tools/ide/tool/ide/IdeToolCommandlet.java#L165

If we do not want to change the directory structure below `getIdeMetadataPath()`, you would need to keep the knowledge of the ".vscode" (and potentially also ".userdata") folder and navigating into that for the update/setup path. Otherwise things might remain even more simple if we just keep the old structure and only relocate from `getWorkspacePath()` to `getIdeMetadataPath()` or is that not possible with this `--user-data-dir` option and we still have other bugs?

hohwille (commented):
@quando632 thanks for quickly taking this up. 👍 I left some quick review feedback.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
quando632: Agreed, changed to `warn`.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
quando632: Good point to double-check. The nested path is handled: `DirectoryMerger` looks up the full target path on every level, so `«workspace»/.vscode` stays as is and only `«workspace»/.vscode/.userdata` is redirected. `VscodeTest.testConfigureWorkspaceMergesUserDataTemplateIntoUserDataFolder` covers this (verified locally: without the redirect it fails).

The reason for the redirect is that one template folder has to end up in two places. ide-settings ships both `.vscode/settings.json`, which VSCode reads from the workspace, and `.vscode/.userdata/User/settings.json`, which has to go to `--user-data-dir`. Relocating everything to `getIdeMetadataPath()` would break the workspace settings. Only adjusting the paths passed to the merger would mean walking the template folders manually to leave `.userdata` out. The existing `merge` signature stays unchanged.

For IntelliJ, `idea.properties` and `.intellij/config` are still written to and read from the workspace consistently, so nothing is broken there.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -102,7 +152,7 @@ protected void configureToolArgs(ProcessContext pc, ProcessMode processMode, Lis
       // already-running IDEasy window. Each project and workspace gets its own profile for isolated auth and settings.
       pc.addArg("--profile=" + getProfileName());
     } else {
-      pc.addArg("--user-data-dir=" + getIdeMetadataPath().resolve("config"));
+      pc.addArg("--user-data-dir=" + getUserDataPath());
```
QuangAnhLe: Line 155(default): VS Code is told to use a custom user-data dir at $`IDE_HOME/.ide/vscode/«workspace»/config. `Its user settings live in .`../config/User/settings.json.` ---> **Strategy B.**

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -102,7 +152,7 @@ protected void configureToolArgs(ProcessContext pc, ProcessMode processMode, Lis
       // already-running IDEasy window. Each project and workspace gets its own profile for isolated auth and settings.
       pc.addArg("--profile=" + getProfileName());
```
QuangAnhLe: Line 153( (opt-in, --profile): VS Code uses its default user-data dir (%APPDATA%\Code on Windows) and just selects a named profile inside it. The profile's settings live in a version-dependent internal path like `<default-user-data-dir>/User/profiles/ideasy-<proj>-<ws>/settings.json.` ---> **Strategy A**

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
```diff
@@ -89,6 +95,50 @@ private String getProfileName() {
     return "ideasy-" + this.context.getProjectName() + "-" + this.context.getWorkspaceName();
   }
 
+  /**
+   * @return the {@link Path} to the VSCode user-data folder passed via {@code --user-data-dir}.
+   */
+  private Path getUserDataPath() {
```
QuangAnhLe: The PR's redirect always points the .vscode/.userdata template at getUserDataPath(), which is hard-coded to the Strategy-B location. Right ?

So the template (telemetry off, update.mode: none, etc.) is always merged into `$IDE_HOME/.ide/.../config/User/settings.json.` Why that fixes the default but not --profile
In **Strategy B**( Line 155), VS Code actually reads `$IDE_HOME/.ide/.../config` → the merged settings take effect. Bug fixed.

But In **Strategy A**( Line 153) , VS Code reads its default user-data-dir and the profile subpath — it never looks at $IDE_HOME/.ide/.../config. So the template is merged into a folder VS Code ignores → user settings are still silently not applied. Same symptom as the ticket.

How widespread is it

`VSCODE_PROFILE_ENABLED` defaults to off, so **Strategy B** is the common case and is fully fixed. ( ok) 
But `Strategy A` is only hit when someone explicitly enables the flag (it exists to keep VS Code's IPC lock at the default location so vscode:// OAuth callbacks / GitHub auth find the running window). For that subset, the bug persists.
Why the "just redirect to the right path" fix is not clean. You can't simply redirect to the profile path, because the profile's settings location is an internal, version-dependent layout of VS Code. Hard-coding it would be fragile and break across VS Code releases.

**Recommendation**

Given it's opt-in and the correct target is unstable, the pragmatic path is:

- At minimum — declare scope: add a PR note that the fix covers the default --user-data-dir mode, that --profile mode is opt-in and still does not apply the settings template, and open a follow-up ticket. This keeps the merge honest.
- If you want to fix it properly: it needs a VS Code-version-aware approach (or documenting that profile-mode users must apply settings separately). That's a distinct behavior and best handled in a follow-up ticket rather than this PR.

cli/src/main/java/com/devonfw/tools/ide/tool/vscode/Vscode.java
hohwille: Thanks for checking and explanation.
I rechecked your changes to DirectoryMerger and you are fully correct.
I was just expecting a different approach but fully understood your solution and it is fine.
Will merge it now so we can do a re-test.

hohwille:
I have started the nightly-build after this merge.
I retested this with the new SNAPSHOT (2026.09.002-09_24_08-SNAPSHOT) and it seems to work.
In the trace log I found these messages:
```
Saved created/updated file D:\projects\devon-ideasy\.ide\vscode\main\config\User\settings.json
Saved created/updated file D:\projects\devon-ideasy\workspaces\main\.vscode\settings.json
```

Is that intended or a bug that we produce the `settings.json` in two different locations?
Should not block the release but I am just trying to understand our code and ensure we have clean behaviour.
If not we can still fix that after the release if the configuration is working fine now.

quando632:
Intended. The two files are different VSCode setting scopes and come from two different templates in the settings repository:

`$IDE_HOME/.ide/vscode/«workspace»/config/User/settings.json` are the user settings, merged from `vscode/workspace/update/.vscode/.userdata/User/settings.json`. They apply to the whole VSCode instance.
`«workspace»/.vscode/settings.json` are the workspace settings, merged from `vscode/workspace/update/.vscode/settings.json. `They apply to the opened folder and take precedence over the user settings.

